PR #9, opened 2015-08-10
Autotools

### LINKED ISSUE #2: basic autoconf usage
First steps in autotools integration is:
1. using a simple configure.ac template to generate a configure script
2. generating Makefiles from Makefile.in templates

This issue won't cover migrating the Makefiles over to automake, just generating them from templates and using some of the more important autoconf variables.

flihp:
This should be sufficient to resolve #2. It gives us basic ./bootstrap && ./configure && make functionality without making significant changes to the structure of the Makefiles themselves. It includes minor platform detection functions for the C / C++ compilers (CC & CXX) but not significant probing for dependencies etc.


wcarthur1:
Changes for autotools look good.

PR #12, opened 2015-08-12
Use directory guard function in place of PHONY for build output direc…

### LINKED ISSUE #5: unnecessary linking of libtpm.a
When running 'make' in a source tree that has already been built libtpm.a is linked even though none of its dependencies have changed. This is usually caused by a bad .PHONY. I think it's related to the DIRS target but haven't tested this theory.

This should be a very easy fix but I'm going to hold off and fix it as part of the autotools 'milestone' just for the sake of simplicity.

flihp:
…tories.

PHONY's don't work well in these cases since a PHONY target is always
considered out of date and rebuilt.

This resolves #5.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1,0 +1,1 @@
+dir_guard=@mkdir -p $(@D)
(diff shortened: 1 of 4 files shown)


flihp:
I took a look at this approach and decided I don't like it. I'm working up a new PR that uses order-only prerequisites instead.

PR #13, opened 2015-08-12
Fix bad path to resourcemgr directory.

flihp:
This gets rid of an error message output to the console and broken
dependency.


wcarthur1:
Looks good to me.

PR #16, opened 2015-08-14
Vpath

### LINKED ISSUE #3: replace debug / production targets in build with VPATH
This may be a bit controversial because it's something used to make developers lives easier. The autotools / GNU philosophy w/r to the build system however is that the audience is the user (person who downloads code and compiles it for their own system) and not the developer / maintainer. This means that specific 'debug' / 'production' targets are discouraged.

Despite removing these custom build targets we can still achieve the same goals / functionality for developers by using the builtin VPATH feature. For this issue to be resolved VPATH must work and documentation describing how to generate debug / production builds (really any build that requires custom compiler flags) needs to go into the README. Hopefully this will keep developers happy and productive wile making it easier for packagers to build this stuff for distros.

flihp:
This resolves #3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,8 @@ configure
 config.log
 config.status
 Makefile
-resourcemgr/debug/
-resourcemgr/release/
+resourcemgr/lib/
+resourcemgr/resourcemgr
 sysapi/lib/
-test/tpmclient/debug/
-test/tpmclient/release
+test/tpmclient/lib/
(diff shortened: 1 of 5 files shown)


wcarthur1:
Kicking this back to you for rebasing ('cause you told me to!)

flihp:
Reopening and assigning to myself for the rebase. The rebased code will show up in this same PR when it's ready.

Issue #4, opened 2015-08-06
Migrate Makefile.in to Makefile.am (automake)


flihp:
Moving from Makefile to Makefile.in templates (as described in #2) is a checkpoint on the road to full automake integration. This means using Makefile.am files and having the configure process generate our Makefile.ins for us. This will cover implementing the standard GNU build targets (install, clean, dist etc).

flihp:
resolved by #22

PR #22, opened 2015-08-28
Automake

flihp:


flihp:
This resolves #4

wcarthur1:
Philip,

Comments:
1.  Re the changes to .gitignore:  those look good for Linux, but not for Windows.  I'm assuming that autotools/automake aren't going to handle the Windows build.  If that's the case, some of the .gitignore things you deleted need to be present for Windows.  BTW, feedback I've received from the TSS WG indicates that they really like the Windows option, so I think we need to keep this.
2.  Is there any way to keep the autogeneration of source and include files in the automake environment?  If not, that's fine; it not a hill worth dying on.  I just really liked it.

flihp:
Autotools will work on Windows under cygwin but I haven't tested it. This would be convenient so we're not maintaining 2 build systems in the same source tree. Not sure how people's preferred tools on Windows will interact with this though.

Adding back the right .gitignore rules to get us back to where we were is probably the right first step.

As to your #2 above autotools doesn't seem to like doing such things. I can rig the configure step up to generate a list of these files for automake to import but doing these sort of things is pretty "non-standard". This could also just be my ignorance though so I'll keep digging into automake to see if this feature exists. Does Visual Studio have similar functionality? It would be a nice way to keep breakage to a minimum when we have multiple people working in different build environments.

wcarthur1:
Philip,

Below…

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
Doesn't look like the .gitignore lines for Windows were added back in.  I thought they were from you comment above.  Do you want to add them back in after I accept these changes?

flihp:
Hey Will. I just rebased this with ignore rules for the VisualStudio bits. Sorry for the delay, took me a while to remember how do stuff in VS again.

wcarthur1:
Philip,

I still see that the new .gitignore is going to remove the ignore rules for VisualStudio.  I’m not sure what you mean by “rebase”, but it doesn’t seem to have resulted in any changes to the .gitignore.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
Added another line to ignore the VS 'Release' build artifacts. Rebased again. Should be OK but testing on the version of VS that you're using would probably be a good idea.

PR #34, opened 2015-09-08
build: Remove libcommon.a, reference common source directly.

### LINKED ISSUE #32: Get rid of libcommon.a
This was a hack to work around some weirdness in the build when migrating to automake. Once #25 is sorted libcommon.a shouldn't be necessary. Odds are it was only necessary in the first place due to my lack of automake skillz.

flihp:
This abstraction leaked on account of having to specify SAPI_CLIENT at
compile time for the tpmclient. By removing libcommon.a and specifying
the source files directly through the *_SOURCES variables we compile
individual source files in the common directory with the different
C(XX)?FLAGS for both the resourcemgr and the tpmclient.

Also worth noting is that the headers for libtpm are no longer included
in the *_SOURCES variables for resourcemgr and tpmclient. This dependency
is automagically accounted for by automake.

This resolves #32

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,18 +27,14 @@
 #;**********************************************************************;
 
 bin_PROGRAMS = resourcemgr/resourcemgr test/tpmclient/tpmclient
-noinst_LIBRARIES = common/libcommon.a sysapi/libtpm.a
-
-common_libcommon_a_CFLAGS  = -I$(srcdir)/sysapi/include/ \
-    -I$(srcdir)/tcti/tpmsockets/ -I$(srcdir)/test/tpmclient/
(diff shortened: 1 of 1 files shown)


wcarthur1:
I think I already did this is some of my recent changes.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
If that's the case, probably best to close this pull request and mark it as a duplicate. AFAIK our master branch still builds libcommon.a though (see: https://github.com/01org/TPM2.0-TSS/blob/master/Makefile.am#L32). Not sure what changes you're referring to above. Can you provide a pointer to a commit?

PR #39, opened 2015-09-10
build: Generate lists of source files at bootstrap time in src_vars.mk.

### LINKED ISSUE #26: source file enumeration
I couldn't figure out how to get automake to enumerate all source files for the build like we were doing in our hand made Makefiles. I've found a few examples of projects that do this at configure time and it seems like a viable solution.

flihp:
This moves the definition of SYSAPI_SRC, COMMON_SRC, LOCALTPM_SRC,
SAMPLE_SRC and TPMSOCKETS_SRC variables to src_vars.mk. This file is
generated at bootstrap time and the list of source files in each
variable generated dynamically. We leave the list of source files for
the applications (resourcemgr and tpmclient) in the Makefile.am due
to the Windows specific files in the tpmclient directory that the
bootstrap script would pickup. We can work around these with #ifdefs
in the future if necessary.

This resolves #26.


wcarthur1:
Need to rebase to latest code.

flihp:
Should be good to go.

PR #41, opened 2015-09-15
resourcemgr: add a command line option "-sharedhandle"

### LINKED ISSUE #40: Add option to allow transient object handle to be used across connections in resourcemgr
In current resourcemgr, transient virtual object handle can only be accessed by the creation connection. This make things hard for the collaboration of standalone tools based upon SAPI.

We can add a command line option "-sharedhandle" to explicitly allow the sharing of transient handles across connections.

gwei3:
close #40

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -1,27 +1,27 @@
 //**********************************************************************;
 // Copyright (c) 2015, Intel Corporation
 // All rights reserved.
-// 
-// Redistribution and use in source and binary forms, with or without 
+//
+// Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
-// 
-// 1. Redistributions of source code must retain the above copyright notice, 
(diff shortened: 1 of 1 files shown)


wcarthur1:
Jimmy,

We'll need to talk about this one.  What you've done makes all handles "shareable" which violates the TSS resource manager specification.  

The right way for applications to "share" an object is to save the context of the object, send it to another application which can then load the context.  This means that there will be two instances of the same object, one owned by each process.

Will

gwei3:
Will,

I will look into the “right” way you suggested. But I still think using a command line option to enable this behavior explicitly can give users one more choice in their solution to just leverage RM to do context management w/o access control.

Jimmy

gwei3:
Cancel the pull request until agreement was made.

wcarthur1:
Let’s talk about it in tomorrow’s meeting.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

PR #47, opened 2015-10-05
Make libtpm and resource manage installable + some cleanup.

flihp:
This is, I think, the last bits necessary to make the Linux build friendly to the expected workflow. With these changes you can build libtpm and the resource manager like:

configure && make && make install

libtpm will be built as a static, and a dynamic library. The resource manager will be dynamically linked to libtpm by default (thought this can be changed at configure time). Also the sysapi header files are installed so other programs can link against the shared object.

Will: can you take a look at the install locations and make sure they're what you want? I opted to use the 'pkg' standard which installs libraries and headers under in the expected locations ($prefix/lib and $prefix/include respectively) in their own directory. Since the project name is 'tpm2.0-tss' this ends up being on that path: $prefix/lib/tpm2.0-tss and $prefix/include/tpm2.0-tss.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -44,5 +44,11 @@ $ ../TPM2.0-TSS/configure \
   CXXFLAGS="-O0 -Wall -Werror -fno-operator-names -fpermissive -ggdb3"
 $ make
 
-The tpm2.0-tss software currently does not have an install target. This
-section will be updated with installation instructions when they are relevant.
+Once you've built the tpm2.0-tss software it can be installed with the usual:
+$ sudo make install
+
+This will install libtpm and the resource manager to locations determined at
+configure time. See the output of ./configure --help for the available
+options. Typically you won't need to do much more than provide an alternative
+--prefix option at configure time, and DESTDIR at install time if you're
+looking to create a binary packag.
--- a/Makefile.am
+++ b/Makefile.am
@@ -30,12 +30,14 @@ include src_vars.mk
 
 ACLOCAL_AMFLAGS = -I m4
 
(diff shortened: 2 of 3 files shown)


flihp:
@wcarthur1 : I just started playing around with writing a simple program to use the tcti/sysapi interface and realized that I need to be able to initialize a TCTI context in order to use the SYSAPI (reading spec here: http://www.trustedcomputinggroup.org/resources/tss_system_level_api_and_tpm_command_transmission_interface_specification). AFAIK the TCTI stuff doesn't get compiled into libtpm currently so there's no way for a user application (compiled outside the TPM2.0-TSS project) to use the resourcemgr or even talk to the TPM2 device node directly.

Since libtpm currently contains only the SYSAPI stuff we may want to just rename this to libsysapi and compile the TCTI bits into libtcti. Alternatively we could roll the whole thing into libtpm. Do you have any opinions on how this should be done?

wcarthur1:
Philip,

TCTI is an interface and, and from a specification viewpoint, really only exists as a header file.

The functions to initialize TCTI instances are communication channel-specific and aren’t part of the SAPI/TCTI specification.

We have two examples of TCTI instances, local and sockets, but these are more like sample code.

My preference:  roll each of these into a separate library, liblocaltcti and libsocketstcti.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
Will take this into account and resubmit. Probably going to make a few issues to track stuff related to dependencies that have polluted the bits that we want to turn into libraries as well as the desired structure.

Issue #38, opened 2015-09-10
libraries and 'install' build target


flihp:
Currently libtpm.a is a 'noinst' library so it's not installed when 'make install' is invoked. If we want distros to package this we'll need to sort out installing it in the right place along with the resource manager.

flihp:
Had a chat with @wcarthur1 yesterday. We agreed that the sapi library alone isn't sufficient, we need to break out the tctis from the repo into libraries as well. The agreed upon structure for libraries is:

* All libraries will be stalled directly into $(libdir). No subdirs (aka no pkg* stuff).
* libtpm2sapi: All code under $(topdir)/sapi goes into a library dedicated to the SAPI. Call this library libtpm2sapi.(a|so). Headers will be isntalled into $(includedir)/tpm2sapi
* libtpm2tctidev: All code under $(topdir)/tcti/localtpm goes into a library dedicated to the TCTI that talks directly to the /dev/tpm* device nodes. Call this library libtpm2tctidev.(a|so). Headers will be installed into $(includedir)/tpm2tctidev
* libtpm2tctisock: All code under $(topdir)/tcti/tpmsockets goes into a library dedicated to the TCTI that talks to the TPM through a socket. Call this library libtpm2tctisock.(a|so). Headers will be installed into $(includedir)/tpm2tctisock

flihp:
Creating an issue for each library to track incremental progress.

PR #49, opened 2015-10-10
Create shared libraries and set install locations.

### LINKED ISSUE #37: libtpm shared object
Build libtpm.a as a shared object. The benefits here are the same as for every other shared object: save space by sharing executable code. Hopefully this is just some more autotools magic.

flihp:
This pull request resolves #37 and #38. To test it do the following:

$ ./bootstrap
$ ./configure --prefix=/usr
$ make -j$(($(nproc)*2)) 2>&1 | tee build.log
$ mkdir destdir && DESTDIR=$(pwd)/destdir make install

This will build the code which will produce the new libraries described in #38. It will then install everything into a subdirectory called 'destdir' in the source tree. This is how most distros build packages but it's useful here so you can see where everything will be installed. You can poke around in this directory to see the libraries landing in destdir/usr/lib and headers in destdir/usr/include.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 *.a
+*.la
+*.lo
 *.o
 *.opensdf
 *.pdb
@@ -6,6 +8,7 @@
 *.suo
 .deps/
 .dirstamp
+.libs
 aclocal.m4
 autom4te.cache/
 [Bb]uild/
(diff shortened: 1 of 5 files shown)


flihp:
Nevermind. This doesn't take into account the pile of dependencies shared between the localtpm and tpmsockets TCTI and the resource manager. Separating the two TCTIs won't be possible till the dependencies on the resource manager are broken up, or we compile those dependencies into each library. I'm going back to the drawing board on this one.

PR #52, opened 2015-10-27
resourcemgr: return error earlier for non-implemented command code

gwei3:
Update function GetNumCmdHandles to return -1 instead of 0 when command
code was not supported by the TPM, and handle the -1 case properly.

This prevent wrong unmarshalling sequence to be applied when command
code with handles > 0 but GetNumCmdHandles return 0 when it is not
supported by TPM.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -85,6 +85,10 @@ int GetNumCmdHandles( TPM_CC commandCode, TPML_CCA *supportedCommands )
             rval = cmdAttributes.cHandles;
         }
     }
+    else
+    {
+        rval = -1;
+    }
 
     return rval;
 }
@@ -1200,7 +1204,12 @@ TSS2_RC ResourceMgrSendTpmCommand(
     }
 
     numHandles = GetNumCmdHandles( currentCommandCode, supportedCommands );
-    
+    if( numHandles < 0 )
+    {
+        responseRval = TPM_RC_COMMAND_CODE;
+        goto SendCommand;
+    }
+
     for( i = 0; i < numHandles; i++ )
     {
         RESMGR_UNMARSHAL_UINT32( command_buffer, command_size, &currentPtr, &( cmdHandles[i].handle), &responseRval, SendCommand );


wcarthur1:
Jimmy, what problem were your trying to fix here?  If the number of command handles is 0 because the command code is unsupported, no handles will be virtualized and the command will be sent to the TPM unaltered, and it will return an error code.  The design rationale in SAPI and TAB/RM is to let the TPM handle errors as much as possible.
Unless you can provide some compelling reason for this change, I plan to reject it.  One of my reasons is that a test suite might want to sent unsupported commands to the TPM--I'd rather have the TPM reject them than the TAB/RM.

wcarthur1:
Jimmy, I added a test for unsupported command code.  It works fine, except for one issue when debug level is bumped up.  I am rejecting your pull request, but will add my own for my changes.

gwei3:
Will, if the unsupported command code itself requires handles, then the caller will marshal handles into the cmd buffer. resourcemgr is trying to unmarshal part of buffer to do so some handling before call lower layer to transmit the cmd buffer. if GetNumCmdHandles returned 0, then the handles in current cmd buffer will not be unmarshalled, and the buffer pointer will still point to the first handle, but then resourcemgr will try to use the wrong buffer pointer(@handle) to unmarshal other type of parameter, which will cause unexpected issues.
In my case, tpm2_hmac was called on a tpm implementation which didn't support it, we expect a TPM_RC_COMMAND_CODE return code from SAPI, but in fact we will get a error 0xC0006(TSS2_RESMGR_CONTINUE_BIT_MISMATCH). It took me quite a lot of time to figure out it is caused by wrong unmarshalling sequence on unsupported command code.

Refer to resourcemgr/resourcemgr.c:1202~1236.

wcarthur1:
Jimmy,

Below…

Can you set “-dbg 3” on your tpmtest command line and send me the output of the resource manager from the error case you mention below?

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
And…can you send me the test case that’s failing this way.

I’m looking at the code and can’t see how you’re getting this problem, so I need the exact error case so that I can duplicate it.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

From: Arthur, Will C
Sent: Thursday, October 29, 2015 8:33 AM
Subject: RE: [TPM2.0-TSS] resourcemgr: return error earlier for non-implemented command code (#52)

Jimmy,

Below…

Can you set “-dbg 3” on your tpmtest command line and send me the output of the resource manager from the error case you mention below?

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
Jimmy,

I fixed the problem.  You were right about the return value needing to be -1.  But I decided in that case to push the unsupported command to the TPM.

Also there was a bug with error code overlap; this caused the error to be reported incorrectly.

I fixed all these issues.  See my latest checkin.

Issue #51, opened 2015-10-14
Tss2_Sys_SetCmdAuths issue


juliantrzeciak:
file: https://github.com/01org/TPM2.0-TSS/blob/master/sysapi/sysapi/authorizations.c 
commit: https://github.com/01org/TPM2.0-TSS/commit/ca8e7f3c705553f95e6ecafe96b291ab57dbb71f)

**1. TSS2_SYS_RC_BAD_VALUE should cause immediate return**
this is "source null pointer" scenario
line 78 change **break** to **return rval**

wcarthur1:
Julian,

I've added tests and fixed all the issues with SetCmdAuths. 
Can you look at my pull request #65 and approve them for merging?

Thanks,
Will

juliantrzeciak:
Will,

I'm affaid this one is not fixed.

Scenario is that in 
`
 TSS2_RC Tss2_Sys_SetCmdAuths(
     TSS2_SYS_CONTEXT            *sysContext,
     const TSS2_SYS_CMD_AUTHS 	*cmdAuthsArray
     )
`
 
the case for cmdAuthsArray==NULL is covered in line num.40

But I was thinking of kind more weird scenario when null-pointer for one of auths member within the TSS2_SYS_CMD_AUTHS 

`
// Input structure for authorization area(s).
typedef struct {
    uint8_t cmdAuthsCount;
    TPMS_AUTH_COMMAND **cmdAuths;
} TSS2_SYS_CMD_AUTHS;
`

I'm sorry for confusion.

now have a some pointer to TSS2_SYS_CMD_AUTHS structure but first of TPMS_AUTH_COMMAND members is null then on line 78 break out of 72-92 'for loop' then set auth from a null member, then disaster.

Best Regards,
Julian

wcarthur1:
Julian,

I looked at this, and initially I agreed with you.

But then I was trying to figure out why my test case didn’t catch this.

What will happen when I break out of the loop is that I will calculate the size of any valid auth structures found before the NULL one (in the case you mention of the first one being NULL, there would be no valid auth structures.  Then I will move existing parameters out of the way to make room for the auth structures (line 106).  Then at line 108, I will see the BAD_VALUE rval and skip the rest and exit.

I do agree that it would be clearer if I checked the rval after exiting the for loop and handled it there.  As soon as I get a chance, I will make that change.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

Issue #50, opened 2015-10-14
Tss2_Sys_GetRspAuths issues


juliantrzeciak:
file: https://github.com/01org/TPM2.0-TSS/blob/master/sysapi/sysapi/authorizations.c 
commit: https://github.com/01org/TPM2.0-TSS/commit/ca8e7f3c705553f95e6ecafe96b291ab57dbb71f)

**1. missing oversize check TSS2_SYS_RC_INVALID_SESSIONS**
for hmac at line 201

**2. skip rspAuth extraction when destination pointer ==0**
according to:
http://www.trustedcomputinggroup.org/files/static_page_files/28CE28FA-1A4B-B294-D0A965038809FE9B/TSS%20system%20API%2001.pdf
TSS System Level API and TPM Command
Transmission Interface Specification
Family "2.0"
Level 00, Revision 01.00
26 January 2015

page 41
>8.7.4.2 Tss2_Sys_GetRspAuths
>...
>This function gets the response authorization data from the sysContext. If an element in the
>rspAuthArray->rspAuths array is NULL, that authorization is skipped in the sysContext.

Scenario: 
sysContext contains 3 rspAuths, 
rspAuthArray->rspAuths[1] == 0 // element is NULL but others not
Tss2_Sys_GetRspAuths(sysContext, rspAuthArray) 
**result**: rspAuthArray->rspAuths[2] is not populated with third rspAuth from sysContext
**expected result**: 
rspAuthArray->rspAuths[0] is populated with first rspAuth from sysContext
rspAuthArray->rspAuths[1] extraction skipped
rspAuthArray->rspAuths[2] is populated with third rspAuth from sysContext

Let me know If I understand it correctly, then I can provide a fix.

**3. TSS2_RC_SUCCESS when rspAuthsArray == NULL**
according to doc it should be:
>TSS2_SYS_RC_BAD_REFERENCE: if sysContext or rspAuthsArray are NULL

update for doc?

**4. not implemented TSS2_SYS_RC_BAD_VALUE: if rspAuthsArray.authsCount is 0**
its covered by either by 
>TSS2_RC_SUCCESS when SYS_CONTEXT->authsCount ==0

or
>TSS2_SYS_RC_INVALID_SESSIONS when ( rspAuthsArray->rspAuthsCount != SYS_CONTEXT->authsCount )

update for doc?

wcarthur1:
Julian,

I've added tests and fixed all the issues with getRspauths. 
[...]

PR #65, opened 2015-11-09
Issue#50

wcarthur1:
Added tests and fixes for issues #50 and #51 and fixed other bugs found:

Fixed bug in RM:  when it gets a zero sized response from the
TPM, it was returning a misleading error code.

Fixed bugs in Tss2_Sys_GetRspAuths:
1.  We weren't doing bounds checking for hmac field. 
2.  We weren't checking rspAuthsArray for NULL.
3.  Returning wrong error code when incomplete authorization structures
    are returned.
4.  For commands that can't take authorization, it was returning success.
5.  Wasn't returning an error when rspAuthsCount was set to 0.

Fixed bugs in Tss2_Sys_SetCmdAuths function:
1. We weren't checking for NULL pointers in cmdAuths array.

Created an offset for level- and implementation-specific error codes per agreed on changes
to SAPI spec.

Added tests for Tss2_Sys_SetCmdAuths and Tss2_Sys_GetRspAuths.

Add power down command at beginning of tpmtest().  This will restart the 
TPM simulator in case previous test exited without cleaning up.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -1607,6 +1607,12 @@ TSS2_RC ResourceMgrReceiveTpmResponse(
                 goto returnFromResourceMgrReceiveTpmResponse;
             }
         }
+
+        if( *response_size == 0 )
+        {
+            rval = TSS2_TCTI_RC_IO_ERROR;
+            goto returnFromResourceMgrReceiveTpmResponse;
+        }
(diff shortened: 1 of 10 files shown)


wcarthur1:
Approved this merge after Julian reviewed my changes.  He found one issue that needed fixing, and I fixed it.

Issue #83, opened 2015-12-08
Need to solve tboot 1.8.3(TPM 2.0) issue


kkamagui:
When TSS is used with the tboot 1.8.3(TPM 2.0), the resource manager of TSS returns an error code to the client. Because of this error code, the client doesn't operate properly. 

```bash
# Error messages of the resource manager
Resource Manager Other CMD Server accepted client
Accept socket:  0x6
Resource Manager TPM CMD Server accepted client
||  Loaded transient object handles: 
||  0x80000000, TpmCmdServer died (TPM CMD), rval: 0x00000000, socket: 0x6.
OtherCmdServer died (Other CMD), socket: 0x7.
```

PR #86, opened 2015-12-11
 Fixing error for using TSS with tboot 1.8.3 (issue #83).

kkamagui:
tboot 1.8.3(TPM 2.0) create an object handle during boot process. However the resource manager of TSS seems to presume this handle as abnormal object and returns an error code to client.

To solve this problems, the resource manager is changed to flush all loaded handles before the resource manager is started.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -611,7 +611,45 @@ endTestForLoadedHandles:
     return rval;
 }    
 
+TSS2_RC FlushAllLoadedHandles()
+{
+    TPMS_CAPABILITY_DATA capabilityData;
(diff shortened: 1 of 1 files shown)


wcarthur1:
Took at look at this.  sorry for taking so long...been on vacation.  One thing that needs to be added:  need to also flush the context for all active sessions in the flush handles function.

wcarthur1:
I assume we can delete this one, since I accepted your more recent changes.

kkamagui:
Hello, wcarthur1.

I will close this pull-request and issue #83 immediately.

Thank you.

======== 
Seunghun Han 
National Security Research Institute

PR #87, opened 2015-12-14
Finished all TCTI and SAPI tests and bug fixes.

wcarthur1:
These changes fix issues #: 71, 74, 75, 76, 77, and 78.  Description of the changes:

    Added changes to properly return MALFORMED_RESPONSE errors.

    Developed tests for local TPM TCTI interface.
    
    fixed bug with local TPM TCTI code:  wasn't returning correct
    size when response_buffer is NULL.
    
    Altered TestTctiAPIs to run with passed in tcti context
    so that I can create different versions of the test for
    different interfaces.

    Got TCTI tests working against resourcemgr/PTT.

    Fixed localtpm tcti code.

    Fixed TCTI test code that forces an IO error.

    Fixed 21 instances of bad size in a TPM2B in tpmclient tests.

    Preliminary changes to localTPM TCTI code.

    Added receive test for NULL responseBuffer pointer.

    Got all tests passing for receive with too small a buffer.

    Added support for receive when response buffer is too small.

    Changed unmarshalling of TPM2B_CREATION_DATA.  Before it was unmarshalled as
    a simple TPM2B, but should have been a complex TPM2B.
    
    Added routines for unmarshalling nested TPM2B's--this new routine doesn't
    check the size field.
    
    Changed Marshal_Simple_TPM2B to return INSUFFIENT_BUFFER instead
    of INSUFFICIENT_CONTEXT.
    
    Added size check to Unmarshal_Simple_TPM2B routine and updated all
    appropriate test and resource manager code.

    Changed most instances of TSS2_SYS_RC_INSUFFICIENT_BUFFER to
    TSS2_SYS_RC_INSUFFICIENT_CONTEXT in order to match SAPI spec.

    Added test for GetRpBuffer when no response handles or params.
    
    Found bug with GetRpBuffer when response has no authorizations.  Fixed it.
    
    Added test for GetRpBuffer when response has no authorizations.


gwei3:
Hi, Will,

I will try to make the review done this week.

PR #88, opened 2015-12-15
[resourcemgr]fix for Tss2_Sys_ContextSave(session)

gwei3:
Execute tpmtest with resourcemgr connected to simulator, then follow the hint to run all tests in 2 - STARTUP TESTS, then run all tests in 10 - START_AUTH_SESSION TESTS, will get failure code 0xCF802(TSS2_RESMGR_FIND_FAILED).

The root cause is that the in memory context will be flushed after running savecontext for a session, so the original code try to prevent resourcemgr automatically evicting the context again via setting foundEntryPtr->status.loaded = 0. But this will cause the final call to EvictEntities failed.

The solution is to avoid calling into EvictEntities for Tss2_Sys_ContextSave(session) case while adding more if conditions.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -2070,6 +2070,8 @@ returnFromResourceMgrReceiveTpmResponse:
     // If command was FlushContext, the entry was already removed from the list.  No eviction
     // necsssary or possible (because no entry exists for this object or sequence anymore).
     if( numHandles &&
+            !( currentCommandCode == TPM_CC_ContextSave &&
+                IsSessionHandle( cmdSavedHandle ) ) &&
             ( currentCommandCode != TPM_CC_FlushContext ) )
     {
         // Create array of handles.


wcarthur1:
Jimmy, please provide me with a branch and commit # for this.  I merged in my latest changes to master branch and tpmtest doesn't run with those.

wcarthur1:
Jimmy,  I'm also trying to understand:
1.  How the DEBUG_GAP_HANDLING #define is being set.  The code seems to think it's being set, but I can't see how.
2.  What changed to cause this error?  This test used to work right, so what changed?

wcarthur1:
One more note:  my tpmclient.cpp test works fine.  So I don't understand why tpmtest is failing, nor the  need for a resource manager change to make tpmtest work.

gwei3:
I have rebased the changes to the latest top of master.

the DEBUG_GAP_HANDLING macro was defined in resourcemgr/resourcemgr.h, so it is always defined if this head file was included. I think this case might not be well tested after tpmtest was included.

tpmclient bypassed the start_auth_session_tests by default, and it will only be tested when "-startAuthSessionTest" option was given.

wcarthur1:
Jimmy,

A couple things here:
1.  We shouldn't be running the tpmtest or tpmclient with DEBUG_GAP_HANDLING set.  My notes in resourcemgr.h say that this will only work against a doctored up version of the simulator.
2.  But...even with that, your bug fix is correct.  You found a test case that hadn't been caught before.  Thanks!!

I will accept the change.
I will also be adding a change to uncomment the #define for DEBUG_GAP_HANDLING.

wcarthur1:
This bug was the result of resource manager changes.  It didn't get caught because my tpmclient doesn't normally run the TestStartAuthSession function.  I'm fixing that.

wcarthur1:
You need to port your tpmtest to the latest changes in the TSS library code and the tpmclient.cpp file.  The best way to do this is to create shared code for tpmtest and tpmclient that incorporates all the code that is common to both.  This will make minimize maintenance in the future.

gwei3:
Sure, let me open another issue to track this.

PR #92, opened 2016-01-04
Fix include preferences

theopolis:
When installing the development headers the build creates:

```
$(PREFIX)/include/tpm2sapi
$(PREFIX)/include/tpm2tcti
```

This does not reflect the project file hierarchy and such the include-path search will fail unless projects explicitly set add `$(PREFIX)/include/tpm2sapi` to their include paths. It makes more sense to reference these headers locally.

### DIFF
--- a/sysapi/include/tpm20.h
+++ b/sysapi/include/tpm20.h
@@ -43,15 +43,15 @@
 #include    <stdlib.h> 
 #include    <string.h> 
 
-#include    <basetypes.h>
-#include    <tpmb.h>
-#include    <implementation.h>
-#include    <tss2_tpm2_types.h>
+#include    "basetypes.h"
+#include    "tpmb.h"
+#include    "implementation.h"
+#include    "tss2_tpm2_types.h"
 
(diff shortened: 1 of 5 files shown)


flihp:
Thanks @theopolis, I'll get this reviewed today.

flihp:
The changes to the include directives look fine. The changes you've made to the body of the tpmsockets.cpp file look unrelated though. Seems sane to have what looks like a debug print function within the DEBUG_SOCKETS ifdef though so this may well be something we want to pull in but probably in a different PR. This change is easy enough so I'll clean this up as part of the merge.

theopolis:
Awesome, thanks for the review and for doing that @flihp, you rock!

flihp:
Made some minor edits to this PR: removed a whitespace cleanup and a change to the location of a call to a debug print functions. I'm all for cleanups to whitespace and debug stuff but that should go into a separate PR. If you're motivated to help clean that sort of stuff up I'll gladly review those changes. Just assign the PR to me.

Issue #107, opened 2016-01-21
global data woes


flihp:
I'm running into a lot of issues with the way the TSS code relies on variables that are declared using 'extern' keyword. This declares a variable without actually defining it. This punts a lot of checks that can be done at compile time to link time. Another side effect is lots of global data and weird dependencies between c files (see dependency between resourcemgr.c and the tcti implementations).

extern is misused in a number of places as well. Specifically the tcti headers where a few functions are defined using the extern keyword.

flihp:
This issue overlaps #105 significantly. It's also a bit of a misnomer in that I don't intend to eradicate all global data in the code but only global data that has an impact on the libraries. Closing this in favor of using #105

PR #108, opened 2016-01-21
Big change to better align with spec for exposed headers and structures.

### LINKED ISSUE #100: don't install util headers
Had a chat with @wcarthur1 today and it looks like my efforts to install the SAPI headers were a bit overly ambitious. Some of the headers we're now installing aren't intended for public consumption. This issue is to track identifying which headers shouldn't be installed and then backing them out of the build / install target.

### LINKED ISSUE #102: consistent naming of tcti headers
The headers for the TCTIs we provide should be named consistently. The library they're built into, as well as the directory they're installed into need to be renamed as part of the overall cleanup effort.

This ticket covers the following reorg:
tpm2tcti/localtpm.h -> tcti/tpmdevice.h
tpm2tcti/tpmsocket.h -> tcti/tpmsocket.h
libtpm2tctidev.so -> libtctidev.so
libtpm2tctisock.so -> libtctisock.so

flihp:
Top level changes:
- Rename the tpm2sapi and tpm2tcti include directory to the preferred tss2
  and tcti respectively.
- Move headers that aren't part of the spec out of the installed system
  directory and remove the tss2_ prefix from them. They're back in the
  sapi include directory. There are still a few files that aren't part of
  the spec (tpm20.h) but we're including these for convenience for the
  time being. This is still very much subject to change but future changes
  will have a much lower impact.
- Rename the libraries produced:
  libtpm2sapi.so -> libtss2.so
  libtpm2tctidev.so -> libtctidev.so
  libtpm2tctisock.so -> libtctisock.so
Minor changes:
- Attempt to get some consistency in our use of either "" or <> form of
  the include directive. For the headers that have moved we're using the
  <> form for system headers including those installed by this package.
  For headers that are internal to the project we're using the "" form.
- Cope with the removal of some 'extern' variables declared in
  previously public headers. I've only touched the cases that were
  required to move the headers currently so more work will be required
  to clean this up completely and that will be tracked as part of issue
  #107.
- Fixup both autotools and visual studio build files.

Resolves #100
Resolves #102


flihp:
Apologies for the "big bang" patch but it was a bit unavoidable. I don't expect anyone to be able to review this line by line (~300 files were changed) so the approach I recommend is:
1) make sure it builds on your platform
2) verify tests work as expected
3) verify the headers and library files land where you expect them to be

This PR will likely break any client consuming the TSS2 so considered your collective selves warned. Please DO NOT MERGE this till I've been able to assess the impact on the tpm2.0-tools repo. I'll be working up a fix sync'd with this PR over the weekend likely. I did however want to get this PR up for testing in parallel to this work. Hopefully we'll be able to sync pushes to both repos next week to minimize breakage.

flihp:
I've been poking at the tpm2.0-tools repo to do a sort of impact assessment. I've managed to break down a few incremental changes that can be made and so will close this PR. I'll have an updated PR with similar but hopefully smaller changes in the near future.

PR #110, opened 2016-01-24
Move TCTI_MAGIC and TCTI_VERSION values to a common include.

### LINKED ISSUE #109: TCTI_MAGIC and VERSION
From the spec:
> The magic value is some unique number, perhaps simply a random number. The SAPI implementation
> will likely not read it. The value is a sanity check, ensuring that the lower TCTI layer receives a context
> that it can interpret, typically one it created.

This value is currently in the tss2_sapi_util.h header which we've resolved to keep internal to the TSS2 implementation. The quote above doesn't stipulate one way or the other as to whether or not this value can be exposed to the consumer of a given TCTI. The current issue however is that the initialization function for our TCTI implementation requires the caller provide this value. When the util header goes away the TCTIs will break. Both the version and the magic value need to be associated with the TCTI headers somehow.

flihp:
These two values must be specified by clients consuming our TCTI
interface. Currently they reside in the tss2_sapi_util.h file that we
include in the tpm2sapi headers. This is a non-standard file that will
be removed soon and so this value must be moved to keep from breaking
client code. Since these two values are used by both the dev and socket
TCTIs I've put them into a common 'magic.h' file.

This resolves #109

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,6 +29,8 @@
 include src_vars.mk
 
 ACLOCAL_AMFLAGS = -I m4
+AM_CFLAGS = -I$(srcdir)/include
+AM_CPPFLAGS = -I$(srcdir)/include
 
 # stuff to build, what that stuff is, and where/if to install said stuff
 sbin_PROGRAMS   = $(resourcemgr)
--- a/include/tpm2sapi/tss2_sysapi_util.h
+++ b/include/tpm2sapi/tss2_sysapi_util.h
@@ -43,9 +43,6 @@ extern "C" {
 
 enum cmdStates { CMD_STAGE_INITIALIZE, CMD_STAGE_PREPARE, CMD_STAGE_SEND_COMMAND, CMD_STAGE_RECEIVE_RESPONSE, CMD_STAGE_ALL = 0xff };
 
-#define TCTI_MAGIC 0x7e18e9defa8bc9e2
-#define TCTI_VERSION 0x1
-
 typedef struct {
     //
     // These are inputs to system API functions.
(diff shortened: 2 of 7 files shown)


flihp:
Assigned this to @wcarthur1 for review.

wcarthur1:
Won't build resource manager under Windows.  Error messages:

z:\tpm2.0-tss-philip\tpm2.0-tss\include\tpm2tcti\tpmsockets.h(35): fatal error C1083: Cannot open include file: 'tpm2tcti/magic.h': No such file or directory
2>  commonchecks.c
2>z:\tpm2.0-tss-philip\tpm2.0-tss\tcti\commontcti\commonchecks.c(31): fatal error C1083: Cannot open include file: 'tpm2tcti/magic.h': No such file or directory
2>  syscontext.c
3>  TpmHmac.c
2>  platformcommand.c
3>  TpmHash.c
2>z:\tpm2.0-tss-philip\tpm2.0-tss\include\tpm2tcti\tpmsockets.h(35): fatal error C1083: Cannot open include file: 'tpm2tcti/magic.h': No such file or directory
2>  getcommands.c

Under Linux, tpmclient and tpmtest don't build.  I fixed those errors, but I can't seem to check in the changes to your repository.  After fixing, tpmclient ran fine, which is what I was trying to verify.

flihp:
Thx for the testing. I didn't have a windows system handy over the weekend, completely forgot about that. I tested the Linux build w/o and issue so I'll try to repro. I'm sure I missed something. Will try to have another iteration out ASAP.

Philip

flihp:
This PR should be fixed up: functional builds for both Linux and Windows. @wcarthur1 have a look when you've got a chance.

PR #112, opened 2016-01-25
platformcommand: Link commands for interacting with the MS simulator …

### LINKED ISSUE #111: Packaging platform commands
In going through our headers as part of #105 I've realized that the socket TCTI exports functions for executing 'platform commands'. These are commands for controlling the Microsoft TPM2.0 simulator. As part of the work I've done for packaging for the library for Linux I missed this code (it was in the 'common' directory) and was instead building it into the binary programs we produce.

This issue covers building these functions into the libtpm2tctisock library. The tpm2.0-tools will appreciate this I think :)

flihp:
…into libtpm2tctisock.

The platformcommand code is now common but consumed through the
libtpm2tctisock library. It should not be linked into binaries directly.
The accompanying command codes have been moved into the tpmsockets.h
file to expose them to clients.

This resolves #111.


flihp:
assigned to @wcarthur1 for review

PR #119, opened 2016-02-10
Incomplete macro __cplusplus

lakshminarayanand:
The macro defined with extern "c" seems to incomplete .
This patch helps to fix the closure of the defined macro


flihp:
Merged manually to resole conflicts. Thanks!

PR #127, opened 2016-02-22
tpmtest: add a missed size initialization

gwei3:
This might cause random failures.


wcarthur1:
Did this actually fix something?  It doesn't look to me like it's necessary, but it doesn't do any harm either.

PR #130, opened 2016-03-03
Tcti init cfg

flihp:
In this PR:
Remove the TSS2_TCTI_DRIVER_INFO struct.
Trim unnecessary stuff from Init and Teardown functions.
Replace configuration string and parsing logic with simple structure for both TCTIs.


flihp:
Updated device TCTI init function to use 'const' type qualifier.

PR #132, opened 2016-03-06
tcti_socket: Remove simulator flag.

### LINKED ISSUE #131: Remove 'simulator' extern declaration from socket TCTI header
This is a uint8_t declared with the 'extern' keyword in the tcti_socket header: https://github.com/01org/TPM2.0-TSS/blob/master/include/tcti/tcti_socket.h#L107
This means that it's defined in some other source file and resolved at link time. Declaring it like we do in the header effectively exposes global data from the application consuming the TCTI to the TCTI.

AFAIK we do this as a way to allow the consuming application to control the behavior of the TCTI. The behavior we're controlling is the way that the TCTI handles 'platform commands'. The resource manager uses this as a mechanism to filter out platform commands when it's talking directly to the device TCTI. This TCTI doesn't have a way to send platform commands to the physical TPM, and we assume the commands are specific to the microsoft simulator.

My take on this is: this is the wrong place to do the check. We expose the 'PlatformCommand' through the socket TCTI interface. If this value is not set then the PlatformCommand becomes an effective noop. If the resource manager wants to filter the messages / commands that it's proxying, then the code should be in the resource manager.

This is a task generated by #107

flihp:
The resourcemgr is a better place to filter PlatformCommands. The TCTI
shouldn't care whether or not the resourcemgr is talking to the
simulator or not. We effectively move the check up a level in the call
scope to the resource manager. The check now happens before the resource
manager calls PlatformCommand.

Consuming applications will no longer be required to define this
variable. The socket TCTI will always send platform commands when the
command socket is connected.

This resolves #131

### DIFF
--- a/include/tcti/tcti_socket.h
+++ b/include/tcti/tcti_socket.h
@@ -101,9 +101,6 @@ TSS2_RC SendSessionEndSocketTcti(
     UINT8 tpmCmdServer
     );
 
-extern UINT8 simulator;
-
-
 // Commands to send to OTHER port.
 #define MS_SIM_POWER_ON         1
 #define MS_SIM_POWER_OFF        2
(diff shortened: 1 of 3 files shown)


flihp:
Rebased on latest master.

PR #142, opened 2016-03-16
test: Integrate cmocka into autotools build, add basic test case for …

### LINKED ISSUE #18: test automation
Automake has support for running test scripts as part of the 'make check' target. Seems like the right place to test the TPM2.0-TSS code. This 'issue' covers the basic setup and maybe an initial test case. Each test migrated / created should have it's own issue for tracking.

flihp:
…device TCTI.

The current test layout I'm trying is a module (.c & .h file) in the
base test/ directory that is a collection of test functions for a
specific module of code from the tpm2.0-tss project. This test module is
named after the module it tests and the suffix '_test' is appended.
There should be NO dependency on the test framework in this module.

This test module is accompanied by a driver that sets up the test
framework with the tests from the module. This driver is given the name
of the module that it tests.

This resolves #18 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,8 @@ test/tpmclient/tpmclient
 .klocwork/
 ipch/
 test/tpmtest/tpmtest
+test-driver
+test-suite.log
+test/tcti_device
+test/*.log
+test/*.trs
--- a/INSTALL
+++ b/INSTALL
@@ -6,6 +6,7 @@ GNU Libtool
 C compiler
 C++ compiler
 C Library Development Libraries and Header Files (for pthreads headers)
+cmocka unit test framework
 
 Currently the tpm2.0-tss is only distributed via GitHub as we have not yet
 produced an official source release. To obtain the tpm2.0-tss sources you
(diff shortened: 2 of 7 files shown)


wcarthur1:
Philip, I looked at this.  Doesn't look like it's actually testing anything yet.  I couldn't find anything in the build output that looked like a test ran.

But looks like a good first inroad to creating a test infrastructure.

flihp:
Test cases are executed when 'make check' is run. I'll add a note about this in the INSTALL file.

As to whether or not the initial test I've added actually tests anything, that's a matter of what you think a test is. It's not testing anything at the scope that the current tpmclient / tpmtest do but that's intentional. The single test in this commit tests whether or not the device TCTI initialization function returns the proper size when the context parameter is NULL. If the call returns anything but the size of the TCTI structure the test case will fail. Testing such a small / contained piece of functionality is what unit tests are all about.

wcarthur1:
Ok, I ran "make check", and it worked.  I had no issue with the test case you created, it was just that I wasn't seeing any test results.

PR #148, opened 2016-03-21
Fixed bugs:  no return values for TCTI and SAPI finalize calls.  Fixed

wcarthur1:
all related code.


flihp:
howdy @wcarthur1, saw this get assigned to me for a second so I took a look at your branch. Looks good. The spec is clear on this so :+1: from me.

PR #151, opened 2016-03-21
Tcti device printf

flihp:
This is a significant departure from the way we currently handle the debugging functions in the TCTIs. The way we've done this up till now was to have a global function pointer set at compile time. This was typically controlled by a pre-processor macro. The global data this approach causes a litany of problems.

The refactoring I've been doing has all been leading up to decoupling the TCTIs from the application code (removing this shared global data) while preserving the current debug messages. This PR is an experiment in allowing the application consuming the TCTI to provide a callback function and some data that the TCTI will invoke in place of the global debug function pointer. This PR only addresses the issue in the device TCTI since it's a more simple case. The socket TCTI comes next.

Very interested in feedback not just on the implementation but on the general approach (user supplied callback). This was the most flexible thing I could think up and it's a pretty common pattern (event driven etc) but that's not to say that I've anticipated every possible contingency and it may be that this approach has some flaws that I haven't anticipated.

### DIFF
--- a/tcti/tcti_device.c
+++ b/tcti/tcti_device.c
@@ -39,6 +39,7 @@
 #include "debug.h"
 #include "commonchecks.h"
 #include <tcti/tcti_device.h>
+#include "logging.h"
 
 #ifdef  _WIN32
 #define ssize_t int
@@ -48,7 +49,7 @@
 
 #define HOSTNAME_LENGTH 200
 
-int (*tpmLocalTpmPrintf)( printf_type type, const char *format, ...) = DebugPrintf;
+const char *deviceTctiName = "device TCTI";
 
 TSS2_RC LocalTpmSendTpmCommand(
     TSS2_TCTI_CONTEXT *tctiContext,       /* in */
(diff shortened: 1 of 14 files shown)


tcti/tcti_device.c
```diff
@@ -251,8 +250,6 @@ TSS2_RC InitDeviceTcti (
     }
     else
     {
-        (*tpmLocalTpmPrintf)(NO_PREFIX, "Initializing %s Interface\n", interfaceName );
```
wcarthur1: Seems like you could init callback before this and use it instead of the tpmLocalTpmPrintf pointer.

flihp: There's a call to TCTI_LOG after the data is populated which is effectively the same (see line 270). The only thing that happens between the call that's removed and the call that's added are a few assignments. You're concerned that something will go wrong between the two locations and you won't get the debug message?

wcarthur1:
No, I was thinking for consistency's sake, use the TCTI_LOG everywhere.  I didn't see the later use of TCTI_LOG to output the initializing message.  In that case, do you really need the first one that uses the function pointer?

flihp:
I agree. But unless I'm missing something I think you're getting confused reading the diff. Your comment is on a line of code that this PR removes. This function pointer is removed completely by this PR.

wcarthur1:
My mistake...looked right at it and didn't see it.

PR #152, opened 2016-03-22
Bug fix (no return value for SAPI and TCTI finalize function).

wcarthur1:


flihp:
Looks great. Tests pass too. Merging ...

PR #153, opened 2016-03-23
Fixed bugs in AddSession and DeleteSession (these never worked right to

wcarthur1:
begin with).

Because AddSession never worked right (the list always had only the latest
created session), fixing these bugs cause a failure in the simple HMAC session
test (HMAC computed incorrectly because GetSessionStruct found the wrong
session (multiple entries with same session handle).  This required fixes
in CmdRspAuthsTests to delete previously created sessions.

### DIFF
--- a/test/common/sample/StartAuthSession.c
+++ b/test/common/sample/StartAuthSession.c
@@ -32,9 +32,11 @@
 
 #define SESSIONS_ARRAY_COUNT MAX_NUM_SESSIONS+1
 
-typedef struct {
+typedef struct _SESSION_LIST_ENTRY *SESSION_LIST_ENTRY_PTR;
+
+typedef struct _SESSION_LIST_ENTRY{
     SESSION session;
-    void *nextEntry;
+    SESSION_LIST_ENTRY_PTR nextEntry;
 } SESSION_LIST_ENTRY;
 
 SESSION_LIST_ENTRY *sessionsList = 0;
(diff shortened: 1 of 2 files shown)


flihp:
Reviewing this one now. Any interest in working up a unit test(s) for this functionality? These sort of tests are a great way to ensure that things work as expected and an even better way to catch regressions.

wcarthur1:
Philip,

This gets tested by my tests for virtualized capabilities (another PR coming down the pike).

Probably a good thing to do, but I don’t think I can spare the time to write very finely granular unit tests for this.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
Another point on this:  this is sample code for applications, not SAPI or RM code.  So I don’t think it requires the rigor of unit tests.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

From: Arthur, Will C
Sent: Thursday, March 24, 2016 1:41 PM
Subject: RE: [TPM2.0-TSS] Fixed bugs in AddSession and DeleteSession (these never worked right to (#153)

Philip,

This gets tested by my tests for virtualized capabilities (another PR coming down the pike).

Probably a good thing to do, but I don’t think I can spare the time to write very finely granular unit tests for this.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
Just read through this. Didn't realize that this code is part of the existing test / example suite. So if I'm reading this right it's a bug fix in what is effectively an integration test that creates multiple sessions with the RM and exercises them in some way.

There is a unit test hiding in there though. I think this would look something like a module that directly invokes whatever functions the resource manager has to track the current sessions / connections from clients. The goal here wouldn't be to create a client process and talk to the RM through the socket interface, but to directly exercise the internal machinery of the RM independent of the TCTI layer and whatever other layers are between the RM and clients.

Just practicing getting my brain to think in unit tests. "Too much work" is a valid answer. Merging this in a second.

wcarthur1:
Below…

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

PR #155, opened 2016-03-23
Debugability enhancements to RPSOURCE_MANAGER_ENTRY structure

wcarthur1:
to make looking at the list's entries easier.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -2475,7 +2475,7 @@ const char *resDeviceTctiName = "device TCTI";
 const char *resSocketTctiName = "socket TCTI";
 TCTI_SOCKET_CONF simInterfaceConfig = {
     DEFAULT_HOSTNAME,
-    DEFAULT_RESMGR_TPM_PORT
+    DEFAULT_SIMULATOR_TPM_PORT
 };
 
 SOCKET simOtherSock;


resourcemgr/resourcemgr.c
```diff
@@ -245,7 +246,7 @@ typedef struct {
                                     //  flushed.
     UINT64 connectionId;            // Used to identify which connection owns the object,
                                     // sequence, or session.
-    void *nextEntry; // Next entry in the list; 0 to terminate list.                                
+    RESOURCE_MANAGER_ENTRY_PTR nextEntry; // Next entry in the list; 0 to terminate list.                             
 } RESOURCE_MANAGER_ENTRY;
```
flihp: From looking just at the structure and it's name it's not entirely clear what a RESOURCE_MANAGER_ENTRY is (to me at least). The structure holds a pile of data and it all seems related to various handles that the RM may be switching in / out so an 'entry' would be a handle then?

resourcemgr/resourcemgr.c
```diff
@@ -245,7 +246,7 @@ typedef struct {
...
 
 RESOURCE_MANAGER_ENTRY *entryList = 0;
```
flihp: On line 215 above you go through the trouble of creating a type for a pointer to a RESOURCE_MANAGER_ENTRY structure. Great that you've replaced the void* in the structure with this but you've got ~20 uses of RESOURCE_MANAGER_ENTRY* in resourcemgr.c. My take on this is that we should pick one approach (RESOURCE_MANAGER_ENTRY* or RESOURCE_MANAGER_ENTRY_PTR) and use it consistently. Mixing these arbitrarily hurts readability.

wcarthur1: Good point.  I can fix that.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

resourcemgr/resourcemgr.c
wcarthur1: RESOURCE_MANAGER_ENTRY is a data structure that keeps all the state data about an object, sequence, or session that the RM needs to track.

The RM maintains and uses a linked list of these structures.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

resourcemgr/resourcemgr.c
```diff
@@ -245,7 +246,7 @@ typedef struct {
                                     //  flushed.
     UINT64 connectionId;            // Used to identify which connection owns the object,
                                     // sequence, or session.
-    void *nextEntry; // Next entry in the list; 0 to terminate list.                                
+    RESOURCE_MANAGER_ENTRY_PTR nextEntry; // Next entry in the list; 0 to terminate list.
```
flihp: There's also a pile of whitespace tacked on to the end of this line. It may just be an artifact from the line you've replaced but since you're fixing the void* thing you might as well do the whitespace cleanup at the same time.

resourcemgr/resourcemgr.c
wcarthur1: I’m going to back off from agreeing with you.

The main reason I created the RESOURCE_MANAGER_ENTRY_PTR structure was for debuggability.  In debugging I was having to walk through the linked list of structures.  When it was a void pointer, I had to cast every instance of this to a RESOURCE_MANAGER_ENTRY_PTR * in order to look at the data.

Outside of the structure, I don’t really care whether RESOURCE_MANAGER_ENTRY * is used or RESOURCE_MANAGER_ENTRY_PTR.  They’re the same thing and there’s really no advantage or disadvantage to using either of them or mixing them.

Forcing a consistent use seems like excessive pickiness to me.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

From: Arthur, Will C
Sent: Thursday, March 24, 2016 2:40 PM
Subject: RE: [TPM2.0-TSS] Debugability enhancements to RPSOURCE_MANAGER_ENTRY structure (#155)

Good point.  I can fix that.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
Upon further looking at this, I was able to removed some (now) unneeded casts by changing all instances of "RESOURCE_MANAGER_ENTRY *" to "RESOURCE_MANAGER_ENTRY_PTR".  
Also took care of the spurious whitespace you noted.

PR #162, opened 2016-03-23
Fixed bug in TestRM (uninitialized sizes for TPM2B's passed to Create…

wcarthur1:
…PrimaryComplete).

Failures noticed when running under Linux.


flihp:
Your commit message says you noticed failures caused by this while testing on Linux. Does this mean it worked against the windows simulator?

flihp:
I ask because running the tpmclient on Linux has never got past the 'TestTctiApis' step for me.

flihp:
Merged. It looks like the sane thing to do even though I can't trigger the test on Linux.

Issue #133, opened 2016-03-07
Test needed for stressing connections to RM


wcarthur1:
Need a test for many connections to RM and simultaneous commands being sent from many different connections.

flihp:
resourcemgr isn't long for this world, won't fix

PR #172, opened 2016-03-31
Added code to clean up when connection between application and resour…

wcarthur1:
…ce manager dies.

This required adding a couple critical sections to the RM and platformcommand.c.

The current test case for this is in TestEvict() when we tear down the other
TCTI context, line #2526 of tpmclient.cpp.

Addition of the critical sections should also fix the case where multiple
connections are trying to access the TPM at the same time.   We don't yet have
a test case for that, which is issue #133.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -93,7 +93,7 @@ LIBRARY_LDFLAGS = -fPIC
 RESOURCEMGR_INC = -I$(srcdir)/include -I$(srcdir)/common \
     -I$(srcdir)/sysapi/include -I$(srcdir)/resourcemgr \
     -I$(srcdir)/test/tpmclient
-RESOURCEMGR_C = resourcemgr/resourcemgr.c
+RESOURCEMGR_C = resourcemgr/resourcemgr.c resourcemgr/criticalsection_linux.c
 
 TCTICOMMON_INC = -I$(srcdir)/include -I$(srcdir)/common \
     -I$(srcdir)/sysapi/include
--- a/resourcemgr/criticalsection.h
+++ b/resourcemgr/criticalsection.h
@@ -1,0 +1,51 @@
+//**********************************************************************;
+// Copyright (c) 2015, Intel Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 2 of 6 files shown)


flihp:
Wanted to summarize the discussion we had about the PR off-line here for completeness:

1. Use of SAPI_CLIENT ifdef in platformcommand makes separating TCTI code from the RM more difficult (something I've been working on for some time) and hurts readability in general. Consider moving this code "up a level" into the RM code that's calling platformcommand
1. Related to the previous point is my ask about the resource we're controlling access to. Thinking of this in the 'critical section' head space makes me look at it as though we're controlling access to a block of code. But after our chat today it looks more like we're controlling access to the TCTI (really the sockets within this TCTI) that sits between the RM and the TPM. What we're doing then is locking this TCTI to be sure that only one thread in the RM is using it at a time. I think this reinforces the point above: lock the resource where it lives and that's in the RM.

flihp:
There was a 3rd point too: the code in criticalsection.c is split between windows/linux mutex code and most of this requires:
```
ifdef __win32
  do something
elseif __linux
  do something else
else
  error unsupported platform
endif
```
Since the majority of the code in this module falls between ifdefs it may be cleaner if we just provide multiple implementations in the form of criticalsection_linux.c / criticalsection_windows.c with a single header. This will allow the different builds to just use the file for their respective platform and save the user from having to turn on / off their ifdef filters when reading the code.

wcarthur1:
Ok, Philip, I updated this with the changes we discussed yesterday.

tcti/platformcommand.c
```diff
@@ -62,7 +62,7 @@ TSS2_RC PlatformCommand(
     // Send the command
     iResult = send( TCTI_CONTEXT_INTEL->otherSock, sendbuf, 4, 0 );
     if (iResult == SOCKET_ERROR) {
-        (*printfFunction)(NO_PREFIX, "send failed with error: %d\n", WSAGetLastError() );
+        DebugPrintf(NO_PREFIX, "send failed with error: %d\n", WSAGetLastError() );
```
flihp: I think this change is an artifact from a rebase? I've got an outstanding PR that gets ride of this function pointer so maybe this snuck in when you were reviewing / testing that work?

flihp:
I've added one comment inline (see above) w/r to come code that I think is a rebase artifact. Also there's a lot of trailing whitespace in this new code. The rest of the changes look great!

tcti/platformcommand.c
wcarthur1: I did rebase, and got some weirdness. But didn’t rebase to your outstanding PR.

What do you want me to do with it?

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
I don’t see the trailing whitespaces.  How are you seeing those?

And…do we really need to be this picky about whitespace?

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
The output from 'git show' on my terminal shows trailing whitespace highlighted in red.

wcarthur1:
Ok, that helps.  Now I can see them.

But that still begs the question:  how much effort do we want to put into whitespace management?  It doesn’t seem worth the effort to me.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

wcarthur1:
Ok, I did all the changes you wanted, I think.

Can you merge this one in as soon as possible so that I can be working on the next one?  I’d rather not have to wait another day or so before you get back from Oregon.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

PR #180, opened 2016-04-01
Fix install spelling error

wcarthur1:


wcarthur1:
Oops--didn't base this correctly.

PR #183, opened 2016-04-04
Fixed uninitialized size fields in output simple TPM2B structures.

wcarthur1:


flihp:
Just out of curiosity: why is it size - 2?

wcarthur1:
TPM2B’s are a sized structure, UIN16 size followed by buffer of size # of bytes.

The size field is not the size of the whole structure, just the size of the buffer, hence sizeof (TPM2B_XXXX) – 2.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
Gotcha. So 2 is sizeof (uint16_t) to account for the structure containing an indication of it's own size. This is a lot like a TLV sequence. Makes sense to me.

wcarthur1:
yes

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
Cool. LGTM. Merging now.

PR #186, opened 2016-04-08
tpmtest: fix one more place needing initializing TPM2B size

gwei3:
I found we have new macro to do the TPM2B structure initialization. I would like to make this fix in first, then I will do the macro replacement later.


wcarthur1:
Jimmy, seems like you ought to fix all of these issues, like I did in my PR #183 and use the new macro.  Just port those changes to tpmtest.

gwei3:
Sure, I will do that soon.

gwei3:
close it and create a new PR.

PR #187, opened 2016-04-08
RM: allow Load to use persistent parent created by other connection

gwei3:
First, set persistent flag in evicted entry to make the persistent flag
functioning as designed.
Then in tpm2_Load case, allow using persistent parent created by other
connection.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -1309,7 +1309,8 @@ TSS2_RC ResourceMgrSendTpmCommand(
                 goto SendCommand;
             }
 
-            if( foundEntryPtr->connectionId != cmdConnectionId )
+            if( foundEntryPtr->status.persistent == 0 &&
+                foundEntryPtr->connectionId != cmdConnectionId )
             {
                 responseRval = TSS2_RESMGR_UNOWNED_HANDLE;
                 goto SendCommand;
@@ -1932,6 +1933,12 @@ TSS2_RC ResourceMgrReceiveTpmResponse(
                                 foundEntryPtr->hierarchy, cmdConnectionId );
                         if( rval != TSS2_RC_SUCCESS )
                             goto returnFromResourceMgrReceiveTpmResponse;
+
+                        // Find entry for the new persistent object and flag it as persistent
+                        rval = FindEntry( entryList, RMFIND_REAL_HANDLE, persistentHandle, &foundEntryPtr );
+                        if( rval != TSS2_RC_SUCCESS )
+                            goto returnFromResourceMgrReceiveTpmResponse;
+                        foundEntryPtr->status.persistent = 1;
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


gwei3:
Will, any comments on this?

resourcemgr/resourcemgr.c
```diff
@@ -1309,7 +1309,8 @@ TSS2_RC ResourceMgrSendTpmCommand(
                 goto SendCommand;
             }
 
-            if( foundEntryPtr->connectionId != cmdConnectionId )
+            if( foundEntryPtr->status.persistent == 0 &&
+                foundEntryPtr->connectionId != cmdConnectionId )
```
wcarthur1: I do not agree with this change.  There is nothing in the RM spec that allows sharing of persistent objects between connections.
What's the problem you're trying to solve here?

resourcemgr/resourcemgr.c
```diff
@@ -1932,6 +1933,12 @@ TSS2_RC ResourceMgrReceiveTpmResponse(
                                 foundEntryPtr->hierarchy, cmdConnectionId );
                         if( rval != TSS2_RC_SUCCESS )
                             goto returnFromResourceMgrReceiveTpmResponse;
+
+                        // Find entry for the new persistent object and flag it as persistent
+                        rval = FindEntry( entryList, RMFIND_REAL_HANDLE, persistentHandle, &foundEntryPtr );
+                        if( rval != TSS2_RC_SUCCESS )
+                            goto returnFromResourceMgrReceiveTpmResponse;
+                        foundEntryPtr->status.persistent = 1;
```
wcarthur1: I agree that this code change or something like it is needed.  I do need to set the persistent bit here, since I use it elsewhere.  

But...what I'd rather do is just eliminate the persistent bit and always look at the handle type when trying to decide if a handle is persistent or not..

gwei3: Yes, I agree that is another to just look at the handle type to judge for whether it is a persistent handle or not. You can decide which way to go.

resourcemgr/resourcemgr.c
gwei3: If the persistent objects can't be shared between different connections to RM, then users will be very restricted to get the benefit of persistency of object since all access to the persistent objects have to be finished in a single connection. How to handle it if people want to create a persistent object in a connection and use it for later connections? Using persistent objet instead of savecontext/loadcontext for a object can avoid saving context files between connections and also make the execution faster.

wcarthur1: agree, but this needs to be changed in the RM spec first.  If you would like to attend one of the TSS WG meetings and present your case for this, we could see what the WG says.

resourcemgr/resourcemgr.c
wcarthur1: I say we look at the handle type.

gwei3: ok. Actually I tried the way to looking at the handle type before I tried to using persistent bit. I choosed the persistent bit way since it make my change looks smaller and easy to understand. I can make new PR to remove the persistant bit.

resourcemgr/resourcemgr.c
gwei3: I can try.

gwei3:
Finished as discussed with will.

PR #192, opened 2016-04-12
tpmtest: fix simple TPM2B initialization issues

gwei3:
fixed one more bug, and updated to use INIT_SIMPLE_TPM2B_SIZE for the initialization.

### DIFF
--- a/test/common/sample/DecryptEncrypt.c
+++ b/test/common/sample/DecryptEncrypt.c
@@ -56,7 +56,7 @@ TSS2_RC GenerateSessionEncryptDecryptKey( SESSION *session, TPM2B_MAX_BUFFER *cf
     
     if( rval == TSS2_RC_SUCCESS )
     {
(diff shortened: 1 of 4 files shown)


wcarthur1:
Jimmy, you also need to do this in the test\common\sample directory.

gwei3:
updated the sample code.

wcarthur1:
Jimmy, good job on sample code:  you caught a couple cases I didn't update to use the macro.
On the other hand:  I looked at tpmtest and you missed a lot of the cases that need to add the initialization of the TPM2B size field.  some examples:  look at lines 633, 639, and 652 of tpmtest.cpp and compare to what I did in tpmclient.cpp.  You'll need to go through the whole of tpmtest.cpp and make sure you did this everywhere I did it in tpmclient.cpp.  Using a good side by side smart compare tool like Araxis Merge or Beyond Compare will help you in this.

gwei3:
Finished.

PR #193, opened 2016-04-14
Fixed bug with DEBUG_SOCKETS

wcarthur1:
When a command code is not valid, the code could cause a seg fault by looking for a string outside the correct range.  Found this in the first test in TestRM when running on Linux.

### DIFF
--- a/common/debug.c
+++ b/common/debug.c
@@ -212,7 +212,17 @@ const char *commandCodeStrings[] =
     "TPM2_PolicyNvWritten" // 18f
 };
 
+char undefinedCommandString[10] = "";
+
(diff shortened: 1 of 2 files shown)


flihp:
Simple enough. What do you think about moving the bounds checking to the strTpmCommandCode function? This would save callers from having to make the check themselves. I'm guessing there's other code that calls this function and that doesn't do this bounds checking. Probably best to solve this for every caller in the function directly?

wcarthur1:
Good catch, Philip.  I updated the code to push the bounds checking into strTpmCommandCode.

PR #195, opened 2016-04-17
Tcti no sapi client

flihp:
The interesting bit here is that this allows us to link the resourcemgr to the TCTIs dynamically after this PR.


wcarthur1:
Doesn't build with -Wall -Werror--unused variable on line #.  I can merge my local branch with a change for this to master, or you can do it.  Let me know.

flihp:
That bit of cleanup is easy enough. I'll sort it out and push directly.

I've also created an issue to get -Wall -Werror set in the build by default. Should help minimize the back and forth on PRs caused by differences in build / test environments.

PR #200, opened 2016-04-22
RM: make physical TPM case working again

### LINKED ISSUE #199: Resource Manager hang
I'm getting this second hand from a few different sources so this may not be completely accurate. Bear with me.

It looks like 3c0debc broke something when the resource manager is connected to a proper TPM instead of the simulator. That's my commit so I'm digging into this in the next few days.

gwei3:
resolve #199.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -2531,7 +2531,11 @@ UINT8 OtherCmdServer( SERVER_STRUCT *serverStruct )
         command = CHANGE_ENDIAN_DWORD( command );
 
         if( !simulator )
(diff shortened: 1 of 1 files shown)


snits:
Tested on my system here and both resourcemgr runs, and tpmtest now runs successfully.

gwei3:
thanks for the quick verification��

wcarthur1:
I'm going to ask Philip to look at this.
Thanks for looking at this.

flihp:
This is a great patch @gwei3. Thanks for taking care of this. I added some additional information to your commit message but that's it.

PR #203, opened 2016-04-22
build: Add -Werror -Wall to default compiler flags.

flihp:
I found that we were using CPPFLAGS where we should have been using
CXXFLAGS in the process of working this out. This has been fixed here as
part of pulling in these new compiler flags.


flihp:
this doesn't work the way I thought it did :(

PR #208, opened 2016-05-02
Got debug messages working in RM.

wcarthur1:
Added command line control of debug messages in RM

Cleaned up uses of RM prefix.  This prefix is ONLY to be used for
TPM commands/responses generated by the RM itself.

Added debug level for commands sent from application.

Enabled debug level control over TPM commands issued during RM initialization.

Added RM prefix to debug messages for TPM commands issued during RM initialization.

Added control over RM prefix to both local and socket TCTI layers.


flihp:
Doesn't build w/o CFLAGS="-DDEBUG "

wcarthur1:
Fixed the build issue and some other issues.

PR #209, opened 2016-05-06
Make data for the libraries queryable through pkg-config.

flihp:
This requires building a *.pc file for each library.


flihp:
merging my own patch, 10 days in the queue is long enough

PR #216, opened 2016-05-25
Update marshal_simple_tpm2b.c

erezgeva:
Cast after subtraction of pointers.
Android GCC compiler complain about.


erezgeva:
Cast after subtraction of pointers.
Android GCC compiler complain about.

flihp:
merged

PR #218, opened 2016-05-27
Rename libraries for clarity & readability

flihp:
This patch renames:
libtss2 -> libsapi
libtctidevice -> libtcti-device
libtctisocket -> libtcti-socket

The first is for clarity: The name 'tss2' covers the whole of the software stack. The library however has only the SAPI code in it. Naming it libsapi makes a lot more sense.
The second two are for readability: libtcti-device.so is more readable than libtctidevice.so. The convention on Linux is to use hyphens to separate words in library names AFAIK.

This patch doesn't change any of the header names since the spec uses '_'s instead of hyphens. It's probably best to follow this convention for consistency with the other headers.

This commit should be coordinated with tpm2.0-tools to minimize breakage.


flihp:
@gwei3 This one too plz?

PR #223, opened 2016-06-06
RM: fix the issue - RM exit after a period of time

### LINKED ISSUE #222: Resourcemgr stop functioning after served a huge number of connections
Below is the failure log:
_Resource Mgr failed to create OTHER command server thread, error #11.  Exiting...
Accept socket:  0x9
Resource Mgr failed to create OTHER command server thread, error #11.  Exiting...
Accept socket:  0x7
Resource Manager Other CMD Server accepted client
Accept socket:  0x9
Resource Mgr failed to create OTHER command server thread, error #11.  Exiting...
Resource Mgr failed to create OTHER command server thread, error #11.  Exiting..._
it exists for both alpha & beta tags.

gwei3:
Add call to pthread_detach to make thread resource releasable.
Fix #222

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -2707,6 +2707,11 @@ UINT32 WINAPI SockServer( LPVOID servStruct )
             printf( "Resource Mgr failed to create OTHER command server thread, error #%d.  Exiting...\n", rval );
             continue;
         }
+        rval = pthread_detach( cmdServerStruct->threadHandle );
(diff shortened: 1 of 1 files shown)


flihp:
Yikes! Reclaiming resources is pretty important. I'm traveling for the next few days but I'll have a chance to build / test this this afternoon.

Classes of bugs like this make me think we should make running the tests under valgrind. memcheck probably would have caught this. Makes me wonder what else we might uncover.

PR #224, opened 2016-06-06
RM: Fix deadloop in physical tpm case

### LINKED ISSUE #221: resourcemgr uses 100% CPU on Ubuntu 16.04 x86_64
I cloned the current TPM2.0-TSS master and built it on an Ubuntu 16.04 x86_64 desktop system. resourcemgr is started as a systemd service and doesn't use any CPU resources. After successfully executing any of the tpm2-tools (e.g. tpm2_listpcrs) the resourcemgr is left behind using up 100% of the CPU cycles. Repetitive tpm2-tool calls will cause the resourcemgr to use all available processor cores.

This behaviour does not occur with the official Ubuntu libtss2 package which apparently is based on a much older version of the TPM2.0-TSS software.

Best regards

Andreas

gwei3:
For other cmd server thread, in physical tpm case, just skipped all cmd and give a successful response. this is not enough. Need to handle TPM_SESSION_END cmd to exit the thread, instead of continue to let the thread in a deadloop state.
Fix #221.

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -2550,6 +2550,9 @@ UINT8 OtherCmdServer( SERVER_STRUCT *serverStruct )
         {
             rval = CHANGE_ENDIAN_DWORD( TSS2_RC_SUCCESS );
             rmSendBytes( serverStruct->connectSock, (unsigned char *)&rval, 4 );
+            if( command == TPM_SESSION_END )
+                break;
+
             continue;
         }
 


flihp:
Will try to get a review done tonight.

flihp:
Pushed but I didn't get a chance to test since I've been on the road. It's only a 2 line patch though and the bug is pretty serious so I'm pushing it a bit blind. Will test over the weekend just for good measure.

Issue #207, opened 2016-04-28
remove final SAPI_CLIENT usage from getcommands


flihp:
There's one final useage of the SAPI_CLIENT flag in the common getcommands file. This doesn't actually look to be used by the SAPI code so it's likely something that can just be removed from the SAPI library build.

PR #228, opened 2016-06-21
No sapi client

flihp:
This patch set removes the last remnants of the SAPI_CLIENT macro.

### DIFF
--- a/common/getcommands.c
+++ b/common/getcommands.c
@@ -33,7 +33,6 @@
 #ifdef SAPI_CLIENT
 TSS2_RC level = TSS2_ERROR_LEVEL( TSS2_APP_ERROR_LEVEL );
 #else
-extern void *(*rmMalloc)(size_t size);
 TSS2_RC level = TSS2_ERROR_LEVEL( TSS2_RESMGR_ERROR_LEVEL );
 #endif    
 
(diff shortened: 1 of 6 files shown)


flihp:
this resolves issue #207

common/getcommands.c
```diff
@@ -47,12 +40,6 @@ TSS2_RC GetCommands( TSS2_SYS_CONTEXT *resMgrSysContext, TPML_CCA **supportedCom
     TSS2_RC rval = TSS2_RC_SUCCESS;
     UINT32 i;
     
-#ifdef SAPI_CLIENT
-    void *(*gcMalloc)(size_t size) = malloc;
-#else    
-    void *(*gcMalloc)(size_t size) = rmMalloc;
```
wcarthur1: Problem here is that this code is used by RM, too, and we needed to return a RM-specific error level.  Unless you accounted for that in the RM code, this error level won't come out right.

wcarthur1:
RM call to getCommands is going to return wrong error level now.

flihp:
Thanks for the review. I've been parsing the spec to figure out how the error codes work these past few days. Will have to revisit these now that I sorta "get it".

On another note: I'm trying to sort out a review process now that you're no longer with Intel. Can I count on you for future reviews or should I classify you as an occasional drop in?

flihp:
Fixed up the error level thing pointed out by @wcarthur1

resourcemgr/resourcemgr.c
```diff
@@ -2939,6 +2939,7 @@ TSS2_RC InitResourceMgr( int debugLevel)
     rval = GetCommands( resMgrSysContext, &supportedCommands );
     if( rval != TPM_RC_SUCCESS )
     {
+        rval += TSS2_RESMGR_ERROR_LEVEL;
```
gwei3: We should not use += to change the error level, it will cause problem if there already be a non-zero error level. I suggest using: 
rval = (rval & (~TSS2_ERROR_LEVEL_MASK)) + TSS2_RESMGR_ERROR_LEVEL;;

flihp: Agree, but I'm realizing that this will also add the wrong layer indicator for errors returned from further down the stack. Honestly I'm tempted to just move the 'getcommands' function into the resourcemgr directly since it's not used anywhere else in the code. That will make this much more simple. Agree / disagree?

gwei3: That is also a good idea. I agree.

flihp:
Updated this PR. @gwei3 can you please take a look?

PR #234, opened 2016-07-08
Tpmtest fix

gwei3:
make tpmtest able to run better on Infineon TPM in real production platforms.


flihp:
Thanks for hacking this up @gwei3. I'll get a review done in the next day or so.

flihp:
I was able to reproduce the error code 0x8006 on my Minnowboard running tpmtest #34. The test that fails is the first. After applying this patch set that first test passes but a later test fails. This time it's the 4th test but w/o reading the code I've no clue what that test actually does. I'll create a separate issue for that failure. Either way this fix at least solves the problem at hand.

Issue #235, opened 2016-07-09
merging of #218 breaks the build


flihp:
#218 didn't account for a number of file path fixups that were necessary. The pkg-config files are needing fixups as well.

flihp:
resolved in #236

PR #236, opened 2016-07-09
Fixup bugs introduced by 3b7e71220c9 by fixing up include directives …

flihp:
…to use sapi/ instead of tss2/


flihp:
this resolves #235

PR #248, opened 2016-07-23
Common prepare prologue deref

### LINKED ISSUE #247: CommonPreparePrologue has a NULL deref bug
https://github.com/01org/TPM2.0-TSS/blob/master/sysapi/sysapi_util/CommandUtil.c#L108

This one's kinda funny: we test to see if the parameter is NULL ... and then we immediately dereference it :) It's not immediately obvious though since we access the NULL pointer through a macro with no parameters and the name of the variable hard coded into it.

Probably best to add a test case for this function while we're fixing this.

flihp:
This resolves #247

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ test/tpmtest/tpmtest
 test-driver
 test-suite.log
 test/getcommands-malloc-mock_unit
+test/CommonPreparePrologue_unit
 test/tcti_device
 test/*.log
 test/*.trs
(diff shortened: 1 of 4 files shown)


flihp:
Thanks for the data Will. This work is very much 'macro' in it's focus. I'm going through function by function in the serialization logic and building unit tests for each. I think your point about the larger problem at hand is a good one. That said I'm not yet looking across the larger call chain. For now I'm working on individual functions and building up macro / unit tests for each. This helps me understand how each is supposed to work, how it actually works and helps build up a body of tests to be sure that they do what they say they do.

In this particular case there was a NULL check on the context that then triggered an immediate dereference. Since this function did the check IMHO it should do it right. After I've got some good coverage for these I'll start leveling my work up a bit to figure out where these sorts of checks actually belong (as far up the call chain as possible) and then go cleaning out the unnecessary checks.

Issue #261, opened 2016-08-09
libcmocka dependency should be conditional


flihp:
This library isn't critical to the proper operation of the TSS build. It's only required for testing. Thus we should detect the presence of libcmocka and enable the unit tests if it's available. If not the unit tests should be disabled.

flihp:
Probably a good idea to provide a configure flag (--with-cmocka) to force a configure time failure if cmocka isn't installed too

flihp:
This is a specific request from people who are annoyed by having to install the unit testing library when they're not running any of the tests. I'm not sure how I feel about this but "you've gotta give the people what they want" or something.

PR #262, opened 2016-08-10
configure.ac: Allow for the unit tests to be enabled selectively.

flihp:
This makes the dependency on libcmocka optional.


flihp:
this resolves #261

PR #273, opened 2016-09-19
Remove semicolons from end of TPMA_* macro definitions.

### LINKED ISSUE #272: semicolons at end of TPMA_* constant definitions
See examples here:
https://github.com/01org/TPM2.0-TSS/blob/master/include/sapi/tss2_tpm2_types.h#L505
https://github.com/01org/TPM2.0-TSS/blob/master/include/sapi/tss2_tpm2_types.h#L626

The semicolon prevents us from using these macros like so:
```
TPMA_ALGORITHM algorithm_attrs = TPMA_ALGORITHM_ASYMMETRIC + TPMA_ALGORITHM_SYMMETRIC;
```

The example above is a bad one since it makes no sense, but it should be possible so that we can set more than one bit at a time.

flihp:
The semicolon prevents us from using these macros like so:

TPMA_ALGORITHM algorithm_attrs = \
    TPMA_ALGORITHM_ASYMMETRIC + TPMA_ALGORITHM_SYMMETRIC;

This resolves #272


wcarthur1:
Looks like a bug in the autogeneration code.  Should be fixed there.

PR #277, opened 2016-09-21
Remove Windows build files.

flihp:
The Windows build needs lots of love to get dll's building and to get an
installer functional. We don't have anyone to do or support the work
though so best to remove these and just add them back if / when someone
motivated to do the task comes along. We can't have these as part of a
release unless we can support whoever may use them.

This resolves #273 


wcarthur1:
I'm going to really hate this.  Even though windows build doesn't work for real applications/normal use, it is very useful for code development.  I would prefer that it just be documented as such (and I think it already is, i.e. windows build only runs against simulator).

David-Challener:
I think this is a bad idea.  Why is it a bug? Why remove existing functionality?

mwiseman32:
Currently, developers need to build this on both Windows and Linux which is an unnecessary step when the primary focus of this project is Linux.

flihp:
Specifically answering the questions from @David-Challener above (hi @David-Challener !): Like @wcarthur1 says, these files are part of one developers debugging workflow. They're not a feature. Keeping a set of debug files in the source tree for each developer just won't scale and each file is a liability for the maintainer.

David-Challener:
But do we not want to have this available on both LInux and Windows?  It
certainly is nice to run the same code from TrouSerS on both.

flihp:
@David-Challener That would be nice but I think you're making my point for me here: These aren't a VS build files that we can support since they just statically like everything into the test applications. These files are part of the debug environment that @wcarthur1 left behind. Having these files in the source tree is confusing people: it gives off the impression that we support Windows when in fact we don't.

I've made a pile of 'issues' for the work necessary for us to support a real Windows build but no one has taken on the work. If you've got some cycles to spare I'd gladly take PRs for those issues. They're organized under [milestone #9](https://github.com/01org/TPM2.0-TSS/milestone/9) if you're interested in helping out.

David-Challener:
THen why don't you simply document what they are, and leave them available
for the next person that has cycles necessary to update them?  Removing
data is seldom a good idea.

flihp:
I appreciate the advice @David-Challener but VCS mostly obsoletes it. A motivated party like yourself will only need to go back into git history and revive these files. Or better yet you could create a topic branch, pull them over and fix them up! Really though my advice would be to start from scratch since last I checked these were for VS 2010 which is pretty old and requires a paid license. Microsoft has iterated several times on the express and now community editions of VS which would be much more appropriate for the audience we're courting.

I'll make a link to the commit that removes these files for you in case you decide to pitch in. That way you can just revert it if you want to start from the existing files.

wcarthur1:
BTW, I just built under Windows and VS2015 last week.  And I plan to continue to do so.  I will keep this build environment going for my own use, but unfortunately, no one else will be able to use it.

I've said it before...developing TPM code running against the simulator and under VS2015 is the quickest and easiest path for TPM code developers.  I currently have a project where I have to port some Linux code from TPM 1.2 to 2.0.  I plan to get the 1.2 code to build under Windows, then add the 2.0 support while developing under Windows then port to Linux.  The debugging environment in Linux is vastly inferior to that under VS/Windows.

The documentation has always made it clear that the Windows support only works with the simulator and is for prototyping purposes.

I sympathize with the purist approach, but at heart I'm a pragmatist and the easiest way to get a job done is what I'm going to do.

So how did this get resolved?  Is the VS support now removed?  Github says that Philip closed this and deleted the flihp:rm-win-build branch.  I don't know if that means the VS files are gone or still there.

Issue #243, opened 2016-07-19
udev rules to deprivilege resourcemgr


flihp:
By default /dev/tpm0 is owned by root. This requires that the resourcemgr run as root as well which is a terrible idea. We need an example udev rule file to cause /dev/tpm0 to be owned by an unprivileged user to allow the resourcemgr to run as the same. This is probably a subtask from #17

flihp:
Again resolved months ago.

PR #279, opened 2016-09-22
Add example systemd service file for resourcemgr and udev rules for t…

### LINKED ISSUE #244: Example systemd unit for resourcemgr
Since a lot of this stuff is distro specific such an example should go into a $(src)/contrib directory. It must be installed by hand (not by ''make install') and should be described in the documentation as a reference / example.

flihp:
…pm device node.

These work as-is on Debian provided there's a system user named 'tss'
already and the PREFIX is '/usr/local'. Integrating these files into
the build shouldn't be too hard but the only systemd module for autoconf
that I can find is out of tree and not in the archive either.

This resolves #244 and #243.

### DIFF
--- a/contrib/resourcemgr.service
+++ b/contrib/resourcemgr.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=TPM2 resource manager & access broker
+Documentation=http://www.github.com/01org/TPM2.0-TSS
+
+[Service]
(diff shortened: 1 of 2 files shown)


flihp:
57b95db8465006da604d159bb71436398c58ab7a

PR #280, opened 2016-09-23
tcti_socket: Move simulator initialization code into the context init…

flihp:
… function.

This is something that every application has to do to be sure that the
simulator has been initialized. Running these commands multiple time has
no negative effect (AFAIK). Might as well just run this each time we
initialize a connection to the simulator. Save everyone else the
headache.


flihp:
resolved in 40e926ded61523c1a6c1011939c918a7b827f089

PR #287, opened 2016-10-12
Fix #286

### LINKED ISSUE #286: resource manager eats up more that 100% CPU if client disconnects without freeing resources
Steps to reproduce:
 - run a lengthy operation using TPM tools (e.g. `tpm2_listpcrs`);
 - CTRL-c the application (timing is important as this should be issued after connection to the TPM is opened and before it is closed by the app);
 - observe resource manager eating up more than 100% and sockets staying in CLOSE_WAIT state forever;

```
root@ubuntu:~# ps -p 88331 -o %cpu,%mem,cmd
%CPU %MEM CMD
 176  0.0 resourcemgr -sim
```
and 
```
root@ubuntu:~# lsof -Pani -p 88331
COMMAND     PID    USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
resourcem 88331 dguerri    3u  IPv4 475745      0t0  TCP 127.0.0.1:52162->127.0.0.1:2322 (ESTABLISHED)
resourcem 88331 dguerri    4u  IPv4 475746      0t0  TCP 127.0.0.1:57148->127.0.0.1:2321 (ESTABLISHED)
resourcem 88331 dguerri    5u  IPv4 475747      0t0  TCP 127.0.0.1:2324 (LISTEN)
resourcem 88331 dguerri    6u  IPv4 475748      0t0  TCP 127.0.0.1:2323 (LISTEN)
resourcem 88331 dguerri    9u  IPv4 476814      0t0  TCP 127.0.0.1:2324->127.0.0.1:35818 (CLOSE_WAIT)
```

As it turns out, the resource manager doesn't close the socket when it receives 0 from recv.
From `man recv`:

```
Return Value

These calls return the number of bytes received, or -1 if an error occurred. 
The return value will be 0 when the peer has performed an orderly shutdown.
```

Instead of detecting that the client has closed the connection, `resoucemgr` keeps calling recv in an infinite for loop (see [common/sockets.cpp](https://github.com/01org/TPM2.0-TSS/blob/master/common/sockets.cpp#L22-L27))

dguerri:
resource manager eats up more that 100% CPU if client disconnects without freeing resources

As it turns out, the resource manager doesn't close the socket when it receives 0 from recv.
From man recv:

  Return Value

  These calls return the number of bytes received, or -1 if an error occurred.
  The return value will be 0 when the peer has performed an orderly shutdown.

Instead of detecting that the client has closed the connection, resoucemgr keeps calling recv in an infinite for loop (see common/sockets.cpp)

### DIFF
--- a/common/sockets.cpp
+++ b/common/sockets.cpp
@@ -22,7 +22,7 @@ TSS2_RC recvBytes( SOCKET tpmSock, unsigned char *data, int len )
     for( bytesRead = 0, length = len; bytesRead != len; length -= iResult, bytesRead += iResult )
     {
         iResult = recv( tpmSock, (char *)&( data[bytesRead] ), length, 0);
-        if (iResult == SOCKET_ERROR)
+        if ((iResult == SOCKET_ERROR) || (!iResult))
             return TSS2_TCTI_RC_IO_ERROR;
     }
 


flihp:
Great patch. Thanks for tracking this down.

For those following along at home who want to reproduce this issue you can just comment out the following lines and rebuild the socket TCTI:

https://github.com/01org/TPM2.0-TSS/blob/master/tcti/tcti_socket.cpp#L256
https://github.com/01org/TPM2.0-TSS/blob/master/tcti/tcti_socket.cpp#L257

This will cause any application using it to finalize the TCTI instance without letting the resourcemgr or simulator know that the client is disconnecting. This will cause the condition described in #286.

PR #292, opened 2016-10-30
Ensure the command size from the client does not exceed the cmdBuffer…

### LINKED ISSUE #291: resourcemgr buffer overflow
The resourcemgr code has a pretty nasty buffer overflow.

The issue boils down to relying on untrusted input. The TpmCmdServer function https://github.com/01org/TPM2.0-TSS/blob/master/resourcemgr/resourcemgr.c#L2284 where the resorucemgr receives TPM command buffers from client connections. There's a bit of a protocol between libtcti_socket and the resourcemgr TpmCommandServer where the client must send:
1. a command identifier (either TPM_SESSION_END or MS_SIM_TPM_SEND_COMMAND)
2. the locality for the command
3. the size of the command buffer
4. finally, the TPM command buffer

The buffer overflow is the result of trusting the size provided by the client in step 3 above. The resourcemgr recv's the size as a uint32 (https://github.com/01org/TPM2.0-TSS/blob/master/resourcemgr/resourcemgr.c#L2366), then blindly reads however many bytes the user specified as the TPM command buffer (https://github.com/01org/TPM2.0-TSS/blob/master/resourcemgr/resourcemgr.c#L2377). The destination buffer for the TPM command in the resourcemgr is a fixed size determined by the TPM (through a call to GetCapability). This is a classic OWASP "top 10" data validation bug (https://www.owasp.org/index.php/Data_Validation).

To exploit this, a client can provide an unusually large value for the TPM command size in step 3 of the protocol and then just dump garbage data into the socket. The resourcemgr will blindly read as much data as the client specifies. Eventually the destination buffer will be overrun and the resourcemgr program corrupted and crashed. No doubt other results may be obtained but simple corrupt / crash is the easiest for demonstration purposes.

I've written a client program to do just this. The output from the resourcemgr crashing looks like this:

```
socket created:  0x3
socket created:  0x4
Client connected to server on port:  2322
Client connected to server on port:  2321
Initializing Resource Manager
maxActiveSessions = 64
gapMaxValue = 65535
socket created:  0x5
bind to IP address:port:  127.0.0.1:2324
Other CMD server listening to socket:  0x5
socket created:  0x6
bind to IP address:port:  127.0.0.1:2323
TPM CMD server listening to socket:  0x6
tpmCmdServerStruct running
Starting SockServer (TPM CMD), socket: 0x6.
Starting SockServer (Other CMD), socket: 0x5.
Accept socket:  0x7
TpmCmdServer
selecting on other socket I think?
after select, before receive sendCmd
  received MS_SIM_TPM_SEND_COMMAND
  received locality: 0  received size: 0xffffffff
In rmRecvBytes, recv failed (socket: 0x7) with error: 14
In rmSendBytes, send failed (socket: 0xdeadbeef) with error: 9
In rmSendBytes, send failed (socket: 0xdeadbeef) with error: 9
In rmSendBytes, send failed (socket: 0xdeadbeef) with error: 9
Segmentation fault
```

My exploit program sends UINT32_MAX as the size and then just repeatedly sends a uint32 (0xdeadbeef) as the data / TPM command buffer. As you can see in the output above from the resourcemgr, the exploit ends up overwriting an internal structure where the resourcemgr keeps socket file descriptors. Eventually the resourcemgr tries to send data to the socket but the fd is now invalid (0xdeadbeef). Eventually the corruption causes the resorucemgr to be terminated.

Fixing this issue is trivial: add a bit of bounds checking to the resourcemgr to ensure that it doesn't read more data than the cmdBuffer can hold.

flihp:
… capacity.

This resolves a buffer overflow that is the result of trusting the
size of the TPM command provided by the client in the TpmCmdServer
thread. The resourcemgr recv's the size as a uint32, and then blindly
reads however many bytes the user specified into the cmdBuffer.
The cmdBuffer in the resourcemgr is a fixed size determined by the
TPM (through a call to GetCapability). This is a classic OWASP "top
10" data validation bug:
https://www.owasp.org/index.php/Data_Validation

We resolve this by comparing the size of the command provided by the
client to the maxCmdSize variable. At the time of initialization,
the resourcemgr queries the TPM for the maximum command buffer size
that it supports and sets maxCmdSize to this value. If the client
attempts to send a command buffer larger than this size, the
resourcemgr will now send a response buffer indicating that
the buffer available is insufficient:
TSS2_TCTI_RC_INSUFFICIENT_BUFFER.

This resolves #291 

Signed-off-by: Philip Tricca philip.b.tricca@intel.com

### DIFF
--- a/resourcemgr/resourcemgr.c
+++ b/resourcemgr/resourcemgr.c
@@ -2372,6 +2372,12 @@ UINT8 TpmCmdServer( SERVER_STRUCT *serverStruct )
             }
 
             numBytes = CHANGE_ENDIAN_DWORD( numBytes );
+            if( numBytes > maxCmdSize )
+            {
+                CreateErrorResponse( TSS2_TCTI_RC_INSUFFICIENT_BUFFER );
+                SendErrorResponse( serverStruct->connectSock );
+                continue;
+            }
 
             // Receive the TPM command bytes from calling application.
             rval = rmRecvBytes( serverStruct->connectSock, (unsigned char *)cmdBuffer, numBytes);


flihp:
@gwei3: can I get a quick review of this when you get a chance? I can provide example exploit code if you'd like to repro.

gwei3:
It is a obvious bug, and the fix looks good.

PR #296, opened 2016-11-18
fix #295

### LINKED ISSUE #295: Resourcemgr does not support multiple processes using mutilple connection
I have a multi-process scene, each process will create a new connection, this scenario will fail to call. After investigation, I found a bug of **resourcemgr**. In **resourcemgr.c** **2383** line, a global variable **cmdBuffer** is used to receive commands. When multiple processes send commands at the same time, the value of this global variable is unknown, resulting in a bug，I think this variable should be a **local variable**

medition:
this is a fix for bug  #295


flihp:
Moved to #299 for minor rewrite.

PR #298, opened 2016-11-18
Tcti device null deref

### LINKED ISSUE #297: NULL deref bug in libtcti-device
The device TCTI initialization function will dereference a null pointer (second parameter) if the first is also NULL. See https://github.com/01org/TPM2.0-TSS/blob/master/tcti/tcti_device.c#L244

flihp:
This resolves #297, adds a unit test to ensure we don't regress and does some cleanup in the unit test. This was the first unit test I wrote so it still needs a pile of cleanup.


flihp:
this PR needs to be broken up into one for the cleanup in the device TCTI unit test and another for the bug fix.

Issue #305, opened 2016-11-29
compiling with std=c99 causes inclusion failure


williamcroberts:
Building tpm2.0-tools with CFLAGS=-std=c99 causes:

In file included from /usr/local/include/sapi/tpm20.h:46:0,
                 from src/debug.c:31:
/usr/local/include/sapi/tss2_tcti.h:66:2: error: #error Info: Platform not supported for TCTI_POLL_HANDLES
 #error Info: Platform not supported for TCTI_POLL_HANDLES

It appears as though the if defined linux || defined unix is not working. Testing locally seems to indicate that __linux__ should be used instead, or included in the ifdef switch.

flihp:
The logic there is a little weird. Not sure why we define that type for the `TCTI_POLL_HANDLES` stuff at all there if we're just going to `#error` out. I guess that's a separate issue all together though. Back to your point:

`linux` is an obsolete symbol (not POSIX) for Linux systems. It should be `__linux__`. Similarly `unix` should be `__unix__`. Our use of `__linux` is equally broken. https://sourceforge.net/p/predef/wiki/OperatingSystems/

I spent a few minutes with this and the `__linux__` stuff is pretty easy to fix. Other stuff goes sideways with `<time.h>` stuff in `criticalsection.h`.

flihp:
Looks like  gcc compiles with `-std=gnu89` by default, might be better to go to `-std=gnu99` instead of straight to `-std=c99`. This stuff won't compile with `-std=c89` after all.

williamcroberts:
Yeah I figured we could flip this later if we want. The C build for tpm2.0-tools was the one htting this #define include issue when I was flipping that over.

flihp:
Gotcha. I've already started playing around with this so I figure cleaning it up a bit is a good idea since I'm already halfway there. I'll update the default CONFIG_SITE file to include `-std=gnu99` and fix up the `__linux/linux` to `__linux__`.

Do you have an opinion on `ifdef __blah__` vs `if defined(__blah__)`?

flihp:
I guess '#ifdef' is limited to a single conditional (can't append logical operators) so maybe it's best to go with `#if defined` so we can handle the logical or.

williamcroberts:
Yeah ifdef as you pointed out is limited, so for compound statements #if defined() is the choice.

flihp:
@williamcroberts : got a PR over in #310

flihp:
[...]

PR #310, opened 2016-12-08
build: build with gnu99 & cleanup OS detection preprocessor macros

flihp:
Our use of __linux / __unix / linux / unix are all obsolete. The modern
equivalent is __linux__ and __unix__. We were also using an inconsistent
mess of '#if defined' '#ifdef' and '#if' when checking for these. We're
now consistently using '#if defined(__blah__)'. This allows us to use
logical operators easily since we're checking for two symbols.

### DIFF
--- a/include/sapi/tss2_tcti.h
+++ b/include/sapi/tss2_tcti.h
@@ -54,11 +54,11 @@ extern "C" {
 #include "tss2_common.h"
 #include <stddef.h>
 
-#if defined _WIN32
+#if defined(_WIN32)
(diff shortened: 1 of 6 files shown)


flihp:
see discussion in #305

flihp:
Thanks for the review @williamcroberts. I've been tied up with some travel but will have this merged tomorrow hopefully.

flihp:
I've added a few lines to the CHANGELOG.md to document these changes and pushed.

PR #312, opened 2016-12-13
adds fix for installed libs not being found

mchlnix:
After following the build instructions and running `sudo make install`, resourcemgr could not find libsapi.so.0. Running `sudo ldconfig` fixed that lookup error.


flihp:
This is great @mchlnix, thanks for clarifying what's likely a common issue. I've broken your message out into a `NOTE:` after the install step with an explicit command. Should be all merged.

PR #318, opened 2016-12-22
[Resource Manager] Use SO_REUSEPORT

### LINKED ISSUE #317: Resource manager doesn't use SO_REUSEPORT
Without that, restarting Resource Manager after it has been stopped while some clients were connected is not possible.

Step to reproduce: 
1) Start resource manager
2) Use some command from TPM2.0-tools (e.g. tpm2_listpcrs)
3) Stop resource manager before the previous command has terminated
4) Try to restart  resource manager

```
dguerri@ubuntu:~$ resourcemgr -sim
socket created:  0x3
socket created:  0x4
Client connected to server on port:  2322
Client connected to server on port:  2321
Initializing Resource Manager
maxActiveSessions = 64
gapMaxValue = 65535
socket created:  0x5
bind to IP address:port:  127.0.0.1:2324
Other CMD server listening to socket:  0x5
socket created:  0x6
bind to IP address:port:  127.0.0.1:2323
TPM CMD server listening to socket:  0x6
Starting SockServer (TPM CMD), socket: 0x6.
Starting SockServer (Other CMD), socket: 0x5.
Accept socket:  0x7
Accept socket:  0x8
Resource Manager Other CMD Server accepted client
TpmCmdServer died (TPM CMD), rval: 0x00000000, socket: 0x7.
OtherCmdServer died (Other CMD), socket: 0x8.
Accept socket:  0x9
Accept socket:  0xa
Resource Manager Other CMD Server accepted client
TpmCmdServer died (TPM CMD), rval: 0x00000000, socket: 0x9.
OtherCmdServer died (Other CMD), socket: 0xa.
^C
```
Rerouce manager has been interrupted with ^C

```
dguerri@ubuntu:~$ resourcemgr -sim
socket created:  0x3
socket created:  0x4
Client connected to server on port:  2322
Client connected to server on port:  2321
Initializing Resource Manager
maxActiveSessions = 64
gapMaxValue = 65535
socket created:  0x5
bind to IP address:port:  127.0.0.1:2324
Other CMD server listening to socket:  0x5
socket created:  0x6
bind failed with error 98
Resource Mgr, upstream interface to applications, failed to init sockets.  Exiting...
```

error 98 means `EADDRINUSE`

dguerri:
Fix #317 

I understand that using SO_REUSEPORT could have security implications as another application can bind to the port used by resource manager in order to capture some of its incoming connections.
In Linux, to prevent unwanted processes from hijacking a port that has already been bound by a server using SO_REUSEPORT, all of the servers that later bind to that port must have an effective user ID that matches the effective user ID used to perform the first bind on the socket.
So that can be mitigate running the resource manager with an ad-hoc user. If it is run as root, root would be required to bind, but in that case there is very little we can do.

Apparently, this don't work in the same way on Windows so I would recommend against using SO_REUSEADDR (there is not SO_REUSEPORT on Windows) for the security issues that it implies.

### DIFF
--- a/common/sockets.cpp
+++ b/common/sockets.cpp
@@ -56,6 +56,9 @@ InitSockets( const char *hostName,
 {
     sockaddr_in otherService;
     sockaddr_in tpmService;
+#ifndef _WIN32
+    int optval = 1;
+#endif
     int iResult = 0;            // used to return function results
 
 #ifdef _WIN32
@@ -80,6 +83,13 @@ InitSockets( const char *hostName,
         return(1);
     }
     else {
+#ifndef _WIN32
+        iResult = setsockopt(*otherSock, SOL_SOCKET, SO_REUSEPORT, &optval, sizeof(optval));
+        if (iResult) {
+          SAFE_CALL( debugfunc, data, NO_PREFIX, "setsockopt failed with error = %d\n", WSAGetLastError() );
+          return(1);
+        }
+#endif
         SAFE_CALL( debugfunc, data, NO_PREFIX, "socket created:  0x%x\n", *otherSock );
         otherService.sin_family = AF_INET;
         otherService.sin_addr.s_addr = inet_addr( hostName );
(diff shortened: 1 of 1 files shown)


dguerri:
Hey, you are welcome! 
That's definitely an interesting use of SO_REUSEPORT that I wasn't aware of :)
My goal was to allow restarting the resource manager after a crash or when it is stopped normally while some connections are still in TIME_WAIT.

The flip side (of both your use case and in mine) is that a rogue process can steal connections meant for the resource manager and do a MITM attack (that's why I added that rationale).

flihp:
I didn't click through to the issue you reference in the message. I looked straight past it. Works for me. Merged in 186e589d7db80e10aff461268c77720957575dcf

PR #322, opened 2017-01-03
tss2_tpm2_types: Fix definition of PCR_LAST.

flihp:
See part 2 of the spec, table 30. In its previous form any use of the
PCR_LAST macro produced an error since IMPLEMENTATION_PCR1 was
undefined. Should have been IMPLEMENTATION_PCR - 1.


wcarthur1:
Probably due to a bug in my autogeneration code.

PR #323, opened 2017-01-10
test: Add integration test for asymmetric encrypt and decrypt use case.

hteh:
Remove asymmetric encrypt and decrypt use case from tpmtest.


flihp:
Did some fixup and merged. You're not building with the `CONFIG_SITE` environment variable set to `$(srcdir)/lib/default_config.site` so your local build isn't failing on the warning (in this case it was an unused variable). See https://github.com/01org/TPM2.0-TSS/blob/master/INSTALL#L23

PR #343, opened 2017-03-01
Replace TCTI invocation through function pointers with tss2_tcti_* macros

flihp:
We made these macros specifically to keep people from doing what we're
doing here. Cleanup and readability.


flihp:
merged in 8c638101cbdf1ddb5e437a0454db67c5a82e1412

PR #346, opened 2017-03-02
libmarshal: Add functions to (un)?marshal TPM_ST and use them.

flihp:
This commit adds the marshalling functions as well as replaces some
convoluted access to said types in the CommandUtil module.

Use of these functions shows that having the Marshal function take a
reference to the type being marshalled makes using constants awkward. It
requires that the constant be assigned to a variable first which isn't
ideal or necessary. Probably worth fixing the API early if this turns
out to be a good change.


flihp:
Broke a unit test with this one

flihp:
Ugh, this belongs in the 'base-types' module.

Issue #309, opened 2016-12-05
CommandUtil.c


williamcroberts:
rval == TSS2_RC_SUCCESS always TRUE. The last branch in the snippet below checks that rval is is OK, however, that is initialized to OK and not changed since function entry. Thus that branch is always taken. Should it be chcking the SYS_CONTEXT->rval value instead?

`
TSS2_RC     rval = TSS2_RC_SUCCESS;
UINT32      responseSize;

if( SYS_CONTEXT->rval != TSS2_RC_SUCCESS )
    return SYS_CONTEXT->rval;
if( cmdAuthsArray != 0 )
{
    SYS_CONTEXT->rval = Tss2_Sys_SetCmdAuths( sysContext, cmdAuthsArray );
}

if( rval == TSS2_RC_SUCCESS )
{
    SYS_CONTEXT->rval = FinishCommand( SYS_CONTEXT, cmdAuthsArray, &responseSize );
`

flihp:
Yeah the outer conditional @ https://github.com/01org/TPM2.0-TSS/blob/master/sysapi/sysapi_util/CommandUtil.c#L241 looks like a n00p to me too. Looks like it should be
```
SYS_CONTEXT-rval
```
instead as you note.

PR #347, opened 2017-03-02
CommandUtil: Check rval from sysContext CommonOneCall.

flihp:
The local variable being checked previously was initialized to the value
in the conditional which made it a no-op. After this change the
variable was unused and thus removed.


flihp:
this resolves #309

flihp:
merged in 362d45fe783dc48b0339087ff8b9cc8bc0b1e4d6

Issue #348, opened 2017-03-02
link unit tests against libraries


flihp:
The current Makefile.am builds in the relevant sources from the SAPI / TCTI libraries directly to the test binary. This is pretty silly since we've got libraries already built. We should link against the relevant libraries instead.

PR #349, opened 2017-03-02
Link unit tests against libsapi instead of building source directly.

flihp:


flihp:
this resolves #348

PR #351, opened 2017-03-05
Cleanup includes of local files.

flihp:
This is ugly but beneficial. All source files that includd local headers
(as in headers in this project) now use the quoted form in the #include
directive. They also reference files from $(srcdir) instead of relying
on the build to setup -I's for each binary / library.


flihp:
merged in dd3b03ce36be9e1021d3d793a6819089ec7b0d29

Issue #217, opened 2016-05-26
hardening compiler flags


flihp:
Select a set of compiler flags to "harden" the build output. Probably best to start with the data / recommandations from http://security.stackexchange.com/questions/24444/what-is-the-most-hardened-set-of-options-for-gcc-compiling-c-c and go from there. 

It's likely that we don't want to shove these flags down our users throats though. Probably best to keep the default flags generic and allow consumers to specify their own at ./configure time. This will probably end up being a documentation exercise with the recommended flags for hardening are documented in the README alongside instructions for passing them through ./confgiure.

flihp:
given the direction from #219 these flags should land in the config.site file

flihp:
PR #355 will resolve this issue.

PR #355, opened 2017-03-08
Flag cleanup

flihp:
Add the remaining preprocessor, compiler and linker flags. Do some cleanup along the way.


flihp:
resolves #217

Issue #356, opened 2017-03-08
marshalling functions for base types should pass by value


flihp:
No reason we can't pass these on the stack. Passing pointers makes things unnecessarily awkward.

flihp:
fixed in 42039cc0d422d234925006296345398aac9abd78

PR #357, opened 2017-03-09
libmarshal: Pass source parameter by value for *_Marshal functions.

flihp:
Passing these by reference made no sense and made using the functions
awkward (couldn't marshal constants).


flihp:
this resolves #356

PR #358, opened 2017-03-09
More flags fixes

flihp:
This also includes a fix for a type-punning issue. The CAST_TO_* macros were a convenient way to convert from a primitive type to a byte array. But as we're dialing up the hardening of the compiler / preprocessor / linker flags GCC gets increasingly strict about things like aliasing. Using `memcpy` is the right way to do this regardless.

Note to future self: enable these flags before you write a single line of code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,8 @@ depcomp
 install-sh
 libtool
 ltmain.sh
-m4/
+m4/libtool*
+m4/lt*.m4
 Makefile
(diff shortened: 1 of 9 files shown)


flihp:
Those macros you added to the tools repo were really useful. I didn't realize how badly I was abusing CONFIG_SITE till I did a few hours of research after reading that thread on the zcash PR you had in your comments. Very helpful.

PR #372, opened 2017-03-11
tpmtest: Use inttypes 64bit format specifier for clock values.

flihp:
long unsigned int is a different size on 64bit and 32bit platforms. Since
the clock values are fixed width we shouldn't be using architecture
dependant format spacifier anyways.


flihp:
this fixes build breakage on 32bit systems

flihp:
merged in f73ea368bbfa11ff9747d5bd91738cf6588297e9

PR #375, opened 2017-03-13
Fix wrong MACROs defined in tss2_tpm2_types.h

### LINKED ISSUE #374: Wrong MACROs defined in tss2_tpm2_types.h
The value for TPM_CLOCK_COARSE_SLOWER, TPM_CLOCK_MEDIUM_SLOWER and  TPM_CLOCK_FINE_SLOWER should be negative instead of positive according to spec.

gwei3:
The value for TPM_CLOCK_COARSE_SLOWER, TPM_CLOCK_MEDIUM_SLOWER and
TPM_CLOCK_FINE_SLOWER should be negative instead of positive according to spec.
Fix #374.


flihp:
merged in 6941fe25eacbc304375f8241f5eb3314d2d099ed

Issue #378, opened 2017-03-14
Unit test for device TCTI to test successful transmit / receive path


flihp:
These are test cases for the common, successful execution of the transmit / receive paths. Will require wrapping `read` and `write` syscalls but should be simple.

flihp:
resolved in tcti-device/socket unit tests

PR #380, opened 2017-03-14
unit: Add tests for successful calls to device TCTI receive / transmit.

flihp:
This required building the device tcti source directly into the unit
test. Linking against libtcti-device borked up the linker wrap magic.


flihp:
this resolves #378

Issue #368, opened 2017-03-09
CHANGE_ENDIAN_* use in authorizations.c


flihp:
Looks like accesses to nonces and hmac parameters and the typical header stuff.

flihp:
missed a few calls, will resubmit

tstruk:
Fixed in 16369f8c

PR #384, opened 2017-03-20
SetCmdAuths: Remove use of CHANGE_ENDIAN_*.

flihp:
This is pretty awkward: marshalling, then unmarshalling the size field
from the header. If possible, maintaining the header values in context
structure (in host byte order), and marshaling them immediately before
transmission would be easier.


flihp:
this resolves #368

flihp:
missed a bunch of stuff in this, will resubmit

PR #385, opened 2017-03-20
Rsp header save

flihp:
First steps toward simplifying the response processing path. These commits add a response header structure to the context to persist the unmarshalled response header. This keeps later code processing the response from having to do a bunch of redundant casting / unmarshalling.


flihp:
Added another patch that closes out #368

PR #387, opened 2017-03-21
configure: Remove check for C++ compiler.

flihp:
This should have been removed when all cpp files were converted to c.


flihp:
resolved in 41a423802433181ecb1e82f35adaef011c9a8cac

PR #389, opened 2017-03-23
Makefile.am: Add missing dependency to libmarshal

webmeister:
Without this fix, when cross-compiling for ARMv7, ld complains
`warning: libmarshal.so.0, needed by sysapi/.libs/libsapi.so, not found (try using -rpath or -rpath-link)`
and subsequently fails to build because of undefined references to the marshal functions.


flihp:
LGTM. Thanks

Issue #270, opened 2016-09-12
TPM2.0 Testing on Ubuntu with Intel's TPM2.0-TSS.


musakontron:
Hi, we are using TPM2.0 (Fw 5.51.2098.2) on our platform:
, following is the error i am getting on our platform

$ ./test/tpmtest/tpmtest
we instantiate the tests suite and then a menu of desired test is displayed 
when it comes to 32 - EC Ephemeral TESTS we get this message:
**EC Ephemeral TESTS:
failing case: FAILED! Ret code s/b: 8000b, but was: 143
TpmCmdServer died (TPM CMD), rval: 0x00000000, socket: 0xa.
OtherCmdServer died (Other CMD), socket: 0xb.**

I made some research adn i found out https://github.com/01org/TPM2.0-TSS/issues/233 is describing the same issue.
Conclusion:
As to Non-NULL owner auth, it is usually set by you when execute some standalone tools such as tpm2_takeownership etc. You need to clear the tpm.
There are some ways to clear the TPM:
1. when the platform auth is NULL, you can run tpm2_takeownership -c.
2. when the platform auth is non-NULL, you need to find out how to clear the tpm via bios settings, hardware jumper, etc.
   i tried to clear it with the bios setup, i still have the failure.
   i tried to clear it using tpm.msc from windows but still the same failure.

At this point i tried:
/tpm2.0-tools/src/tpm2_takeownership -c
**Start to clear the Hierarchy auth....
ClearControl Failed ! ErrorCode: 0x9a2**

I disabled the option in BIOS setup (PH Randomization in TPM menu) and after that i was able to take ownership:
I performed  TPM Clear  from the BIOS and reboot the system.
When rebooting on OS (ubuntu), i reran 
root@ubuntu:~/TPM2.0-TSS# tpm2_takeownership -c
**Accept socket: 0x8
Resource Manager Other CMD Server accepted client
Accept socket: 0x9
Resource Manager TPM CMD Server accepted client

Start to clear the Hierarchy auth....
TpmCmdServer died (TPM CMD), rval: 0x00000000, socket: 0x9.
OtherCmdServer died (Other CMD), socket: 0x8.**

$ /usr/local/sbin/resourcemgr
$ ./test/tpmtest/tpmtest
we instantiate the tests suite and then a menu of desired test is displayed 
when it comes to 32 - EC Ephemeral TESTS we get this message:

**EC Ephemeral TESTS:
[...]

PR #392, opened 2017-03-30
Check for not-implemented commands in "tpmtest"

DomiNic5787:
Due to "TCG PC Client Platform TPM Profile (PTP) Specification Level 00 Revision 00.43" folllowing commands are optional:
* TPM_CC_ChangeEPS
* TPM_CC_ChangePPS
* TPM_CC_EncryptDecrypt
* TPM_CC_EC_Ephemeral

Signed-off-by: Dominic Grauvogl <dominicmanuel.grauvogl@infineon.com>

### DIFF
--- a/test/tpmtest/tpmtest.c
+++ b/test/tpmtest/tpmtest.c
@@ -88,6 +88,8 @@
 TPM_CC currentCommandCode;
 TPM_CC *currentCommandCodePtr = &currentCommandCode;
 
+TPML_CCA *supportedCommands;
+
 //----
 //char errorString[200];
 //----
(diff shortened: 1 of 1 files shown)


DomiNic5787:
This commit should fix issue #270.

flihp:
Thanks for digging into this @DomiNic5787. Assigned to myself for review.

test/tpmtest/tpmtest.c
```diff
@@ -8684,6 +8799,9 @@ void InitTpmTest()
 
     Tss2_Sys_Startup ( sysContext, TPM_SU_CLEAR );
     CheckTpmType();
+
+    GetCommands( sysContext, &supportedCommands );
```
flihp: You return a TSS2_RC from GetCommands but you it's not checked here. This could result in the supportedCommands pointer being uninitialized when you go to check it later?

test/tpmtest/tpmtest.c
```diff
@@ -458,6 +460,95 @@ void GetTpmVersion()
...
+    {
+        numCommands = capabilityData.data.tpmProperties.tpmProperty[0].value;
+    }
+    else
+    {
+        goto returnFromGetCommands;
+    }
+
+    // Allocate memory for them
+    *supportedCommands = (TPML_CCA *)malloc( numCommands * sizeof( TPMA_CC ) + sizeof( UINT32 ) );
```
flihp: What's the reason for the dynamic allocation here? How badly does a static global TPMS_CAPABILITY_DATA structure bloat the binary?

test/tpmtest/tpmtest.c
```diff
@@ -458,6 +460,95 @@ void GetTpmVersion()
...
+    if( !*supportedCommands )
+    {
+        rval = TSS2_BASE_RC_INSUFFICIENT_BUFFER + TSS2_RESMGR_ERROR_LEVEL;
```
flihp: I know that this is a rare condition but if a memory allocation in this test program fails the error that will be returned to the user will indicate that the error came from the resource manager. This is really misleading and could lead to some really painful (and useless) bug hunts in the resource manager. If the application fails to allocate memory there's little that the caller can do other than report the issue to the caller / user and exit. I'd think a simple `printf` with information about the failure would suffice. In this case the caller will probably just want to terminate the program.

test/tpmtest/tpmtest.c
```diff
@@ -458,6 +460,95 @@ void GetTpmVersion()
...
+{
+    UINT32 i;
+    UINT8 rval = 0;
+
+    for( i = 0; i < supportedCommands->count; i++ )
+    {
+        if( (TPM_CC)( supportedCommands->commandAttributes[i].commandIndex ) == commandCode )
+        {
+            rval = 1;
+			break;
```
flihp: whitespace

test/tpmtest/tpmtest.c
```diff
@@ -458,6 +460,95 @@ void GetTpmVersion()
...
+//
+// Searches in a list for command attributes bit field for a command code.
+// Assumes that the supportedCommands structure has been populated by
+// calling GetCommands beforehand.
+//
+// Returns:
+//  1, if found.
+//  0, if not found.  This means that the TPM doesn't support this command.
+//
+UINT8 IsCommandImpl( TPM_CC commandCode, TPML_CCA *supportedCommands )
```
flihp: The semantics of this call (using the "Is" prefix) makes me think it should return a boolean value. Doing so will make code that calls the function much easier to read:

```
if (IsCommandImplemented (TPM_CC_ChangeEPS, supportedCommands))
{
    /* do stuff */
}
```

DomiNic5787:
Thanks for the review @flihp.
I will provide a new patch set as soon as possible.

DomiNic5787:
@flihp: I reworked your change requests on this commit some days ago. Can you please have a look on it. Thanks!

flihp:
This was merged a while back in a43c48fe781db54158a69f3e058a6740d8883e30.

PR #395, opened 2017-04-04
Android Support and Warning Fixes

williamcroberts:


Android.mk
```diff
@@ -0,0 +1,31 @@
...
+LOCAL_COMMON_SRC = $(wildcard common/*.c)
+
+LOCAL_LIBMARSHAL_SRC := marshal/base-types.c log/log.c
+
+###################################
+include $(CLEAR_VARS)
+
+LOCAL_MODULE := libtss2.0
+LOCAL_MODULE_TAGS := optional
+LOCAL_SRC_FILES := $(LOCAL_SYSAPI_SRC) $(LOCAL_SYSAPIUTIL_SRC) $(LOCAL_COMMON_SRC)
```
williamcroberts: I think this is blank here, this library looks awfully small.

williamcroberts:
Intresting, Android has some TPM2 project already: https://android-review.googlesource.com/#/admin/projects/platform/external/tpm2

flihp:
Wonder if that's the same stack Google built for ChromeOS

flihp:
I'm going to take the patch with the header guard fix in it manually. We can chat about a plan to support the Android build stuff in the mean time.

PR #396, opened 2017-04-06
do not adjust loop variables in case of send/recv being interrupted

d-e-s-o:
When send/recv indicate an error (i.e., return -1) because the call was
interrupted, we should not use the return value to adjust the loop
variables representing the bytes send/received and the remaining bytes
to send/receive.
This change moves the modification of the loop variables from the loop
header into the body where they are skipped in case the send/recv calls
were interrupted (that is, they returned an error and errno is set to
EINTR).


flihp:
Great fix. Merged

PR #400, opened 2017-04-08
make bootstrap binaries configurable via environment

d-e-s-o:
The bootstrap script hard-codes the name of the autoreconf binary. This
hard-coding makes it unfeasible to run in a sandboxed environment where
those binaries are located at non-standard paths. Typically in
makefiles, binaries can be overwritten through environment variables.
This change adjusts the bootstrap script to honor the AUTORECONF
variable if present in the environment but is able to fall back to the
previous values if such variables are not present.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -13,6 +13,7 @@ src_listvar () {
 }
 
 VARS_FILE=src_vars.mk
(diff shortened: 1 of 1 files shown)


flihp:
LGTM. Thanks @d-e-s-o !

PR #401, opened 2017-04-17
fix off-by-one error in CheckOverflow

d-e-s-o:
The overflow check as provided by CheckOverflow contains an off-by-one
error causing an out-of-bounds read by one byte to not be detected
properly. The used size should be calculated as
  usedSize = (INT64)(nextData - buffer) + size;
and not
  usedSize = (INT64)(nextData - buffer) - 1 + size;

For the latter case, consider the attempt of reading the very first two
bytes from a buffer of one byte size. 'usedSize' will be 1 (0x1000 -
0x1000 - 1 + 2). Because 'usedSize' is not greater than 'bufferSize'
(also 1), we will not detect the out-of-bounds read.
This change fixes the issue by removing the addend of -1.

### DIFF
--- a/common/checkoverflow.c
+++ b/common/checkoverflow.c
@@ -34,7 +34,7 @@ TSS2_RC CheckOverflow( UINT8 *buffer, UINT32 bufferSize, UINT8 *nextData, UINT32
     TSS2_RC rval = TSS2_RC_SUCCESS;
     INT64 usedSize;
 
-    usedSize = (INT64)(nextData - buffer) - 1 + size;
+    usedSize = (INT64)(nextData - buffer) + size;
     if( usedSize > (INT64)bufferSize )
     {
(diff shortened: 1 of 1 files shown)


d-e-s-o:
Any update on this pull request?

flihp:
Sorry @d-e-s-o I've been buried under another task. Will get this done in the next few days.

flihp:
Thanks for taking the time to dig into this @d-e-s-o. This code is slowly being phased out but you're definitely fixing a legit scenario. A quick search shows that the only remaining place this function is called from is the resourcemgr which will be removed / replaced entirely in the next week-ish (this is the rock I've been hiding under). Either way, merging to fix the issue in the interim.

This will actually be most useful in the 1.x release branch that we've created recently. It's looking like we'll be doing a few releases based on the 1.0 code for a few customers since master has / will be breaking API compatibility with the 1.0 code. This function is still heavily used there.

PR #404, opened 2017-04-24
tss2_tcti.h: Align return types from TCTI invocation macros with spec.

flihp:
The spec is a inconsistent w/r to the return type for a NULL context.
The inconsistency is between the example macro to invoke a function
pointer from those in the context structure in section 7.3.2 and the
description of these functions in 7.3.3.2.1 through 7.3.3.2.6.

This patch aligns with the "normative" text in the descriptions of the
functions and now returns BAD_REFERENCE instead of BAD_CONTEXT for NULL
contexts provided said functions.


flihp:
merge this into master after merging into 1.x

flihp:
merged into 1.x, merged back to master

PR #405, opened 2017-04-28
Tss2_Sys_Execute: Replace absurdly large timeout with TSS2_TCTI_TIMEO…

lamalexck:
<This is a cherry-pick of 68876a on the main branch>

…UT_BLOCK.

The spec states that a call to Tss2_Sys_Execute should be equivalent
to a call to Tss2_Sys_ExecuteAsync followed by a call to
Tss2_Sys_Finish with a timeout of -1 aka TSS2_TCTI_TIMEOUT_BLOCK.
See section 8.7.3.3 in spec titled "TSS System Level API and TPM
Command Transmission Interface Specification" version "Family 2.0,
Level 00, Revision 01.00"


flihp:
merged into 1.x

PR #407, opened 2017-05-09
Update bootstrap to use autoreconf as in master branch

lamalexck:


flihp:
I'll cherry-pick these from master

PR #412, opened 2017-05-22
configure: Undefine FORTIFY_SOURCE first, then define it

PeterHuewe:
On my hardened gentoo system _FORTIFY_SOURCE is already defined,
thus it makes sense to undefine it first then define it, not vice
versa.

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>


flihp:
LGTM. Thanks @PeterHuewe

PeterHuewe:
@flihp This somehow went missing again from master branch?!
And it is not present in the history of the file?

flihp:
Hmmm. I may have dropped this fix by mistake while cherry picking stuff into the 1.x branch and then merging it back. The only real conflicts were in the autotools stuff. Not sure why this would cause the commit to disappear but there are a lot of things about git that I don't understand.

When I get back from holiday I'll fix this up. Thanks for your vigilance. I probably wouldn't have noticed otherwise.

flihp:
Did a bit of digging on this finally. This patch was applied in 290840c841646c24d43527b1af284d76efc046c9. It's still visible in the files history. It does however look like it was reverted by a merge of the `1.x` branch back into `master` (between 3d1ad8d and 6e58f75). These branches have diverged significantly (especially after moving `tpmclient` into the test harness) and conflicts are becoming increasingly difficult to resolve accurately. We may be down to cherry-picking only from master.

Anyways I'm about to re-apply your original patch. It'll be redundant in the history but I'll make note of the issue in the resultant commit.

PR #413, opened 2017-05-22
Make TPM2.0-TSS compilable with clang and add clang to travis

PeterHuewe:
Some people prefer using clang to compile their code.
Some minor modifications are needed to make the code pass the compilers stricter settings.
Once it builds we can also add this new configuration to travis.


flihp:
LGTM: thanks @PeterHuewe

PR #418, opened 2017-05-26
Fixed bug in KDFA function:  if requested key size in bytes rounded u…

wcarthur1:
…p was

greater than the hash size, the key would be generated incorrectly.

Found this bug in January, but unable to submit pull request until some legal issues wrt open source were resolved.  Sharing this now so that others don't spend the painful day's effort it cost me to find this bug.

### DIFF
--- a/test/common/sample/kdfa.c
+++ b/test/common/sample/kdfa.c
@@ -87,7 +87,7 @@ TPM_RC KDFa( TPMI_ALG_HASH hashAlg, TPM2B *key, char *label,
     {
         // Inner loop
 
-        i_Swizzled = CHANGE_ENDIAN_DWORD( i );
+        i_Swizzled = CHANGE_ENDIAN_DWORD( i++ );
         *(UINT32 *)tpm2b_i_2Ptr = i_Swizzled;
 
         j = 0;


flihp:
Two things @wcarthur1 
1. I always assumed (incorrectly?) that this function was taken straight from the TPM part 4 spec. If we're making changes here then this isn't the case? Or was this just an error in the transcription?
2. Any chance I can get you to describe (or write) a test case to trigger / catch the issue? This is a pretty central component, would be nice to get test coverage.

Thanks!

flihp:
Also thanks for fighting the good fight on this one. I know how hard it is to get the military industrial complex to give back to OSS. Thanks for sticking it out.

wcarthur1:
Philip,

It was an error in transcription.

Will

wcarthur1:
Re test case, call the KDFa:
TPM_RC KDFa( TPMI_ALG_HASH hashAlg, TPM2B *key, char *label,
    TPM2B *contextU, TPM2B *contextV, UINT16 bits, TPM2B_MAX_BUFFER  *resultKey )

with SHA256 as the hashAlg with bits greater than 32*8.  For the rest of the parameters just select something.  You'll get a different resultKey with the bad code than you'll get with my fix.

flihp:
thanks for the details, I'll queue this up

flihp:
Apologies for the delay. Got stuck elsewhere.

PR #419, opened 2017-05-27
sort input files

bmwiedemann:
when building packages (e.g. for openSUSE Linux)
(random) filesystem order of input files
influences ordering of functions in the output,
thus without the patch, builds (in disposable VMs) would differ.

See https://reproducible-builds.org/ for why this matters.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -8,7 +8,7 @@ src_listvar () {
     suffix=$2
     var=$3
 
-    find "${basedir}" -name "${suffix}" | tr '\n' ' ' | (echo -n "${var} = " && cat)
+    find "${basedir}" -name "${suffix}" | LC_ALL=C sort | tr '\n' ' ' | (echo -n "${var} = " && cat)
     echo ""
 }
 


flihp:
I've been following the reproducible build stuff from a distance but obviously haven't got the details right :smile: Many thanks for the help @bmwiedemann.

I'm merging this into the 1.x release branch then back into master so here's hoping I don't screw this up too badly.

williamcroberts:
Funny, similar patches just hit the SE Linux mailing list as well. This person has been quite busy..

bmwiedemann:
@williamcroberts yep. I've gone through the whole openSUSE Tumbleweed diffs and fixed >90% = dozens of this kind of issues in a few weeks. See [the blog from the last week](https://reproducible.alioth.debian.org/blog/posts/109/) and [the one before](https://reproducible.alioth.debian.org/blog/posts/108/) for an incomplete list.

https://github.com/01org/TPM2.0-TSS/commit/dd43b828f3d9b516236e521ae0ff19cb4b11e620 seems to be on master, so closing this PR.

btw: the fun part is how different the ways are to contribute to projects, with github usually being the easiest and fastest and 'subscribe to the development mailing list' being among the less appreciated ones for 100x one-time contributers like me.

Issue #397, opened 2017-04-06
TCTI instantiation example


flihp:
The wiki needs a page with example code documenting how a TCTI should be instantiated.

flihp:
Been slow going on the TSS with efforts to get the tpm2-abrmd out the door in full swing. Lost track of this for a bit. See PR #421 for the proposed solution.

Issue #398, opened 2017-04-06
TCTI overview docs


flihp:
The wiki needs a page giving an overview of what a TCTI is, what it does and how to user one.

danglus:
The TSS spec doc includes a good overview:
https://trustedcomputinggroup.org/tss-system-level-api-tpm-command-transmission-interface-specification/

flihp:
Been slow going on the TSS with efforts to get the tpm2-abrmd out the door in full swing. Lost track of this for a bit. See PR #421 for the proposed solution.

PR #421, opened 2017-06-05
Man pages for basic TCTI functions.

flihp:


flihp:
This resolves #397 and #398

PR #426, opened 2017-06-13
Remove references to config.site from INSTALL

tstruk:
We should be using the system config.site


flihp:
Merged. Thanks for giving the docs a sanity check @tstruk. I completely forgot about all that CONFIG_SITE stuff.

PR #428, opened 2017-06-14
More cleanups

tstruk:
@flihp
- Cleanup the usage of extern "C"
- Only use goto where is make sense


tstruk:
@williamcroberts Done. Thanks for spotting the typo.

flihp:
Review from @williamcroberts accepted. LGTM. Doing away with the `if (null check) { return -1; } else { /* tons of code */ }` is a welcome change :smile:

PR #432, opened 2017-06-16
configure; Remove dependency / check for pthreads.

flihp:
The resourcemgr was using pthreads but it was removed. No need for this
library any longer.


flihp:
This needs to be merged back into master as well

flihp:
resolves #432

Issue #382, opened 2017-03-14
Ensure integration tests get fresh simulator


flihp:
As it stands now the travis config starts the simulator and then runs the integration tests in order. In the event that one of the tests leaves a bunch of stuff / objects lying around in the simulator it may effect subsequent tests: resource exhaustion, attempts to create objects / NV indices that are already defined etc.

Ideally there's something that we can do to reset the TPM completely to get a fresh instance. This of course depends on the features exposed by the simulator. First thing to do is look for interface to clear & reset the simulator. This must remove all objects including things that would survive a power cycle.

If this isn't possible we may have to fall back on cleaning out all of this stuff by hand.

flihp:
see #433

PR #433, opened 2017-06-19
Int log compiler

flihp:
Enable parallel execution of the integration tests by introducing a test wrapper script that starts an instance of the simulator for each integration test.


flihp:
resolves #382

flihp:
Fixed up & rebased to address issues found with the original integration log compiler script over in the tabrmd:

PR #447, opened 2017-06-30
README: Added documentation images for easy access

pdxjohnny:


pdxjohnny:
Personally I think this is easier to find the images if they are in the README.md. Nice to have all on one page IMO, this is a super subjective PR, I just think its nice like this. No offense taken if you close it.

flihp:
This is a great patch. Those diagrams were left behind by my predecessor so they're not quite accurate any longer but getting them on the front of the README is definitely the right thing to do. Thanks!

PR #455, opened 2017-07-05
Master merge 1.x

flihp:
This is effectively a n00p for master. Just merging the 1.x release branch back into master after a pile of cherry-picking.


flihp:
ugh, github wants to create a new merge commit even though the last commit in this series is exactly that. Will push this directly after checks pass.

williamcroberts:
This is why you should always rebase changes onto the branch when possible, cleaner history.

flihp:
what do you mean?

PR #458, opened 2017-07-12
Introduce Marshaling library

tstruk:
Add generic Marshaling library for TPM types


marshal/base-types.c
```diff
@@ -75,26 +71,36 @@ type##_Marshal ( \
...
+            break; \
+        case 8: \
+            src = HOST_TO_BE_64(src); \
+            break; \
+\
+    } \
+    memcpy (&buffer [local_offset], &src, sizeof(src)); \
     if (offset != NULL) { \
-        *offset = local_offset + sizeof (src); \
+        *offset = local_offset + sizeof(src); \
```
flihp: This is another style thing: I typically include a space after the function name and before the opening paren / `(`. I do the same for built-ins like `sizeof`. Is this unusual? Whatever the result we should document this in the CONTRIBUTING.md.

marshal/base-types.c
```diff
@@ -128,9 +134,25 @@ type##_Unmarshal ( \
          (uintptr_t)buffer, \
          (uintptr_t)dest, \
          local_offset); \
-    type tmp = 0; \
+\
     memcpy (&tmp, &buffer [local_offset], sizeof (tmp)); \
```
flihp: Above you removed the space between `sizeof` and `(` but not here. If we're going to change the style we should change them all (at least in this module).

marshal/tpma-types.c
```diff
@@ -0,0 +1,182 @@
...
+#include <inttypes.h>
+#include <string.h>
+
+#include "sapi/marshal.h"
+#include "sapi/tpm20.h"
+#include "tss2_endian.h"
+#include "log.h"
+
+#define TPMA_MARSHAL(type) \
+TSS2_RC type##_Marshal(type src, uint8_t buffer[], \
```
flihp: The declaration in the header and the definition in the implementation file use a different style. I commented on a similar change made in the last commit. I'd prefer to keep these consistent unless there's a compelling reason not to.

marshal/tpms-types.c
```diff
@@ -0,0 +1,701 @@
...
+    TPMS_PCR_SELECT *pcrSelect = (TPMS_PCR_SELECT *)ptr;
+    UINT32 i;
+    TSS2_RC ret;
+
+    if (ptr == NULL) {
+        LOG (WARNING, "src param is NULL");
+        return TSS2_TYPES_RC_BAD_REFERENCE;
+    }
+
+	ret = UINT8_Marshal(pcrSelect->sizeofSelect, buffer, buffer_size, offset);
```
flihp: There are some weird whitespace errors at the front of a few lines here. These are probably left over from copy / paste from the old implementation.

tstruk:
Why was this closed? I'm working on the comments now.

include/sapi/implementation.h
```diff
@@ -1331,36 +1325,14 @@ typedef  UINT32             TPM_CC;
 					  )
 
 #define VENDOR_COMMAND_ARRAY_SIZE   ( 0					\
-				      + CC_Vendor_TCG_Test		\
-				      )
+				      + CC_Vendor_TCG_Test)
```
tstruk: I really think the closing bracket in a new line looks strange and I agree that we should formalize the style. I will leave this out for now.

include/sapi/implementation.h
```diff
@@ -1331,36 +1325,14 @@ typedef  UINT32             TPM_CC;
...
-				      MAX(ALG_SHA384 * SHA384_DIGEST_SIZE, \
-					  MAX(ALG_SM3_256 * SM3_256_DIGEST_SIZE, \
-					      MAX(ALG_SHA512 * SHA512_DIGEST_SIZE, \
-						  0 ))))))
-
-#if MAX_DIGEST_SIZE == 0 || MAX_HASH_BLOCK_SIZE == 0
-#error "Hash data not valid"
-#endif
+#define MAX_HASH_BLOCK_SIZE SHA512_BLOCK_SIZE
+#define MAX_DIGEST_SIZE SHA512_DIGEST_SIZE
```
tstruk: This login looks strange because 64 (SHA1_BLOCK_SIZE) will always be smaller than 128 (SHA512_BLOCK_SIZE). Since the library will support all the algs, ALG_* for all SHAs macros are set to YES this logic didn't make much sense. If we add a new algorithm this MAX_HASH_BLOCK_SIZE will have to be updated anyway.

include/sapi/tss2_tpm2_types.h
```diff
@@ -614,12 +603,12 @@ typedef union {
...
 	unsigned int continueSession : 1;	/* SET 1 In a command this setting indicates that the session is to remain active after successful completion of the command. In a response it indicates that the session is still active. If SET in the command this attribute shall be SET in the response.CLEAR 0 In a command this setting indicates that the TPM should close the session and flush any related context when the command completes successfully. In a response it indicates that the session is closed and the context is no longer active.This attribute has no meaning for a password authorization and the TPM will allow any setting of the attribute in the command and SET the attribute in the response.This attribute will only be CLEAR in one response for a logical session. If the attribute is CLEAR the context associated with the session is no longer in use and the space is available. A session created after another session is ended may have the same handle but logically is not the same session.This attribute has no effect if the command does not complete successfully.  */
 	};
-	UINT32 val;
+	UINT8 val;
 } TPMA_SESSION;
 
 #else
 
-typedef uint32_t	TPMA_SESSION;
+typedef uint8_t	TPMA_SESSION;
```
tstruk: Good point. Changed to TPMA_* types.

marshal/base-types.c
tstruk: Agree. I left the spaces as they were. We can update the style changes in separate commits after we agree on what we want.

marshal/tpma-types.c
tstruk: Yes, that was based on personal custom. Changed to consisted style with the header.

marshal/tpms-types.c
tstruk: Right, this should be fixed now.

flihp:
This PR was inactive for ~2 weeks with no indication that you were still working on it. Easy enough to re-open it but we do need to get you sufficient privileges to do so yourself. I'll get on that now.

flihp:
weird. I can't reopen it either. It looks like Github won't let that happen since you've force a push to your branch since it was closed. I guess this wasn't a permission issue after all.

tstruk:
So do you want me to create a new PR?

flihp:
It looks like you'll have to. The issue is documented here: https://github.com/isaacs/github/issues/361

tstruk:
The new PR is #483

PR #459, opened 2017-07-13
Tpmclient mv to integration

flihp:
This patch set ports a particularly obnoxious test over to use the integration test harness (making 3 new tests). Once this is done there's a bit of cleanup followed by the porting of the `tpmclient` test harness to be executed like any other test case by the automake-based integration test harness.


test/tpmclient/tpmclient.c
```diff
@@ -287,7 +286,7 @@ void Delay( UINT16 delay)
 		   __FUNCTION__, __LINE__);				\
     }									\
     									\
-    Delay(demoDelay);							\
+    Delay(0);							\
```
tstruk: Why not remove this call altogether?

test/tpmclient/tpmclient.c
```diff
@@ -310,7 +309,7 @@ TPM2B_AUTH nullSessionHmac;
       DebugPrintf( NO_PREFIX, "\tPASSED! (%s@%u)\n",			\
 		   __FUNCTION__, __LINE__);				\
     }									\
-    Delay(demoDelay);							\
+    Delay(0);							\
```
tstruk: Why not remove this call altogether?

test/tpmclient/tpmclient.c
flihp: At first I did but it broke something. I started to investigate but didn't want to hold up the larger effort on account of fixing this. The breakage was very strange and only broke for clang not GCC so I'm expecting that it will take a while to track down. I think you're right though: the delay isn't adding anything useful to the program and should be removed in future work.

test/tpmclient/tpmclient.c
flihp: see comment above

flihp:
two questions: which two commits do you want me to squash? I'm having a hard time telling with the way github structures the comment thread. did you mean 30d4432 and 2224eda?

flihp:
second question is w/r to the tpmclient "script". This is generated by the build so you probably just need to clean and rebuild.

tstruk:
> did you mean 30d4432 and 2224eda?
No, sorry, I meant 30d4432 and e0512c7

flihp:
squashed

PR #461, opened 2017-07-18
Autogenerate list of source files

tstruk:
Automatically generate src_vars.mk every time files have been added or deleted.


flihp:
This has gone quiet. Feel free to open it back up if you'd like to discuss further.

tstruk:
The git dependency is a valid point, but this supposed to be for development only. It is helpful when you add remove files, which I'm doing now.
For release we can exclude it from the tar ball the same way we exclude the .git dir. Nobody suppose to remove/add files from the release packet anyway. I understand the objections and I'm not pushing it. I can simply have this locally on my repo. I can reopen if you see it valid.

PR #464, opened 2017-07-20
WIP: tpmclient to libtest_utils migration

pdxjohnny:
I'm moving a few things into unit tests and I need to move some of the functions from the tpmclient into the test_utils library so that we can use those functions for unit and integration tests as well. This PR will be a few patches moving things into the libtest_utils library so that we can use these test/helper functions in the other testing, not just tpmclient.

Thoughts / discussion would be good I'm not sure if this is the best way to do this but I like the idea of consolidating these helper functions into some kind of "testing" library.


flihp:
You're heads in the right place for sure. I went a different way by porting the tpmclient program over to use the integration test harness in #459. This removed the need for `InitSysContext` function since there's already similar functionality in https://github.com/01org/TPM2.0-TSS/blob/master/test/integration/context-util.c.

So you've obviously picking up on the libtest_util thing but I'm trying to phase out the stuff that's in `tpmclient` and not move it into libtest.

PR #471, opened 2017-07-21
1.x merge

flihp:
merging 1.x back into master after the tpmclient integration harness stuff


flihp:
groan, github can't handle merge commits that it doesn't create? weird. Will push directly. Sorry all.

PR #474, opened 2017-07-24
tpmclient: Fix and enable the hmac tests.

flihp:
The hmac tests only requrie the primary key referenced via the global
'handle2048rsa' handle variable. The hmac tests are run after the
TestCreate function (which loads this required key and one more) but
having this additional key loaded cause the hmac tests to fail with the
RC 0x902 aka 'TPM_RC_OBJECT_MEMORY'. Unloading this additional key manuall
allows the hmac tests to run successfully.

After the hmac tests run we flush the TPM and then re-run the TestCreate
function. This ensures that the keys referenced by the global handle2048rsa
and loadedSha1KeyHandle variables are loaded.

### DIFF
--- a/test/tpmclient/tpmclient.int.cpp
+++ b/test/tpmclient/tpmclient.int.cpp
@@ -4685,6 +4685,7 @@ void HmacSessionTest()
                     symmetric.keyBits.aes = 128;
                     symmetric.mode.aes = TPM_ALG_CFB;
 
+                    EndAuthSession (nvSession);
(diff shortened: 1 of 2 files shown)


test/tpmclient/tpmclient.int.cpp
```diff
@@ -6479,6 +6480,11 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
     TestDictionaryAttackLockReset();
 
     TestCreate();
+    rval = Tss2_Sys_FlushContext( sysContext, loadedSha1KeyHandle );
```
tstruk: Can we remove these spaces after/before the braces

flihp: We should start a larger conversation w/r to coding style. The original author used an undocumented style that's very foreign to OSS code. We need to document the style that we want to use going forward and set some rules around how we handle the migration. Personally when I add *new* code I use the style documented in the tools and tabrmd repo. When fixing up *existing* code I've tried to mimic the style used previously as best I can.

PR #475, opened 2017-07-24
Update location of algorithm ids

williamcroberts:
implementation.h stated that Table 2 contains the algorithm identifiers from the TCG Algorithm registry. The current document, here: https://trustedcomputinggroup.org/wp-content/uploads/TCG_Algorithm_Registry_Rev_1.24.pdf shows it in Table 3.


williamcroberts:
Actually the whole thing is out of date, and needs to be updated. For example, TPM_ALG_SHA3_256 is missing.

PR #478, opened 2017-07-25
add missing algorithms

williamcroberts:
Per the TCG Algorithm Registyry for TPM2.0, the header
implementation.h was missing the following algorithms:
* TPM_ALG_SHA3_256 - 0x27
* TPM_ALG_SHA3_384 - 0x28
* TPM_ALG_SHA3_512 - 0x29

Also, the definition is in table 3 not table 2.

The specification can be found at:
https://trustedcomputinggroup.org/wp-content/uploads/TCG_Algorithm_Registry_Rev_1.24.pdf

Fixes: 476

### DIFF
--- a/include/sapi/implementation.h
+++ b/include/sapi/implementation.h
@@ -318,7 +318,7 @@
     ((MAX_RSA_KEY_BYTES/2)*(3+CRT_FORMAT_RSA*2))
 #define  MAX_VENDOR_BUFFER_SIZE           1024
 
-// From TCG Algorithm Registry: Table 2 - Definition of TPM_ALG_ID Constants
+// From TCG Algorithm Registry: Table 3 - Definition of TPM_ALG_ID Constants
 
 typedef  UINT16             TPM_ALG_ID;
 #define  TPM_ALG_ERROR               (TPM_ALG_ID)(0x0000)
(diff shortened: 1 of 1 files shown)


williamcroberts:
@tstruk @flihp I'm not sure what the criteria is for enabling these via the
```
#define ALG_SHA3_384 YES
```

directives, do we want these enabled?

include/sapi/implementation.h
```diff
@@ -444,6 +444,18 @@ typedef  UINT16             TPM_ALG_ID;
...
+#endif
+#define ALG_SHA3_256_VALUE 0x27
+#if defined ALG_SHA3_384 && ALG_SHA3_384 == YES
+#define TPM_ALG_SHA3_384 (TPM_ALG_ID)0x28
+#endif
+#define ALG_SHA3_384_VALUE 0x28
+#if defined ALG_SHA3_512 && ALG_SHA3_512 == YES
+#define TPM_ALG_SHA3_512 (TPM_ALG_ID)0x29
+#endif
+#define ALG_SHA3_512_VALUE 0x29
```
williamcroberts: Do we want to enable the TPM_ALG defines?
```
#define ALG_SHA3_XXX YES
```

flihp:
The stuff in `implementation.h` is a bit convoluted. AFAIK the code followed the example in the TPM spec that was really intended for use by those implementing TPMs (not TSS). So those were meant to allow a TPM implementer to selectively enable whatever algorithms their TPM supports.

AFAIK the TSS is not the right place to determine which algorithms are supported. Higher level components can query supported algorithms for whatever TPM they're talking to using `GetCapability` if they need to.

The question though is when to make these change. I've been holding off waiting for the TCG to publish the new spec that describes the headers. My plan was to just convert the right parts of this spec into `.h` files and use these to replace the existing headers.  I figured keeping the header changes to one big commit would make rolling back if things go sideways easier.

We can do this in bits and pieces if necessary if you need it. I can't say how long it will take the header spec to make it through the TCG review process though. Could be months from what I'm told.

PR #485, opened 2017-08-02
WIP: test: Resource Manager integration tests

pdxjohnny:
Tore out TestRM from tpmclient and created integration tests for it.


flihp:
I put off looking at this since the travis tests are failing but after looking at those logs it seems like the build hung. After forcing travis to try again it looks like these tests are causing the hang? Maybe I'm missing something there. Either way these tests were disabled when we removed the resource management daemon from this repo (it's over at https://github.com/01org/tpm2-abrmd now) so they're likely testing stuff that shouldn't work unless the tabrmd is in use. So I'd say it's most likely that they should move over to the tpm2-abrmd repo.

flihp:
No motion on this in a while so I'm closing. Feel free to re-open if you make progress on this.

PR #486, opened 2017-08-02
man: update error message in man page example

cardoe:
The man page example implied that these functions connected using the
tabrmd but these are to bypass the tabrmd and go directly to the device
or the socket.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>


flihp:
Yeah that's not right. I think you just busted me for copy / pasting error messages from libtcti-tabrmd. Good catch & thanks for the patch @cardoe

PR #491, opened 2017-08-04
sapi: Add Tss2_Tcti_Finalize() to the API

### LINKED ISSUE #490: Helper macros for functions in tss2_tcti.h are not friendly to FFIs
When trying to use this toolstack from another language you must do so via a FFI (foreign function interface). Unfortunately the opposite function to `InitSocketTcti()` and `InitDeviceTcti()` appears to be a macro called `tss2_tcti_finalize()`. By using this approach its not possible to call this from another language that is calling into the C API. There really needs to be wrapper functions that do these actions and have them exported by the library. There are a few other TCTI macros in here but at this point I am not sure if I would have to call them if I intend on only using the SAPI from my FFI.

cardoe:
Added an explicit Tss2_Tcti_Finalize() to the API to avoid issues with
languages trying to wrap up this API via a FFI. By exposing this method
only via a macro on an opaque structure it becomes impossible to call it
from an FFI. But by providing an explicit function it can now be called.
I left the existing tss2_tcti_finalize() macro because some code
(specifically the unit tests) link to the TCTI libraries only. In these
cases the new function would not be available. The function cannot be
provided by the TCTI library since many programs link all of them in
when they use the SAPI and that would cause a conflict. So both are
available to preserve compatibility. Refs #490.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>

### DIFF
--- a/include/sapi/tss2_tcti.h
+++ b/include/sapi/tss2_tcti.h
@@ -155,6 +155,11 @@ TSS2_TCTI_POLL_HANDLE *handles, size_t *num_handles);
 
 typedef TSS2_TCTI_CONTEXT_COMMON_V1 TSS2_TCTI_CONTEXT_COMMON_CURRENT;
 
+// function that is equivalent to the tss2_tcti_finalize macro
+TSS2_RC Tss2_Tcti_Finalize(
+    TSS2_TCTI_CONTEXT *tctiContext
+    );
+
 #ifdef __cplusplus
 }
 #endif
--- a/sysapi/sysapi/Tss2_Tcti_Finalize.c
+++ b/sysapi/sysapi/Tss2_Tcti_Finalize.c
@@ -1,0 +1,47 @@
+//**********************************************************************;
+// Copyright (c) 2015, 2016, Intel Corporation
+// Copyright (c) 2017, Star Lab Corp.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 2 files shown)


flihp:
Hey @cardoe. Thanks for putting this together. I'm not sure where this should live though. Since it's not part of the SAPI spec I'm hesitant to put it in that library. But since there's no single TCTI library where we can put this things get a bit weird.

What would you think about creating a separate library specifically for enabling bindings for the TCTI layer? It would contain Tss2_Tcti_* functions like the finalize you've provided but one for each function. Each would effectively be a wrapper to invoke the function pointer in the underlying structure. Effectively we'd be creating a library that provides similar functionality to the tss2_tcti_* macros. It would be tiny and I'd expect most consumers would link statically against it.

Also where do your bindings live? What's the language (rust)? Would love to take a look so I can understand this stuff better. Never done any language bindings myself.

cardoe:
Yeah I like that idea. Maybe something like `libtctiutil.a`? What I've got so far isn't usable at all but I've gone ahead and pushed it publicly at https://github.com/starlab-io/tss-sapi Unfortunately unions are unsafe by their very nature and much of the public API I've run into thus far uses unions so I've got to rethink my approach. e.g. https://github.com/01org/tpm2-tss/blob/master/include/sapi/tss2_tpm2_types.h

flihp:
Just posted to the related issue (#490). Gonna shut this down to keep the discussion in one place.

PR #493, opened 2017-08-04
implementation.h: Remove undef of TRUE and FALSE macros.

flihp:
This fixes a bug caused by 01e1e12965d747f5d95760baf931850e69294334
which removed the macro defining TRUE and FALSE but didn't remove the
preceding undef. This causes these macros to be undefined by the TSS2
headers even though they may have been defined by some other header
(like glib in the case of the tabrmd).


flihp:
I'm pushing this through ahead of other PRs to fix the build breakage that it causes in the tabrmd.

PR #494, opened 2017-08-04
Alg fixes

williamcroberts:
The alg defines didn't go as I had planned :-P
I have two patches, one to normalize SHA3 defines, and the other one to get rid of those silly casts. Those casts can actually hide errors BTW. I can squash these down if you care.


flihp:
Thanks for cleaning this up @williamcroberts. The spacing on those was killing me 😄

PR #497, opened 2017-08-09
TSS2_Sys_LoadExternal: Fix marshaling error

williamcroberts:
Commit 60dcc re-places the old marshaling code with the newer
marshaling code. This commit, introduced a breakagae in
tpm2_loadexternal tool when a private key was not specified.

In the case of a NULL private key, the Tss2_Sys_LoadExternal
call should set the size field of the Private Key to 0.

Fixes: #496

### DIFF
--- a/test/tpmclient/tpmclient.int.c
+++ b/test/tpmclient/tpmclient.int.c
@@ -3293,7 +3293,10 @@ void TestUnseal()
     INIT_SIMPLE_TPM2B_SIZE( name );
     rval = Tss2_Sys_LoadExternal ( sysContext, 0, 0, &outPublic,
             TPM_RH_PLATFORM, &loadedObjectHandle, &name, 0 );
-    CheckFailed( rval,  TSS2_TYPES_RC_BAD_REFERENCE);
+    CheckPassed( rval );
+
+    rval = Tss2_Sys_FlushContext( sysContext, loadedObjectHandle );
+    CheckPassed( rval );
 
     INIT_SIMPLE_TPM2B_SIZE( name );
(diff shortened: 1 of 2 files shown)


williamcroberts:
Per section 12.3.1 of [Part 3 Commands](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf), it states, "This command is used to load an object that is not a Protected Object into the TPM. The command allows loading of a public area or both a public and sensitive area."

williamcroberts:
WTF, now its the tpm unseal test failing:
```
	passing case:  	FAILED!  TPM Error: 0x902
 (TestUnseal@3301)
```

test/tpmclient/tpmclient.int.c
```diff
@@ -3293,7 +3293,7 @@ void TestUnseal()
     INIT_SIMPLE_TPM2B_SIZE( name );
     rval = Tss2_Sys_LoadExternal ( sysContext, 0, 0, &outPublic,
             TPM_RH_PLATFORM, &loadedObjectHandle, &name, 0 );
-    CheckFailed( rval,  TSS2_TYPES_RC_BAD_REFERENCE);
+    CheckPassed( rval );
```
tstruk: Do we need to call Tss2_Sys_FlushContext() after this?

williamcroberts:
description: out of memory for object contexts

test/tpmclient/tpmclient.int.c
williamcroberts: That would probably do it

PR #499, opened 2017-08-14
fix auto-generated pkgconfig files for libsapi

liuqun:
fix auto-generated pkgconfig files to allow us to install libsapi.a into a customized directory, e.g. /opt/xxx

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -243,6 +243,7 @@ TESTS_LDADD = $(noinst_LTLIBRARIES) $(lib_LTLIBRARIES)
 %.pc : %.pc.in
 	$(call make_parent_dir,$@)
 	sed -e "s,[@]VERSION[@],$(PACKAGE_VERSION),g; \
+	        s,[@]libdir[@],$(libdir),g; \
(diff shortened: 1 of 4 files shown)


liuqun:
the previous /opt/xxx/lib/pkgconfig/*.pc doesn't work for me, because:
1. "Cflags" was improperly set to sub-directories  "include/tcit/" and "include/sapi/"
    While the correct directory should be just the @includedir@ itself.
2. "Libs" should provide a gcc link option "-L@libdir@" in case we're going to install libsapi.a into a customized directory

See the original issue I've posted on tpm2-tools: https://github.com/01org/tpm2-tools/issues/402

flihp:
LGTM. Thanks @liuqun

liuqun:
Many 3rd party projects are still using `#include <tpm20.h>`
Since our previous changes on sapi.pc may break their code, I suggest we might just place another fake "tpm20.h" at the top level of "@@includedir@@"
For example:
```
/*
 * This "tpm20.h" is placed at top level for those 3rd party code who might use `#include <tpm20.h>` currently
 */
#include <sapi/tpm20.h>
```

PR #503, opened 2017-08-15
build: Switch travis from virtual machine to containers.

flihp:
This gets us a faster turn around time on each build:
https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments
Additionally it allows developers to use a local travis coantainer to
build / test locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
-language: cpp
+language: c
 compiler:
   - gcc
   - clang
 
-# This is a lie: we don't need sudo but this is required to get an
-# Ubuntu image with a libc that isn't ancient, and with cmocka libs.
-sudo: required
+sudo: false
 dist: trusty
 
 env:


.travis.yml
```diff
@@ -1,11 +1,11 @@
-language: cpp
+language: c
 compiler:
   - gcc
   - clang
 
 # This is a lie: we don't need sudo but this is required to get an
 # Ubuntu image with a libc that isn't ancient, and with cmocka libs.
```
tstruk: This comment can go now, right.

tstruk: We still need to install cmake to be able to build cmocka, but it looks like it works even with "sudo = false"

flihp: Yeah that comment was obsolete as soon as travis started supporting a docker & trusty environment. When we started using travis the trusty instance was VMs only. I'll strip this out before pushing. Good catch by the way :+1:

PR #504, opened 2017-08-15
build: Fold output from travis commands that generate a lot of output.

flihp:
This just cleans up the build output for us. This isn't really a
supproted feature but it has been stable and used by a number of
projects according to this thread:
https://github.com/travis-ci/travis-ci/issues/2285

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,14 @@ addons:
     branch_pattern: coverity_scan
 
 install:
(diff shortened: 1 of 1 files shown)


flihp:
I'm closing this since similar changes in the tabrmd travis-ci seem to be causing the build to hang. I haven't got to the root cause of that issue but If it's related I don't want to spread more pain and suffering on account of broken builds.

PR #505, opened 2017-08-16
Ptt capabilities

flihp:
The documentation for this feature is now up on the wiki: https://github.com/01org/tpm2-tss/wiki/PTT-GetCapability

This PR includes code provided to us by another Intel team with fixups by myself (required to get it to compile) as well as an additional commit I've added to prevent the unmarshalling function from ignoring properties when multiple are requested.

### DIFF
--- a/sysapi/sysapi_util/Unmarshal_TPML_INTEL_PTT_PROPERTY.c
+++ b/sysapi/sysapi_util/Unmarshal_TPML_INTEL_PTT_PROPERTY.c
@@ -33,6 +33,8 @@ void Unmarshal_TPML_INTEL_PTT_PROPERTY(
 	TPML_INTEL_PTT_PROPERTY *intelPttProperty
 )
 {
+	UINT32 i;
+
 	if (SYS_CONTEXT->rval != TSS2_RC_SUCCESS)
 		return;
 
@@ -40,7 +42,9 @@ void Unmarshal_TPML_INTEL_PTT_PROPERTY(
 		return;
 
 	Unmarshal_UINT32(SYS_CONTEXT->tpmOutBuffPtr, SYS_CONTEXT->maxResponseSize, &(SYS_CONTEXT->nextData), &intelPttProperty->count, &(SYS_CONTEXT->rval));
-	Unmarshal_UINT32(SYS_CONTEXT->tpmOutBuffPtr, SYS_CONTEXT->maxResponseSize, &(SYS_CONTEXT->nextData), &intelPttProperty->property, &(SYS_CONTEXT->rval));
-
+	for( i = 0; i < intelPttProperty->count; ++i )
+	{
+		Unmarshal_UINT32(SYS_CONTEXT->tpmOutBuffPtr, SYS_CONTEXT->maxResponseSize, &(SYS_CONTEXT->nextData), &intelPttProperty->property[i], &(SYS_CONTEXT->rval));
+	}
 	return;
 }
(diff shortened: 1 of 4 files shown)


sysapi/sysapi_util/Unmarshal_TPML_INTEL_PTT_PROPERTY.c
```diff
@@ -0,0 +1,50 @@
...
+* INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+* CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+* ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+* THE POSSIBILITY OF SUCH DAMAGE.
+***********************************************************************/
+
+#include <sapi/tpm20.h>
+#include "sysapi_util.h"
+
+void Unmarshal_TPML_INTEL_PTT_PROPERTY(
```
tstruk: Shouldn't this be called Unmarshal_TPML_VENDOR_PTT_PROPERTY?

sysapi/sysapi_util/Unmarshal_TPMU_CAPABILITIES.c
```diff
@@ -87,6 +87,11 @@ void Unmarshal_TPMU_CAPABILITIES(
 			Unmarshal_TPML_ECC_CURVE( sysContext, &capabilities->eccCurves );
 			break;
 #endif
+#ifdef TPM_CAP_VENDOR_PROPERTY
```
tstruk: Don't we need to #define the TPM_CAP_VENDOR_PROPERTY in tss2_tpm2_types.h?

sysapi/sysapi_util/Unmarshal_TPML_INTEL_PTT_PROPERTY.c
flihp: The type being unmarshalled is called `TPML_INTEL_PTT_PROPERTY`, the unmarshal function was named to match. Are you suggesting the name of the type be changed or just the name for the unmarshal function?

sysapi/sysapi_util/Unmarshal_TPMU_CAPABILITIES.c
flihp: it is: https://github.com/01org/tpm2-tss/blob/master/include/sapi/tss2_tpm2_types.h#L264

sysapi/sysapi_util/Unmarshal_TPML_INTEL_PTT_PROPERTY.c
tstruk: Just the function. Are we not going to reused to unmarshal other vendors PTT Capabilities?

sysapi/sysapi_util/Unmarshal_TPMU_CAPABILITIES.c
tstruk: Sorry, I missed that

flihp: It's kinda weird that it was already defined but it came straight from the table from the spec so the original author must have put it in the headers from completeness.

sysapi/sysapi_util/Unmarshal_TPML_INTEL_PTT_PROPERTY.c
flihp: Ideally we'd be able to use the same function for other TPM vendor capabilities but the spec doesn't define the type so this one is specific to Intel's PTT. It's very likely that a different TPM would use a different type, likely a tagged property unlike the raw UINT32 used by PTT.

So you're right: this commit only adds support for PTT and it probably won't work with other vendor specific properties. I figure we can make this more dynamic in the future if we get any requests for support from other vendors. No need to make it too complicated in the first iteration.

tstruk: fair enough

PR #508, opened 2017-08-16
build: Fix 'dist' and 'distcheck' target.

flihp:
There are a few small changes here but all are required to get proper
release tarballs made by dist / distcheck.

The project name in configure.ac needed to be updated to account for the
project name change from TPM2.0-TSS to tpm2-tss. Also the options passed
to the configure script need to be propagated to the distcheck target.
This is accomplished by setting the DISTCHECK_CONFIGURE_FLAGS when the
configure script generates the Makefile.

In Makefile.am we add all of the log files generated by the test harness
to CLEAN_FILES to ensure they don't end up in the distribution tarball.
We also add the integration test log compilation script to EXTRA_DIST to
ensure it makes it into the distribution. Finally we add the log.h file
to marshal_libmarshal_la_SOURCES as it won't make it into the release
tarball otherwise.

Finally we add explicit execution of the distcheck target to the
.travis.yml file to ensure future breakage is detected early.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ script:
   - mkdir ./build
   - pushd ./build
   - ../configure --enable-unit --with-simulatorbin=$(pwd)/../ibmtpm532/src/tpm_server
-  - make -j$(nproc)
+  - make -j$(nproc) distcheck
   - make -j$(nproc) check
   - |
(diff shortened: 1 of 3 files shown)


flihp:
this will need to be cherry-picked into the 1.x branch for use in the next release

PR #514, opened 2017-08-25
Static version string

flihp:


configure.ac
```diff
@@ -1,5 +1,8 @@
 AC_INIT([tpm2-tss],
-        [m4_esyscmd_s([git describe --tags --always --dirty])])
+        [2.0.0-dev],
```
williamcroberts: Are you talking the "source release"  via make tarball? Couldn't you  hook the build and have git generate a VERSION file?

williamcroberts: I guess this command fails... here during bootstrap?

williamcroberts: What if you did a git hook that updated a VERSION file with each change, this way you can just cat the version file here, and zips from github and tarballs from make will have the VERSION file.

flihp: Sorry I didn't give a whole lot of details as to the source of the issue. The use case is allowing people to regenerate the `configure` script from a source release. Currently the autoconf file depends on running a git command to generate the version number from the git tag. I'm sure you already figured this out but I wanted to get it here for completeness.

The static version string in this PR was the most simple solution, and guaranteed to work. I'm still not a big fan. The other approach that I think will work is to generate a `VERSION` file using the same `git` command in the `bootstrap` script. The `autoconf` file can then just grab the contents of this file and use it as the version string. This is effectively what we already do for the file lists.

I'm gonna merge this PR first and then give the `bootstrap` approach a try.

williamcroberts: Yeah that sounds good.

williamcroberts: The only think, is that won't work if someone gets the source via github (no .git directory). depends on whether or not you care. Maybe if the bootstrap fails you can just jam a string in version saying that it wasn't configured properly.

flihp: Is there a way to prevent github from producing the .zip files associated with each tag? I'm generating release tarballs using `distcheck` and uploading these as part of the release. They effectively obsolete the github source zip feature.

williamcroberts: I don't think so. Yesterday casually pursuing topics on this, a few projects closed out bugs saying that builds from the download feature of github are not supported and they need to use official releases or build from git.

flihp: Yeah that's annoying, bigly. With the static version string the .zip archives will work but I don't think the loss of convenience is worth supporting them. I think generating a `VERSION` file in `./bootstrap` is the right balance. I'll follow suite in the tabrmd as well.

PR #517, opened 2017-08-26
fix spelling of "synchronous" in a couple of places

d-e-s-o:
The word "synchronous" (and variations of it) is lacking an 'h' in a
couple of places. Fix them up.


flihp:
Good eye @d-e-s-o. Thanks!

PR #521, opened 2017-08-31
INSTALL.md: Add note on how to build from source in Fedora

martinezjavier:
The INSTALL.md file has instructions on how to build tpm2-tss from source
in other distros, add also a note on the steps needed to do it in Fedora.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -33,6 +33,15 @@ $ sudo apt -y install \
   autoconf
 ```
 
+## Fedora
+
+There is a package already, so the package build dependencies information can be
+used to make sure that the needed packages to compile from source are installed:
+
+```
+$ sudo dnf builddep tpm2-tss
+```
+
 # Building From Source
 ## Bootstrapping the Build
 To configure the tpm2.0-tss source code first run the bootstrap script, which


INSTALL.md
```diff
@@ -33,6 +33,15 @@ $ sudo apt -y install \
   autoconf
 ```
 
+## Fedora
+
+There is a package already, so the package build dependencies information can be
+used to make sure that the needed packages to compile from source are installed:
```
webmeister: Arch Linux also has [existing packages](https://aur.archlinux.org/packages/?K=tpm2). Maybe just add a list of all those (major) distros, where packages already exist that can be (re)used?

martinezjavier: @webmeister sounds like a good idea. I thought this file was more about building from source than installing from packages so that's why I added that command.

But I think that the "Ubuntu" section can be moved after the "Building From Source" section and instead an "Install From Packages" version could be added that contains commands to install packages for distros (since Ubuntu also has tpm2-* packages).

webmeister: There are two things that could be listed here, for each distribution:

1. The package names of all dependencies, that need to be installed, in order to build from source (this is what is currently present for Ubuntu).
2. The package name of the TSS package, so that this can be used to retrieve all dependencies without explicitly listing them (this is what you add for Fedora).

Maybe for point 1 it would be sufficient to have a generic list of projects/versions that this project depends on, letting the distros figure out how exactly those packages are called in their repositories? Then we'd only need point 2 for every distribution, unless no package exists yet in which case we could fall back to point 1.

If packages already exist for Ubuntu, then `apt-get build-dep` should be equivalent to your `dnf builddep`.

martinezjavier: @webmeister as you said, apt-get build-dep should be enough for Ubuntu, unless a new version of tpm2-tss needs a package for build that still is not a build dependency in the package due being updated.

So I would just remove the package list for Ubuntu and replace it for apt-get build-dep as you mentioned.

webmeister: Sounds good to me :)

flihp: Thanks for working through this. I like the approach. The ideal end state is end users consuming packages from their distro. Now that the majority of the distros out there do have packages using these to get dependencies installed is definitely the right way to go.

Issue #285, opened 2016-10-07
tpmclient tests don't pass when run against 1.36 simulator


wcarthur1:
In a couple cases, the simulator returns TPM_RC_RETRY.  This just means that the command needs to be retried.  I've fixed it on my local version of tpmclient.  When I get a chance, I'll create a pull request for this.

flihp:
resolved in #522

flihp:
whoops, forgot to close this when we merged #522

PR #522, opened 2017-09-04
Use latest simulator in test harness

flihp:
Not much to it: just need to retry when the RC says to.


flihp:
this resolves issue #285

flihp:
groan ... getting a test failure from clang now. Rebasing this a few times trying fixes.

flihp:
OK. That wasn't as bad as I thought it would be. Should be good to merge.

test/integration/sapi-util.h
```diff
@@ -35,6 +35,16 @@
 #define TSS2_APP_ERROR(base_rc)   (TSS2_APP_ERROR_LEVEL | base_rc)
 #define TSS2_APP_RC_BAD_REFERENCE  TSS2_APP_ERROR (TSS2_BASE_RC_BAD_REFERENCE)
 /*
+ */
+#define TSS2_RETRY_EXP(expression)                         \
+    ({                                                     \
+        TSS2_RC __result = 0;                              \
+        do {                                               \
+            __result = (expression);                       \
+        } while ((__result & 0x0000ffff) == TPM_RC_RETRY); \
```
williamcroberts: There's no predefined mask for getting out the retry code?

test/tpmclient/SetLocality.c
```diff
@@ -27,6 +27,7 @@
 
 #include "sapi/tpm20.h"
 #include "sample.h"
+#include "../integration/sapi-util.h"
```
williamcroberts: Why not set -I via the makefile?

flihp: I had a version that did that. I figured a relative path here wasn't much worse. The end goal is to get rid of this file regardless. Really it should be built into the utility library in the integration directory. Just haven't got around to it yet.

test/integration/sapi-util.h
flihp: I had expected to get beat up for the empty comment above the macro (didn't see this till I made the PR). And the headers we ship now don't have a mask for this. There's likely one in the next iteration of the specs working their way through TCG. If not we'll get one added. I've got a TODO for making an RC decoding library like `strerror` (that should have a function to extract the base RC) I just haven't had the cycles to whip up a prototype yet.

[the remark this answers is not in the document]
flihp: Good question. The only documentation for this RC is in part 2 and part 3 of the TPM spec. In both cases (part 2 table 16, part 3 table 4) the descriptive text is the same:

```
the TPM was not able to start the command
```
Not much to go on. But based on my experimentation with the simulator, the second try always succeeds so I figured this code was sufficient. Not sure it's appropriate but I modeled this after `TEMP_FAILURE_RETRY`. This relies on the `EINTR` errno which is a rare and temporary AFAIK. But again, this may not be appropriate and will largely depend on how TPM manufacturers use the RC, which may not be the same as the way the simulator uses it..

All that said, the patch adds the macro to a header in the test harness. It'll won't be installed on a users systems, it should only be consumed by the test cases / harness and so if one of the test cases ends up stuck in an infinite loop I'm not too concerned. If we move it to a public header I'd say we would want better data on how the RC is being used in the wild.

Issue #531, opened 2017-09-12
pack struct breaks abi


williamcroberts:
Commit https://github.com/01org/tpm2-tss/commit/82a2ff9cec3169d33e1ad4433b074d399002f1c0 breaks the abi between tss users and abrmd. Unless all users of the library add pack-struct, then the compiler emits different offsets/padding for the structures members as well as sizeof(struct).

Don't pack public structus shared. This notably hurts when calling from abrmd into the tcti stuffs that use the struct TCTI_SOCKET_CONF.

I'm getting tons of memory issues like this:
```
==16314== Memcheck, a memory error detector
==16314== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==16314== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==16314== Command: tpm2-abrmd --tcti=socket
==16314== 

** (tpm2-abrmd:16314): WARNING **: BILL CONFIG LOGCB: (nil)

==16314== Thread 2:
==16314== Conditional jump or move depends on uninitialised value(s)
==16314==    at 0x5FBE90E: vfprintf (vfprintf.c:1631)
==16314==    by 0x5FC5898: printf (printf.c:33)
==16314==    by 0x583E6AD: InitSocketTcti (tcti_socket.c:557)
==16314==    by 0x4177FD: tcti_socket_initialize (tcti-socket.c:191)
==16314==    by 0x412BC9: tcti_initialize (tcti.c:54)
==16314==    by 0x4056BD: init_thread_func (tabrmd.c:604)
==16314==    by 0x5AB2BB4: ??? (in /lib/x86_64-linux-gnu/libglib-2.0.so.0.4800.2)
==16314==    by 0x5D5A6B9: start_thread (pthread_create.c:333)
==16314==    by 0x60773DC: clone (clone.S:109)
``
Note that I have modified tss and abrmd with some print statements to see where the issue is.

Reverting this fixes my problem, revert coming upstream.

tstruk:
On 09/12/2017 12:04 PM, William Roberts wrote:
> Commit 82a2ff9 <https://github.com/01org/tpm2-tss/commit/82a2ff9cec3169d33e1ad4433b074d399002f1c0> breaks the abi between tss users and abrmd. Unless all users of the library add pack-struct, then the compiler emits different offsets/padding for the structures members as well as sizeof(struct).

Just created a PR to fix this:
https://github.com/01org/tpm2-tss/pull/532
[...]

PR #532, opened 2017-09-12
Add a proper structure packing

tstruk:


williamcroberts:
@tstruk Why does it even matter that they are packed, do you need them with no padding for serialization?

flihp:
If I'm understanding this correctly we can't support little endian systems w/o packing structures. The old implementation does what you suggest (no packing, with the serialization commands accounting for the size difference) but it won't work for LE systems.

flihp:
I should take a back seat here to @tstruk since he's much closer to the type marshaling library and I'm about to head to the airport  and will be out of comms range for a while.

tstruk:
@williamcroberts The problem without packing was the actual size of a type is different to what the spec specifies because the compiler sticks few bytes here and there. This causes serialization code building invalid commands as it mainly uses on sizeof() to jump between types and calculation of the offset in the command. The old code had a hack where some of the types were defined as 4 bytes instead of 2 or 1, but used Marshal_UINT8() to build commends anyway. This was only worked in LE systems but would likely not work on BE.

williamcroberts:
@tstruk sizeof compiler primitives is always the same on a given arch. ie sizeof(uint16_t) will always be 2.

sizeof(struct) can, and will vary based on padding, but the marshaling code shouldn't be using the sizeof a struct for anything, as you should be de-serializing at primitives.

williamcroberts:
I would restore #531 and abandon this. For now the pragma is supported. You shouldn't need to worry about structure padding if the serialization code is correct. This comes from Android code that runs on BE and LE systems that has to marshall data across a binder interface.

tstruk:
@williamcroberts completely agree with sizeof(uint_16), but take for instance a primitive TPMA_LOCALITY. 
What should the sizeof(TPMA_LOCALITY) be? According to the spec it is 1 byte, but before is was defined as 4 bytes and serialized with Mashal_UINT8(), which only worked by luck.

flihp:
We also have to deal with odities like TPMA_SESSION: https://github.com/01org/tpm2-tss/blob/master/include/sapi/tss2_tpm2_types.h#L592. I think the desire to provide a friendly interface to the field names we were set up for this packing issue since it ends up being a union.

Personally I'd advocate for simplifying this to the primitive type and macros for bit field selection.

flihp:
The situation I've identified above is just a variation on the one that @tstruk identified for the locality attribute type. Still I'm for getting rid of the union, use a straight UINT8 primitive, replace the bit field accessors with macros.

tstruk:
@williamcroberts also the issue is as bit more complicated. It hits us when we have more complex types like for instance TPMS_CREATION_DATA and we need to serialize this while converting LE -> BE.

Take for instance:
TPMS_ALG_PROPERTY:

struct TPMS_ALG_PROPERTY {
	TPM_ALG_ID	alg; /* UINT16 */
	TPMA_ALGORITHM	algProperties; /* UINT32 */
};

Gcc, without packing, adds 2 bytes of padding after the 'alg' so the sizeof TPMS_ALG_PROPERTY is really 8.
When we initialize it like this:

alg.alg = TPM_ALG_ECDSA;
alg.algProperties.asymmetric = 1;
alg.algProperties.signing = 1;

on LE we have this in the memory:

00 18 00 00 01 01 00 00, where 

00 18 - alg = TPM_ALG_ECDSA
00 00 - padding
01 01 00 00 - algProperties = TPMA_ALGORITHM_ASYMMETRIC | TPMA_ALGORITHM_SIGNING;

Then we serialize it, first goes alg: 00 18
then we increment the offset by alg - sizeof(UINT16) == 2 so it points to the padding.
Then we serialize algProperties, 4 bytes, we have '00 00 01 01' and in the buffer 
insted of 01 01 00 00.

flihp:
If having packed structures exposed in the headers is bad form then it sounds like we may need to re-think our strategy for serializing the types that follow this pattern. It's not as clean and definitely a pain but marshaling the members of this structure individually looks like it will resolve the issue?

tstruk:
On 09/12/2017 02:23 PM, Philip Tricca wrote:
> If having packed structures exposed in the headers is bad form then it sounds like we may need to re-think our strategy for serializing the types that follow this pattern. It's not as clean and definitely a pain but marshaling the members of this structure individually looks like it will resolve the issue?

This is what we had before. Do we want to go back to the old method?

tstruk:
On 09/12/2017 02:23 PM, Philip Tricca wrote:
> If having packed structures exposed in the headers is bad form then it sounds like we may need to re-think our strategy for serializing the types that follow this pattern. It's not as clean and definitely a pain but marshaling the members of this structure individually looks like it will resolve the issue?
> 

From the other hand having to pack structures is not uncommon.
That's why we gcc and others support attribute((packed)).
It is completely transparent to user who access struct members
in the usual way (ptr->member).

williamcroberts:
I'm assuming your serializing code is doing this:

```
BYTE *b = (BYTE *)&some_struct;
UINT8 a = unmarshal8(b);
b += sizeof(a);
UINT8 c = unmarshal8(b);
...
```
If that's the case, then padding **does** byte you in the ass (pun intended).

With that said, you can use offsetof to maneuver around padding, with the annoyance
of using field names.

Looking at TPMA_LOCALITY. wasn't sizeof screwy because it was unsigned int? I see now its unsigned char, thankfully on Linux sizeof(char) == 1, so it and the uninioned UINT8 val are ok.

Requiring structs packed via the API does put a hindrance with portability. I **have** had bugs in tool chains where attribute packed between versions changed (this was a long time ago when arm was new). 

I would just avoid the compiler extension and use offsetof(type, name) myself or go with attribute packed as proposed in this patchset.

tstruk:
On 09/12/2017 04:14 PM, William Roberts wrote:
> I'm assuming your serializing code is doing this:
> 
> |BYTE *b = (BYTE *)&some_struct; UINT8 a = unmarshal8(b); b += sizeof(a); UINT8 c = unmarshal8(b); ... |
> 
> If that's the case, then padding *does* byte you in the ass (pun intended).

More less so, with the difference that instead of some_struct we have uint8_t buffer[]
and instead of BYTE *b we have size_t *offset as specified by the TSS 2.0 Marshaling/Unmarshaling API Spec.
 
> Looking at TPMA_LOCALITY. wasn't sizeof screwy because it was unsigned int? I see now its unsigned char, thankfully on Linux sizeof(char) == 1, so it and the uninioned UINT8 val are ok.
> 
> Requiring structs packed via the API does put a hindrance with portability. I *have* had bugs in tool chains where attribute packed between versions changed (this was a long time ago when arm was new).

I'm not sure what you mean by "Requiring structs packed via the API". We don't require anything via API.
We just enforce the compiler to align structure members in memory the way we want.
This does not affect API users in any way.

> 
> I would just avoid the compiler extension and use offsetof(type, name) myself or go with attribute packed as proposed in this patchset.
> 
I thought about using offsetof, but it is a gcc extension and tss needs to be pure c99 so we can't use it.

martinezjavier:
@tstruk the Linux kernel has an [offset() implementation as a macro](http://elixir.free-electrons.com/linux/latest/source/include/linux/stddef.h#L18) for compilers that don't have a offsetof built-in as an extension.

```
#define offsetof(TYPE, MEMBER)  ((size_t)&((TYPE *)0)->MEMBER)
```

martinezjavier:
And in case that can't be used by the kernel being GPLv2 (I'm not sure if using this macro can be considered derivative work since there isn't too many ways to implement something like offsetof), the edk2 project (whose license is BSD 2-Clause) also [has one](https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Base.h#L801).

```
define OFFSET_OF(TYPE, Field) ((UINTN) &(((TYPE *)0)->Field))
```

williamcroberts:
@tstruk for packing s/API/ABI/g

@martinezjavier Most experienced C programmers know about casting a NULL pointer to get to size... In fact the tpm2-tools has a similar usage:
```#define BUFFER_SIZE(type, field) (sizeof((((type *)NULL)->t.field)))```

williamcroberts:
I’m ok with both, I would just leave attribute packed on the headers and understand that not all compilers will
support this. Windows is a land of unkown.

williamcroberts:
First off, offsetof is in stddef.h and part of ANSI C. Second, packing the struct is an extension.

https://en.wikipedia.org/wiki/Offsetof

tstruk:
@williamcroberts after giving it a second thought I think that you are right and we shouldn't need packing. So I'll close this and could you please reopen #533.

williamcroberts:
@tstruk there is a first time for everything (me being right)

tstruk:
Revert of commit 82aaff9 done in #533. No packing should be required

PR #533, opened 2017-09-12
Revert "configure: use -fpack-struct instead of pragmas"

williamcroberts:
This reverts commit 82a2ff9cec3169d33e1ad4433b074d399002f1c0.

Packing the public structures causes an ABI mismatch
between consumers of TSS public structures.

Fixes: #531


tstruk:
That should be fixed in this PR:
https://github.com/01org/tpm2-tss/pull/532

PR #543, opened 2017-09-16
configure: Add compiler flag to ignore int-in-bool-context errors fro…

flihp:
…m GCC 7.

The offending code has already been removed from the master branch.
We can't remove these identifiers from the 1.x branch w/o breaking
backward compatability. GCC 7 doesn't like how we use them though so
we've gotta ignore the warnings it spits out.


martinezjavier:
@flihp we are using the same flag though.

Maybe you could also use -Wno-int-in-bool-context instead of -Wno-error=int-in-bool-context?

I chose that flag since I didn't see the value in having the build warning for a known issue.

flihp:
Makes sense to me. I didn't know you could do that.

flihp:
updated PR to account for suggestion from @martinezjavier

PR #551, opened 2017-09-22
build: Cause warnings to be errors when checking for compiler flag su…

flihp:
…pport.

This was the original intent I just didn't understand how the macro works.
Turns out that warnings are ignored so if a compiler flag that's not supported
only causes a warning then it's considered "supported" and will be included
in CFLAGS with the macro as it was. But since we make all warnings errors
when we build, the unsupported flag causes a build failure. With this patch
causes unsupported, optional compiler flags to be dropped.


flihp:
This is a cherry-pick of the fix from the 1.x branch.

flihp:
I'm not sure about this. At build time if `-Werror,-Wall` effects the preprocessor then I'd say it should be used in the test here. I suppose it couldn't hurt either way. I'll give it a test first and if it works as expected I'll update the PR and push.

flihp:
Sorry, didn't realize this was for the cherry-pick to master, not the original commit. Since it's a cherry-pick we shouldn't do any re-writes so let's add this as a new commit after the merge.

PR #557, opened 2017-09-29
build: Cause warnings to be errors when checking for preprocessor fla…

flihp:
…g support.

This is a continuation of the work from ce8b05ed45c83064f559727527ef3ab9a56c8796.
It may be that the preprocessor will issue a warning for an unsupported flag
instead of an error. Since we build with -Werror & -Wall we must ensure that
all warnings are treated as errors when we're checking for supported flags.


flihp:
This suggestion came out of a previous discussion on #551. Considering the review by @tstruk there as an 'ack' here given how minor the change is.

Issue #540, opened 2017-09-13
Convert SAPI api to libmarshal


tstruk:
Remove all wrapper macros, and convert all sapi code the libmarshal

tstruk:
This has been resolved with #560

PR #560, opened 2017-10-03
Convert SAPI to the MU API

tstruk:
Start conversion with CreatePrimary


sysapi/sysapi/EncryptParam.c
```diff
@@ -93,9 +93,12 @@ TSS2_RC Tss2_Sys_SetEncryptParam(
...
             else
             {
-                CopyMem( (uint8_t *)currEncryptParamBuffer, encryptParamBuffer,
-                        encryptParamSize,
-                        SYS_CONTEXT->tpmInBuffPtr + SYS_CONTEXT->maxCommandSize );
+                if (currEncryptParamBuffer + encryptParamSize >
+                        SYS_CONTEXT->tpmInBuffPtr + SYS_CONTEXT->maxCommandSize)
+                    return TSS2_SYS_RC_INSUFFICIENT_CONTEXT;
+
+               memmove((UINT8 *)currEncryptParamBuffer,
```
flihp: You're casting away the `const` here. If I'm remembering memmove right this is the pointer to the location where the data is moved (the destination) so I'm wondering if the const is appropriate at all?

sysapi/sysapi/Tss2_Sys_CreatePrimary.c
```diff
@@ -30,117 +30,152 @@
...
-    {
-        rval = CommonOneCall( sysContext, cmdAuthsArray, rspAuthsArray );
+    rval = CommonOneCall(sysContext, cmdAuthsArray, rspAuthsArray);
 
-        if( rval == TSS2_RC_SUCCESS )
-        {
-            rval = Tss2_Sys_CreatePrimary_Complete( sysContext, objectHandle, outPublic, creationData, creationHash, creationTicket, name );
-        }
-    }
+    if (rval)
```
flihp: Your whitespace is a bit inconsistent in these calls. I prefer to have no whitespace between the function call and the check on the return value.

sysapi/sysapi/Tss2_Sys_CreatePrimary.c
```diff
@@ -30,117 +30,152 @@
...
+        return rval;
 
+    rval = Tss2_Sys_CreatePrimary_Complete(sysContext, objectHandle, outPublic,
+                                           creationData, creationHash,
+                                           creationTicket, name);
     return rval;
-}
 
+}
```
flihp: Whitespace between return and the closing parenthesis departs from the norm.

sysapi/sysapi_util/unmarshal_simple_tpm2b.c
```diff
@@ -29,67 +29,35 @@
...
+        return;
     }
-}
 
+    *rval = UINT16_Unmarshal(outBuffPtr, maxResponseSize, nextData, outTPM2B ? &outTPM2B->size : NULL);
+
+    if (*rval)
+        return;
+
+    for (i = 0; i < length; i++) {
```
flihp: This may be best left to another commit but these for loops that do byte by byte copy operations have always seemed wasteful to me. What do you think about just using a memcopy here? Looks to me like we've already got the size of all relevant data (data size, destination buffer size etc) to do so safely.

sysapi/sysapi_util/unmarshal_simple_tpm2b_no_size_check.c
```diff
@@ -28,60 +28,30 @@
...
-                {
-                    // Let low level function deal with NULL output pointer.
-                    Unmarshal_UINT16( outBuffPtr, maxResponseSize, nextData, 0, rval );
-                }
+    if (!outBuffPtr || !nextData) {
+        *rval = TSS2_SYS_RC_BAD_REFERENCE;
+        return;
+    }
+
+    length = BE_TO_HOST_16(*(UINT16 *)(outBuffPtr + *nextData));
```
flihp: What do you think about using the appropriate unmarshalling function here? That would get us the missing boundary check here for free.

sysapi/sysapi_util/unmarshal_simple_tpm2b.c
```diff
@@ -29,67 +29,35 @@
...
-                        if( outTPM2B != 0 )
-                        {
-                            Unmarshal_UINT8( outBuffPtr, maxResponseSize, nextData,  &( outTPM2B->buffer[i] ), rval );
-                        }
-                        else
-                        {
-                            // Let low level function deal with NULL output pointer.
-                            Unmarshal_UINT8( outBuffPtr, maxResponseSize, nextData, 0, rval );
-                        }
+    length = BE_TO_HOST_16(*(UINT16 *)(outBuffPtr + *nextData));
```
flihp: What do you think about using the appropriate unmarshalling function here? That would get us the missing boundary check for free.

sysapi/sysapi_util/unmarshal_simple_tpm2b_no_size_check.c
```diff
@@ -28,60 +28,30 @@
...
 
-                        if( *rval != TSS2_RC_SUCCESS )
-                        {
-                            break;
-                        }
-                    }
-                }
-            }
-        }
+    for (i = 0; i < length; i++) {
```
flihp: Same comment as above w/r to memcpy

sysapi/sysapi/EncryptParam.c
tstruk: This is because the Tss2_Sys_GetEncryptParam() expects const uint8_t ** so we have to cast somewhere. I'm not sure which place is better to cast memmove or Tss2_Sys_GetEncryptParam.
For now I will change the declaration to uint8_t *currEncryptParamBuffer

sysapi/sysapi/Tss2_Sys_CreatePrimary.c
tstruk: Do you mean blank line?

sysapi/sysapi/Tss2_Sys_CreatePrimary.c
tstruk: Blank lines removed

sysapi/sysapi_util/unmarshal_simple_tpm2b_no_size_check.c
tstruk: We could, but I'm planning to remove all the tree (un)marshal_simple_tpm2b(_no_size_check).c
files when the conversion is done, and use the appropriate TPM2B_<TYPE>_Marshal/TPM2B_<TYPE>_Umarshal functions, so this is only temporary anyway.

sysapi/sysapi_util/unmarshal_simple_tpm2b.c
tstruk: Right, but as I said, this is only temporary and will be replaced the the proper TPM2B_TYPE_Marshal calls.

sysapi/sysapi/Tss2_Sys_CreatePrimary.c
flihp: yes

sysapi/sysapi_util/unmarshal_simple_tpm2b_no_size_check.c
flihp: Understood. Comment withdrawn.

tstruk:
this resolves #540

Issue #567, opened 2017-10-10
Generate AUTHORS file from git history


flihp:
We haven't been great at maintaining the contents of this file. It should be something we generate from git history with a command like:
```
git log --format='%aN <%aE>' | sort -f | uniq
```

PR #568, opened 2017-10-10
build: Generate AUTHORS file from git history.

flihp:
Automating this is the only way to keep it up to date. The script generates
the list of authors from the git history while removing duplicates and
sorting the whole thing by frequency with the most frequent strings at the
top. It also filters out github 'noreply' email addresses.


flihp:
this resolves #567

PR #574, opened 2017-10-14
tcti: Repurpose TCTI_CONTEXT_INTEL macro as a cast macro and local va…

flihp:
…riable.

The TCTI_CONTEXT_INTEL macro was pretty unusual. It made assumptions about
the scope it was called from (variable name). These sort of things are
brittle and IMHO should be avoided. Replacing this with a macro that casts
the pointer from TSS2_TCTI_CONTEXT to a local instance of a
TSS2_TCTI_CONTEXT_INTEL pointer.

### DIFF
--- a/tcti/commonchecks.c
+++ b/tcti/commonchecks.c
@@ -1,5 +1,5 @@
 //**********************************************************************;
-// Copyright (c) 2015, Intel Corporation
+// Copyright (c) 2015, 2017 Intel Corporation
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
@@ -38,18 +38,20 @@ TSS2_RC CommonSendChecks(
     uint8_t           *command_buffer     /* in */
     )
 {
+    TSS2_TCTI_CONTEXT_INTEL *tcti_intel = tcti_context_intel_cast (tctiContext);
+
     if( tctiContext == NULL || command_buffer == NULL )
     {
         return TSS2_TCTI_RC_BAD_REFERENCE;
     }
 
-    if (TCTI_CONTEXT_INTEL->previousStage == TCTI_STAGE_SEND_COMMAND)
+    if (tcti_intel->previousStage == TCTI_STAGE_SEND_COMMAND)
     {
         return TSS2_TCTI_RC_BAD_SEQUENCE;
     }
 
-    if (TCTI_CONTEXT_INTEL->magic != TCTI_MAGIC ||
-        TCTI_CONTEXT_INTEL->version != TCTI_VERSION)
+    if (tcti_intel->magic != TCTI_MAGIC ||
+        tcti_intel->version != TCTI_VERSION)
     {
         return TSS2_TCTI_RC_BAD_CONTEXT;
     }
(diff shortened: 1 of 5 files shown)


tcti/tcti.h
```diff
@@ -53,7 +53,7 @@
 #define TCTI_LOG_DATA(ctx)     ((TSS2_TCTI_CONTEXT_INTEL*)ctx)->logData
 #define TCTI_LOG_BUFFER_CALLBACK(ctx) ((TSS2_TCTI_CONTEXT_INTEL*)ctx)->logBufferCallback
 #define TCTI_CONTEXT ((TSS2_TCTI_CONTEXT_COMMON_CURRENT *)(SYS_CONTEXT->tctiContext))
-#define TCTI_CONTEXT_INTEL ((TSS2_TCTI_CONTEXT_INTEL *)tctiContext)
+#define TCTI_CONTEXT_INTEL(ctx) ((TSS2_TCTI_CONTEXT_INTEL*)ctx)
```
williamcroberts: Why not use static inlines, same code as a macro *WITH* type checking.

flihp: Something like `static inline TSS2_TCTI_CONTEXT_INTEL* cast_function (TSS2_TCTI_CONTEXT* ctx)`

williamcroberts: Yep....

tstruk:
We should probably do the same with the SYS_CONTEXT in sapi

tcti/commonchecks.c
```diff
@@ -38,18 +38,20 @@ TSS2_RC CommonSendChecks(
...
-    if (TCTI_CONTEXT_INTEL->previousStage == TCTI_STAGE_SEND_COMMAND)
+    if (tcti_intel->previousStage == TCTI_STAGE_SEND_COMMAND)
     {
         return TSS2_TCTI_RC_BAD_SEQUENCE;
     }
 
-    if (TCTI_CONTEXT_INTEL->magic != TCTI_MAGIC ||
-        TCTI_CONTEXT_INTEL->version != TCTI_VERSION)
+    if (tcti_intel->magic != TCTI_MAGIC ||
+        tcti_intel->version != TCTI_VERSION)
```
williamcroberts: Call me crazy, but id move these MAGIC and VERSION checks to the cast routine....

williamcroberts:
```
ptr = cast(old)
if (!ptr) {
  return BAD_REF;
}
```

williamcroberts: in the cast routine:

```
type *cast (old) {
  new = (type *)old;
  if (!new || new->nagic != TCTI_MAGIC || new->VERSION != TCTI_VERSION) {
     return NULL;
  }
  return new;
}
```

flihp: even better that the current implementation checks the MAGIC and VERSION fields after it accesses private data in the context structure ... whoops. Should probably fix that too.

flihp: ugh I'm just now realizing why I didn't make this change in the first PR: the TCTI functions are still in one big `if` block. I really want to remove this before making any significant changes to the body of the functions.

@williamcroberts are you OK with merging this PR as-is so that I can restructure these functions before taking on the MAGIC & VERSION checks?

williamcroberts: Thats fine...

PR #577, opened 2017-10-17
test: flush the handlers after use

jiazhang0:
The unused handlers may exhaust the RAM of a real TPM hardware with
such a failure:

PASS: test/integration/asymmetric-encrypt-decrypt.int
PASS: test/integration/create-primary-rsa-2048-aes-128-cfb.int
PASS: test/integration/create-keyedhash-sha1-hmac.int
FAIL: test/integration/evict-ctrl.int

test/integration/sapi-util.c:82:create_primary_rsa_2048_aes_128_cfb():
CreatePrimary RSA 2048, AES 128 CFB
test/integration/sapi-util.c:100:create_primary_rsa_2048_aes_128_cfb():
CreatePrimary FAILED! Response Code : 0x902

It is fine to clean up the unused handlers before exit.

Signed-off-by: Jia Zhang <qianyue.zj@alibaba-inc.com>

### DIFF
--- a/test/integration/asymmetric-encrypt-decrypt.int.c
+++ b/test/integration/asymmetric-encrypt-decrypt.int.c
@@ -114,12 +114,16 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
     session_data.hmac.t.size = 0;
 
     rc = TSS2_RETRY_EXP (Tss2_Sys_Create(sapi_context, sym_handle, &sessions_data, &in_sensitive, &in_public, &outside_info, &creation_pcr, &out_private, &out_public, &creation_data, &creation_hash, &creation_ticket, &sessions_data_out));
-    if (rc != TPM_RC_SUCCESS)
+    if (rc != TPM_RC_SUCCESS) {
+        Tss2_Sys_FlushContext( sapi_context, sym_handle );
         print_fail("Create FAILED! Response Code : 0x%x", rc);
+    }
 
     rc = Tss2_Sys_Load(sapi_context, sym_handle, &sessions_data, &out_private, &out_public, &loaded_sym_handle, &name, &sessions_data_out);
-    if (rc != TPM_RC_SUCCESS)
+    if (rc != TPM_RC_SUCCESS) {
+        Tss2_Sys_FlushContext( sapi_context, sym_handle );
         print_fail("Load FAILED! Response Code : 0x%x", rc);
+    }
     print_log( "Loaded key handle:  %8.8x\n", loaded_sym_handle );
 
     input_message.t.size = strlen(message);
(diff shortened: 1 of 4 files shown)


martinezjavier:
@jiazhang0 as explained in the project's [README](https://github.com/01org/tpm2-tss/blob/master/README.md), the tests are not meant to be run on a real TPM chip.

They are only to exercise the TCTI and SAPI code and must be run only with the TPM simulator.

jiazhang0:
@martinezjavier I see that. But is it better to improve the test to adapt the real hardware? These fixes don't introduce any side effects and nobody will quickly recognize these changes are made for the real hardware without extra descriptions.

jiazhang0:
I meant the fixes themselves completes the testcases with a whole flow of creating, using and releasing. The real TPM hardware can also benefit from these changes by the way.

martinezjavier:
I'm not that familiar with the goal of the tpm2-tss testd to answer your question.

I just wanted to mention the caveat explained in the readme and point out that some tests may even be destructive for the TPM (i.e: the ones for the NV).

What I do think is that you need to explain a different motivation than the one mentioned in the commit message about a real TPM running of available space for transient objects.

For example, sayung that flushing the transient objects makes the tests to be a better code example for the SAPI sounds reasonable for me.

OTOH, most SAPI users would probably access the TPM through a resource manager so explicitly flushing objects would be less common.

webmeister:
I'd say it is good style for the tests to clean up after themselves.

> I just wanted to mention the caveat explained in the readme and point out that some tests may even be destructive for the TPM (i.e: the ones for the NV).

They might not be intended to be used that way, but if they can, all the better. If the tests just work everywhere, this is certainly easier than explaining to people all the time that no, they have not found a bug, they just shouldn't use that particular combination of tests and device.

To cause a hardware TPM to fail due to NV wearout, you'd need to run the tests for a long time, so I do not see a problem with running those tests occasionally during normal development.

martinezjavier:
@webmeister agreed, makes sense.

jiazhang0:
@flihp I would like to do that.

In addition, I plan to allow all tests, including tpm2-abrmd's, running on a real hardware. The prompted warning should be printed before the test. Currently, tpm2-tools has this goal. Not sure this idea is applicable to TPM2-TSS and tpm2-abrmd.

webmeister:
> In addition, I plan to allow all tests, including tpm2-abrmd's, running on a real hardware.

I like that. It gives us an easy way to have some indication that the tss/abrmd/tools projects and a specific TPM implementation are compatible with each other.

flihp:
> In addition, I plan to allow all tests, including tpm2-abrmd's, running on a real hardware. The prompted warning should be printed before the test. Currently, tpm2-tools has this goal. Not sure this idea is applicable to TPM2-TSS and tpm2-abrmd.

AFAIK neither the tss or the tabrmd actively prevent execution of the test applications against a real TPM. We just automate everything to run against the simulator and so the defaults are to use the simulator. I think you should be able to run any of the existing tests from the integration test harness against a physical TPM by setting the `TPM20TEST_TCTI_NAME` environment variable to the right string for the TCTI you want. I think "device" will work for what you're after.

The code that deals with all of the environment variables in the test harness is here: The integration tests include code that checks an environment variable to control which TCTI is used: https://github.com/01org/tpm2-tss/blob/master/test/integration/test-options.c. We don't have docs for this stuff but most / all of the plumbing is there I think.

jiazhang0:
@flihp I have did that. But I think we could have a configuration or commandline or something else to do that rather than manually modification. In addition, I still have big trouble on tpmclient. This integration test makes many assumption about being using simulator. Anyway, I will continue contributions to improve the testcases for real hardware use.

webmeister:
> AFAIK neither the tss or the tabrmd actively prevent execution of the test applications against a real TPM.

Yes, we've been running those tests against hardware TPMs for some time already, but last time I checked not all of them were successful (probably due to assuming a simulator backend). So I appreciate these efforts to make the tests run everywhere :)

flihp:
>  In addition, I still have big trouble on tpmclient. This integration test makes many assumption about being using simulator. Anyway, I will continue contributions to improve the testcases for real hardware use.

@jiazhang0 the `tpmclient` application is a huge mess. Work to clean it up would be much appreciated. Currently it has several places where it assumes the downstream TCTI is the simulator TCTI. Any place it's calling `PlatformCommand` is exactly this. I know @tstruk is doing some work currently in `tpmclient` so you all should coordinate as best you can.

jiazhang0:
@flihp This is the first work for the purpose of testcase improvement on a real TPM hardware.
https://github.com/01org/tpm2-abrmd/pull/223

I will then do the similar thing in TPM2-TSS.

flihp:
Similar work went into the tabrmd a while back: https://github.com/intel/tpm2-abrmd/pull/226. This PR needs to be updated to take a similar approach. Please re-open this PR when it's ready.

PR #579, opened 2017-10-19
sysapi: fix Tss2_Sys_GetCommandCode

tstruk:
The spec says: [1] "This function gets the command code for the command.
The command code is returned as an array of bytes in big endian order.
This array can be used for calculating the cpHash for a command or rpHash
for a response."

The problem is that our current implementation desn't return the command
code in big endian order, although it keeps it in big endian.
What happens is it is swapped three times:

First time in CommonPreparePrologue() in CommandUtil.c, line139:
SYS_CONTEXT->commandCodeSwapped = HOST_TO_BE_32(commandCode);

Then second time in Tss2_Sys_GetCommandCode()
*(UINT32 *)*commandCode = BE_TO_HOST_32(SYS_CONTEXT->commandCodeSwapped);

And the third time in tpmclient (TpmCalcPHash.c) in TpmCalcPHash():
*(UINT32 *)hashInputPtr = BE_TO_HOST_32(*(UINT32 *)cmdCodePtr);

This wrong. We should keep it in host byte order and only swap it once
to BE in Tss2_Sys_GetCommandCode(). This is fixed in this patch.

[1]. https://trustedcomputinggroup.org/wp-content/uploads/TSS_SAPI_v1.1_r21_Public_Review.pdf

### DIFF
--- a/sysapi/include/sysapi_util.h
+++ b/sysapi/include/sysapi_util.h
@@ -75,7 +75,7 @@ typedef struct {
     // These are set by system API and used by helper functions to calculate cpHash,
     // rpHash, and for auditing.
     //
-    TPM_CC commandCodeSwapped;
+    TPM_CC commandCode;
     UINT32 cpBufferUsedSize;
     UINT8 *cpBuffer;
     UINT32 *rspParamsSize;  // Points to response paramsSize.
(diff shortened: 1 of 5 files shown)


tstruk:
Thanks for comments. Yes, changed the prototype to what it was and got the fix in a new updated commit. Will push it as soon as the travis ci passes.

tstruk:
That should be good to go now.

PR #580, opened 2017-10-19
remove duplicate access ptrs from sys context struct

tstruk:
The context command/response buffer is accessed by
SYS_CONTEXT->tpmInBuffPtr/SYS_CONTEXT->maxCommandSize
SYS_CONTEXT->tpmOutBuffPtr/SYS_CONTEXT->maxResponseSize
pointers/limiters for tpm commands and tpm responses respectively.
Even though they are separate, they point to the same buffer
and have the same size values always. They are also used quite loosely,
and the tpmInBuffPtr seems to be used most of the times to
access both commands and responses.
This patch replaces them with a single pairtpm commands and
SYS_CONTEXT->cmdBuffer/SYS_CONTEXT->maxCmdSize to handle both
tpm commands and tpm responses.

### DIFF
--- a/sysapi/include/sysapi_util.h
+++ b/sysapi/include/sysapi_util.h
@@ -58,17 +58,9 @@ typedef struct _TPM20_ErrorResponse {
 #pragma pack(pop)
 
 typedef struct {
-    //
-    // These are inputs to system API functions.
-    //
     TSS2_TCTI_CONTEXT *tctiContext;
-
-    // In and out buffers can be the same for a minimalized memory footprint implementation.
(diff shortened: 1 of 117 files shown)


flihp:
I just did a skim over this and I'm 100% on-board with this conceptually. Pretty hard to review a ~1000 line patch though. Since this was likely a `sed` find / replace job I'm happy to approve the patch. Looks like it needs to be rebased though.

tstruk:
Yes, mostly sed s/// job. I should have mentioned that only the sysapi_util.h header file and the CommandUtil.c should be reviewed. This has been rebased now. Thanks

PR #582, opened 2017-10-20
Tcti cleanup

flihp:
This is a small collection of patches that start to clean up the TCTI implementation. There should be no functional changes, only reorganization, reformatting and a 1 or 2 new functions.


flihp:
This came from a suggestion by @williamcroberts in #574. The check for the MAGIC value works best in the `tcti_checks` function and not the cast though given the current code structure. It's not quite what @williamcroberts suggested but I think it's pretty close.

tcti/tcti.c
```diff
@@ -33,43 +33,62 @@
...
+TSS2_RC tcti_common_checks (
+    TSS2_TCTI_CONTEXT *tcti_context
+    )
+{
+    TSS2_TCTI_CONTEXT_INTEL *tcti_intel;
+
+    tcti_intel = tcti_context_intel_cast (tcti_context);
+    if (tcti_context == NULL) {
+        return TSS2_TCTI_RC_BAD_REFERENCE;
+    } else if (tcti_intel->magic != TCTI_MAGIC ||
```
tstruk: else not really needed here

tcti/tcti_socket.c
```diff
@@ -271,17 +273,17 @@ TSS2_RC SocketSetLocality(
...
-    } else if (tcti_intel->status.locality != locality) {
-        if (tcti_intel->status.commandSent == 1) {
-            return TSS2_TCTI_RC_BAD_SEQUENCE;
-        } else {
-            tcti_intel->status.locality = locality;
-        }
+    rc = tcti_common_checks (tctiContext);
+    if (rc != TSS2_RC_SUCCESS) {
+        return rc;
+    } else if (tcti_intel->status.commandSent == 1) {
```
tstruk: else not needed here

tcti/tcti.c
```diff
@@ -33,43 +33,62 @@
...
-    TSS2_TCTI_CONTEXT_INTEL *tcti_intel = tcti_context_intel_cast (tctiContext);
+    TSS2_TCTI_CONTEXT_INTEL *tcti_intel;
+    TSS2_RC rc;
 
-    if (tctiContext == NULL || command_buffer == NULL) {
+    tcti_intel = tcti_context_intel_cast (tctiContext);;
+    rc = tcti_common_checks (tctiContext);
+    if (rc != TSS2_RC_SUCCESS) {
+        return rc;
+    } else if (command_buffer == NULL) {
```
tstruk: same here

tcti/tcti.c
```diff
@@ -33,43 +33,62 @@
...
-    TSS2_TCTI_CONTEXT_INTEL *tcti_intel = tcti_context_intel_cast (tctiContext);
+    TSS2_TCTI_CONTEXT_INTEL *tcti_intel;
+    TSS2_RC rc;
 
-    if (tctiContext == NULL || response_size == NULL) {
+    tcti_intel = tcti_context_intel_cast (tctiContext);
+    rc = tcti_common_checks (tctiContext);
+    if (rc != TSS2_RC_SUCCESS) {
+        return rc;
+    } else if (response_buffer == NULL || response_size == NULL) {
```
tstruk: and here

flihp:
I played around with doing this both ways. The use of `else if` is an attempt to give the compiler a hint that the second conditional will only ever be tested if the first fails. I'm guessing that the compiler is smart enough to figure this out though so I'm with you. Will update and push.

Issue #523, opened 2017-09-06
CompareTPM2B routine in tpmclient.int.cpp only compares first byte over and over again


wcarthur1:
Not much more I can say about this.  Good news is that this wasn't masking any failures in tpmclient.

tstruk:
This is fixed by #583

PR #583, opened 2017-10-20
tpmclient: fix CompareTPM2B

tstruk:
The current CompareTPM2B only compares the first byte.


tstruk:
This fixes  #523

Issue #515, opened 2017-08-25
Port PTT capabilities stuff to master


flihp:
This was added to the 1.x branch to make it into the next 1.x release. To make it into the 2.x releases (when they happen) we need to port it over to use libmarshal.

tstruk:
Closed with #591

PR #591, opened 2017-10-23
Add support for Intel PTT capability query

tstruk:
Add support for Intel PTT specific capability query.
Ported from 1.x branch PR: #505 "Ptt capabilities"


tstruk:
This solves #515

PR #600, opened 2017-11-02
TPM2B simplification

tstruk:
Remove unions.

### DIFF
--- a/include/sapi/tpmb.h
+++ b/include/sapi/tpmb.h
@@ -39,21 +39,14 @@ typedef struct {
 } TPM2B;
 
 #define TPM2B_TYPE1(name, bytes, bufferName)  \
-    typedef union {                           \
-    struct  {                                 \
+    typedef struct {                          \
         UINT16  size;                         \
         BYTE    bufferName[bytes];            \
-    } t;                                      \
-    TPM2B   b;                                \
     } TPM2B_##name
 
 #define TPM2B_TYPE2(name, type, bufferName)   \
-    typedef union {                           \
-    struct  {                                 \
+    typedef struct {                          \
         UINT16  size;                         \
         type bufferName;                      \
-    } t;                                      \
-    TPM2B   b;                                \
-    } TPM2B_##name
-
+    } TPM2B_##name;
 #endif
--- a/marshal/tpm2b-types.c
+++ b/marshal/tpm2b-types.c
@@ -53,18 +53,18 @@ TSS2_RC Tss2_MU_##type##_Marshal(type const *src, uint8_t buffer[], \
         LOG (WARNING, "buffer and offset parameter are NULL"); \
         return TSS2_TYPES_RC_BAD_REFERENCE; \
     } else if (buffer == NULL && offset != NULL) { \
-        *offset += sizeof(src->t.size) + src->t.size; \
+        *offset += sizeof(src->size) + src->size; \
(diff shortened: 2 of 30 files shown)


gwei3:
Why do we need to do this? It will impact all apps using TSS2.

flihp:
The whole `b` and `t` thing in our structures didn't survive the standardization effort. These fields are pretty awkward when writing code and accessing the contents of the structure through the `.t` member is always the right thing to do. So in the spec that we're using to standardize the TSS2 headers these fields are gone: https://trustedcomputinggroup.org/wp-content/uploads/TCG-TSS-2.0-Overview-and-Common-Structures-Specification-Version-0.90-Revision-02.pdf

There are a number of changes that will end up defining our 2.0 release that break API/ABI like this one. unfortunately our early adopters will suffer a bit on account of it. I think accounting for this change should be as easy as running a sed script over your code: `sed -i 's/\.[bt]\./\./g' $(SRCDIR)/.c`?

That said we should probably announce the change on the mailing list to give everyone a chance to prepare.

flihp:
@tstruk : we should probably hold off on merging this until @williamcroberts has the next release branch for the tools set up and we understand the impact this change will have on the tools (and for the tabrmd too I guess). My untested `sed` script above won't be sufficient for a lot of cases since it doesn't handle C designated initializers and other cases that span multiple lines. Also see above note about announcing this change on the mailing list. Giving people time to "brace for impact" is probably the right thing to do here.

tstruk:
@flihp Sure, I can put it on hold until everybody are ready.
FYI I have patches for tools and abrmd for this:
https://github.com/tstruk/tpm2.0-tools/commit/754ad8cab4cc7abddede8ab7db22f2948a50f8be
https://github.com/tstruk/tpm2-abrmd/commit/f1a9118007213ea4e685f58923787814282fe5c3

williamcroberts:
Very happy with this, ill branch 3.X now so master can move ahead.

williamcroberts:
3.X branched from master: https://github.com/01org/tpm2-tools/tree/3.X

Go ahead and submit anything, 3.X may break until I fix it up to a fixed version of tss, don't worry about that, just keep my master working.

martinezjavier:
> 3.X branched from master: https://github.com/01org/tpm2-tools/tree/3.X

Very happy to see the 3.X already branched, one step closer to a release :smiley: 

> Go ahead and submit anything, 3.X may break until I fix it up to a fixed version of tss, don't worry about that, just keep my master working.

Is still the plan to allow the 3.X release to be compatible with the tpm2-tss 1.x branch? If so I could help next week to make sure that happen (my paternity leave ends on Monday so I should have more time again to dedicate to the tpm2 projects).

williamcroberts:
@martinezjavier yes, we're going to be targeting v1.3 rc0 is available here: https://github.com/01org/tpm2-tss/releases/tag/1.3.0-rc0

martinezjavier:
@williamcroberts great, thanks a lot for the confirmation.

gwei3:
Sounds good, thanks for the explanation and actions. @flihp .

flihp:
I think we're ready for this one to be merged. @tstruk : you said you have patches / PRs ready for the tools and tabrmd. Are you ready to queue those up?

tstruk:
@flihp yes, I will create PR for tools and rm today.

Issue #608, opened 2017-11-09
test unit: Failed to build with libcmocka-dev<=0.4.1 on Debian 8.0(Jessie)


liuqun:
https://github.com/intel/tpm2-tss/blob/fb93568596419c4ac20031db898963abb0aca0ec/configure.ac#L21-L25

Failed to build tpm2-tss with libcmocka-dev<=0.4.1 on Debian 8.0 or older.
Old release of libcmocka-dev<=0.4.1 on Debian 8.0 do not support "CMUnitTest" or "cmocka_unit_test".
In the old release of /usr/inclue/cmocka.h, "unit_test" was defined without prefix "cmocka_" and 
"UnitTest" was defined without prefix "CM":

```
typedef struct UnitTest {
    const char* name;
    UnitTestFunction function;
    UnitTestFunctionType function_type;
} UnitTest;

/** Initializes a UnitTest structure. */
#define unit_test(f) { #f, f, UNIT_TEST_FUNCTION_TYPE_TEST }
```

```
struct CMUnitTest {
    const char *name;
    CMUnitTestFunction test_func;
    CMFixtureFunction setup_func;
    CMFixtureFunction teardown_func;
    void *initial_state;
};

/** Initializes a CMUnitTest structure. */
#define cmocka_unit_test(f) { #f, f, NULL, NULL, NULL }
```

```
$ cat /etc/debian_version 
8.0
$ apt show libcmocka-dev
Package: libcmocka-dev
Source: cmocka
Version: 0.4.1-2
Maintainer: David Prévot <taffit@debian.org>
Installed-Size: 62.5 kB
Depends: libcmocka0 (= 0.4.1-2)
Recommends: cmocka-doc
Homepage: http://cmocka.org/
Priority: extra
Section: libdevel
Download-Size: 13.0 kB
APT-Manual-Installed: yes
APT-Sources: http://mirrordirector.raspbian.org/raspbian/ jessie/main armhf Packages
Description: development files for the CMocka unit testing framework
 This package contains the library development files for CMocka.
 .
 The idea of CMocka is that a test application only requires the
 standard C library and the CMocka library itself to minimize conflicts
 with standard C headers especially on a lot of different platforms.
 .
 Features:
  * Easy to use framework to write unit tests
  * Support for mock objects
  * Testing of memory leaks, buffer overflows and underflows
  * Platform and compiler independent
```

[...]

PR #611, opened 2017-11-09
build: Check for cmocka version >= 1.0 @ configure time.

flihp:
This is to keep developers with an old version of cmocka from having to
deal with compile time errors when we can catch this at configure time.


flihp:
This issue was identified in #608. This commit should be cherry-picked into the 1.x branch once it's merged.

PR #614, opened 2017-11-10
implementation.h: drop the casts and parens on Algs

williamcroberts:
The TPM_ALG_XXX defines cast the values as TPM_ALG_ID types.
There is really no need for this additional layer.

This cast, actually causes compilers to ignore the case when
things are oversized. For example, this code will cause no
overflow warning:

uint16_t x = (uint16_t)0x111111111111;

Where as this code will:
uint16_t x = 0x111111111111;

warning: large integer implicitly truncated to unsigned type [-Woverflow]

### DIFF
--- a/include/sapi/implementation.h
+++ b/include/sapi/implementation.h
@@ -322,149 +322,149 @@
 // From TCG Algorithm Registry: Table 2 - Definition of TPM_ALG_ID Constants
 
 typedef  UINT16             TPM_ALG_ID;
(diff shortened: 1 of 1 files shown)


williamcroberts:
@flihp can I get this in 1.3.0?

flihp:
I'm taking @tstruk approving this review for the patch against the 1.x branch as approval to include this in the next release. I don't have any objections either. Merging, will get another RC out today when I do an RC0 for the tabrmd.

PR #617, opened 2017-11-13
sapi: remove SYS_CONTEXT macro

tstruk:
The SYS_CONTEXT macro, which assumed that a local variable named
"sysContext" will be valid in a given block of code, is replaced
by syscontext_cast() inline function.
This should prevent errors caused by local variable eclipse etc.
The external Tss2_Sys_*() functions take the same TSS2_SYS_CONTEXT * type
and all internal helpers pass _TSS2_SYS_CONTEXT_BLOB *.
The syscontext_cast() does the conversion between the two types.

### DIFF
--- a/sysapi/include/sysapi_util.h
+++ b/sysapi/include/sysapi_util.h
@@ -100,9 +100,25 @@ typedef struct {
     size_t nextData;
 } _TSS2_SYS_CONTEXT_BLOB;
 
-#define SYS_CONTEXT ((_TSS2_SYS_CONTEXT_BLOB *)sysContext)
-#define SYS_RESP_HEADER ((TPM20_Header_Out *)(SYS_CONTEXT->cmdBuffer))
-#define SYS_REQ_HEADER ((TPM20_Header_In *)(SYS_CONTEXT->cmdBuffer))
+struct TSS2_SYS_CONTEXT;
+
+static inline _TSS2_SYS_CONTEXT_BLOB *
+syscontext_cast(TSS2_SYS_CONTEXT *ctx)
+{
+    return (_TSS2_SYS_CONTEXT_BLOB*) ctx;
+}
+
+static inline TPM20_Header_Out *
+resp_header_from_cxt(_TSS2_SYS_CONTEXT_BLOB *ctx)
+{
+    return (TPM20_Header_Out *)ctx->cmdBuffer;
+}
+
+static inline TPM20_Header_In *
+req_header_from_cxt(_TSS2_SYS_CONTEXT_BLOB *ctx)
+{
+    return (TPM20_Header_In *)ctx->cmdBuffer;
+}
 
 typedef struct {
     TPM2_CC commandCode;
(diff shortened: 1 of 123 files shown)


tstruk:
This is a fix for https://github.com/intel/tpm2-tss/issues/578

sysapi/include/sysapi_util.h
```diff
@@ -103,9 +103,16 @@ typedef struct {
...
+struct TSS2_SYS_CONTEXT;
+
+static inline _TSS2_SYS_CONTEXT_BLOB *
+syscontext_cast(TSS2_SYS_CONTEXT *ctx)
+{
+    return (_TSS2_SYS_CONTEXT_BLOB*) ctx;
+}
+
+#define SYS_RESP_HEADER(ctx) ((TPM20_Header_Out *)(ctx->cmdBuffer))
+#define SYS_REQ_HEADER(ctx) ((TPM20_Header_In *)(ctx->cmdBuffer))
```
williamcroberts: Why not make these static inline? Same performance as a macro *with* type checking.

sysapi/include/sysapi_util.h
```diff
@@ -117,52 +124,40 @@ typedef struct {
...
-TSS2_RC CommonComplete( TSS2_SYS_CONTEXT *sysContext );
+TPM_RC ConcatSizedByteBuffer(TPM2B_MAX_BUFFER *result, TPM2B *addBuffer);
+void InitSysContextFields(_TSS2_SYS_CONTEXT_BLOB *ctx);
+void InitSysContextPtrs(_TSS2_SYS_CONTEXT_BLOB *ctx, size_t contextSize);
+TSS2_RC CompleteChecks(_TSS2_SYS_CONTEXT_BLOB *ctx);
+TSS2_RC CommonComplete(_TSS2_SYS_CONTEXT_BLOB *ctx);
 
 TSS2_RC  CommonOneCallForNoResponseCmds(
-    TSS2_SYS_CONTEXT *sysContext,
+    _TSS2_SYS_CONTEXT_BLOB *ctx,
```
williamcroberts: Is it possible to flipp the paradigm here and left everything coming into the routines as a TSS2_SYS_CONTEXT pointer and casted them to something else, perhaps a TSS2_INTERNAL_SYS_CONTEXT pointer and just renamed the structures?
 I'm not sure if this is possible due to some API breakage.

sysapi/include/sysapi_util.h
tstruk: You are right. These should be static inlines as well. I think I just focused on SYS_CONTEXT too much.

sysapi/include/sysapi_util.h
tstruk: All the external Tss2_Sys_*() functions take TSS2_SYS_CONTEXT * opaque as per the spec.
We cast it too the _TSS2_SYS_CONTEXT_BLOB *, which is used internally.
We should not expose the internal/private _TSS2_SYS_CONTEXT_BLOB.
Is this what you had in mind?

tstruk:
@williamcroberts Uploaded a new version. Could you have a look please. Thanks.

sysapi/include/sysapi_util.h
williamcroberts: I think it boils down to the use of _ prefix to denote the internal cast of the opaque structure. But that appears to already be in place.

tstruk:
Rebased on top of https://github.com/intel/tpm2-tss/pull/624 @flihp please have a look.

PR #625, opened 2017-11-17
SAPI: Add _complete calls for all TPM functions.

AndreasFuchsSIT:
Previously TPM functionality that does not have return
parameters did not have a SAPI _Complete call associated.
This changes with the most recent spec.

This patch adds according _Complete calls that were
missing.

### DIFF
--- a/sysapi/include/sysapi_util.h
+++ b/sysapi/include/sysapi_util.h
@@ -145,11 +145,6 @@ void InitSysContextPtrs(_TSS2_SYS_CONTEXT_BLOB *ctx, size_t contextSize);
 TSS2_RC CompleteChecks(_TSS2_SYS_CONTEXT_BLOB *ctx);
 TSS2_RC CommonComplete(_TSS2_SYS_CONTEXT_BLOB *ctx);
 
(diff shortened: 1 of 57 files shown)


AndreasFuchsSIT:
Ok, sorry for the problems... But I cannot recreate the failure...
I have clang version 3.8.0-2ubuntu4 which runs fine...
Travis runs clang version 3.9.0...

Error is: FAIL: test/integration/self-test.int

Which does not help me much, and I cannot recreate...
Will try to investigate further next week, but if anybody has an idea, let me know...

AndreasFuchsSIT:
Funny enough, the make distcheck run on travis gives a
PASS: test/integration/self-test.int

I'll see what I find...

flihp:
I think there's a race in the test harness. It's likely in the bit that starts the simulator. The port selected may already be in use or something like that.

AndreasFuchsSIT:
Awesome.. Thanks for the help...
Looking forward to the merge... :+1:

flihp:
#617 from @tstruk went in before this. Caused a pile of conflicts :(

AndreasFuchsSIT:
Reworked to fit on top of latest master...

PR #626, opened 2017-11-20
tcti: Fixed tcti function macros to use CamelCase

AndreasFuchsSIT:
As per spec, we need to use CamelCase for tcti function macros.

The two patches change this for the header macros and for the tests.
It however leaves the lower-case version for compatibility until tools and resmgr are also updated.

This patchset applies ontop of #625 which is the reason for the third commit. Just ignore it.


AndreasFuchsSIT:
The first commit will NOT cause the build to fail, since we added defines mapping the old names to the new ones. Also in order to give ResMgr and tpm-tools time for transition.

Please reassess...

tstruk:
Right, sorry I missed the wrappers. Still, instead of adding wrappers I think we should just squeeze the two commits into one.

AndreasFuchsSIT:
Updated and pushed...

PR #627, opened 2017-11-20
tpm-types: Switch TPMAs from structs to defines

AndreasFuchsSIT:
Per spec this changes TPMAs to be used as flags with defines instead of bitfields throughout the project.
For the moment, applications can -DTPM_BITFIELD_LE to use the old behavior.
TODO: Once they are updated the bitfield structs need to be removed from the headers.

Note: This PR is actually just one patch. It is based off of #626

### DIFF
--- a/test/integration/asymmetric-encrypt-decrypt.int.c
+++ b/test/integration/asymmetric-encrypt-decrypt.int.c
@@ -62,12 +62,12 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
     in_public.publicArea.type = TPM2_ALG_RSA;
     in_public.publicArea.nameAlg = TPM2_ALG_SHA256;
     *(UINT32 *)&(in_public.publicArea.objectAttributes) = 0;
-    in_public.publicArea.objectAttributes.restricted = 1;
-    in_public.publicArea.objectAttributes.userWithAuth = 1;
-    in_public.publicArea.objectAttributes.decrypt = 1;
-    in_public.publicArea.objectAttributes.fixedTPM = 1;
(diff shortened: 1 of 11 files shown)


tstruk:
You probably want to create https://github.com/intel/tpm2-tss/pull/627/commits/94bd389a0a5484aa50fe349672dc203c72cd4249 on a separate branch to keep the commits from different PRs separate.

AndreasFuchsSIT:
Yes, we should definitively do so.
I kept them for the moment, until ResMgr and tpm-tools are known to not require them anymore.
In the end, we should only have the non-struct/-union versions left over...

AndreasFuchsSIT:
Updated to be 1 commit only...

test/integration/asymmetric-encrypt-decrypt.int.c
```diff
@@ -100,13 +100,12 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
 
     // First clear attributes bit field.
     *(UINT32 *)&(in_public.publicArea.objectAttributes) = 0;
-    in_public.publicArea.objectAttributes.restricted = 0;
-    in_public.publicArea.objectAttributes.userWithAuth = 1;
-    in_public.publicArea.objectAttributes.decrypt = 1;
-    in_public.publicArea.objectAttributes.sign = 1;
```
martinezjavier: You missed a `in_public.publicArea.objectAttributes |= TPMA_OBJECT_SIGN` here.

AndreasFuchsSIT: Awesome catch... Thanks, fixed...

PR #629, opened 2017-11-20
1.x coverity fixups

flihp:


flihp:
If it's easier on the eyes we can always do the check in the loop after the call to the marshalling function and break out of the loop if there's an RC in rval `!= TSS2_RC_SUCCESS`. I try to avoid `break` if possible though so I went with using the loop terminating condition.

tstruk:
We could just return, but thinking about it now - are we not supposed to reset the SYS_CONTEXT->nextData to what in was at the fn entry in case of an error in the middle of (un)marshaling? The new MU spec requires the implementation to do so.

flihp:
Like you say, there's nothing that requires we do so until the libmu code. I'd say we should minimize the changes we make here to keep the behavior consistent. All of this will be fixed in 2.0.

tstruk:
ok, so that should do it.

PR #631, opened 2017-11-21
move CheckPassed and CheckFailed to integration folder 

furtnerthomas:
move CheckPassed and CheckFailed to integration folder in order to make it available for integration tests

### DIFF
--- a/test/integration/test.h
+++ b/test/integration/test.h
@@ -1,5 +1,35 @@
 #include "sapi/tpm20.h"
 
+#define CheckPassed(rval) {				\
+    							\
+    DebugPrintf( NO_PREFIX, "\tpassing case:  " );			\
+    if ( rval != TPM2_RC_SUCCESS) {					\
+      ErrorHandler( rval);						\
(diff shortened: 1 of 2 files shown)


furtnerthomas:
Just saw that moving CheckPassed to test.h is not sufficient in order to use it e.g. in get-random.int.c:

In file included from test/integration/get-random.int.c:3:0:
test/integration/get-random.int.c: In function ‘test_invoke’:
test/integration/test.h:5:5: error: implicit declaration of function ‘DebugPrintf’ [-Werror=implicit-function-declaration]
     DebugPrintf( NO_PREFIX, "\tpassing case:  " );   \
     ^

Should I move all test helper functions like "DebugPrintf()" to test.c?

tstruk:
The CheckPassed and CheckFailed are only used in tpm_client and they eventually will be removed after we clean it up. The integ tests do simple if (rc != TSS2_RC_SUCCESS) and print_fail() in case of  failure.

PR #632, opened 2017-11-22
Add support for code-coverage

AndreasFuchsSIT:
Add support for code coverage measurement.

Also a minor tweak to build check_PROGRAMS independently running them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,8 @@ script:
   - pushd ./build
   - ../configure --enable-unit --with-simulatorbin=$(pwd)/../ibmtpm/src/tpm_server
   - make -j$(nproc) distcheck
-  - make -j$(nproc) check
+  - make -j$(nproc) check-programs
+  - make -j1 check
   - cat test-suite.log
   - |
     for LOG in $(ls -1 test/unit/*.log); do
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,10 @@ install:
   - make
   - make install
   - cd ../../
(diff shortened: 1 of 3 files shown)


AndreasFuchsSIT:
Ok, so travis-ci does not support this...
Under Ubuntu ax_code_coverage.m4 is part of autoconf-archive

flihp:
It's not pretty, but downloading an autoconf-archive release tarball (much in the same way we do the tpm2 simulator) is probably your best bet here.

AndreasFuchsSIT:
Either that or adding the m4 macro into the m4/ directory.
However that macro is apparently under LGPL...
Can we deal with multi-license in this project ?

martinezjavier:
@AndreasFuchsSIT that's a very good point. I think the answer then is that we can't include the macro it in the project repo...

The opposite wouldn't have been a problem (if tpm2-tss was licensed under LGPL and the autoconf macro was licensed under BSD 2-clause) since the derivative work would still had been LGPL.

So @flihp proposed workaround seems like the best way to solve this for me.

AndreasFuchsSIT:
In that case, I have a script for travis:
```
wget http://ftpmirror.gnu.org/autoconf-archive/autoconf-archive-2017.09.28.tar.xz
sha256sum autoconf-archive-2017.09.28.tar.xz | grep -q 5c9fb5845b38b28982a3ef12836f76b35f46799ef4a2e46b48e2bd3c6182fa01
tar xJf autoconf-archive-2017.09.28.tar.xz
cp autoconf-archive-2017.09.28/m4/ax_code_coverage.m4 m4/
```

AndreasFuchsSIT:
Sorry, I was unaware of the .travis.yml file...
Test is working now...

PR #633, opened 2017-11-22
sysapi: fix GetRpBuffer

AndreasFuchsSIT:
GetRpBuffer would return garbage if called before
complete. This patch fixes this.
Also eliminates the field rpBuffer from SYS_CONTEXT along
the way.

### DIFF
--- a/sysapi/sysapi/EncryptParam.c
+++ b/sysapi/sysapi/EncryptParam.c
@@ -35,7 +35,7 @@ TSS2_RC Tss2_Sys_GetEncryptParam(
     const uint8_t **encryptParamBuffer)
 {
     _TSS2_SYS_CONTEXT_BLOB *ctx = syscontext_cast(sysContext);
-    TPM2B *encryptParam;
+    uint8_t *offset;
 
     if (!encryptParamSize || !encryptParamBuffer || !ctx)
         return TSS2_SYS_RC_BAD_REFERENCE;
@@ -47,13 +47,15 @@ TSS2_RC Tss2_Sys_GetEncryptParam(
         BE_TO_HOST_16(resp_header_from_cxt(ctx)->tag) == TPM2_ST_NO_SESSIONS)
         return TSS2_SYS_RC_NO_ENCRYPT_PARAM;
 
-    /* Get first parameter and return its size and a pointer to it. */
-    ctx->rpBuffer = (UINT8 *)ctx->rspParamsSize;
-    /* Skip over params size field. */
-    ctx->rpBuffer += 4;
-    encryptParam = (TPM2B *)ctx->rpBuffer;
-    *encryptParamSize = BE_TO_HOST_16(encryptParam->size);
-    *encryptParamBuffer = encryptParam->buffer;
+    /* Get first parameter, interpret it as a TPM2B and return its size field
+     * and a pointer to its buffer area. */
+    offset = ctx->cmdBuffer
+            + sizeof(TPM20_Header_Out)
+            + ctx->numResponseHandles * sizeof(TPM2_HANDLE)
+            + sizeof(TPM2_PARAMETER_SIZE);
+
+    *encryptParamSize = BE_TO_HOST_16(*((UINT16 *)offset));
+    *encryptParamBuffer = offset + sizeof(UINT16);
 
     return TSS2_RC_SUCCESS;
 }
--- a/sysapi/sysapi/GetRpBuffer.c
+++ b/sysapi/sysapi/GetRpBuffer.c
@@ -34,6 +34,7 @@ TSS2_RC Tss2_Sys_GetRpBuffer(
     const uint8_t **rpBuffer)
 {
     _TSS2_SYS_CONTEXT_BLOB *ctx = syscontext_cast(sysContext);
+    TSS2_RC rval;
 
     if (!ctx || !rpBufferUsedSize || !rpBuffer)
         return TSS2_SYS_RC_BAD_REFERENCE;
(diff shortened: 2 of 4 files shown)


sysapi/sysapi_util/CommandUtil.c
```diff
@@ -159,16 +158,11 @@ TSS2_RC CommonComplete(_TSS2_SYS_CONTEXT_BLOB *ctx)
         rval = Tss2_MU_UINT32_Unmarshal(ctx->cmdBuffer,
```
tstruk: If we are not saving anything here then the comment above is irrelevant. Please remove

sysapi/sysapi_util/CommandUtil.c
```diff
@@ -159,16 +158,11 @@ TSS2_RC CommonComplete(_TSS2_SYS_CONTEXT_BLOB *ctx)
...
                                         ctx->maxCmdSize,
                                         &ctx->nextData,
-                                        &ctx->rpBufferUsedSize);
+                                        NULL);
         if (rval)
```
tstruk: No need for this check here

sysapi/sysapi/GetRpBuffer.c
```diff
@@ -44,8 +45,32 @@ TSS2_RC Tss2_Sys_GetRpBuffer(
...
-    *rpBuffer = ctx->rpBuffer;
-    *rpBufferUsedSize = ctx->rpBufferUsedSize;
+    /* Calculate the position of the response parameter section within the TPM
+     * repsponse as well as its size. Structure is:
+     * Header (tag, responseSize, responseCode)
+     * handle(if Command has handles)
+     * parameterSize (if TPM_ST_SESSIONS), size of rpArea
+     * rpArea
+     * Sessions (if TPM_ST_SESSIONS) */
+	size_t offset = 0;
```
tstruk: Indentation TABs -> spaces

sysapi/sysapi/EncryptParam.c
```diff
@@ -35,7 +35,7 @@ TSS2_RC Tss2_Sys_GetEncryptParam(
     const uint8_t **encryptParamBuffer)
 {
     _TSS2_SYS_CONTEXT_BLOB *ctx = syscontext_cast(sysContext);
-    TPM2B *encryptParam;
+    void *offset;
```
tstruk: UINT16 *offset instead of void *

sysapi/sysapi/GetRpBuffer.c
tstruk: No need to assign it to 0, just offset = sizeof(TPM20_Header_Out); below

sysapi/sysapi/GetRpBuffer.c
```diff
@@ -44,8 +45,32 @@ TSS2_RC Tss2_Sys_GetRpBuffer(
...
+	offset += sizeof(TPM20_Header_Out); // Skip over the header
```
tstruk: Please keep format of the comments consistant

flihp:
@tstruk : if you're requesting changes to this PR you should indicate as much in the `review` interface. It's obvious from your comments above what you want but it still looks like the PR is awaiting your review (doesn't show "changes requested" from the top level PR list).

sysapi/sysapi/EncryptParam.c
AndreasFuchsSIT: I will change to `uint8_t *offset` since it refers to the buffer...

sysapi/sysapi/GetRpBuffer.c
AndreasFuchsSIT: yep

AndreasFuchsSIT: ok, sure

sysapi/sysapi/GetRpBuffer.c
AndreasFuchsSIT: @tstruk sorry, but what do you mean here ?

sysapi/sysapi_util/CommandUtil.c
AndreasFuchsSIT: Changed to "Skipping over"

sysapi/sysapi/GetRpBuffer.c
martinezjavier: @AndreasFuchsSIT I think he means to use C `/* */` style comments instead of C++ `//` ones.

sysapi/sysapi_util/CommandUtil.c
AndreasFuchsSIT: 1. I don't agree, since the TPM response comes from outside. An 12-byte TPM response would make this call fail.
1. More general: We'd end up with a unused variable or unchecked return value which is unnice. So I personally favour all returns a checked, even if it is unused.

@tstruk I'd like to leave the check in, if that's ok with you.

AndreasFuchsSIT:
Pushed with 4oo6 fixes, see above.

sysapi/sysapi/GetRpBuffer.c
AndreasFuchsSIT: ok, sure... pushed...

sysapi/sysapi_util/CommandUtil.c
tstruk: I don't see it as unchecked return value because in either case it is up to the caller of CommonComplete() to check it. In my opinion return rval; followed by return rval; doesn't make much sense. And we can always remove the rval if it is not needed.

tstruk:
@flihp ok will use change requested in the future.

sysapi/sysapi_util/CommandUtil.c
tstruk: Having said that I like this change and I can remove the redundant check and the unused variable later.

AndreasFuchsSIT: Oh, sorry... I completely missed out on the double returning...

I guess I'd personally leave it in, just because people might add something before the return. But go ahead and make your choice. I don't care...

tstruk: I wouldn't worry about the future. If people will change it then we will see if that's ok or not.

AndreasFuchsSIT: ok, changed and pushed...

tstruk: If you have pushed to the old PR then it's too late because it has been merged already.
Could you please create a new PR for the new change?

AndreasFuchsSIT: sone #644

PR #634, opened 2017-11-22
sysapi: Update SetCmdAuths and GetRspAuths to new parameter type

AndreasFuchsSIT:
Since the last iteration of the specification there is just a struct
with an array directly embedded.
This is a large change, but removes a lot of complexity.

Note that this applies on top of #627 and did not make sense separately...

### DIFF
--- a/include/sapi/tss2_sys.h
+++ b/include/sapi/tss2_sys.h
@@ -55,17 +55,14 @@ typedef struct _TSS2_SYS_OPAQUE_CONTEXT_BLOB TSS2_SYS_CONTEXT;
 // Input structure for authorization area(s).
 //
 typedef struct {
-    uint8_t cmdAuthsCount;
-    TPMS_AUTH_COMMAND **cmdAuths;
-} TSS2_SYS_CMD_AUTHS;
+    uint16_t          count;
+    TPMS_AUTH_COMMAND auths[3];
+} TSS2L_SYS_AUTH_COMMAND;
 
-//
-// Output structure for authorization area(s).
-//
 typedef struct {
-    uint8_t rspAuthsCount;
-    TPMS_AUTH_RESPONSE **rspAuths;
-} TSS2_SYS_RSP_AUTHS;
+    uint16_t           count;
+    TPMS_AUTH_RESPONSE auths[3];
+} TSS2L_SYS_AUTH_RESPONSE;
 
 
 //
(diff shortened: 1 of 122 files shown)


AndreasFuchsSIT:
The travis failure seems to be the race-condition we already had the other day....
I'll rebase to master and push to trigger a new build...

include/sapi/tss2_sys.h
```diff
@@ -55,17 +55,14 @@ typedef struct _TSS2_SYS_OPAQUE_CONTEXT_BLOB TSS2_SYS_CONTEXT;
 // Input structure for authorization area(s).
 //
 typedef struct {
-    uint8_t cmdAuthsCount;
-    TPMS_AUTH_COMMAND **cmdAuths;
-} TSS2_SYS_CMD_AUTHS;
+    uint16_t          count;
```
williamcroberts: i'm noticing changes using stdint,h directly rather than UINIT16, are the defined sizes in the tss going to be replaced by stdint directly at some point?

Im just thinking consitency here, why define UINT16 if it's not going to be used in leiu of uint16_t?

flihp: This is a long standing debate and I think I asked the same question when I started participating in the TSS working group. The answer I got boils down to "This is the way the TPM spec does things and we follow the TPM spec as closely as possible". IMHO it's an unnecessary level of indirection which is what I think you're getting at.

flihp: Sorry I misunderstood your point. I think I'm with you now: all of the other structures from the spec use UINT* so why are we using the stdint.h types here. I think @AndreasFuchsSIT was just using the closest type to the one we already had in the structure.

AndreasFuchsSIT: Well, I took it from the TSS Spec https://trustedcomputinggroup.org/wp-content/uploads/TSS_SAPI_v1.1_r21_Public_Review.pdf page 42.

@flihp @williamcroberts I don't know, why we made it such, but for the moment it is, I suppose.

Shall I change it nevertheless ?

flihp: @williamcroberts : you asked the question originally as part of your review. Is the justification given sufficient or are you requesting changes?

tstruk: @williamcroberts this is still waiting for your input

tstruk:
@AndreasFuchsSIT do you want to rebase and push again your https://github.com/intel/tpm2-tools/pull/645? I will create a PR for tpm2-abrmd for this.

AndreasFuchsSIT:
I can only do so in about 20 hours...
So do as you like... ;-)

PR #636, opened 2017-11-27
[RFC] Make: Add silent rules

AndreasFuchsSIT:
IMHO, Make's output is just too cluttered... So I never see any warnings...

RFQ: Would other also like to add silent rules as proposed in this patch ?

P.S. `make V=1` still gives you the old output...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -304,20 +304,20 @@ test_integration_system_api_int_SOURCES = test/integration/main.c \
 TESTS_LDADD = $(noinst_LTLIBRARIES) $(lib_LTLIBRARIES)
 
 AUTHORS :
-	git log --format='%aN <%aE>' | grep -v 'users.noreply.github.com' | sort | \
+	$(AM_V_GEN)git log --format='%aN <%aE>' | grep -v 'users.noreply.github.com' | sort | \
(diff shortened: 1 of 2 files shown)


flihp:
Is the `-Werror` in configure.ac not working?

AndreasFuchsSIT:
-Werror does work... However:
- If you build with -j 48, the errors may be a few pages up, because of parallelism...
- If you scan-build for clang then warnings do not trigger errors and --status-bugs only triggers at the very end...
- As a matter of personal preference, I prefer it... ;)

So just consider it an RFC...

flihp:
I guess I've always been of the opinion that build output should be verbose. Better to redirect stdout / stderr if you want to keep the two streams separate. @tstruk : do you have an opinion?

tstruk:
Yes, I like the new output. It's what I'm used to, and we can still get all the verbose output with V=1. I'm +1 for this change.

drbild:
FWIW, `./configure --enable-silent-rules` already gives silent output.  I don't know enough about `autotools` to know if these changes bring any additional improvement.

AndreasFuchsSIT:
It make silent-rules the default, if autotools is new enough to support it.

It also wraps some Makefile targets into a macro that silences its outputs and presents a "Generating" instead...

PR #638, opened 2017-11-27
sysapi: Fixes about GetCommandCode

AndreasFuchsSIT:
This patch updates the function signature to reflect the
latest specification version.
Further, it make the GetCommandCode function safe to use
with non-aligned arrays.
Also it fixes the tests to account for correct endianess.

### DIFF
--- a/sysapi/sysapi/GetCommandCode.c
+++ b/sysapi/sysapi/GetCommandCode.c
@@ -31,7 +31,7 @@
 
 TSS2_RC Tss2_Sys_GetCommandCode(
     TSS2_SYS_CONTEXT *sysContext,
-    UINT8 (*commandCode)[4])
+    UINT8 *commandCode)
 {
     _TSS2_SYS_CONTEXT_BLOB *ctx = syscontext_cast(sysContext);
 
@@ -41,7 +41,8 @@ TSS2_RC Tss2_Sys_GetCommandCode(
     if (ctx->previousStage == CMD_STAGE_INITIALIZE)
         return TSS2_SYS_RC_BAD_SEQUENCE;
 
-    *(TPM2_CC *)commandCode = HOST_TO_BE_32(ctx->commandCode);
+    TPM2_CC tmp = HOST_TO_BE_32(ctx->commandCode);
+    memcpy(commandCode, (void *)&tmp, sizeof(tmp));
 
     return TSS2_RC_SUCCESS;
 }
--- a/include/sapi/tss2_sys.h
+++ b/include/sapi/tss2_sys.h
@@ -148,7 +148,7 @@ TSS2_RC Tss2_Sys_Execute(
 //
 TSS2_RC Tss2_Sys_GetCommandCode(
     TSS2_SYS_CONTEXT *sysContext,
-    UINT8 (*commandCode)[4]
+    UINT8 *commandCode
     );
 
 TSS2_RC Tss2_Sys_GetRspAuths(
--- a/sysapi/include/sysapi_util.h
+++ b/sysapi/include/sysapi_util.h
@@ -64,7 +64,7 @@ typedef struct {
     // These are set by system API and used by helper functions to calculate cpHash,
     // rpHash, and for auditing.
     //
-    TPM2_CC commandCode;
+    TPM2_CC commandCode;    // This is in host endianess
     UINT32 cpBufferUsedSize;
     UINT8 *cpBuffer;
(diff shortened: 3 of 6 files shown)


sysapi/sysapi/GetCommandCode.c
```diff
@@ -41,7 +41,8 @@ TSS2_RC Tss2_Sys_GetCommandCode(
     if (ctx->previousStage == CMD_STAGE_INITIALIZE)
         return TSS2_SYS_RC_BAD_SEQUENCE;
 
-    *(TPM2_CC *)commandCode = HOST_TO_BE_32(ctx->commandCode);
+    TPM2_CC tmp = HOST_TO_BE_32(ctx->commandCode);
+    memcpy(commandCode, (void *)&tmp, sizeof(tmp));
```
tstruk: No need to cast to void *.
Isn't a direct assignment better than calling a memcpy on 4 bytes?

flihp:
As I remember it the type isn't TPM_CC since the data it's necessarily in host byte order. If the command code isn't in host byte order then it won't be a valid command code. IMHO this is a bit rigid but it's valid.

AndreasFuchsSIT:
Also just for convenience.

In the intended usecase, you use this in order to build up a cpBuffer for a cpHash. These cpBuffers are uint8_t[]s so this way, you can just `getCommandCode(ctx, &cpBuffer[offset]);`...

sysapi/sysapi/GetCommandCode.c
AndreasFuchsSIT: If the 4 bytes of memory behind commandCode are not page-aligned, i.e. cross a page border, then the assignment may fail, if I remember correctly.

The thing about (void *), shall I change this ?

tstruk: If that was the case then all direct assignments of ints would have to be change to memcpy.
I think direct assignment is good here and we don't need the tmp variable (which BTW can also be unaligned).

AndreasFuchsSIT: The thing is that you may not cast a uint8_t[] into a uint32_t, because the former is 1-byte-aligned, whereas the latter is 4-byte-aligned.

I may be wrong about this, but I thought that an uint32_t would never go over page-borders, whereas in a uint8_t[]-array different fields of the array can be on different pages... I have run something like this on some Coretex-M3 at some point... But hardly remember.

P.S. in x86 I think the CPU can handle cross-page-primitive types. But others cannot.

So yes, whenever a byte-array is cast into a multi-byte integer, these places need to be changed...

Also see https://wiki.sei.cmu.edu/confluence/display/c/EXP36-C.+Do+not+cast+pointers+into+more+strictly+aligned+pointer+types

tstruk: but now it is uint32_t across the board and has nothing to do with uint_8_t[4].
We don't do any conversion between uint_8_t[4] and uint32_t now right?

AndreasFuchsSIT: `(void *)&tmp` basically goes from 4-byte-aligned to unaligned.

`*(TPM2_CC *)commandCode` used to go from unaligned to 4-byte-aligned, which was bad...

but now we are ok

tstruk: So do you want to update it to:
*(TPM2_CC *)commandCode = HOST_TO_BE_32(ctx->commandCode);
?

AndreasFuchsSIT: No, `*(TPM2_CC *)commandCode` is something we must not do.

It breaks on certain platforms and it is against SEI recommendations.
I stand with my proposed patch.

tstruk: For my own education, could you please point me to the SEI recommendations that says that we should not do direct variable assignment.

AndreasFuchsSIT: This one: https://wiki.sei.cmu.edu/confluence/display/c/EXP36-C.+Do+not+cast+pointers+into+more+strictly+aligned+pointer+types
Note especially EXP36-C-EX1 on that page.

The thing is not about the assignment, it is about casting a uint8_t* to a uint32_t*, i.e. "casting to a more strictly aligned type". Though the examples do not exactly match.

Also the second example here: https://wiki.sei.cmu.edu/confluence/display/c/EXP39-C.+Do+not+access+a+variable+through+a+pointer+of+an+incompatible+type
(Though this does not result in CPU-Exceptions, so I value the first one higher)

PR #639, opened 2017-11-27
marshal: remove the fake na member from TPMU types

tstruk:
There is no need to add the fake members to TPMU. We can use the
first member of the union instead. That way we don't have to
polute the TPM types.

### DIFF
--- a/marshal/tpmu-types.c
+++ b/marshal/tpmu-types.c
@@ -135,7 +135,7 @@ static TSS2_RC marshal_keyedhash(BYTE const *src, uint8_t buffer[],
 }
 
 
-static TSS2_RC marshal_null(char const *src, uint8_t buffer[],
+static TSS2_RC marshal_null(void const *src, uint8_t buffer[],
                             size_t buffer_size, size_t *offset)
 {
     return TSS2_RC_SUCCESS;
@@ -234,8 +234,8 @@ static TSS2_RC unmarshal_keyedhash(uint8_t const buffer[], size_t buffer_size,
     return unmarshal_tab(buffer, buffer_size, offset, dest, sizeof(TPM2B_DIGEST));
 }
 
-static TSS2_RC unmashal_null(uint8_t const buffer[], size_t buffer_size,
-                             size_t *offset, char *dest)
+static TSS2_RC unmarshal_null(uint8_t const buffer[], size_t buffer_size,
+                              size_t *offset, void *dest)
 {
     return TSS2_RC_SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


williamcroberts:
Looks like we duplicated work: https://github.com/intel/tpm2-tss/pull/640

marshal/tpmu-types.c
```diff
@@ -308,10 +307,12 @@ TSS2_RC Tss2_MU_##type##_Marshal(type const *src, uint32_t selector, uint8_t buf
...
-#define TPMU_MARSHAL2(...) TPMU_MARSHAL(__VA_ARGS__, -1, ADDR, na, marshal_null, -2, ADDR, na, marshal_null,\
-                                        -3, ADDR, na, marshal_null, -4, ADDR, na, marshal_null, -5, ADDR, na, marshal_null, \
-                                        -6, ADDR, na, marshal_null, -7, ADDR, na, marshal_null, -8, ADDR, na, marshal_null, \
-                                        -9, ADDR, na, marshal_null, -10, ADDR, na, marshal_null)
+#define TPMU_MARSHAL2(type, sel, op, m, fn, ...) \
+    TPMU_MARSHAL(type, sel, op, m, fn, __VA_ARGS__, -1, ADDR, m, marshal_null, \
+                 -2, ADDR, m, marshal_null, -3, ADDR, m, marshal_null, \
+                 -4, ADDR, m, marshal_null, -5, ADDR, m, marshal_null, \
+                 -6, ADDR, m, marshal_null, -7, ADDR, m, marshal_null, \
+                 -8, ADDR, m, marshal_null, -9, ADDR, m, marshal_null)
```
williamcroberts: clever, reuse m as they will always have 1 member.. duh.

marshal/tpmu-types.c
```diff
@@ -361,10 +362,11 @@ TSS2_RC Tss2_MU_##type##_Unmarshal(uint8_t const buffer[], size_t buffer_size, \
     return ret; \
 }
 
-#define TPMU_UNMARSHAL2(...) TPMU_UNMARSHAL(__VA_ARGS__, -1, na, unmashal_null, -2, na, unmashal_null,\
-                                            -3, na, unmashal_null, -4, na, unmashal_null, -5, na, unmashal_null, \
-                                            -6, na, unmashal_null, -7, na, unmashal_null, -8, na, unmashal_null, \
-                                            -9, na, unmashal_null, -10, na, unmashal_null)
+#define TPMU_UNMARSHAL2(type, sel, m, fn, ...) \
+    TPMU_UNMARSHAL(type, sel, m, fn, __VA_ARGS__, -1, m, unmashal_null, \
+            -2, m, unmashal_null, -3, m, unmashal_null, -4, m, unmashal_null, \
```
williamcroberts: unmarshal_null

marshal/tpmu-types.c
```diff
@@ -308,10 +307,12 @@ TSS2_RC Tss2_MU_##type##_Marshal(type const *src, uint32_t selector, uint8_t buf
     return ret; \
 }
 
-#define TPMU_MARSHAL2(...) TPMU_MARSHAL(__VA_ARGS__, -1, ADDR, na, marshal_null, -2, ADDR, na, marshal_null,\
-                                        -3, ADDR, na, marshal_null, -4, ADDR, na, marshal_null, -5, ADDR, na, marshal_null, \
-                                        -6, ADDR, na, marshal_null, -7, ADDR, na, marshal_null, -8, ADDR, na, marshal_null, \
-                                        -9, ADDR, na, marshal_null, -10, ADDR, na, marshal_null)
+#define TPMU_MARSHAL2(type, sel, op, m, fn, ...) \
+    TPMU_MARSHAL(type, sel, op, m, fn, __VA_ARGS__, -1, ADDR, m, marshal_null, \
```
williamcroberts: You don't need this op field. You can just move it into "m".

What I did is just expose src/dest in the macros and then you send the member and all through.


Ie:
```
TPMU_MARSHAL(FOO, BAR1, &dest->bar, marshal_bar, BAR2, &dest->bar2, marshal_bar2,...)
```

I think it's simpler and more flexible but is a larger change.

PR #642, opened 2017-11-28
Headers: Fixing macros for error levels (to rc layers)

AndreasFuchsSIT:
The names of error macros changed with the latest spec.
Notably, ERROR_LEVEL turned to RC_LAYER.
Also adding RCs for the MUAPI.
Also banning non-standard APP_RC_LAYER into the tpmclient
test suit.


martinezjavier:
@AndreasFuchsSIT out of interest, what's the latest version of the spec you are taking these from?

I thought the latest version was TSS Family "2.0" Level 00, Revision 01.00 26 January 2015.

@flihp should I re-post my kernel patch with this naming convention instead? I based on the names of the defines in the SAPI header (I just renamed TSS2 to TPM2 so it's consistent with the rest of the driver).

flihp:
I was about to reply to your patch and suggest this. The problem is that we're making changes to these specs in parallel. The specs unfortunately used the term "error", "response", and "return" codes interchangeably, as well as "level" and "layer". We're in the process of cleaning this up now and everything will be "response code" and "response layer" going forward.

martinezjavier:
@flihp yeah, I found it confusing at the beginning and mixed the concepts. I'll change according to this PR.

While being there, do you still want me to remove the unused defines o do you think that could provide some value as documentation? I'm really OK with either so ill do whatever you prefer.

@AndreasFuchsSIT sorry for hijacking your PR!

AndreasFuchsSIT:
@martinezjavier no problem, make yourself at home... ;-)

I'm basing my patches on the latest draft (https://trustedcomputinggroup.org/specifications-public-review/) where a lot of cleanups and clarifications were made.
Hope this helps.

martinezjavier:
@AndreasFuchsSIT ah, it's a draft. That explains why I couldn't find a newer than the latest published version I had. Thanks for the info!

AndreasFuchsSIT:
IMHO, we should implement the draft right away. The release of a final can take a few months, but we should already get the advantages of the updates.

martinezjavier:
@AndreasFuchsSIT I don't know how conservative the tpm2-* projects are with regard to stuff in draft specs, but I do agree with you that if something is unlikely to change, then we can just implement it.

I should take a look to the draft specs, so far I've been mostly focusing on the latest published versions.

AndreasFuchsSIT:
Is this a positive review then ?

martinezjavier:
@AndreasFuchsSIT I'm not sure if your question was for to me or the other but yes, LGTM after your explanation. Sorry for not explicitly approving the PR before.

flihp:
I'm working up a patch to the tools to minimize breakage.

flihp:
well this one broke the build somehow

AndreasFuchsSIT:
I guess those occurrences were new since I sent this PR...

Thanks for fixing...

flihp:
np

PR #643, opened 2017-11-28
bootstrap: use portable printf instead of echo

drbild:
The `-n` flag is not supported by the `echo` built-in of some versions
of `sh`. Use the portable `printf` instead.

Signed-off-by: David R. Bild <david.bild@xaptum.com>


flihp:
LGTM. Thanks @drbild !

PR #645, opened 2017-11-28
configure.ac: Use fixed version for stable branch

martinezjavier:
Having the version generated from a git tag could be useful for dev branches
and when people build from a git repository. But is not correct for released
tarballs, since attempting to execute the git command will fail and won't be
a version defined, i.e:

$ ./bootstrap
fatal: Not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

$ grep -i version lib/sapi.pc
Version:

So after a branch is released, just use a fixed version number:

$ grep -i version lib/sapi.pc
Version: 1.3.0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,4 @@
-AC_INIT([tpm2-tss],
-        [m4_esyscmd_s([git describe --tags --always --dirty])])
+AC_INIT([tpm2-tss], [1.3.0])
 AC_CONFIG_MACRO_DIR([m4])
 AC_PROG_CC
 AC_PROG_CXX


flihp:
I'm a bit confused. We use git and tags to get the version number. If you've building from git you can check out the appropriate tag, do the whole `./bootstrap && ./configure` dance and the `PACKAGE_VERSION` will be set appropriately. So like you say for dev and builds from git this works as expected.

For the release tarballs we use the `dist` target from automake. What this does for us is package up the *generated* `./configure` script which already has the `PACKAGE_VERSION` set. This works because I generate the release tarballs from a checked out git repo so people consuming the release tarballs don't have to.

In your example you run the `./bootstrap` script which isn't packaged in our release tarballs. Is is possible that you're using the github generated `Source.zip` files instead of the signed tarballs?

martinezjavier:
@flihp my bad. I missed that you also have release tarballs in https://github.com/intel/tpm2-tss/releases. I thought there were only source code tarballs there, since that's the case for the tpm2-tools (https://github.com/intel/tpm2-tools/releases).

@williamcroberts I think you should also do the same and publish the release tarball generated by the`make dist` target.

I'll fix the packages to point the source to https://github.com/intel/tpm2-tss/releases/download/1.3.0-rc2/tpm2-tss-1.3.0-rc2.tar.gz instead of https://github.com/intel/tpm2-tss/archive/1.3.0-rc2.tar.gz as is the case now.

I'm closing this PR now, sorry for the noise.

martinezjavier:
@flihp there's a bug in the generated tarball though, the LICENSE file is missing. I'll send a PR for this.

PR #647, opened 2017-11-29
build: detect ld version-script flag support

drbild:
Some linkers do not support the --version-script flag.  Only include
it if the linker supports it.

Signed-off-by: David R. Bild <david.bild@xaptum.com>

### DIFF
--- a/m4/ld-version-script.m4
+++ b/m4/ld-version-script.m4
@@ -1,0 +1,53 @@
+# ld-version-script.m4 serial 3
+dnl Copyright (C) 2008-2014 Free Software Foundation, Inc.
+dnl This file is free software; the Free Software Foundation
+dnl gives unlimited permission to copy and/or distribute it,
+dnl with or without modifications, as long as this notice is preserved.
+
+dnl From Simon Josefsson
+
+# FIXME: The test below returns a false positive for mingw
+# cross-compiles, 'local:' statements does not reduce number of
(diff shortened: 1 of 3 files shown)


m4/ld-version-script.m4
```diff
@@ -0,0 +1,53 @@
+# ld-version-script.m4 serial 3
+dnl Copyright (C) 2008-2014 Free Software Foundation, Inc.
+dnl This file is free software; the Free Software Foundation
+dnl gives unlimited permission to copy and/or distribute it,
+dnl with or without modifications, as long as this notice is preserved.
+
```
tstruk: Is this compatible with GPL2?

drbild: IANAL, but according to Fedora, yes: https://fedoraproject.org/wiki/Licensing/FSF_Unlimited_License

tstruk: @flihp @martinezjavier any thoughts?

martinezjavier: @tstruk I'm not sure why you are asking for GPLv2 compatibility since the tpm2-tss project is licensed under BSD.

But yes, as @drbild said I believe that's OK to include this macro since the FSFUL is a very permissive license and so BSD compatible (I'm not a lawyer though and license compliance is always confusing for me).

tstruk: @martinezjavier sorry I meant BSD. Thanks for input.

PR #649, opened 2017-11-29
Support building on OS X

drbild:
This patch series adds support for building on Mac OS X by cleaning up some imports and using the BSD method for blocking SIGPIPE.

Unit tests (`make check`) do not build because the OS X linker does not support the `--wrap` flag.  Some unit tests depend on symbol wrapping, and there is not an easy way to mimic this behavior on OS X. 

The primary goal of this change is to enable OS X users to experiment with TPM 2 via the TSS and simulator, not to enable development of the TSS itself on OS X. So the lack of unit test support should be acceptable.  Regardless, this patch series does increase portability, at least incrementally.

N.B. Building on OS X requires PRs #647 and #648 as well.

### DIFF
--- a/tcti/sockets.c
+++ b/tcti/sockets.c
@@ -75,6 +75,9 @@ InitSockets( const char *hostName,
     struct sockaddr_in otherService = { 0 };
     struct sockaddr_in tpmService = { 0 };
     int iResult = 0;            // used to return function results
+#if defined(TSS2_USE_SO_NOSIGPIPE)
+    int no_sigpipe_val = 1;
+#endif
 
     *otherSock = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
     if (*otherSock == INVALID_SOCKET)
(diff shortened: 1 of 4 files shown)


tcti/sockets.c
```diff
@@ -104,6 +107,26 @@ InitSockets( const char *hostName,
...
     }
 
+#if defined(TSS2_USE_SO_NOSIGPIPE)
+    iResult = setsockopt(*otherSock, SOL_SOCKET, SO_NOSIGPIPE, (void*)&no_sigpipe_val, sizeof(no_sigpipe_val));
+    if (iResult)
+    {
+        SAFE_CALL( debugfunc, data, NO_PREFIX, "setting SO_NOSIGPIPE failed with error = %d\n", WSAGetLastError() );
+        closesocket( *otherSock );
+        WSACleanup();
+        return(1);
```
tstruk: No need for the brackets around 1

tcti/sockets.c
```diff
@@ -104,6 +107,26 @@ InitSockets( const char *hostName,
...
+    }
+
+    iResult = setsockopt(*tpmSock, SOL_SOCKET, SO_NOSIGPIPE, (void*)&no_sigpipe_val, sizeof(no_sigpipe_val));
+    if (iResult)
+    {
+        SAFE_CALL( debugfunc, data, NO_PREFIX, "setting SO_NOSIGPIPE failed with error = %d\n", WSAGetLastError() );
+        closesocket( *otherSock );
+        WSACleanup();
+        return(1);
```
tstruk: No need for the brackets around 1

tcti/sockets.c
```diff
@@ -104,6 +107,26 @@ InitSockets( const char *hostName,
...
+    }
+#endif
+
```
tstruk: git apply complains about trailing whitespaces here" 
<stdin>:95: trailing whitespace.

tcti/sockets.c
drbild: Fixed.  Both styles are used in the existing code in that function.

tcti/sockets.c
drbild: Fixed.  Both styles are used in the existing code in that function.

tcti/sockets.c
drbild: Fixed.

PR #658, opened 2017-12-06
Use _MU_ RC layer in place of _TYPES_ layer removed in e66804b9ad58.

flihp:


flihp:
Merged this to fix failures from #642 despite the build failure here. It seems to be related to the new gcov / coveralls stuff.

Issue #657, opened 2017-12-05
coveralls integration breaks builds on private forks


flihp:
#654 adds much needed coveralls integration for our gcov data. Unfortunately something about the authorization mechanism causes builds pushed to private forks that are hooked up to travis-ci in their own right to break. An example is here: https://github.com/flihp/tpm2-tss/tree/remove-debug-prefix

@williamcroberts : have you had to deal with this for the tools? I'm tempted to try to filter submission to coveralls using the URL of the repo being built but that seems like a hack.

flihp:
This seems to have spread to the PRs, not just private forks. @williamcroberts : have you seen this in the tools stuff? Any ideas? If this was just for private repos I'd just skip coveralls if the credentials variable isn't set but I don't think we can do that for PRs.

flihp:
doesn't look like a new issue: https://github.com/lemurheavy/coveralls-public/issues/703

flihp:
merged #665 which resolves this. existing PRs need to be rebased to get rid of the scary red 'x' next to their build status 😄

PR #660, opened 2017-12-08
build: Include README.md and friends in the release tarball.

flihp:


flihp:
This PR is failing for what looks to be the same issue with the coveralls stuff. It gives the same error message / behavior as documented in #657

PR #698, opened 2017-12-21
Attached Component commands

tstruk:
This PR adds support for Attached Component commands.


tstruk:
In this PR I have dropped the table numbers from the comments in include/sapi/tss2_tpm2_types.h.
These can change between revisions. For example we have a commnet:
/* Table 205  Definition of TPMS_CREATION_DATA Structure <OUT> */
but in the 1.46 rev it is: "Table 215 - Definition of TPMS_CREATION_DATA Structure <OUT>"
We should either update them or drop them completely.

PR #699, opened 2017-12-21
marshal: fix compiler errors for non GNU compatible compilers

tstruk:
Compilers that don't support GNU extensions complain about
missing initializer.


tstruk:
See: https://github.com/intel/tpm2-tools/issues/743

flihp:
LGTM. If we can get test tested with the tool chain where the error was seen (like you ask in https://github.com/intel/tpm2-tools/issues/743) all the better.

PR #702, opened 2017-12-29
Remove trailing semicolon from TPM2B_TYPE2 macro

flihp:


flihp:
Worth noting is that the default config for visual studio is what uncovered this for me. You can induce the same behavior / error with GCC by supplying the `-pedantic` compiler flag.

AndreasFuchsSIT:
On the other hand, the missing semicolons mess up indent...
It has its own patch to add the missing ones: ca26fed

flihp:
stuff like https://github.com/intel/tpm2-tss/commit/ca26fed63cb10500f854038e0d359af3e7e2b158 will need to go away. Stinks that it screws up indent but it produces code that annoys `-pedantic` compilers which I'd consider a higher priority.

PR #703, opened 2017-12-29
Define TPMS_EMPTY structure to have a BYTE array with a single member.

flihp:
If define as an array with 0 members this causes pedantic C compilers to
complain of the invalid 0 sized array (disallowed by c99).

### DIFF
--- a/include/sapi/tss2_tpm2_types.h
+++ b/include/sapi/tss2_tpm2_types.h
@@ -1185,7 +1185,7 @@ typedef	TPM2_ST TPMI_ST_COMMAND_TAG;
 
 /* Table 68  Definition of TPMS_EMPTY Structure <INOUT> */
 typedef	struct {
-	BYTE empty[0];	 /* a structure with no member  */
+	BYTE empty[1];	 /* a structure with no member  */
 } TPMS_EMPTY;
 
 /* Table 69  Definition of TPMS_ALGORITHM_DESCRIPTION Structure <OUT> */


flihp:
The latest public review doc for the TSS2 headers defines TPMS_EMPTY in the same way: https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-Library-Part-2_Structures-01.46_PubRev.pdf

include/sapi/tss2_tpm2_types.h
```diff
@@ -1185,7 +1185,7 @@ typedef	TPM2_ST TPMI_ST_COMMAND_TAG;
 
 /* Table 68  Definition of TPMS_EMPTY Structure <INOUT> */
 typedef	struct {
-	BYTE empty[0];	 /* a structure with no member  */
+	BYTE empty[1];	 /* a structure with no member  */
```
tstruk: Shouldn't this be just BYTE empty[]; ?

AndreasFuchsSIT: Thing about `BYTE empty[]` is, that it's incomplete and sizeof() is not defined.
Since C99 disallows `[0]`, I do favour `[1]`. I just don't want to get into trouble with `malloc`s on TPMS_EMPTY in generated code.

flihp: The links from @AndreasFuchsSIT pretty much covers the issue. Not sure we can quantify the complexity this will cause in "generated code" since we don't yet have any though. In any case I think the complexity caused by this one type can be absorbed by code generator so I'd like to leave this as-is if @tstruk is OK with it.

tstruk: Yes, BYTE empty[1]; looks good to me.

PR #704, opened 2017-12-29
Master fix test race

flihp:
This resolves the race condition worked around in 24fb1306e0204c931bb0cab63d04c6534d49bb78 so we revert 24fb1306e0204c931bb0cab63d04c6534d49bb78 after we've fixed it.

### DIFF
--- a/script/int-log-compiler.sh
+++ b/script/int-log-compiler.sh
@@ -157,20 +157,47 @@ TEST_NAME=$(basename "${TEST_BIN}")
 SIM_LOG_FILE=${TEST_BIN}_simulator.log
 SIM_PID_FILE=${TEST_BIN}_simulator.pid
 SIM_TMP_DIR=$(mktemp --directory --tmpdir=/tmp tpm_server_XXXXXX)
-for i in $(seq 5); do
-    SIM_PORT=$(od -A n -N 2 -t u2 /dev/urandom | awk -v min=1024 -v max=65535 '{print ($1 % (max - min)) + min}')
-    simulator_start ${SIM_BIN} ${SIM_PORT} ${SIM_LOG_FILE} ${SIM_PID_FILE} ${SIM_TMP_DIR}
-    if [ $? -eq 0 ]; then
-        break;
-    else
-        echo "Failed to start simulator on port ${SIM_PORT}, retrying"
+PORT_MIN=1024
+PORT_MAX=65534
+BACKOFF_FACTOR=2
+BACKOFF_MAX=6
+BACKOFF=1
+for i in $(seq ${BACKOFF_MAX}); do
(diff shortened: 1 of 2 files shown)


script/int-log-compiler.sh
```diff
@@ -157,20 +157,43 @@ TEST_NAME=$(basename "${TEST_BIN}")
...
-    simulator_start ${SIM_BIN} ${SIM_PORT} ${SIM_LOG_FILE} ${SIM_PID_FILE} ${SIM_TMP_DIR}
-    if [ $? -eq 0 ]; then
-        break;
-    else
-        echo "Failed to start simulator on port ${SIM_PORT}, retrying"
+PORT_MIN=1024
+PORT_MAX=65534
+BACKOFF_FACTOR=2
+BACKOFF=1
+for i in $(seq 10); do
```
tstruk: Do we really need to wait that long? When all probes fail then it will take ~16 minutes to run. I would do (seq 6) max which will make it ~1 minute max.

Makefile.am
```diff
@@ -42,8 +42,6 @@ TEST_EXTENSIONS = .int
 INT_LOG_COMPILER = $(srcdir)/script/int-log-compiler.sh
 INT_LOG_FLAGS = --simulator-bin=$(SIMULATOR_BIN)
 
-check-programs: $(check_PROGRAMS)
```
AndreasFuchsSIT: Would you mind keeping the build target ? It doesn't hurt and you can have static code analysis of tests without requiring tests to be run.

script/int-log-compiler.sh
flihp: Reasonable. I just never did the math to see how long the maximum wait time would be. 10 was just a "nice round number" :smile: A minute sounds reasonable.

script/int-log-compiler.sh
```diff
@@ -157,20 +157,43 @@ TEST_NAME=$(basename "${TEST_BIN}")
...
+    SIM_PORT_DATA=$(od -A n -N 2 -t u2 /dev/urandom | awk -v min=${PORT_MIN} -v max=${PORT_MAX} '{print ($1 % (max - min)) + min}')
```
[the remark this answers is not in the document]
flihp: no reason we can't make this change ...

Makefile.am
flihp: sure, will update PR

flihp:
Commits have been fixed up to include the requested changes. The PR was approved despite the requested changes so I'll merge as soon as the build succeeds.

flihp:
also worth noting: The original patch to work around the race condition was only partial. The `distcheck` target run by travis-ci before `check` runs the test harness in parallel too. The fact that we haven't been bitten by this race again since the work-around was just a coincidence :smile_cat:

PR #653, opened 2017-11-30
[RFC] Reworked logging

AndreasFuchsSIT:
So far there were 4 logging frameworks in the project. This patch series reworks it to just one with the following feature:

- Compile time setting of max log level
- Runtime setting of log level via env variable
- Setting of log levels per module

PR #707, opened 2018-01-04
tcti: Remove unused log data from device TCTI config structure.

flihp:
These fields have been unused since
43a0d76bf4e8f50b164a8be89102f1de03d46fa6.


flihp:
this is going in faster than usual on account of it fixing an issue with #653 / #705

PR #708, opened 2018-01-05
common TCTI init

### LINKED ISSUE #690: Define protocol for dlopening and querying TCTI for common init
The common initialization function should have a function prototype something like:
```
TSS2_RC (*Tss2_Tcti_Init) (TSS2_TCTI_CONTEXT *context, size_t *size, char *config);
```
To allow users to discover information about the TCTI module we include a common info function with the prototype:
```
TSS2_RC (*Tss2_Tcti_Info)(TSS2_TCTI_INFO *info);
```

The TCTI header (tss2_tcti.h) should define a type for each, as well as string representations of the symbol for each:
- `#define TSS2_TCTI_INFO "Tss2_Tcti_Info"`
- `#define TSS2_TCTI_INIT "Tss2_Tcti_Init"`

Currently we need an interface to query a TCTI for information about the TCTI. This should include probably 3 fields:
1) name - simple string name for the module( ex. tcti-device)
2) description - a sentence or two describing what the TCTI module is / does
3) config_help - a help message describing the format of the configuration string
Additionally a function pointer to the modules common init function.

Each TCTI module should expose these symbols if the intended use case allows (may not make sense for things like UEFI or SGX). Not sure what to do about Windows `LoadLibrary` / `LoadLibraryEx`?

flihp:
This resolves #690. Next step is to remove the old init functions but the test harnesses for the various projects need to be updated first. I've also got some example code in a topic branch for the tools: https://github.com/flihp/tpm2-tools/tree/tcti-init-v3 but for the time the old interface is still in use.

### DIFF
--- a/include/sapi/tss2_tcti.h
+++ b/include/sapi/tss2_tcti.h
@@ -1,5 +1,5 @@
 //**********************************************************************;
-// Copyright (c) 2015, Intel Corporation
+// Copyright (c) 2015 - 2017, Intel Corporation
 //
 // Copyright 2015, Andreas Fuchs @ Fraunhofer SIT
 //
@@ -164,6 +164,23 @@ TSS2_TCTI_POLL_HANDLE *handles, size_t *num_handles);
 } TSS2_TCTI_CONTEXT_COMMON_V1;
 
 typedef TSS2_TCTI_CONTEXT_COMMON_V1 TSS2_TCTI_CONTEXT_COMMON_CURRENT;
+
+#define TCTI_INFO_SYMBOL "Tss2_Tcti_Info"
+
+typedef TSS2_RC (*TSS2_TCTI_INIT_FUNC) (
+    TSS2_TCTI_CONTEXT *tctiContext,
+    size_t *size,
+    const char *config
+    );
+
+typedef struct {
+    const char *name;
+    const char *description;
+    const char *config_help;
+    TSS2_TCTI_INIT_FUNC init;
+} TSS2_TCTI_INFO;
+
+typedef const TSS2_TCTI_INFO* (*TSS2_TCTI_INFO_FUNC) (void);
 
 #ifdef __cplusplus
 }
--- a/tcti/tcti_device.c
+++ b/tcti/tcti_device.c
@@ -41,6 +41,8 @@
 #define LOGMODULE tcti
 #include "log/log.h"
 
+#define TCTI_DEVICE_DEFAULT "/dev/tpm0"
+
 TSS2_RC LocalTpmSendTpmCommand(
     TSS2_TCTI_CONTEXT *tctiContext,
     size_t command_size,
@@ -233,4 +235,31 @@ TSS2_RC InitDeviceTcti (
     }
 
     return rval;
+}
+
+TSS2_RC Tss2_Tcti_Device_Init (
+    TSS2_TCTI_CONTEXT *tctiContext,
+    size_t *size,
+    const char *conf
+    )
+{
+    const char *dev_path = conf != NULL ? conf : TCTI_DEVICE_DEFAULT;
+    TCTI_DEVICE_CONF dev_conf = {
+        .device_path = dev_path,
+    };
+
+    return InitDeviceTcti (tctiContext, size, &dev_conf);
+}
+
+const static TSS2_TCTI_INFO tss2_tcti_info = {
+    .name = "tcti-device",
+    .description = "TCTI module for communication with Linux kernel interface.",
+    .config_help = "Path to TPM character device: \"/dev/tpm0\" is default.",
+    .init = Tss2_Tcti_Device_Init,
+};
+
+const TSS2_TCTI_INFO*
+Tss2_Tcti_Info (void)
+{
+    return &tss2_tcti_info;
 }
--- a/tcti/tcti_device.map
+++ b/tcti/tcti_device.map
@@ -1,6 +1,8 @@
 {
     global:
         InitDeviceTcti;
+        Tss2_Tcti_Device_Init;
+        Tss2_Tcti_Info;
     local:
         *;
 };
--- a/tcti/tcti_socket.c
+++ b/tcti/tcti_socket.c
@@ -25,10 +25,14 @@
  * THE POSSIBILITY OF SUCH DAMAGE.
  **********************************************************************/
 
+#include <inttypes.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/time.h>
 #include <inttypes.h>
+
+#include <uriparser/Uri.h>
 
 #include "sapi/tpm20.h"
 #include "sapi/tss2_mu.h"
@@ -39,6 +43,8 @@
 #include "tss2_endian.h"
 #define LOGMODULE tcti
 #include "log/log.h"
+
+#define TCTI_SOCKET_DEFAULT "tcp://127.0.0.1:2321"
 
 static TSS2_RC tctiRecvBytes (
     TSS2_TCTI_CONTEXT *tctiContext,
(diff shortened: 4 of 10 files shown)


include/sapi/tss2_tcti.h
```diff
@@ -165,6 +165,23 @@ TSS2_TCTI_POLL_HANDLE *handles, size_t *num_handles);
...
+    );
+
+typedef struct {
+    const char *name;
+    const char *description;
+    const char *config_help;
+    TSS2_TCTI_INIT_FUNC init;
+} TSS2_TCTI_INFO;
+
+typedef TSS2_TCTI_INFO* (*TSS2_TCTI_INFO_FUNC) (void);
```
AndreasFuchsSIT: I was wondering if this can/may/should be
```
typedef const TSS2_TCTI_INFO* (*TSS2_TCTI_INFO_FUNC) (void);
```

Would also have to be changed in all instatiations below...

tcti/tcti_socket.c
```diff
@@ -528,3 +534,65 @@ TSS2_RC InitSocketTcti (
...
+        goto out;
+    }
+    rc = InitSocketTcti (tctiContext, size, &sock_conf, 0);
+out:
+    uriFreeUriMembersA (&uri);
+    return rc;
+}
+
+/* public info structure */
+static TSS2_TCTI_INFO tss2_tcti_info __attribute__ ((unused)) = {
```
AndreasFuchsSIT: I think this should not require `__attribute__ ((unused))`. See also tss2_tcti_device 's info struct.

tstruk: Right, it is used in 597

tstruk: Should be const static?

tcti/tcti_socket.c
```diff
@@ -528,3 +534,65 @@ TSS2_RC InitSocketTcti (
...
+    .name = "tcti-socket",
+    .description = "TCTI module for communication with the Microsoft TPM2 Simulator.",
```
tstruk: Is that only MS TPM2 Simulator? What about a TPM on a remote platform (which is not yet supported)

tcti/tcti_socket.c
```diff
@@ -528,3 +534,65 @@ TSS2_RC InitSocketTcti (
...
+    if (range > HOST_NAME_MAX) {
+        rc = TSS2_TCTI_RC_BAD_VALUE;
+        goto out;
+    }
+    strncpy (hostname, uri.hostText.first, range);
+    sock_conf.hostname = hostname;
+
+    /* extract port number */
+    range = uri.portText.afterLast - uri.portText.first;
+    if (range < 5 && range > 0) {
```
tstruk: should be if (range <= 5 && range > 0)

tcti/tcti_socket.c
```diff
@@ -528,3 +534,65 @@ TSS2_RC InitSocketTcti (
...
+        strncpy (port, uri.portText.first, range);
```
tstruk: Need to verify if port is not greater than 65535? 99999 will not fit into uint16_t.

tcti/tcti_socket.map
```diff
@@ -2,6 +2,8 @@
     global:
         InitSocketTcti;
```
tstruk: Should the InitSocketTcti be private now?

tcti/tcti_device.c
```diff
@@ -234,3 +236,30 @@ TSS2_RC InitDeviceTcti (
...
+{
+    const char *dev_path = conf != NULL ? conf : TCTI_DEVICE_DEFAULT;
+    TCTI_DEVICE_CONF dev_conf = {
+        .device_path = dev_path,
+    };
+
+    return InitDeviceTcti (tctiContext, size, &dev_conf);
+}
+
+static TSS2_TCTI_INFO tss2_tcti_info = {
```
tstruk: const static?

tcti/tcti_device.map
```diff
@@ -1,6 +1,8 @@
 {
     global:
         InitDeviceTcti;
```
tstruk: Should InitDeviceTcti be private now?

tcti/tcti_socket.c
```diff
@@ -528,3 +534,65 @@ TSS2_RC InitSocketTcti (
...
+        sscanf (port, "%" SCNu16, &sock_conf.port);
```
tstruk: if (sscanf(port ...) != EOF)? or will the uriParseUriA() do it?

tcti/tcti_device.map
flihp: Eventually I'm hoping that it'll go away completely. Removing the old API at the same time the new one is introduced seemed a bit drastic though. To make the transition the tools and tabrmd would need PRs to go in at the same time. Since we don't have those PRs ready and @williamcroberts is out of town I figured we could avoid breakage by keeping the old API in place for a bit, though it should be removed before the 2.0 release.

tcti/tcti_socket.map
flihp: Still in use by the tools, tabrmd and others down stream. See previous comment on device TCTI.

include/sapi/tss2_tcti.h
flihp: works for me. PR updated.

tcti/tcti_socket.c
flihp: Good catch. This was a left over artifact from v2. PR updated.

tcti/tcti_device.c
flihp: PR updated.

tcti/tcti_socket.c
flihp: PR updated.

tcti/tcti_socket.c
flihp: Good catch. I'm putting it in a `uint32_t` now first and checking for `> UINT16_MAX` before putting it into a `uint16_t` for the call to `connect. PR updated.

tcti/tcti_socket.c
flihp: PR updated.

tcti/tcti_socket.c
flihp: `uriParseUriA` catches some invalid port stuff. It'll return an error if the port string has non-numeric characters in it. But it won't detect port numbers outside the valid range (1 to UINT16_MAX). So if there's anything in the port string (aka `range <= 5 && range > 0`) then it's going to be an unsigned integer and `sscanf` won't return EOF.

All that said checking for the return value makes sense just on the principle of CYA. I'll map these errors to to TSS2_TCTI_RC_BAD_VALUE. PR updated.

tcti/tcti_socket.c
flihp: The socket TCTI implements a protocol on top of the normal TPM2 command / response messages. This protocol was defined by Microsoft and published as part of their TPM2 reference code / implementation that's in the TCG specs. You can use this TCTi to communicate with a local simulator or one on a remote system currently (not sure if we consider this "supported").

You could use this same protocol to communicate with any TPM so long as some "middle-ware" implements the protocol on top of the underlying TPM (this is what we did in the old `resourcemgr`) but I wouldn't recommend it. I'm thinking of trying out something with protocol buffers in the tabrmd but haven't started: https://github.com/intel/tpm2-abrmd/issues/138

flihp:
Thanks for the review @tstruk & @AndreasFuchsSIT . The PR has been updated and I believe I've addressed all issues raised. If you agree @tstruk please merge, if not let me know which parts I screwed up and I'll fix it up.

include/sapi/tss2_tcti.h
```diff
@@ -165,6 +165,23 @@ TSS2_TCTI_POLL_HANDLE *handles, size_t *num_handles);
 
 typedef TSS2_TCTI_CONTEXT_COMMON_V1 TSS2_TCTI_CONTEXT_COMMON_CURRENT;
 
+#define TCTI_INFO_SYMBOL "Tss2_Tcti_Info"
```
AndreasFuchsSIT: I was wondering if this should be namespaced similar to the rest; i.e. TSS2_TCTI_INFO_SYMBOL since it is part of tss2_tcti.h?

Apologies for not finding it in the first run.

include/sapi/tss2_tcti.h
```diff
@@ -165,6 +165,23 @@ TSS2_TCTI_POLL_HANDLE *handles, size_t *num_handles);
...
+
+typedef TSS2_RC (*TSS2_TCTI_INIT_FUNC) (
+    TSS2_TCTI_CONTEXT *tctiContext,
```
martinezjavier: The project's coding standard says that variables names must be in lowercase and multiple names separated with an underscore character. I know this `tctiContext` camelCase name comes from the TSS SAPI and TCTI spec and it's all over the code base.

I wonder though if these exceptions should be mentioned in the coding standard doc, or is there any attempt to clean those in the spec and code base to make it consistent?

include/sapi/tss2_tcti.h
```diff
@@ -1,5 +1,5 @@
 //**********************************************************************;
```
martinezjavier: s/exstract/extract in the commit message.

tcti/tcti_device.c
```diff
@@ -234,3 +236,30 @@ TSS2_RC InitDeviceTcti (
...
+        .device_path = dev_path,
+    };
+
+    return InitDeviceTcti (tctiContext, size, &dev_conf);
+}
+
+const static TSS2_TCTI_INFO tss2_tcti_info = {
+    .name = "tcti-device",
+    .description = "TCTI module for communication with Linux kernel interface.",
+    .config_help = "Path to TPM character device: \"/dev/tpm0\" is default.",
```
martinezjavier: Instead of hardcoding /dev/tpm0, I think you should use TCTI_DEVICE_DEFAULT here:

```
.config_help = "Path to TPM character device: \""TCTI_DEVICE_DEFAULT"\" is default.",
```

tcti/tcti_socket.c
```diff
@@ -528,3 +534,84 @@ TSS2_RC InitSocketTcti (
...
+out:
+    uriFreeUriMembersA (&uri);
+    return rc;
+}
+
+/* public info structure */
+const static TSS2_TCTI_INFO tss2_tcti_info = {
+    .name = "tcti-socket",
+    .description = "TCTI module for communication with the Microsoft TPM2 Simulator.",
+    .config_help = "Connection URI in the form tcp://ip_address[:port]. Default is tcp://127.0.0.1:2321.",
```
martinezjavier: Same comment than with the device TCTI, I think this should be:

```
.config_help = "Connection URI in the form tcp://ip_address[:port]. Default is "TCTI_SOCKET_DEFAULT".",
```

tcti/tcti_device.map
williamcroberts: That seems fine, one patch adding them in a backwards compatible way, one patch removing them. This pretty much lets me kill off the tcti init in the tools, as I provided a common interface, so I am glad to get rid of that.

flihp: I'm thinking we'll have a TCTI utility library that provides proper functions for the TCTI function pointers (the macros we have now are good but they're hostile to language bindings) as well as the `dlopen` magic for loading. I'll pick this up after this PR is merged. Should probably get a ticket in the issue tracker too.

tstruk:
Looks ok. Merging it since all other reviewers approved.

include/sapi/tss2_tcti.h
flihp: Already merged, but will have PR for this change in a few moments.

include/sapi/tss2_tcti.h
flihp: this one I can't fix since it's been merged, I'm working on the others though

tcti/tcti_device.c
flihp: Makes perfect sense to me. For this to make sense `TCTI_DEVICE_DEFAULT` will need to be in a public header and currently it's only used internally. This is something that can go into $srcdir/include/tcti/tcti_device.h easy enough though.

tcti/tcti_socket.c
flihp: Comment is same as the device TCTI: makes sense but these macros must be exposed in the TCTI's header.

include/sapi/tss2_tcti.h
flihp: Coding standard updated to account for this situation: https://github.com/intel/tpm2-tss/pull/714

flihp: For completeness: the coding standard has been updated to include a section documenting "exceptions" to our stated policy: 45a711e60c19edb6243f90d9e34f97f57521f2ff. Specifically that TCG uses weird naming conventions for function 
names / prototypes and to remain compliant with the specs we must use their naming.

PR #711, opened 2018-01-09
tcti: Add 'TSS2_' prefix to TCTI_INFO_SYMBOL macro.

flihp:
The TSS2_ prefix is how the TCG specs attempt prevent conflicts with symbols
/ macros from other libraries / headers. Since this macro is in a public
header it should have this prefix.


flihp:
see review comment here: https://github.com/intel/tpm2-tss/pull/708#discussion_r160108181

PR #714, opened 2018-01-09
doc: Document exceptions to function and variable naming convention.

flihp:
The TCG specs uses camelCase and Some_Other_Weird_Form heavily. To
maintain compatibility we must make exceptions for this stuff.


flihp:
NP. If you hadn't pointed this out the doc never would have been updated. Adding like to your original comment for context: https://github.com/intel/tpm2-tss/pull/708#discussion_r160150909

PR #716, opened 2018-01-10
build: Have supported glibc versions before 2.19.

lugong:
When using glibc < 2.19, the use of endian.h requires the
old macro _BSD_SOURCE instead of the new one _DEFAULT_SOURCE.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,6 +57,7 @@ AX_ADD_COMPILER_FLAG([-fpic])
 AX_ADD_COMPILER_FLAG([-fPIC])
 
 AX_ADD_PREPROC_FLAG([-D_DEFAULT_SOURCE])
+AX_ADD_PREPROC_FLAG([-D_BSD_SOURCE])
 
 AC_ARG_WITH([loglevel],
             [AS_HELP_STRING([--with-loglevel={none,error,warning,info,debug,trace}],


lugong:
In fact, when I prepare this PR, I encountered another compiling issue with my glibc (2.17). The error log is as below. It needs the macro _POSIX_C_SOURCE to include the declaration of HOST_NAME_MAX.
CC       tcti/tcti_libtcti_socket_la-tcti_socket.lo
tcti/tcti_socket.c: In function 'Tss2_Tcti_Socket_Init':
tcti/tcti_socket.c:565:19: error: 'HOST_NAME_MAX' undeclared (first use in this function)
                   ^
tcti/tcti_socket.c:565:19: note: each undeclared identifier is reported only once for each function it appears in
tcti/tcti_socket.c:565:10: error: unused variable 'hostname' [-Werror=unused-variable]
          ^
How about replacing three macros _DEFAULT_SOURCE, _BSD_SOURCE and POSIX_C_SOURCE with the one macro _GNU_SOURCE?
Here is a link to give some description about these. https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html

lugong:
@flihp 
In fact, when I prepare this PR, I encountered another compiling issue with my glibc (2.17). The error log is as below. It needs the macro _POSIX_C_SOURCE to include the declaration of HOST_NAME_MAX.
CC tcti/tcti_libtcti_socket_la-tcti_socket.lo
tcti/tcti_socket.c: In function 'Tss2_Tcti_Socket_Init':
tcti/tcti_socket.c:565:19: error: 'HOST_NAME_MAX' undeclared (first use in this function)
^
tcti/tcti_socket.c:565:19: note: each undeclared identifier is reported only once for each function it appears in
tcti/tcti_socket.c:565:10: error: unused variable 'hostname' [-Werror=unused-variable]
^
How about replacing three macros _DEFAULT_SOURCE, _BSD_SOURCE and POSIX_C_SOURCE with the one macro _GNU_SOURCE?
Here is a link to give some description about these. https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html

flihp:
Sorry for missing this question @lugong. I wasn't ignoring the question, just got lost working elsewhere and not checking my notifications enough :(

We've already fixed the `_POSIX_C_SOURCE` issue thanks to your work so that's been addressed. My preference would be to enable the feature test macros as we need them and keep from pulling in GNU specific stuff implicitly with a 'catch-all' macro like `_GNU_SOURCE`.

PR #718, opened 2018-01-11
Tpmu doc and format

flihp:
I spent some time recently figuring out how the macros that generate the TPMU_* marshalling functions work. The existing comments around these macros were helpful, but still left a bunch of work to the reader. I'm hoping to save the next person who comes along a bit of time by filling in some of the less obvious details.


flihp:
Looks like I messed up some of the macros with that last commit. Fixing up now.

PR #719, opened 2018-01-11
[RFC] Add (M|Unm)arshal functions for TPMU_NAME.

flihp:
!!DO NOT MERGE!!

This is an RFC with hopes to get some consensus on what to do about
marshalling functions for the TPMU_NAME type. This union type is a
special case that I wasn't aware of till now. It's a special case
because the spec doesn't define a selector for it (see table 10.5.2 in
the part 2). Instead the member that defines the name (either a hash or
a handle) is determined by circumstance: entities that are defined by a
public area (objects and indexes) the name is the hash of the public
structure. For entities not defined by a public area, the name is a
handle (described in 10.5.1 of part 2).

To be compliant with the TSS2 marshalling API spec we must provide
functions for (M|Unm)arshalling TPMU_NAME types but the circumstantial
handling (as opposed to selector-based like all other TPMU_* types)
makes this difficult (impossible?). As it's written now, the marshalling
spec declares the (Unm|M)arshal functions for the TPMU_NAME type with a
selector. This makes little sense since there are no selectors.

So far our implementation of libmu doesn't use this function internally
(why we have no implementation of it). The only place this may be needed
is in the marshalling of the TPM2B_NAME type. In this case we currently
marshal the TPMU_NAME as a byte array which AFAIK will make our
(M|Unm)arshaling of a TPMU_NAME that's a handle wrong.

This patch is obviously wrong since I'm using dummy values 1 and 0 as
the selectors for digest and handle members respectively. It's not clear
however what the right solution is since the (M|Unm)arshalling spec
seems to ignore this special case. Hopefully this RFC will get the
discussion started and we can sort out a solution which will probably
require providing feedback to the TCG TSS WG.

### DIFF
--- a/include/sapi/tss2_mu.h
+++ b/include/sapi/tss2_mu.h
@@ -1604,6 +1604,22 @@ Tss2_MU_TPMU_PUBLIC_PARMS_Unmarshal(
     TPMU_PUBLIC_PARMS *dest);
 
 TSS2_RC
+Tss2_MU_TPMU_NAME_Marshal(
+    TPMU_NAME const *src,
+    uint32_t         selector,
+    uint8_t          buffer[],
+    size_t           buffer_size,
+    size_t          *offset);
+
+TSS2_RC
+Tss2_MU_TPMU_NAME_Unmarshal(
+    uint8_t const buffer[],
+    size_t        buffer_size,
+    size_t       *offset,
+    uint32_t      selector,
+    TPMU_NAME    *dest);
+
+TSS2_RC
 Tss2_MU_TPMU_PUBLIC_ID_Marshal(
     TPMU_PUBLIC_ID const *src,
     uint32_t       selector_value,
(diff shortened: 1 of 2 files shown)


flihp:
Cc @AndreasFuchsSIT

AndreasFuchsSIT:
My proposal (what we've done in the past) is to use a selector derived from size comments on TPM2B_NAME but with a default handling this time around:
```
switch (selector) {
    case 4: return TPM_HANDLE_unmarshal();
    default: return TPMT_HA_unmarshal();
}
```

In practice, I don't really think that anyone will ever use that function. But if they do, this should be convenient, size they can just pass the surrounding TPM2B_NAME's size field as selector; i.e.
```
TPMU_NAME_unmarshal(&tpm2b_name->buffer[0],     /* buffer */
    tpm2b_name->size,                           /* buffer_size */
    NULL                                        /* offset */,
    tpm2b_name->size,                           /* selector */
    &tpmu_name);
```

flihp:
According to TSS-WG thisproposal was discussed. The proposal didn't make it into the current draft (Unm|M)arshalling spec though which seems to be the source of the problem. There are two tasks here then:
1. Get this special case documented in the spec (errata). This won't change the function prototype but should at least be an informational note describing how we define the selector for this function.
2. Get an implementation of this in our code. This should be rather trivial with the selector defined.

I'll make an issue for this with all of the relevant data and close the PR. Thanks @AndreasFuchsSIT

PR #721, opened 2018-01-19
[V2] tcti-socket: Fix build failure due to undeclared HOST_NAME_MAX

jiazhang0:
V2:
- Change to define _POSIX_SOURCE instead of _GNU_SOURCE.

 
The macro HOST_NAME_MAX is defined when _POSIX_SOURCE is defined.
Otherwise, the following build failure may occur.

tcti/tcti_socket.c: In function 'Tss2_Tcti_Socket_Init':
tcti/tcti_socket.c:565:19: error: 'HOST_NAME_MAX' undeclared (first use
in this function)

[This failure only happens in master branch]

Signed-off-by: Jia Zhang <zhang.jia@linux.alibaba.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,6 +58,7 @@ AX_ADD_COMPILER_FLAG([-fPIC])
 
 AX_ADD_PREPROC_FLAG([-D_DEFAULT_SOURCE])
 AX_ADD_PREPROC_FLAG([-D_BSD_SOURCE])
+AX_ADD_PREPROC_FLAG([-D_POSIX_SOURCE])
 
 AC_ARG_WITH([loglevel],
             [AS_HELP_STRING([--with-loglevel={none,error,warning,info,debug,trace}],


tcti/tcti_socket.c
```diff
@@ -25,6 +25,7 @@
  * THE POSSIBILITY OF SUCH DAMAGE.
  **********************************************************************/
 
+#define _GNU_SOURCE
```
flihp: `limits.h` will declare `HOST_NAME_MAX` when `_POSIX_C_SOURCE` is defined. `_GNU_SOURCE` is a superset of `_POSIX_C_SOURCE` and we don't need all of the extra stuff that comes with it. Feature test macros should be defined project-wide through the build environment. There are two examples here: https://github.com/intel/tpm2-tss/blob/master/configure.ac#L60

jiazhang0:
@flihp Thanks for your comment. Plz review V2.

flihp:
perfect @jiazhang0, thanks for fixing this up

PR #739, opened 2018-01-31
Use __attribute__((packed)) instead of #pragma pack

PeterHuewe:
Some of our embedded compiler-derivates seem to have a problem with #pragma pack and ignore it,
yielding to wrong sizes and total failure of the tss `__attribute__((packed))` works though.

Imho this is also much cleaner, since it puts the packing to the affected datastructures.

Signed-off-by: Peter Huewe <peter.huewe@infineon.com>

### DIFF
--- a/sysapi/include/sysapi_util.h
+++ b/sysapi/include/sysapi_util.h
@@ -34,25 +34,23 @@ enum cmdStates {CMD_STAGE_INITIALIZE,
                 CMD_STAGE_RECEIVE_RESPONSE,
                 CMD_STAGE_ALL = 0xff };
 
-#pragma pack(push, 1)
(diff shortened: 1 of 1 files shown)


tstruk:
Clang doesn't like it because it causes an unaligned pointer. I'm not sure why it doesn't complain about it with the pragma version, since the result should be the same, but it looks like for clang this will need the
AX_ADD_COMPILER_FLAG([-Wno-address-of-packed-member]) flag in configure.ac

flihp:
Strange that clang on Linux balks at this but Visual Studio + Clang doesn't. Could be a version thing.

flihp:
This went stale.

PR #743, opened 2018-02-02
Create loaded

tstruk:
Add support for Create Loaded command

### DIFF
--- a/test/integration/create-loaded.int.c
+++ b/test/integration/create-loaded.int.c
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2018, Intel Corporation
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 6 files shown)


test/integration/create-loaded.int.c
```diff
@@ -0,0 +1,118 @@
...
+                                &auth_rsp);
+    if (rc == TPM2_RC_SUCCESS) {
+        LOG_INFO("success object handle: 0x%x", object_handle);
+    } else {
+        LOG_ERROR("CreateLoaded FAILED! Response Code : 0x%x", rc);
+        exit(1);
+    }
+
+    memset(&out_public, '\0', sizeof(out_public));
+    memset(&name, '\0', sizeof(name));
```
flihp: can you, or don't you already do this when the structure is initialized?

tstruk: I agree that we can have some common init helpers, but in this case it is not redundant.
Both of them will be populated with data by the Tss2_Sys_CreateLoaded(). I want to clean them fresh before calling to Tss2_Sys_ReadPublic()

flihp:
I just tried to use the UI to fixup the conflict #747 caused when it was merged. Looks like it made a merge commit instead of fixing it up into your commit.

flihp:
oh man github is just a writhing pile of merge commits. I'll fix this up manually to get it out of the queue.

flihp:
replaced by #748

PR #747, opened 2018-02-02
build: Link against libcrypt, not libssl.

flihp:
The crypto operations are lin libcrypto, not libssl.


flihp:
I guess some distros include flags for both libcrypt and libssl in the libssl pkg-config metadata. Other distros (I'm on Debian) limit libssl pkg-config data to only libssl (-lssl). Simple solution is to use only libcrypt since we're not really using libssl anyways.

Issue #749, opened 2018-02-05
Visual Studio 64-bit build support


dantpm:
The fix to issues #420, #670, and #673 added support with  Microsoft Visual Studio "solution" files for the Marshaling and System API libraries. However, the fix is 32-bit.

The proposed fix is to also support 64-bit builds. The source should work with both 32- and 64-bit builds.  Although the TPM 2.0 simulator is 32-bit on MS Windows, the socket interface is independent of LP32/LP64 issues (as is the interface with a real TPM 2.0).

Proposed  fix
--------------------------- marshal/tss2-mu.vcxproj
       <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">..\lib\tss2-mu.def</ModuleDefinitionFile>
+      <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">..\lib\tss2-mu.def</ModuleDefinitionFile>
--------------------------- sysapi/tss2-sapi.vcxproj
       <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">include;..\include;..\marshal;$(SolutionDir);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">include;..\include;..\marsh
al;$(SolutionDir);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
. . .
       <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">..\lib\tss2-sapi.def</ModuleDefin
itionFile>
+      <AdditionalLibraryDirectories Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">$(OutDir)</AdditionalLibrar
yDirectories>
+      <ModuleDefinitionFile Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">..\lib\tss2-sapi.def</ModuleDefinit
ionFile>
--------------------------------- tpm2-tss.sln
-VisualStudioVersion = 15.0.27130.2020
+VisualStudioVersion = 15.0.27130.2027

flihp:
With #753 merged can this be closed?

dantpm:
Yes. Fixed 2018-02-08 with #753 merge.

PR #750, opened 2018-02-06
Add Visual Studio 64-bit build rules

dantpm:
Add Visual Studio 64-bit build rules to solution files.

Fixes issue #749.


dantpm:
Tested with Microsoft Visual Studio 15.5.6 on Windows 10. Built 64-bit libraries for both marshall and sysapi.

PR #753, opened 2018-02-06
Add Visual Studio 64-bit build rules

dantpm:
Visual Studio 64-bit build support.

Fixes issue #749.

### DIFF
--- a/tpm2-tss.sln
+++ b/tpm2-tss.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27130.2020
+VisualStudioVersion = 15.0.27130.2027
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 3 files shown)


dantpm:
Tested with a Visual Studio build of 64-bit marshal and sapi libraries.

flihp:
Taking a look at this now @dantpm. I'll flip the switch in AppVeyor too so we're building both configurations.

tpm2-tss.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27130.2020
+VisualStudioVersion = 15.0.27130.2027
```
flihp: Gah. I've missed working with Visual Studo so much! Did a quick search to see if someone has come up with a clever way to keep VS from changing this field every time someone opens the file. I didn't find anything useful but I did find this: https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/13908807-do-not-automatically-change-the-visualstudioversio

Looks like people have reported this annoyance to Microsoft and the Visual Studio team. They just don't seem to care to fix it. For now it's not really an issue but if we start branching & merging a bunch I can see this file starting to thrash.

PR #767, opened 2018-02-14
Add OpenSSL dependency to INSTALL.md

DomiNic5787:


DomiNic5787:
It is a test suite-only dependency, because it is only used by the test suite, isn't it?
If it is meant to only list unit test suite dependencies here, then it has to be mentioned in the description above.

tstruk:
OpenSSL is only required for the integration tests, which is enabled by the --with-simulatorbin option

DomiNic5787:
see #770

flihp:
Good catch @DomiNic5787.

flihp:
PR #771 resolves the the issue w/r to building libtest when the integration tests are disabled. Now a simple `./bootstrap && ./configure && make` should not require libcrypto or any of its headers. Still, the documentation needs a note added to indicate that libcrypto & headers are required when the integration tests are built (enabled with `--with-simulatorbin`).

I left the documentation fix out of my PR since this one is already open. @DomiNic5787: will you please fix up this PR accordingly?

flihp:
thanks @DomiNic5787

Issue #770, opened 2018-02-16
libssl-dev missing in INSTALL.md dependencies


DomiNic5787:
Reproduced on Ubuntu 16.04:
* Installing all required dependencies described in INSTALL.md
* ./bootstrap
* ./configure
* make

```
CC       test/integration/test-options.lo
test/integration/sapi-util.c:27:25: fatal error: openssl/sha.h: No such file or directory
```

This issue is related to 15220e89ff86087afc8661f3f2b7bac71a83649d.
libssl-dev was added to .travis.yml in 57d372e092b413ca3a2a9dd0ec9e02b221bab99f.
But it is missing in INSTALL.md sice then.

flihp:
This is a good catch. OpenSSL is only used in the integration tests but it looks like the test library is being built regardless of whether or not the integration tests are enabled.

flihp:
looks like this can be closed ...

PR #771, opened 2018-02-16
build: Build test library only when integration tests are enabled.

flihp:
This library uses libcrypto and if we build it when the integration
tests aren't enabled it creates a dependency on libcrypto when one
doesn't really exist.


flihp:
This resolves a bug uncovered in #770. We still need to add libcrypto to the list of dependencies but it's only required for the integration tests.

PR #773, opened 2018-02-17
Add test environment sanity check to int-log-compiler.sh. Fixes #764

### LINKED ISSUE #764: script/int-log-compiler.sh needs to detect degraded OS environments

dantpm:
Merged conflict with upstream INSTALL.md file.


dantpm:
This fails the Travis CI, so the netstat test needs to be replaced: 

FAIL: test/integration/asymmetric-encrypt-decrypt
=================================================
Command netstat not listing ports; exiting

Closing this pull request.

PR #774, opened 2018-02-20
TCTI refactor part 1

flihp:
This is the "low hanging fruit" in the TCTI modules. Most of this is re-organizing some of the build bits. The rest implements a common write function shared between the device and socket TCTIs. This is easier than messing with the read path since it's not an async call. The build structure introduced here will make writing and testing the async read stuff easier though.

### DIFF
--- a/include/tcti/common.h
+++ b/include/tcti/common.h
@@ -28,16 +28,10 @@
 #ifndef TCTI_COMMON_H
 #define TCTI_COMMON_H
 
-#include "sapi/tpm20.h"
-
 #if defined (__GNUC__)
 #define COMPILER_ATTR(...) __attribute__((__VA_ARGS__))
 #else
 #define COMPILER_ATTR(...)
 #endif
-
-typedef enum { NO_PREFIX = 0, RM_PREFIX = 1 } printf_type;
-typedef int (*TCTI_LOG_CALLBACK)( void *data, printf_type type, const char *format, ...);
-typedef int (*TCTI_LOG_BUFFER_CALLBACK)( void *useriData, printf_type type, UINT8 *buffer, UINT32 length);
 
 #endif /* TCTI_COMMON_H */
--- a/include/tcti/common.h
+++ b/include/tcti/common.h
@@ -1,30 +1,29 @@
-//**********************************************************************;
-// Copyright (c) 2016, Intel Corporation
-// All rights reserved.
-//
-// Redistribution and use in source and binary forms, with or without
-// modification, are permitted provided that the following conditions are met:
-//
-// 1. Redistributions of source code must retain the above copyright notice,
-// this list of conditions and the following disclaimer.
-//
-// 2. Redistributions in binary form must reproduce the above copyright notice,
-// this list of conditions and the following disclaimer in the documentation
-// and/or other materials provided with the distribution.
-//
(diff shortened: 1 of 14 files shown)


tcti/tcti.h
```diff
@@ -52,6 +55,13 @@
 
 #define TCTI_CONTEXT ((TSS2_TCTI_CONTEXT_COMMON_CURRENT *)(SYS_CONTEXT->tctiContext))
 
+#define TEMP_RETRY(exp) \
+({  int __ret; \
+    do { \
+        __ret = exp; \
+    } while (__ret == -1 && __ret == EINTR); \
```
tstruk: Should be while (__ret == -1 || __ret == EINTR);

tcti/tcti.c
```diff
@@ -96,3 +100,31 @@ TSS2_RC tcti_receive_checks (
...
+                  fd);
+        written = TEMP_RETRY (write (fd,
+                                     (const char*)&buf [written_total],
+                                     size - written_total));
+        if (written >= 0) {
+            LOG_DEBUG ("wrote %zd bytes to fd %d", written, fd);
+            written_total += (size_t)written;
+        } else {
+            LOG_ERROR ("failed to write to fd %d: %s", fd, strerror (errno));
+            return written_total;
```
tstruk: return written_total ? written_total : -1;
According to man 3 write 0 is not a valid return value from write(). It should be either number of bytes written (if any) or -1.

tcti/platformcommand.c
```diff
@@ -50,16 +43,17 @@ TSS2_RC PlatformCommand(
 {
     TSS2_TCTI_CONTEXT_INTEL *tcti_intel = tcti_context_intel_cast (tctiContext);
     int iResult = 0;            // used to return function results
-    char sendbuf[] = { 0x0,0x0,0x0,0x0 };
+    uint8_t sendbuf[] = { 0x0,0x0,0x0,0x0 };
```
tstruk: add space after commas like it is in the next line

include/tcti/common.h
```diff
@@ -1,43 +1,36 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
 #ifndef TCTI_COMMON_H
 #define TCTI_COMMON_H
 
-#include "sapi/tpm20.h"
-
 #if defined (__GNUC__)
```
tstruk: Do we need this file at all now?
Looks like the COMPILER_ATTR is also (re)defined in log/log.h.
Let's put it to include/sapi/tpm20.h and get rid of the include/tcti/common.h altogether.

tcti/tcti.c
flihp: I think you want `man 2 write`. Linux allows for a 0 return from `write` when nothing is written.

include/tcti/common.h
flihp: This is one of the eventual goals but I'm deferring this till a later cleanup patch. Hopefully things will be a bit more clear after some more internal cleanup.

flihp:
Thanks for the quick review @tstruk. Fixups pushed addressing everything except your comment w/r to `write` and the possibility of a 0 return value. Let me know how you want to proceed on that one.

PR #776, opened 2018-02-21
sapi: add TPM2PolicyAuthorizeNV command

tstruk:
Add support for TPM2_CCPolicyAuthorizeNV command.


include/sapi/tss2_tpm2_types.h
```diff
@@ -260,6 +260,7 @@ typedef  UINT32             TPM2_CC;
 #define  TPM2_CC_PolicyNvWritten               (TPM2_CC)(0x0000018f)
 #define  TPM2_CC_PolicyTemplate                (TPM2_CC)(0x00000190)
 #define  TPM2_CC_CreateLoaded                  (TPM2_CC)(0x00000191)
+#define  TPM2_CC_PolicyAuthorizeNV             (TPM2_CC)(0x00000192)
```
williamcroberts: Not for this patch-series, but do we really need this casts?

tstruk: Not really needed. This is how the command codes are defined in the spec.

Issue #768, opened 2018-02-15
make fails after 'distclean; ./configure'


pscheie:
When I run './configure; make; make install' on CentOS7, everything works as expected.  (It seems to work even if autoconf-archive is not installed.)  However, if I then run 'make distclean; ./configure; make', I get this error, as was reported in issue #728 (which is closed and I wasn't sure if adding a comment would reopen it or not):
...
sed -i -e "s,[@]VERSION[@],1.3.0,g;" man/man3/InitDeviceTcti.3
make: *** No rule to make target `man/man3/InitSocketTcti.3', needed by `all-am'.  Stop.

Is this expected behavior?

flihp:
Thanks for reporting this @pscheie. We're gonna roll out another 1.x release in the next few days and I'll get a fix for this into it.

flihp:
This is fixed in master via #777 :slot_machine:, it's queued up to get into 1.x via #779.

flihp:
#779 is merged, this will be in the next minor release. I'll have an RC out for that momentarily.

PR #777, opened 2018-02-22
build: Distribute the source for the InitSocketTcti man page.

flihp:
Before this patch we were shipping the man page after our build
transforms. This means running one of the clean targets will put the
source tree in a state where the build will fail since the source files
for the man page get 'cleaned'.


flihp:
this resolves #768 and needs to be cherry-picked into the 1.x branch before the next release.

PR #778, opened 2018-02-22
Tcti socket write cleanup

flihp:


flihp:
Getting the granularity right for these patches was hard. For the last PR I started out with some really ugly patches that had way too much stuff in them and then had to go back to refactor them into something sensible. This time around I went for smaller patches to save myself from having to rebase and refactor. Just for the sake of time I'm going to push these as-is and I'll try to get the balance better next time.

PR #780, opened 2018-02-23
Spec cleanup

flihp:
This brings our implementation of the 'v1' TCTI context structure up to speed with the current spec. We also add some of the new convenience macros / types from the new draft spec. Next will be an update to the v2 structure in the current draft.


flihp:
This changes API / ABI and will require a coordinated PR in the tpm2-abrmd (and maybe tools). It also needs to be rebased. I'll get all that situated before merging this.

flihp:
groan, cmocka website is down -> travis is failing

flihp:
cmocka is back up, now sourceforge is down ... and scan.coverity.com is still down after ~3 days

Issue #674, opened 2017-12-13
tss2_common.h


flihp:
Ensure file name, contents and install location match the spec.

PR #787, opened 2018-02-26
sapi: change version to the latest defined by spec

tstruk:
Currently the spec defines ABI version to TSS2_API_VERSION_1_2_1_108.
Update rest of the headers respectively.

### DIFF
--- a/esapi/esapi/Esys_ActivateCredential.c
+++ b/esapi/esapi/Esys_ActivateCredential.c
@@ -26,9 +26,9 @@
  ******************************************************************************/
 
 #include <sapi/tpm20.h>
-#ifndef TSS2_API_VERSION_1_1_1_1
+#ifndef TSS2_API_VERSION_1_2_1_108
 #error Version missmatch among TSS2 header files !
-#endif /* TSS2_API_VERSION_1_1_1_1 */
+#endif /* TSS2_API_VERSION_1_2_1_108 */
 #include "esys_types.h"
 #include <esapi/tss2_esys.h>
 #include "esys_iutil.h"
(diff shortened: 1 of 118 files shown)


Makefile.am
```diff
@@ -148,7 +148,7 @@ EXTRA_DIST = \
     lib/sapi.pc.in \
     man/man-postlude.troff \
     man/InitDeviceTcti.3.in \
-    man/man3/InitSocketTcti.3 \
```
martinezjavier: I don't think this belong to this commit, right? Instead you should cherry-pick from 1.x branch commit 784695d69c6e ("build: Distribute the source for the InitSocketTcti man page.").

martinezjavier:
@tstruk there seems to be a lot of style cleanups interleaved with the actual changes. Can you please split those in a different commit, otherwise makes it hard to read the commit diff.

Makefile.am
tstruk: Yes, that's some leftovers from Philip's PR that somehow got into my branch. Let me rebase. Thanks for catching that @martinezjavier

tstruk:
That garbage issue should be fixed now

tstruk:
@martinezjavier there are some style fixes and ABI version number updates. It's all of the header cleanup activity that we want to get this week to fix the open issue against headers starting from #674. I'll try to separate them for the next headers.

martinezjavier:
@tstruk fair enough. The changes looks good to me, but I still think that at least the commit message should mention the style cleanups. Something like:

```
Currently the spec defines ABI version to TSS2_API_VERSION_1_2_1_108.
Update rest of the headers respectively.

While being there, do some style cleanups and use C style comments instead of C++.
```

tstruk:
Comment updated.

PR #789, opened 2018-02-27
build: Fail configure script if unit tests enabled, but cmocka dep. c…

flihp:
…heck fails.

This also puts the dependency information into the INSTALL.md file to
give users a chance to get things right before the configure script
fails on them.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,7 +20,7 @@ The following are dependencies only required when building test suites.
 * Integration test suite (see ./configure option --with-simulatorbin):
 ** OpenSSL
 * Unit test suite (see ./configure option --enable-unit):
-** cmocka unit test framework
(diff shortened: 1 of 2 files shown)


flihp:
I took a look @ the dependency checking in autoconf and the dependencies in in the INTSALL.md as part of doing this. 100% with you on the need to plug some of these gaps but it goes way beyond cmocka. For example: we have a dependency on openssl by way of the integration test harness now but we have *no* check for it @ configure time as far as I could tell (may have missed something). I'm about to make an issue to track this.

flihp:
Also thanks for the quick review!

flihp:
whoops, I misspoke about the openssl dependency. That one works as expected. Still, I just created #790 to track this cleanup.

PR #791, opened 2018-02-27
INSTALL.md: Fix broken formatting / bullet for OpenSSL dependency.

flihp:


flihp:
This feels like a pretty silly patch but the broken formatting causes the layout to render OpenSSL dependency bullet on the same line as the previous dependency. This makes the the dependency pretty hard to find.

PR #805, opened 2018-02-28
tcti: Generalize version struct and add to INFO struct as first member.

flihp:
This addresses feedback from TSS2 WG: The INFO structure should be
versioned just like we do the context structure. The two version
structures must be identical for both structures.

### DIFF
--- a/include/sapi/tss2_tcti.h
+++ b/include/sapi/tss2_tcti.h
@@ -151,7 +151,9 @@ typedef TSS2_RC (*TSS2_TCTI_SET_LOCALITY_FCN) (
 typedef struct {
     uint64_t magic;
     uint32_t version;
-} TSS2_TCTI_CONTEXT_VERSION ;
+} TSS2_TCTI_VERSION ;
+
+typedef TSS2_TCTI_VERSION TSS2_TCTI_CONTEXT_VERSION;
 
 /* current version #1 known to this implementation */
 typedef struct {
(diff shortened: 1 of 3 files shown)


AndreasFuchsSIT:
I don't think that the info-struct requires the magic field...
Its purpose is different, i.e. binding the vfunc table to the context passed into it. So only the uint for version is actually required.

include/sapi/tss2_tcti.h
```diff
@@ -151,7 +151,9 @@ typedef TSS2_RC (*TSS2_TCTI_SET_LOCALITY_FCN) (
 typedef struct {
     uint64_t magic;
```
williamcroberts: Oh I didn't know we we're embedding a magic value, do we really need this, they tend to be superfluous...

flihp: yeah but it's in the spec so I just went with it ¯\_(ツ)_/¯

flihp: also allows someone to associate the info struct from a TCTI with the context returned by the init function. Not sure how useful this is but it's something.

AndreasFuchsSIT: I'd still vote for leaving it out (in code and in the spec)...
But we can discuss this in the WG...

flihp: If the plan is to take this up in the WG can you please approve this PR? We can update the code once the WG weighs in.

PR #809, opened 2018-02-28
log: simplify log level handling

tstruk:
Reuse existing loglevel values defined as enum log_level
instead of introducing new character mappings.
This will allow easy compile time enabing and disabling
of different log level by checking if the MAXLOGLEVEL is
greater of equal to configured level.


flihp:
the current love-letter we're getting from sourceforge w/r to the simulator tarball:
````
We're sorry -- the Sourceforge site is currently in Disaster Recovery mode, and currently requires
the use of javascript to function.  Please check back later.
````
We may want to consider using a mirror. I set one up on my domain over the weekend since sourceforge was down then as well: http://mirror.twobit.us/tpm2-deps/

tstruk:
@AndreasFuchsSIT The configure script accepts --with-maxloglevel option, which can be set to either  none, error, warning, info, debug, trace. The MAXLOGLEVEL is an internal variable that will be set accordingly. Even if user would set it as:
$ export MAXLOGLEVEL=10 the actual value that will be used in the build will be overwritten by the configure option.

tstruk:
@flihp should we switch to the mirror?

AndreasFuchsSIT:
I know... I am talking about alternative build systems than autotools, VS or embedded usage.
If you just use the log.h you will get a #error "MAXLOGLEVEL not set". If you -DMAXLOGLEVEL then it will set it to 1, i.e. Error, which is not the desired behaviour. 

FWIW: Unless you specifically denote the MAXLOGLEVEL you want with -DMAXLOGLEVEL='E' or similar, you should continue to get #error s...

Given this patch, that desired behaviour (desired by me at least) is not achieved any more. Therefore, I dislike this PR.

tstruk:
@AndreasFuchsSIT so we can skip the default value and have the build failing if not explicitly set some something that we support.

AndreasFuchsSIT:
Does 0 have a special meaning or occurrence similar to 1 ?
If not, I'll approve and merge...

tstruk:
0 means not defined, which is in case when no -DMAXLOGLEVEL or is provided.

AndreasFuchsSIT:
So omission of MAXLOGLEVEL means no logging ? That's bad...
Omission should either go to level trace, or to an error.

tstruk:
Yes, sorry, if no MAXLOGLEVEL is provided we will fail with:
#error "LOGMODULE must be set before including log/log.h"
-DMAXLOGLEVEL=0 will switch the logging off.

tstruk:
Right, this is the error I had in mind:
#error "Unknown MAXLOGLEVEL"

flihp:
I just merged #813. If this gets rebased you should get a green light from the CI.

PR #813, opened 2018-02-28
Travis: Deps from mirror instead of upstream

AndreasFuchsSIT:


flihp:
I just created #816 for a long-term solution. Till now this bandage will hold better than sourceforge hopefully.

PR #814, opened 2018-02-28
script: Add Sanity Checks to int-log-compiler.sh

### LINKED ISSUE #764: script/int-log-compiler.sh needs to detect degraded OS environments

dantpm:
Clarify that uriparser and libgcrypt dependencies require libraries and
header files.
Fix typos in INSTALL.md such as missing blank lines and markup.

Fixes #764.

### DIFF
--- a/script/int-log-compiler.sh
+++ b/script/int-log-compiler.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #;**********************************************************************;
-# Copyright (c) 2017, Intel Corporation
+# Copyright (c) 2017 - 2018, Intel Corporation
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


martinezjavier:
It also lacks a Signed-off-by tag.

dantpm:
Yes...it had both of those.  But in the usual round of a dozen merges and rebases or so, it lost that....
I'll resubmit (if it allows me).  Rebase works when there is no other commit in between.

AndreasFuchsSIT:
You seem to have a merge in your history once more...
Please never merge and be careful with pull... I recommend --ff-only...

dantpm:
I will create a new pull request tomorrow, since "git rebase" fails. Thanks for the help anyway!

Screen shot shown of one of hundreds of "git rebase" failures (in this case a simple "git rebase -i HEAD~2):

virtualbox_tumalo_28_02_2018_16_55_25-git-rebase-error

AndreasFuchsSIT:
@dantpm git push only allows for fast-forward stuff. You have to git push --force if you want to rewrite history...

Use with caution however, because --force allows you to completely screw up your remote...

Closing the PR per committer request...

dantpm:
Thanks---I will Google Fast Forward Merge or Rebase or Push.

Yes, I found that merges are to be avoided.  I am in the habit of keeping my repository sync'ed with the upstream main branch.

Thanks for the "git push --force" tip.  I found that I could not update the message/history if there are no changes (it would error out with a message about the github branch being 1 commit ahead, but it was the same commit that I had with only a different commitmessage.

Merge and rebase is a problem with files such as INSTALL.md that change multiple times a day in the middle of my rebase.

PR #815, opened 2018-03-01
Add TPMU NAME

tstruk:


marshal/tpmu-types.c
```diff
@@ -625,3 +625,18 @@ TPMU_UNMARSHAL2(TPMU_PUBLIC_PARMS,
...
+
+TPMU_MARSHAL2(TPMU_NAME,
+    0, ADDR, handle, marshal_null,
+    sizeof(TPM2_HANDLE), VAL, handle, Tss2_MU_UINT32_Marshal,
+    sizeof(TPM2_ALG_ID) + TPM2_SHA1_DIGEST_SIZE, ADDR, digest, Tss2_MU_TPMT_HA_Marshal,
+    sizeof(TPM2_ALG_ID) + TPM2_SHA256_DIGEST_SIZE, ADDR, digest, Tss2_MU_TPMT_HA_Marshal,
+    sizeof(TPM2_ALG_ID) + TPM2_SHA384_DIGEST_SIZE, ADDR, digest, Tss2_MU_TPMT_HA_Marshal,
+    sizeof(TPM2_ALG_ID) + TPM2_SHA512_DIGEST_SIZE, ADDR, digest, Tss2_MU_TPMT_HA_Marshal)
+TPMU_UNMARSHAL2(TPMU_NAME,
+    0, handle, unmarshal_null,
```
AndreasFuchsSIT: IMHO the case of size 0 must not be handled by the TPMU_NAME_Unmarshal function.
The spec gives TPMU_NAME with two options. The mentioning of size=0 is on the TPM2B description.
Same for the marshaling

So the surrounding application needs to do this.
```
TPM2B_NAME_Unmarshal(buffer, size, NULL, &tpm2b_name);
if (tpm2b_name.size != 0)
    TPMU_NAME_Unmarshal(&tpm2b_name.buffer, tpm2b_name.size, NULL, &tpmu_name)
```

tstruk: I have added a 0 case because the spec specifically call it in part 2, 10.5.3 TPM2B_NAME:
"If size is four, then the Name is a handle. If size is zero, then no Name is present. Otherwise...."

AndreasFuchsSIT: But that's for the TPM2B, which has a byte[sizeof(TPMU_NAME)] not a TPMU_NAME as a member field.

If the union should provide a null-case itself, then it would be specified by a line with "null" and a "NULL-selector. See e.g. Table 71-Definition of TPMU_HA Union.

tstruk:
0 case removed now.

PR #807, opened 2018-02-28
script: Add Sanity Checks to int-log-compiler.sh

dantpm:
Clarify that uriparser and libgcrypt dependencies require libraries and
header files.
Fix typos in INSTALL.md such as missing blank lines and markup.

Fixes #764.

PR #785, opened 2018-02-26
Add sanity checks to int-log-compiler.sh. Fixes #764

dantpm:
This addresses comments and issues from a previous pull request, #775

Tested with "make check-TESTS".
Negative tested by removing PID file and also killing TPM simulator process.
Minor corrections in INSTALL.md file.
Fixes #764.

Resubmitting this pull request as the previous pull request #783 failed the travis build failed

PR #783, opened 2018-02-23
Add sanity checks to int-log-compiler.sh. Fixes #764

dantpm:
This addresses comments and issues from a previous pull request, #775 
https://github.com/tpm2-software/tpm2-tss/pull/775/

Tested with "make check-TESTS".
Negative tested by removing PID file and also killing TPM simulator process.
Minor corrections in INSTALL.md file.
Fixes #764.

PR #775, opened 2018-02-21
Add sanity checks to int-log-compiler.sh. Fixes #764.

dantpm:
Tested with "make check-TESTS".
Negative tested by removing PID file and also killing TPM simulator process.

Minor corrections in INSTALL.md file.

PR #818, opened 2018-03-01
script: Add Sanity Checks to int-log-compiler.sh

### LINKED ISSUE #764: script/int-log-compiler.sh needs to detect degraded OS environments

dantpm:
int-log-compiler.sh needs to detect degraded OS environments.
Specifically "Bash for Ubuntu for Windows" partially emulates /proc,
so the ps and netstat commands fail to show processes and open sockets.

Fixes #764.


dantpm:
Previous pull requests #764 #775 #785 #765 #783 #807 #814

Issue #760, opened 2018-02-11
esapi pkgconfig file


flihp:
We need one. It should go in $(srcdir)/lib with the rest of them.

flihp:
Moved this over to the 2.0 cleanup tasks since we need it for the release

AndreasFuchsSIT:
Given the #827 merge, issue is done...

PR #827, opened 2018-03-05
Pkg config debt paydown

flihp:


flihp:
this resolves #760 + some extra cleanup

PR #836, opened 2018-03-07
Tcti cleanups

tstruk:
Tcti cleanups


flihp:
The statement in the commit message for cabf124:
```
The tcti_receive_checks doesn't allow resonse_buffer ptr to be NULL
so the checks after that don't make much sense.
```
is missing something. I think the check you're referring to is here: https://github.com/tpm2-software/tpm2-tss/blob/master/tcti/tcti.c#L94 and it doesn't allow the `response_buffer` AND the `size` parameter to be `NULL` at the same time.

flihp:
I guess my point is that per the spec, it's perfectly valid for the response buffer to be NULL so long as the size parameter isn't. This is used as a way to provide the caller with a way to query the required size for a call to receive. I added a test case for this in the device TCTI unit test but not the socket one. I'll get a PR together to add this test.

flihp:
I've added a test case for the situation described in my comment above: #839

tstruk:
Right, the sentence about tcti_receive_checks() isn't quite right. My point was that we should put the checks for resonse_buffer == NULL at the top of the funct instead of checking for it number of times below (in line 261 and line 295)

flihp:
Those checks are an unfortunate consequence of the previous authors use of `goto` and will have to stick around till that's refactored. I've added a test for the use case I mentioned above: https://github.com/tpm2-software/tpm2-tss/pull/839 and this code breaks that test.

flihp:
926cbe1 still needs to be filled in a bit. The test you've added is really just another assertion added to an existing test. This should probably be a separate test. You've also limited your change to the socket TCTI when the same change should be made to the device TCTI too.

PR #837, opened 2018-03-07
Coding standard include

flihp:
This adds a minor clarification to the whitespace section as well as additional requirements for the way we include headers.

### DIFF
--- a/doc/coding_standard_c.md
+++ b/doc/coding_standard_c.md
@@ -54,7 +54,8 @@ code would benefit from some restructuring.
 ## Whitespace
 All indentation must be spaces, not tabs. Lines are indented in multiples of
 4 spaces. Each line of code and documentation will end with a non-whitespace
-character.
+character. There must *not* be any whitespace between the last line of code
+or documentation in a file and the end of the file.
 
 ## Naming Variables, Functions and Other Stuff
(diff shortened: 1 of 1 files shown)


doc/coding_standard_c.md
```diff
@@ -163,11 +163,65 @@ within the source even if they are already included in that header. This
...
+header: `example.h`
+```
+/*
+ * BSD license block
+ */
+#ifndef HEADER_GUARD_H
+#define HEADER_GUARD_H
+
+#include <stdint.h>
+#include <stdio.h>
```
tstruk: stdio.h is a bad example here. No one should need it in TSS.

flihp: I get your point but no one should need `foo/bar.h` from further down in the example either :stuck_out_tongue:

I guess my point is that this is a slippery slope: The header guard doesn't match the file name. The headers for internal modules aren't real headers ... I'm just not sure where the bottom will be though.

Are you willing to limit your requested change to picking a more realistic header instead of `stdio.h` or do you intend for this to be more far-reaching?

flihp: cool thanks. Just fixed this up. Merging once I get a clean build

PR #841, opened 2018-03-08
[RFC] cleanup tpm2 type headers

flihp:
This patch series removes the `tpm2.h` file and combines the `tss2_tpm2_types.h` and `sys_api_part3.h` files. What I'm not certain about is what to do about the `TPM2B` type. We use this as a generic interface to all other `TPM2B_*` types but I can't find it in any of the TSS2 specs. Thoughts?


include/sapi/tss2_common.h
```diff
@@ -47,6 +47,11 @@ typedef int32_t     INT32;
 typedef uint64_t    UINT64;
 typedef int64_t     INT64;
 
+typedef struct {
+    UINT16 size;
+    BYTE buffer[1];
+} TPM2B;
+
```
AndreasFuchsSIT: I say we move this into an internal header, since it's not part of the public API...

flihp: I'm with you on this point but I couldn't come up with convenient place / header to put it. Definitely open to suggestions. It may be that we want to create a common place for this stuff since it's used across the type marshalling library, sapi and esapi. The only other code that's used so widely is the logging framework and that has it's own home under $(srcdir)/log. The `TPM2B` type doesn't belong there but we may just want to put both in $(srcdir)/util or something?

tstruk: sysapi/include/sysapi_util.h?

flihp: My issue is that it's used by esapi as well. I know that ESAPI uses SAPI internally but it should only use the SAPI through public interfaces. Having ESAPI and the marshalling library code depend on one of the internal SAPI util headers feels wrong to me.

flihp:
I've made an issue for the TPM2B thing: https://github.com/tpm2-software/tpm2-tss/issues/844

include/sapi/tss2_common.h
tstruk: But do have some headers that are used by everything (actually that's the sysapi_util.h)
Eventually we can put ti into src/sysapi_util.h and add things like TPM2B into it.

PR #843, opened 2018-03-08
tcti: don't send MS_SIM_CANCEL_OFF on every receive

tstruk:
TCTI receive sends a platform CANCEL_OFF command on every response
read, which isn't necessary. This changes it to only send CANCEL_OFF
if a CANCEL_ON command has been sent beforehand.

### DIFF
--- a/tcti/tcti.h
+++ b/tcti/tcti.h
@@ -109,6 +109,11 @@ typedef struct {
     tpm_header_t header;
     uint8_t locality;
 
+    /* Flag indicating if a command has been cancelled.
+     * This is a temporary flag, which will be changed into
+     * a tcti state when support for asynch operation will be added */
+    bool cancel;
+
     /* Sockets if socket interface is being used. */
     SOCKET otherSock;
     SOCKET tpmSock;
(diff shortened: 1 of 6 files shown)


tcti/tcti.h
```diff
@@ -100,6 +100,7 @@ typedef enum {
     TCTI_STATE_FINAL,
     TCTI_STATE_TRANSMIT,
     TCTI_STATE_RECEIVE,
+    TCTI_STATE_CANCEL,
```
flihp: You've added a new state to the state machine but you've added no information about what the state is for, how we get into this state or how we get out of it.

tstruk: Actually a new state wasn't the best idea. I have added a new flag which indicates if a previous command has been cancelled. I have added a test to cover this scenario and when paying with it I noticed that we faked a CANCEL rc from TPM into something else. This has been fixed also now.

AndreasFuchsSIT:
I have to admit, that I prefered the state-based approach. Since cancel() can only ever happen in state SENT and state CANCEL is accepted by receive().
But I'm not too keen on this either way...

tstruk:
@AndreasFuchsSIT the reason for the flag is that the state represents the state of the tcti interface and controls what operations can be performed in different state. The cancel is on top of that. It can happen at any time/stage via the platform command mechanism as well as with the Tss2_Tcti_Cancel() call and it changes the state of the, not the interface. We can perform cancel and still send additional commands. In this case the TPM just returns TPM2_RC_CANCELED (for some operations).

PR #845, opened 2018-03-08
tcti: Remove common.h

flihp:
This was no longer required when we removed the deprecated init
functions and the compiler attributes that marked them as deprecated.


flihp:
sorry for the churn here, my dev system had an older version of this file installed which caused my local builds to pass by falling back to the installed copy. should be good now

PR #849, opened 2018-03-09
tcti: Rename 'socket' TCTI to 'mssim'.

flihp:
The protocol implemented by this TCTI is specific to the TPM2 simulator
code from the TCG specs. This code is writtin & maintained by Microsoft.


liuqun:
We forgot to update pkg-config script tcti-mssim.pc:
```
$ pkg-config --libs tcti-mssim
-ltcti-socket
```
It is still referring to libtcti-socket.so
https://github.com/tpm2-software/tpm2-tss/blob/76555ac69f7215707f4338877cfd9aec18e9354a/lib/tcti-mssim.pc.in#L7

PR #852, opened 2018-03-12
libtcti-mssim: Fix regression in default conf string selection.

flihp:
When a `NULL` conf string is provided we're supposed to select a "sane"
default for the caller. Instead we're returning an error. This was
caused by a silly mistake made in checking the 'conf' parameter for
`NULL` in the wrong place.

This patch removes that check. We also add a TRACE level logging
statement to make future debugging a bit easier and a test case to
ensure a NULL conf string doesn't cause this regression again.

### DIFF
--- a/tcti/tcti-mssim.c
+++ b/tcti/tcti-mssim.c
@@ -508,13 +508,13 @@ Tss2_Tcti_Mssim_Init (
     char hostname[HOST_NAME_MAX + 1] = { 0 };
     uint16_t port = TCTI_SOCKET_DEFAULT_PORT;
 
+    LOG_TRACE ("tctiContext: 0x%" PRIxPTR ", size: 0x%" PRIxPTR ", conf: %s",
+               (uintptr_t)tctiContext, (uintptr_t)size, uri_str);
     if (tctiContext == NULL && size == NULL) {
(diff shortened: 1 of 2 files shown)


tcti/tcti-mssim.c
```diff
@@ -508,13 +508,13 @@ Tss2_Tcti_Mssim_Init (
     char hostname[HOST_NAME_MAX + 1] = { 0 };
     uint16_t port = TCTI_SOCKET_DEFAULT_PORT;
 
+    LOG_TRACE ("tctiContext: 0x%" PRIxPTR ", size: 0x%" PRIxPTR ", conf: %s",
+               (uintptr_t)tctiContext, (uintptr_t)size, uri_str);
     if (tctiContext == NULL && size == NULL) {
         return TSS2_TCTI_RC_BAD_VALUE;
-    } else if( tctiContext == NULL ) {
+    } else if (tctiContext == NULL) {
```
flihp: Didn't mean for this formatting fix to sneak in here. It's a fix we'd want but if you want me to refactor this into another commit just say the word.

tstruk: No big deal.

PR #857, opened 2018-03-14
GetHandle fixes

tstruk:
Fixes for two commands in getNumHandle


tstruk:
Yes, just sending it for review3. I will rebase on top of the reorg changes.

tstruk:
Rebased and merged since it has been already approved

PR #858, opened 2018-03-15
Util io refactor

flihp:
This PR moves all code interacting with file descriptors and `SOCKET`s from `$(srcdir)/src/tss2-tcti` to `$(srcdir)/src/util. There's a bit of clean up in related code as well.


flihp:
The names that have changed here are in function prototypes (I think you're referring to the parameters that have changed). None of the names of the types have changed, I'm just using a more appropriate type in a few places. The `_INTEL` suffix on the TCTI structure is left over from Will's tenure.

To your point about having separate structures for each TCTI: Prying the two apart has been a bit of a process and I've been refactoring various components to make separating the TCTI structures a clean operation (your reviews along the way have been very helpful). Previous PRs have set up a structured state machine, removed all of the random flag values in the structure (some where used by one TCTI, some the other) and now this PR cleans up all of the I/O code and moves it to our shiny new `$(srcdir)/src/tutil` module.

All of this work has been leading up to what I think is the next logical step: separating the two TCTI structures. If you like, I can include that work in this PR as well but I've been trying to keep from generating too many PRs with more than 2-3 patches since they get quite difficult to review after that point. I don't think this change will be difficult now that the code modules are more cleanly separated / defined.

Either way let me know what your preference is and I'll do my best to accommodate.

src/util/io.c
```diff
@@ -30,11 +30,65 @@
 #include <string.h>
 #include <unistd.h>
 
-#include "sockets.h"
-#include "tcti.h"
-#define LOGMODULE tcti
+#include "io.h"
+#define LOGMODULE io
```
tstruk: Having to many log modules may start to be problematic. Now we get the trance for what is being sent after we set TSS2_LOG=tcti+TRACE. After that we will need to have TSS2_LOG=tcti+TRACE,io+TRACE, which isn't too bad, but adding more may take some time to setup. We should keep the number of log modules to minimum.

flihp: I was thinking about this when I made the change but hadn't considered the state space explosion. I've never used anything more than `TSS2_LOG=all+TRACE` but that's not to say the issue doesn't exist. How about we just use a single logging module for everything in the util library? That way we have one for each library and one additional one for the utility code?

tstruk: That would work for now since we will only have the io and the log itself in utils, but with more code utils this will defeat the purpose of logs filtering. It should rather be a logical separation.

flihp: I'll flip this back to `tcti` for the log module since that's the only place where this code is used. We can deal with this issue when it becomes an issue. Will fixup and push.

tstruk: sounds good

AndreasFuchsSIT:
@flihp I guess doing in subsequent PRs is the better way now... ;-)

Issue #861, opened 2018-03-15
rename tpm20.h to tss2.h


flihp:
Providing a convenience header is nice but `tpm20.h` is the wrong name now that we've rebranded everything to `tss2`. This should be a simple rename but a lot of one line changes in every source file that includes this header. It may be a good opportunity to have source files just include the file that they need since we don't force people to use only this header any longer.

flihp:
with the way we've cleaned up the headers this convenience header is really not all that convenient. It's also promoting some bad habbits since it pulls in a pile of system headers. I've got an RFC PR ready to go that removes it completely. We'll see how bad that review goes.

flihp:
#862

liuqun:
A candidate C/C++ API header naming style:
```
#include <tpm2dot0/tss.h>
```
or:
```
#include <intel_tpm20_tss.h>
```

PR #862, opened 2018-03-15
headers: Remove tpm20.h header.

flihp:
This "convenience" header isn't part of the spec. It also pulled in a
number of headers that had nothing to do with the TPM or the spec:
stddef.h, stdint.h, stdlib.h, string.h etc. Now that we've cleaned up
the headers users consuming the following libraries only need to include
the related header:
- libtss2-mu, include tss2_mu.h
- libtss2-sys, include tss2_sys.h
- libtss2-esys, include tss2_esys.h

This obsoletes the tpm20.h file completely. All other changes in this
commit are fixing up our source code to include the right headers and
cleaning up other places where we were leaning on tpm20.h header to
include system headers for us (aka undoing the sloppy structure that
this "convenience" header encouraged / required).

There are also a number of source files that had no newline at the end
of the file. The script that I used to fix up the tpm20.h include in
these files added this, as most editors would do as well. I'm leaving
these in place baring objections during code review.


flihp:
I'm going to leave this open till @AndreasFuchsSIT has a chance to look at it on account of the tss2-esys files getting newlines. This change should probably go into whatever templating engine they're using.

flihp:
this resolves #861

PR #864, opened 2018-03-15
Update ABI defines

tstruk:


include/tss2/tss2_sys.h
```diff
@@ -44,10 +44,9 @@ extern "C" {
 #endif
 
 #define TSSWG_INTEROP 1
-#define MAX_NON_VENDOR_SPECIFIC 0x20000000
-#define TSS_SAPI_FIRST_FAMILY 1
+#define TSS_SAPI_FIRST_FAMILY 2
 #define TSS_SAPI_FIRST_LEVEL 1
-#define TSS_SAPI_FIRST_VERSION 1
+#define TSS_SAPI_FIRST_VERSION 108
```
AndreasFuchsSIT: I think, these have no place in the "official header".
The spec instead defines `#define TSS2_ABI_VERSION_CURRENT {1, 2, 1, 108}`

flihp: see section 3.6 of the tss2-sys spec: https://trustedcomputinggroup.org/wp-content/uploads/TSS_SAPI_Version-1.1_Revision-22_review_030918.pdf

tstruk:
@flihp @AndreasFuchsSIT  PR updated. I have a followup PR for Tools and ABRMD

liuqun:
https://github.com/tpm2-software/tpm2-tss/blob/20181514f4d13f3582802f97fb43759c1270ec59/include/tss2/tss2_common.h#L60

PR #865, opened 2018-03-15
Refactor the internal sys context struct

tstruk:


src/tss2-sys/sysapi_util.h
```diff
@@ -164,4 +145,4 @@ void TeardownSysContext(TSS2_SYS_CONTEXT **ctx);
 #ifdef __cplusplus
 }
 #endif
-#endif  // TSS2_SYSAPI_UTIL_H
+#endif
```
AndreasFuchsSIT: Maybe we should add "no uncommented #endif" to the coding conventions ?

tstruk: I'm not really keen of the #endif comments. Vi can find the corresponding #ifdefs, #elses and #endifs with <SHIFT>+<%>

AndreasFuchsSIT: @tstruk good for vi... Github does not support this though... ;-/
Approving anyways...

Issue #844, opened 2018-03-08
where to put TPM2B


flihp:
The TPM2B type is currently exposed through the public headers: https://github.com/tpm2-software/tpm2-tss/blob/master/include/sapi/tss2_common.h#L50

This type is meant to be an internal interface / abstraction used to access other TPM2B structures and shouldn't be in the installed headers. We don't have a convenient location for this type though. It's used by libtss2-mu, sapi and esapi and so putting it in a common utility directory would be preferred but we don't have one. The 'log' directory is the closest thing we have.

PR #866, opened 2018-03-16
headers: Move TPM2B type from tss2_common.h to $(srcdir)/src/util.

flihp:
This type isn't in the common header spec, or any TSS2 spec so it
definitely doesn't belong in any of the headers that get installed.
Since this is a type used by tss2-mu, sys and esys it now lives in
$(srcdir)/src/util/tpm2b.h.

To distribute this header in source tarballs via `make dist` it must be
added to the util library's SOURCE variable. To make this easier in the
future this commit adds code to the `bootstrap` script that populates
the UTIL_SRC variable with all source files under `$(srcdir)/src/util.`

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -93,8 +93,7 @@ tss2_HEADERS = \
 libutil = libutil.la
 noinst_LTLIBRARIES += $(libutil)
 libutil_la_CFLAGS = $(AM_CFLAGS)
-libutil_la_SOURCES = src/util/log.c src/util/log.h src/util/tss2_endian.h \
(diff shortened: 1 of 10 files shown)


flihp:
this resolves #844

AndreasFuchsSIT:
Ok, I tried to resolve the merge conflicts here on github, but apparently this is not meant to be used for rebasing...
Please rebase and push...

flihp:
I've never been able to make the merge web UI work the way I want / expect either. The rebase was easy enough. Will merge when the build turns green.

PR #873, opened 2018-03-19
Fix ESAPI includes

AndreasFuchsSIT:
Fix the includes of ESAPI and its tests.

### DIFF
--- a/test/integration/main-esapi.c
+++ b/test/integration/main-esapi.c
@@ -1,15 +1,42 @@
-#include <stdbool.h>
+/*******************************************************************************
+ * Copyright 2017, Fraunhofer SIT sponsored by Infineon Technologies AG
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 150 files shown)


test/integration/main-esapi.c
```diff
@@ -130,7 +157,7 @@ tcti_proxy_initialize(
     /* Init TCTI context */
     memset(tcti_proxy, 0, sizeof(*tcti_proxy));
     TSS2_TCTI_MAGIC (tctiContext) = TCTI_PROXY_MAGIC;
-    TSS2_TCTI_VERSION (tctiContext) = TCTI_VERSION;
+    TSS2_TCTI_VERSION (tctiContext) = 2;
```
tstruk: Why?

AndreasFuchsSIT: Because TCTI_VERSION is something internal to the tss2-tcti thing.
If anyone updates the tss2-tcti implementations to a version 3 I cannot expect them to know about this testcase that needs to implement the new function as well.
Thus, since I'm implementing v2 of the struct, I'm setting the version to 2...

AndreasFuchsSIT:
I rebased to master, which required a little change to the last patch.

test/integration/main-esapi.c
tstruk: Shouldn't this be:
TSS2_TCTI_VERSION (tctiContext) = TCTI_PROXY_VERSION;

AndreasFuchsSIT: Of course. You are right...
I'll fix this first thing tomorrow in the office...

AndreasFuchsSIT:
Ok, updated...

PR #882, opened 2018-03-21
ESAPI Change the default TCTI initialization

AndreasFuchsSIT:
Two very minor commit first that I just split out before the actual changes.

The ESAPI gains a new behavior for initializing the
default TCTI.

- The configure flag is now interpreted as a file
  instead of as a partial symbol name.
- If a default tcti is configured only this file is
  loaded.
- If no default tcti is configured, esys will attempt
  to load a tcti in the following order
  - Search for a libtss2-tcti-default.so
  - Search for a libtss2-tcti-tabrmd.so
  - Attempt the linked Tcti_Device for "/dev/tpmrm0"
  - Attempt the linked Tcti_Device for "/dev/tpm0"
  - Attempt the linked Tcti_Mssim for "localhost:2321"

@williamcroberts I'd like to get your opinion from the tpm2-tools perspective as well.

### DIFF
--- a/src/tss2-esys/esys_tcti_default.c
+++ b/src/tss2-esys/esys_tcti_default.c
@@ -1,58 +1,221 @@
-#include "tss2_mu.h"
-#include "tss2_sys.h"
-#include "tss2_tcti_device.h"
-#include "tss2_tcti_mssim.h"
+/*******************************************************************************
+ * Copyright 2017-2018, Fraunhofer SIT sponsored by Infineon Technologies AG
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ *******************************************************************************/
+
+#include <stdlib.h>
 #include <errno.h>
-#include <stdio.h>
 #include <string.h>
 #include <inttypes.h>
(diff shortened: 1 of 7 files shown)


src/tss2-esys/esys_tcti_default.c
```diff
@@ -1,56 +1,214 @@
...
+
+    handle = dlopen(file, RTLD_NOW);
+    if (handle == NULL) {
+        LOG_WARNING("Could not load TCTI file: %s", file);
+        return TSS2_ESYS_RC_BAD_REFERENCE;
+    }
+
+    infof = (TSS2_TCTI_INFO_FUNC) dlsym(handle, TSS2_TCTI_INFO_SYMBOL);
+    if (infof == NULL) {
+        LOG_ERROR("Info not found in TCTI file: %s", file);
```
tstruk: dlclose()?

src/tss2-esys/esys_tcti_default.c
```diff
@@ -1,56 +1,214 @@
...
+        LOG_ERROR("tcticontext must not be NULL");
+        return TSS2_TCTI_RC_BAD_REFERENCE;
+    }
+    *tcticontext = NULL;
+#ifdef ESYS_TCTI_DEFAULT_MODULE
+
+#ifdef ESYS_TCTI_DEFAULT_CONFIG
+    const char *config = _XSTR(ESYS_TCTI_DEFAULT_CONFIG);
+#else /* ESYS_TCTI_DEFAULT_CONFIG */
+    const char *config = "";
```
tstruk: should be const char *config = NULL;

src/tss2-esys/esys_tcti_default.c
```diff
@@ -1,56 +1,214 @@
...
+#endif /* ESYS_TCTI_DEFAULT_CONFIG */
+
+    LOG_DEBUG("Attempting to initialize TCTI defined during compilation: %s:%s",
+              ESYS_TCTI_DEFAULT_MODULE, config);
+    return tcti_from_file(ESYS_TCTI_DEFAULT_MODULE, config, tcticontext);
+
+#else /* ESYS_TCTI_DEFAULT_MODULE */
```
tstruk: Is this else needed. If ESYS_TCTI_DEFAULT_MODULE is defined we call tcti_from_file(ESYS_TCTI_DEFAULT_MODULE,..) and return.

src/tss2-esys/esys_tcti_default.c
```diff
@@ -1,56 +1,214 @@
...
- * Initialize a TCTI instance.
- * The caller is returned a TCTI context structure that is allocated by this
- * function. This structure must be freed by the caller.
- */
-TSS2_RC
-get_tcti_default(TSS2_TCTI_CONTEXT ** tcticontext)
+#define ARRAY_SIZE(X) (sizeof(X)/sizeof(X[0]))
+
+struct {
+    char *file;
```
tstruk: #ifndef NO_DL around file?

AndreasFuchsSIT: @tstruk I think I don't know what you mean ?

src/tss2-esys/esys_tcti_default.c
AndreasFuchsSIT: Exactly. Thus code-scanners could complain about unreachable code after return. Thus the #else...

src/tss2-esys/esys_tcti_default.c
```diff
@@ -1,56 +1,214 @@
...
+        } else {
+            LOG_ERROR("Errorous entry in standard TCTIs");
+            return TSS2_ESYS_RC_GENERAL_FAILURE;
+        }
+
+    }
+#endif /* ESYS_TCTI_DEFAULT_MODULE */
+
+    LOG_ERROR("No standard TCTI could be loaded");
+    return TSS2_ESYS_RC_NOT_IMPLEMENTED;
```
AndreasFuchsSIT: But I guess this needs to go before the #endif...

src/tss2-esys/esys_tcti_default.c
AndreasFuchsSIT: Done

src/tss2-esys/esys_tcti_default.c
AndreasFuchsSIT: done

src/tss2-esys/esys_tcti_default.c
tstruk: Just saying that we could also wrap the char *file member with #ifndef NO_DL:
```
struct {
    #ifndef NO_DL
    char *file;
    #endif
    TSS2_TCTI_INIT_FUNC init;
    TSS2_RC	+    char *conf;
    get_tcti_default(TSS2_TCTI_CONTEXT ** tcticontext)	+    char *description;
};
```

AndreasFuchsSIT: ok, done...

AndreasFuchsSIT:
@tstruk hope this is ok now...

PR #884, opened 2018-03-23
IPv6 Client Socket Support

### LINKED ISSUE #876: support IPv6 in mssim TCTI
Pretty much every distro now supports dual IPv4 and 6 stacks. We should accept conf strings in the mssim TCTI not just with v4 IP addresses, but also v6 like: `tcp://::1:2321/`, also see https://www.ietf.org/rfc/rfc2732.txt

dantpm:
Add IPv6 client socket support.
This results in dual-stack IPv4/IPv6 support for client socket connections.

Fixes #876.

### DIFF
--- a/src/util/io.c
+++ b/src/util/io.c
@@ -26,15 +26,20 @@
  */
 #include <errno.h>
 #include <inttypes.h>
+#include <limits.h>
+#include <netdb.h>
 #include <netinet/in.h>
 #include <string.h>
 #include <unistd.h>
+#include <stdio.h>
 
 #include "tss2_tpm2_types.h"
 
 #include "io.h"
 #define LOGMODULE tcti
 #include "util/log.h"
+
+#define MAX_PORT_STR_LEN    sizeof("65535")
 
 ssize_t
 read_all (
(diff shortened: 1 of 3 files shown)


src/util/io.c
```diff
@@ -140,33 +143,55 @@ socket_close (
     return TSS2_RC_SUCCESS;
 }
 
+static void uinttostring(unsigned int num, char *str) {
```
AndreasFuchsSIT: use itoa() or snprintf(), but don't reinvent the wheel please...

src/util/io.c
```diff
@@ -140,33 +143,55 @@ socket_close (
...
+
 TSS2_RC
 socket_connect (
     const char *hostname,
     uint16_t port,
     SOCKET *sock)
 {
-    struct sockaddr_in sockaddr = { 0 };
+    static const struct addrinfo hints = { .ai_socktype = SOCK_STREAM,
+        .ai_protocol = IPPROTO_TCP };
```
AndreasFuchsSIT: Just for clearity, please set .ai_family = AF_UNSPEC

src/util/io.c
```diff
@@ -140,33 +143,55 @@ socket_close (
...
-    sockaddr.sin_family = AF_INET;
-    sockaddr.sin_addr.s_addr = inet_addr (hostname);
-    sockaddr.sin_port = htons (port);
-
-    LOG_DEBUG ("Connecting socket %d to hostname %s on port %" PRIu16,
-               *sock, hostname, port);
-    ret = connect (*sock, (struct sockaddr*)&sockaddr, sizeof (sockaddr));
-    if (ret == -1) {
+    
+    LOG_DEBUG ("Connecting TCP stream socket to host %s on port %" PRIu16,
```
AndreasFuchsSIT: Please put some the LOG_DEBUG with the actual target node inside the loop instead. I think that gives way more information...

If you set the hints.ai_flags = AI_CANONIDN then p->ai_canonname should also be set correctly. Never used it myself, but should work...

dantpm: I added it, but, although it is useful, be warned that AI_CANONIDN is a non-standard glib-specific extension.

src/util/io.c
dantpm: OK, although generally libraries should not report errors, including using fprintf(). That's up to the application. But I see the rule is broken already for tpm2-tss, so I replaced it with snprintf().

Also itoa() does not exist in libc (although atoi does). Function itoa may exist in other libraries, but it is nonstandard.

src/util/io.c
dantpm: OK. That's the default (0), as with all fields, but it does make it explicit.

src/util/io.c
dantpm: OK. Because AI_CANONIDN is non-standard. It is a glib-specific extension, I had to add a #define __USE_GNU to break the standards enforcement in netdb.h.

src/util/io.c
```diff
@@ -146,27 +152,41 @@ socket_connect (
...
     SOCKET *sock)
 {
-    struct sockaddr_in sockaddr = { 0 };
+    static const struct addrinfo hints = { .ai_socktype = SOCK_STREAM,
+        .ai_family = AF_UNSPEC, .ai_protocol = IPPROTO_TCP,
+        .ai_flags=AI_CANONIDN };
+    struct addrinfo *retp = NULL;
+    struct addrinfo *pprev = NULL;
+    struct addrinfo *p;
+    char port_str[MAX_PORT_STR_LEN];
```
tstruk: char port_str[MAX_PORT_STR_LEN + 1]; sizeof doesn't count the terminating NULL.

dantpm: Yes it does count the terminating NUL character.  You may be confusing sizeof() with strlen().  Here's an example:

#include <stdio.h>
#include <string.h>
void main() {
        printf("strlen of \"foo\" is %lu\n", strlen("foo"));
        printf("sizeof of \"foo\" is %lu\n", sizeof("foo"));
}
(glacier ~/tmp) ./foo
strlen of "foo" is 3
sizeof of "foo" is 4

src/util/io.c
AndreasFuchsSIT: Well the actual point was to bring this LOG-Statement inside the `for`-loop, so we see which IPs the lib is attempting to connect to...
I guess a __USE_GNU addition would not be the nicest...
Google gives us an alternative: `inet_ntop` see below...

src/util/io.c
```diff
@@ -146,31 +152,48 @@ socket_connect (
...
-    sockaddr.sin_port = htons (port);
-
-    LOG_DEBUG ("Connecting socket %d to hostname %s on port %" PRIu16,
-               *sock, hostname, port);
-    ret = connect (*sock, (struct sockaddr*)&sockaddr, sizeof (sockaddr));
-    if (ret == -1) {
+    
+    LOG_DEBUG ("Connecting TCP stream socket to host %s on port %" PRIu16,
+               hostname, port);
+    for (p = retp; p != NULL; pprev = p, p = p->ai_next) {
```
AndreasFuchsSIT: FWIW, Somewhere at the top with attribute unused because of MAXLOGLEVEL=NONE:
```
char buff[255] __attribute__((unused));
```
Here:
```
LOG_DEBUG("Connecting to %s", inet_ntop(p->ai_family, p->ai_addr, &buff[0], sizeof(buff)));
```

src/util/io.c
```diff
@@ -146,31 +152,48 @@ socket_connect (
...
+        socket_close(sock);
+    }
+    freeaddrinfo (retp);
+    if (p == NULL) {
         LOG_WARNING ("Failed to connect to host %s on port %" PRIu16 ", errno "
-                     "%d: %s", hostname, port, errno, strerror (errno));
-        socket_close (sock);
+                     "%d: %s", (pprev != NULL && pprev->ai_canonname != NULL) ?
+                      pprev->ai_canonname : hostname,
+                      port, errno, strerror (errno));
```
AndreasFuchsSIT: For this purpose, I'd say that the flag `AI_CANONNAME` and printing `retp->ai_canonname` are more meaningful.

Of course only before the freeaddrinfo, as it should have been here anyways....

src/util/io.c
dantpm: I added this, inside the for() loop.
Connection attempt failures to multi-homed hosts are not an error, but since this is just a debug message it's OK to be in the loop.

dantpm:
I moved the LOG_DEBUG message inside the loop to print every hostname connection attempted.

I also removed the GNU-specific, non-standard  AI_CANONNAME and ai_canonname  and used inet_ntop() as suggested.

dantpm:
Rebased with upstream and addressed all comments.

AndreasFuchsSIT:
@dantpm after #910 this is now conflictig... Sorry, you'll have to rebase again...

dantpm:
I just rebased yet again. It has the changes from #910.

dantpm:
@tstruk, Philip just mentioned that @AndreasFuchsSIT is on vacation.  The last message he left is that I needed to do another rebase before integration, but never did a pull.
All the issues are addressed.

Could you look at it and do a pull?

tstruk:
@dantpm merged. One minor comment. In io.c line 192 comment uses C++ style:
+            break; // socket connected OK
Please use C style comments.
I will update this one.

dantpm:
I fixed it and rebased again.

Issue #879, opened 2018-03-20
Startup error with SLB9670 device


svenschwermer:
I have already opened an issue in the abrmd repo, but I'm suspecting this to be a problem in tss. For reference: https://github.com/tpm2-software/tpm2-abrmd/issues/394

I can't issue the startup command (via abrmd or tools) to the SLB9670 using the latest from master. This works with the software from the 1.x tags. When I try running the startup command, the TPM responds with error 0x144. In text form, this is `the value of authorizationSize is out of range or the number of octets in the Authorization Area is greater than required`. I have tried the kernels 4.9.80 and 4.14.27 (targeted /dev/tpm0 and /dev/tpmrm0).

AndreasFuchsSIT:
Hmmm... The thing is, I would expect this to fail since the kernel already did a StartUp, just with a different RC, i.e. 0x100 instead of 0x144...

What about other tools commands, such as tpm2_getrandom ? Do they work ?

AndreasFuchsSIT:
Could you then please call `TSS2_LOG=all+trace tpm2_startup` with the latest master ?

svenschwermer:
```shell
$ sudo TSS2_LOG=all+trace tools/tpm2_startup -T device
trace:marshal:src/tss2-mu/base-types.c:197:Tss2_MU_TPM2_ST_Marshal() Marshalling TPM2_ST from 0x7e862546 to buffer 0x1261518 at index 0x0
trace:marshal:src/tss2-mu/base-types.c:197:Tss2_MU_TPM2_ST_Marshal() offset parameter non-NULL, updated to 2
trace:marshal:src/tss2-mu/base-types.c:189:Tss2_MU_UINT16_Marshal() Marshalling UINT16 from 0x7e86257e to buffer 0x1261518 at index 0xa
trace:marshal:src/tss2-mu/base-types.c:189:Tss2_MU_UINT16_Marshal() offset parameter non-NULL, updated to 12
trace:tcti:src/tss2-tcti/tcti-device.c:64:tcti_device_transmit() sending 12 byte command buffer: (size=12):
	80010000000c0000
	01440001
trace:tcti:src/util/io.c:78:write_all() writing 12 bytes starting at 0x1261518 to fd 3
trace:tcti:src/util/io.c:83:write_all() wrote 12 bytes to fd 3
debug:tcti:src/tss2-tcti/tcti-device.c:107:tcti_device_receive() Header not yet received, reading 10 byte header from fd 3
[...]

PR #890, opened 2018-03-26
tcti-device: Copy response header to output buffer in receive function.

flihp:
This commit resolves a bug in the device TCTI where the receive function
read the response header from the TPM but neglects to copy this to the
buffer provided by the caller. The unit tests for this module were
unfortunately not catching this misbehavior.

For this fix we add a function to the tcti module to assist in
marshalling the response header back out to the caller provided buffer.
The unit test for the device TCTI has also been updated to compare the
response buffer received by the client to the pre-canned response buffer
that the mock framework uses when mocking the 'read' function.

### DIFF
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -120,8 +120,8 @@ tcti_device_receive (
                   tcti_intel->header.size);
     }
 
-    *response_size = tcti_intel->header.size;
     if (response_buffer == NULL) {
+        *response_size = tcti_intel->header.size;
         LOG_DEBUG ("response_buffer is null, returning size: %zd", *response_size);
         return TSS2_RC_SUCCESS;
     }
@@ -131,6 +131,8 @@ tcti_device_receive (
                      *response_size, tcti_intel->header.size);
         return TSS2_TCTI_RC_INSUFFICIENT_BUFFER;
     }
+    *response_size = tcti_intel->header.size;
+
     rc = header_marshal (&tcti_intel->header, response_buffer);
     if (rc != TSS2_RC_SUCCESS) {
         return rc;
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -112,7 +112,7 @@ tcti_device_receive (
             goto trans_state_out;
         }
         LOGBLOB_DEBUG (header_buf, TPM_HEADER_SIZE, "Response header received");
-        rc = parse_header (header_buf, &tcti_intel->header);
+        rc = header_unmarshal (header_buf, &tcti_intel->header);
         if (rc != TSS2_RC_SUCCESS) {
             return rc;
         }
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -131,9 +131,13 @@ tcti_device_receive (
                      *response_size, tcti_intel->header.size);
         return TSS2_TCTI_RC_INSUFFICIENT_BUFFER;
     }
+    rc = header_marshal (&tcti_intel->header, response_buffer);
+    if (rc != TSS2_RC_SUCCESS) {
+        return rc;
+    }
     /* Read the rest of the response, minus the header that we already jave. */
     size = read_all (tcti_intel->devFile,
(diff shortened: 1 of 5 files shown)


flihp:
this resolves the issue discussed in #879

src/tss2-tcti/tcti-device.c
```diff
@@ -131,9 +131,13 @@ tcti_device_receive (
                      *response_size, tcti_intel->header.size);
         return TSS2_TCTI_RC_INSUFFICIENT_BUFFER;
     }
+    rc = header_marshal (&tcti_intel->header, response_buffer);
```
tstruk: Do we need the header_buf at all? We could just read the header into to the response_buffer and 
get the header from it: parse_header (response_buffer, &tcti_intel->header);
This way we would not need the header_marshal at all.

martinezjavier:
@flihp while reading the code to review this, I noticed that another flaw in the[ tcti_device_receive()](https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-tcti/tcti-device.c#L82) logic:

```
    *response_size = tcti_intel->header.size;
    if (response_buffer == NULL) {
        LOG_DEBUG ("response_buffer is null, returning size: %zd", *response_size);
        return TSS2_RC_SUCCESS;
    }
    if (*response_size < tcti_intel->header.size) {
        LOG_WARNING ("Size of user supplied response buffer %zd is less than "
                     "the size of the response buffer: %" PRIu32,
                     *response_size, tcti_intel->header.size);
        return TSS2_TCTI_RC_INSUFFICIENT_BUFFER;
    }
```

I believe you want to set ` *response_size = tcti_intel->header.size` after you have checked for `*response_size < tcti_intel->header.size`.

I also think that found an issue in [read_all()](https://github.com/tpm2-software/tpm2-tss/blob/master/src/util/io.c#L40):

```
    do {
        recvd = TEMP_RETRY (read (fd, &data [recvd_total], size));
        if (recvd < 0) {
            LOG_WARNING ("read on fd %d failed with errno %d: %s",
                         fd, errno, strerror (errno));
            return recvd_total;
        }
        LOGBLOB_DEBUG (&data [recvd_total], recvd, "read %zd bytes from fd %d:", recvd, fd);
        recvd_total += recvd;
        size -= recvd;
    } while (size > 0);
```

If read(2) returns 0, then this will be an infinite loop. Probably that will be a bug, but still is a case that should be handled.

Maybe you can fix those as part of this pull request? Or if you want I can propose fixes as a follow-up.

src/tss2-tcti/tcti-device.c
AndreasFuchsSIT: Yes, because an application may call tcti_receive() with buffer==NULL.

flihp:
thanks @martinezjavier. The first issue you point out is one that I introduced in this patch. Thought I was being clever but I missed the issue you noted. Looks like another unit test waiting to be written. The second issue is on the read path and wasn't introduced here (I hope) so that should probably get it's own issue and PR.

martinezjavier:
@flihp both issues are not related to your patch in fact, and present in current master. I noticed when reading the current code to understand what your change was about. So mentioned you wanted to do some preparatory patches since you were already touching that part of the code.

But as mentioned those can be done as a follow-up, I can write patches for these after your changes are merged.

src/tss2-tcti/tcti-device.c
flihp: When this code was being refactored I tried both of these approaches. Unmarshalling and storing the header this way just looked better than unmarshalling the size field from a buffer when we needed it. If we're looking to be more efficient we could get away with using 6 instead of 10 bytes since we don't do anything with the RC but I didn't see any harm in storing the whole header.

flihp:
You're right. I remember moving around the code that deals with the response size but I must have moved it back. I'll sort that one out here since we've already got the context for it. Will be in a second commit though.

flihp:
Based on review comments I've added a new commit that fixes another bug identified by @martinezjavier above in the setting of the `response_size` parameter in the `receive` function. This commit also adds a test case to ensure the proper behavior is preserved. Additionally I've renamed the `parse_header` function so that it's consistent with the `header_marshal` function added in the first commit.

Issue #772, opened 2018-02-16
coverity auth tokens aren't working


flihp:
travis-ci integration with coverity is borked.

flihp:
this is blocked on the coverity scan service being down "for scheduled maintenance" ... but it's been down for ~5 days now ☹️

flihp:
scan.coverity.com is going on ~3 weeks of down-time. This is a really bad look for the service but I don't know of any alternatives other than using the clang sanitizers. How comparable these are is probably debatable but it's better than nothing.

williamcroberts:
I know this is terrible

flihp:
got an automated message from coverity scan site. They've been back up for a few days now. Kicking off a scan ...

flihp:
Well, at least we're getting "access denied" errors now: https://travis-ci.org/tpm2-software/tpm2-tss/jobs/358717142. I can log into the UI as well so that's a good sign.

flihp:
Travis can talk to coverity now according to: https://travis-ci.org/tpm2-software/tpm2-tss/jobs/358720794#L1072, doesn't look like it did the scan though.

flihp:
```
The repository is not yet known by Travis CI. Please sync your repository list on the Travis CI service, enable it for sync, and return to this page for configuration details.
```

http://scan.coverity.com goes away for a month and it forgets who we are ... typical

flihp:
Got a scan to upload to coverity service. Travis is still failing though since coverity seems to have rearranged where their stuff runs during the CI build.

flihp:
Finally got a clean travis build on the `coverity_scan` branch. Working up a PR for the required changes ...

flihp:
PR #893 gets coverity back

PR #893, opened 2018-03-27
ci: Get coverity running again accounting for changes after the outage.

flihp:
It looks like there have been some upstream changes in the way coverity
and travis are integrated. Previously the coverity scan build was done
after all other build / test stuff. This made sense since we only want
to fix issues found by coverity for builds that actually build. After
the scan.coverity.com service came back online, the coverity builds
happen before all other tasks in the CI.

Given all of this we have to change a few things around. This commit
adds the necessary commands to bootstrap the source tree, then do a
build instrumented with the coverity stuff in a VPATH. Doing the
coverity build in a VPATH is necessary as it makes the other builds
done by the CI *after* the scan build possible. Autotools doesn't like
configuring a VPATH build from a source tree that has already been
configured.

All scan tokens appear to have been refreshed and so we've created a new
encrypted variable for it. Finally the scan service is also rejecting
access if the `name` for the project is reported as anything other than
`01org/TPM2.0-TSS`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ dist: trusty
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


flihp:
this resolves #772

tstruk:
Looks like the recommended way to resolve the autotools issue is to modify the script section with:

script:
  - if [[ "${COVERITY_SCAN_BRANCH}" == 1 ]];
      then
        echo "Don't build on coverty_scan branch.";
        exit 0;
    fi
  - make

.travis.yml
```diff
@@ -30,10 +31,10 @@ addons:
     - libssl-dev
   coverity_scan:
     project:
-      name: "01org/tpm2-tss"
+      name: "01org/TPM2.0-TSS"
       description: Build submitted via Travis-CI
     notification_email: philip.b.tricca@intel.com
-    build_command_prepend: "make clean"
+    build_command_prepend: "./bootstrap && mkdir cov-build && pushd cov-build && ../configure"
```
tstruk: no need to mkdir here. Just ./bootstrap && ./configure
and modify script: with this
'''
script:
  - if [[ "${COVERITY_SCAN_BRANCH}" == 1 ]];
      then
        echo "Don't build on coverty_scan branch.";
        exit 0;
    fi
  - make
'''

flihp:
Recommended by who? Got a link?

flihp:
I get it. What I did was preserve all the other build / test stuff which, given the purpose of the branch, is wasteful. So when `COVERITY_SCAN_BRANCH` is set the normal build path just exits since the coverity stuff is already done. That'll save ~7 minutes.

Traveling tomorrow, will fixup once I reach my destination.

flihp:
I rebased this with a 'wip' tag on a commit to test code based on feedback from @tstruk. The wip commit was squashed but it still shows up in the PR GUI for some reason though it's no longer present. To get a clean PR I'm going to close this one and re-open a new one from the coverity_scan branch in the tpm2-tss repo.

PR #894, opened 2018-03-27
Some reworks about libtss2-mu

### LINKED ISSUE #847: Initialize TPMS_ and TPML_ to 0 on _Unmarshal ?
There is this weirdness that TPM2B_s' `dest` parameters need to be initialized, because their `size`-field is tested for `==0`. This is because of the sad "underallocation logic".

For TPMS_ and TPML_ this is not the case from the spec. Therefore it should be possible to pass in uninitialized variables as `dest`.

Currently, if a TPMS_ or TPML_ contains a TPM2B_, the outer unmarshaling will call the inner unmarshalling directly. This again leads to access to uninitialized memory.

Instead, TPMS_ and TPML_ unmarshalling should memset the `dest` parameter first and then call the embedded unmarshaling.

AndreasFuchsSIT:
Fixes #847 

Fixes some new bug in base-type marshaling as well.

### DIFF
--- a/src/tss2-mu/tpml-types.c
+++ b/src/tss2-mu/tpml-types.c
@@ -113,7 +113,6 @@ TSS2_RC Tss2_MU_##type##_Unmarshal(uint8_t const buffer[], size_t buffer_size, \
     size_t  local_offset = 0; \
     UINT32 i, count = 0; \
     TSS2_RC ret = TSS2_RC_SUCCESS; \
-    type local_dst; \
 \
     if (offset != NULL) { \
         LOG_TRACE("offset non-NULL, initial value: %zu", *offset); \
@@ -142,9 +141,6 @@ TSS2_RC Tss2_MU_##type##_Unmarshal(uint8_t const buffer[], size_t buffer_size, \
          (uintptr_t)dest, \
          local_offset); \
 \
-    if (dest == NULL) \
-        dest = &local_dst; \
-\
     ret = Tss2_MU_UINT32_Unmarshal(buffer, buffer_size, &local_offset, &count); \
     if (ret) \
         return ret; \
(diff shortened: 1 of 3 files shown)


src/tss2-mu/tpml-types.c
```diff
@@ -113,7 +113,6 @@ TSS2_RC Tss2_MU_##type##_Unmarshal(uint8_t const buffer[], size_t buffer_size, \
     size_t  local_offset = 0; \
     UINT32 i, count = 0; \
     TSS2_RC ret = TSS2_RC_SUCCESS; \
-    type local_dst; \
```
tstruk: This is still needed. The Marshaling spec says that is dest is NULL and offset is not NULL the unmarshaling function needs to set the offset to size that would be needed to unmarshal given type. In this case it needs the local_dst var to unmarshal it and to obtain the size.

src/tss2-mu/tpms-types.c
```diff
@@ -39,210 +39,129 @@
...
+         " at index 0x%zx", (uintptr_t)&src, (uintptr_t)buffer, \
+         offset?*offset:0xffff); \
+\
+    if (src->sizeofSelect > TAB_SIZE(src->pcrSelect)) { \
+        LOG_ERROR("sizeofSelect value %"PRIu8"/%zi too big", src->sizeofSelect, \
+                  TAB_SIZE(src->pcrSelect)); \
+        return TSS2_SYS_RC_BAD_VALUE; \
+    } \
+\
+    before \
```
tstruk: This "before" looks weird. Can we change it to something that will look more like a macro:
```
#define TPMS_PCR_MARSHAL(type, tag) \
                                  size_t buffer_size, size_t *offset)	+TSS2_RC \
{	+Tss2_MU_##type##_Marshal(const type *src, uint8_t buffer[], \
    TPMS_PCR_SELECT *pcrSelect = (TPMS_PCR_SELECT *)ptr;
                         size_t buffer_size, size_t *offset) \
{ \
....
ret = tag; \
...
}\
```
and then:
```
TPMS_PCR_MARSHAL(TPMS_PCR_SELECT,  TSS2_RC_SUCCESS)
TPMS_PCR_MARSHAL(TPMS_PCR_SELECTION, \
Tss2_MU_TPMI_ALG_HASH_Marshal(src->hash, buffer, buffer_size, &local_offset))
TPMS_PCR_MARSHAL(TPMS_TAGGED_PCR_SELECT, \
Tss2_MU_UINT32_Marshal(src->tag, buffer, buffer_size, &local_offset))
```

src/tss2-mu/tpml-types.c
AndreasFuchsSIT: Yes, and that still happens... We merely call the subfield functions with dest=NULL as well...

src/tss2-mu/tpms-types.c
AndreasFuchsSIT: sure... we can do that...

AndreasFuchsSIT:
@tstruk hope this is ok now. IMO the dest==NULL thing works the way I implemented it...

PR #901, opened 2018-03-28
Rework tests wrt transient objects

AndreasFuchsSIT:
This is the very first step towards having the tests run on actual TPMs.
In this step we
- Fix leaking integration tests
- Move TPM_startup into the simulator test harness
- Add a guard around integrations tests to look for transent objects before  and after the test.

### DIFF
--- a/test/integration/asymmetric-encrypt-decrypt.int.c
+++ b/test/integration/asymmetric-encrypt-decrypt.int.c
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <inttypes.h>
 
 #include "tss2_tcti.h"
 #include "tss2_sys.h"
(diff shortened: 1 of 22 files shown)


martinezjavier:
@AndreasFuchsSIT I've been told in the past that the tpm2-tss tests are not suitable to be run on real TPM devices due their destructive nature. And in fact here's a big scary warning in https://github.com/tpm2-software/tpm2-tss/blob/master/README.md#test-suite.

Could you please elaborate a little bit about this in the commit message? Are this warning only referring to the unit tests or also the integration tests?

AndreasFuchsSIT:
@martinezjavier That warning is still totally true...
I'm starting the effort to enable running the testsuit on real TPMs.
I'll have to do a few more things, but then I'll also edit the README.md accordingly...

martinezjavier:
@AndreasFuchsSIT got it, thanks a lot for the clarification.

test/integration/asymmetric-encrypt-decrypt.int.c
```diff
@@ -140,6 +141,17 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
     LOG_INFO("Decrypt successed.");
 
     LOG_INFO("Asymmetric Encryption and Decryption Test Passed!");
+
+    rc = Tss2_Sys_FlushContext(sapi_context, sym_handle);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOG_ERROR("Tss2_Sys_FlushContext failed with 0x%"PRIx32, rc);
+        return 99;
```
tstruk: Could we add some comments about the 99 & 77 return codes?

AndreasFuchsSIT: done

AndreasFuchsSIT:
@tstruk updated...

PR #911, opened 2018-03-29
test: esapi: enable esapi single tests

tstruk:
All ESAPI tests are all passing fine as they run it the harness,
but when run as a single tests they all fail with error code 0x00000100:
"TPM not initialized by TPM2_Startup or already initialized"

This is bacause the Tss2_Sys_Startup is only called from main in
test/integration/main.c, but not from main in test/integration/main-esapi.c

Add Esys_Startup() into test/integration/main-esapi.c to enable
invocation of the esapi test as single tests.

### DIFF
--- a/test/integration/main-esapi.c
+++ b/test/integration/main-esapi.c
@@ -231,12 +231,18 @@ TSS_SAPI_FIRST_VERSION };
         return 1;
     }
     rc = Esys_Initialize(&esys_context, tcti_context, &abiVersion);
-    if (rc != TSS2_RC_SUCCESS && rc != TPM2_RC_INITIALIZE) {
+    if (rc != TSS2_RC_SUCCESS) {
         LOG_ERROR("Esys_Initialize FAILED! Response Code : 0x%x", rc);
         return 1;
     }
-    rc = Esys_SetTimeout(esys_context, TSS2_TCTI_TIMEOUT_BLOCK);
+    rc = Esys_Startup(esys_context, TPM2_SU_CLEAR);
     if (rc != TSS2_RC_SUCCESS && rc != TPM2_RC_INITIALIZE) {
+        LOG_ERROR("Esys_Startup FAILED! Response Code : 0x%x", rc);
+        return 1;
+    }
+
+    rc = Esys_SetTimeout(esys_context, TSS2_TCTI_TIMEOUT_BLOCK);
+    if (rc != TSS2_RC_SUCCESS) {
         LOG_ERROR("Esys_SetTimeout FAILED! Response Code : 0x%x", rc);
         return 1;
     }
--- a/test/integration/main.c
+++ b/test/integration/main.c
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 
 #define LOGMODULE test
+#include "tss2_sys.h"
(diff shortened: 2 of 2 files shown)


test/integration/main-esapi.c
```diff
@@ -231,16 +231,22 @@ TSS_SAPI_FIRST_VERSION };
...
     rc = Esys_Initialize(&esys_context, tcti_context, &abiVersion);
-    if (rc != TSS2_RC_SUCCESS && rc != TPM2_RC_INITIALIZE) {
+    if (rc != TSS2_RC_SUCCESS) {
         LOG_ERROR("Esys_Initialize FAILED! Response Code : 0x%x", rc);
         return 1;
     }
-    rc = Esys_SetTimeout(esys_context, TSS2_TCTI_TIMEOUT_BLOCK);
+    rc = Esys_Startup(esys_context, TPM2_SU_CLEAR);
     if (rc != TSS2_RC_SUCCESS && rc != TPM2_RC_INITIALIZE) {
         LOG_ERROR("Esys_SetTimeout FAILED! Response Code : 0x%x", rc);
```
dantpm: Change error message from "Esys_SetTimeout FAILED!" to "Esys_Startup FAILED!"

AndreasFuchsSIT:
I wonder why you get this problem...

test/integration/main-esapi.c
tstruk: Done

tstruk:
@AndreasFuchsSIT Bacause the Tss2_Sys_Startup is only called from main in
test/integration/main.c, but not from main in test/integration/main-esapi.c
Restart your tpm simulator and then try to run a esapi test. You will see the problem.

tstruk:
@AndreasFuchsSIT extracting TPM2_Startup into int-log-compiler.sh will not help because you still will not be able to run a single esapi test after the simulator is restarted. If fact it will make it worse because we will not be able to call any test outside of the test harness.

tstruk:
Just start your simulator manually and run:
$ test/integration/esys-hmacsequencestart.int
ERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:164:Esys_CreatePrimary() Esys Finish ErrorCode (0x00000100) 
ERROR:test:test/integration/esys-hmacsequencestart.int.c:121:test_invoke_esapi() Error: CreatePrimary ErrorCode (0x00000100)

tstruk:
I have added Tss2_Sys_Startup() back to test/integration/main.c also.

AndreasFuchsSIT:
I think this is a bad idea...
The idea was, that a test should be run against a "startet TPM". Starting the TPM should be done outside of the test, so it can also run with a real TPM without spamming StartUps...

just call test/helpers/tpm_startup before you start the test...

AndreasFuchsSIT:
@tstruk I would have prefered some discussion before you revert parts of PR #901 ...

tstruk:
Sending a Startup command before every test doesn't do any harm. The TPM will just return "already started" response. From the other hand each test should be a self contained, and should not depend on any other test, nor assume any state of the TPM. We have that in tpm_client already and we don't want to make the same mistake.

PR #914, opened 2018-03-30
ci: Get coverity running again accounting for changes after the outage.

flihp:
We now short-circuit the build if the COVERITY_SCAN_BRANCH is set in the
'script' section of .travis.yml. At this point the coverity build has
already happened and there's no point in running the rest of our typical
build / test cycle.

The number of `jobs` used in the coverity build has been reduced from
`$(nproc)*2` to just `$(nproc)` to align with the rest of the builds.
All scan tokens appear to have been refreshed and so we've created a new
encrypted variable for it. Finally the scan service is also rejecting
access if the `name` for the project is reported as anything other than
`01org/TPM2.0-TSS`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ dist: trusty
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


flihp:
This is a cleaned up version of #893 done from a branch in the tpm2-tss repo.

PR #919, opened 2018-04-02
tss2-sys: Tss2_Sys_Finalize now returns void, remove '_Complete'.

flihp:
According to the spec `Tss2_Sys_Finalize` returns void. Also the spec
does not include an async version of the Finalize function and so the
`Tss2_Sys_Finalize_Complete` function makes no sense.


flihp:
this resolves #919

Issue #887, opened 2018-03-26
Add separate TCTI_MAGIC for tcti_mssim and tcti_dev


tstruk:
Do we need to have two separate TCTI_MAGICs. Can one initialize a tcti object with Tss2_Tcti_Device_Init() and then call tcti_platform_command() on it? Also does the tcti_platform_command() function need to validate TCTI_MAGIC in similar way to how transfer() or recv() do?

flihp:
I don't think this makes sense till we give each TCTI a separate context structure. I should have made a ticket for creating separate structures back when @AndreasFuchsSIT asked about it in some other thread a few weeks back. I'll do that now and link the issue to this one.

PR #921, opened 2018-04-03
tcti: Use separate structures for the device and mssim TCTI contexts.

flihp:
Using the same structure for both the device and mssim TCTI was lazy. We
were also using the same magic number for both modules which had the
potential to cause confusion if one library was passed a TCTI
initialized by the other. This patch resolves both of these issues by
breaking the old TSS2_TCTI_CONTEXT_INTEL structure up into the
TSS2_TCTI_COMMON_CONTEXT and a structure specific to the device and
mssim TCTIs.

To accomplish this we rename the old 'tcti' module to 'tcti-common'.
This module defines a context structure containing all of the data
common to both the device and mssim TCTIs. This includes the v2
structure as well as a header structure, the enumeration to track the
state machine as well as the locality. We also create new header files
for the device and mssim TCTI. In these headers we define structures
specific to each module that hold the file descriptor for the device
TCTI and the socket file descriptors for the mssim TCTI.

Another significant change in this commit is the way we're now casting
the TCTI context structures passed into the functions for each TCTI
module. Functions in the device and mssim modules that are passed the
opaque TSS2_TCTI_CONTEXT structure start by first checking the magic
number in the context to be sure it's the right one for the module.
If this succeeds they then "down-cast" the structure to the
TSS2_TCTI_COMMON_CONTEXT and then use this structure to perform all of
the state checks. This reduces the number of unsafe cast operations that
we have to perform.

Finally there are a number of changes in this commit that qualify as
cleanup. This includes some name changes for variable and macro names to
align with the name changes.

### DIFF
--- a/src/tss2-tcti/tcti-mssim.c
+++ b/src/tss2-tcti/tcti-mssim.c
@@ -36,26 +36,61 @@
 #include <uriparser/Uri.h>
 
 #include "tss2_mu.h"
-
 #include "tss2_tcti_mssim.h"
-#include "tcti.h"
+
+#include "tcti-mssim.h"
+#include "tcti-common.h"
 #define LOGMODULE tcti
 #include "util/log.h"
 
 #define TCTI_SOCKET_DEFAULT_CONF "tcp://127.0.0.1:2321"
 #define TCTI_SOCKET_DEFAULT_PORT 2321
 
+/*
+ * This function wraps the "up-cast" of the opaque TCTI context type to the
+ * type for the mssim TCTI context. The only safeguard we have to ensure this
+ * operation is possivle is the magic number in the mssim TCTI context.
+ * If passed a NULL context, or the magic number check fails, this function
+ * will retrun NULL.
+ */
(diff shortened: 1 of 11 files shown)


flihp:
this resolves #921 and #887

src/tss2-tcti/tcti-mssim.c
```diff
@@ -525,39 +568,40 @@ Tss2_Tcti_Mssim_Init (
         return rc;
     }
 
-    tpmSock = &tcti_intel->tpmSock;
-    otherSock = &tcti_intel->otherSock;
+    tcti_mssim = (TSS2_TCTI_MSSIM_CONTEXT*)tctiContext;
+    tcti_common = tcti_common_context_cast (tctiContext);
```
tstruk: or better:
tcti_common = &tcti_mssim->common;

flihp:
How about this @tstruk? Not exactly what you were asking for but I think it cleans up the part you called out.

tstruk:
That will work.

Issue #80, opened 2015-12-01
Write test case for getPollHandles 


wcarthur1:

tstruk:
Fixed with caff75411 and 3f2489b1c

PR #924, opened 2018-04-04
test: Add GetPollHandle test for both TCTIs.

flihp:
Since neither of the TCTIs support async I/O yet both test cases just
make sure that they return the NOT_IMPLEMENTED response code. This will
likely be the case for the device TCTI indefinitely since the kernel
would have to support async I/O over the TPM device node. The mssim
TCTI could support this in the future however and so this test will need
to be updated when that work happens.

### DIFF
--- a/test/unit/tcti-device.c
+++ b/test/unit/tcti-device.c
@@ -107,6 +107,23 @@ tcti_device_teardown (void **state)
 
 }
 /*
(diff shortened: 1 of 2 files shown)


flihp:
this resolves #80 ... today is "find an old issue and fix it day" in case you were wondering :smile:

Issue #912, opened 2018-03-30
Doxygen generates empty documentation files


dantpm:
After the integration of #752, doxygen documents are supposed to be supported.
However, that is not the case. Running the following:

`./configure --enable-doxygen-doc && make doxygen-doc`

generates empty files under `doc/doxy/` and `doc/html/` in the repository with just header and title content:

```
$ make doxygen-doc
  DXGEN  doc/Doxyfile
warning: tag INPUT: input source `./include/esapi' does not exist
warning: tag INPUT: input source `./esapi' does not exist
warning: source ./include/esapi is not a readable file or directory... skipping.
warning: source ./esapi is not a readable file or directory... skipping.
/home/dano/tpm/git/tpm2-tss/doc/doxygen.dox:118: warning: missing title after \d
efgroup Esys_ActivateCredential
/home/dano/tpm/git/tpm2-tss/doc/doxygen.dox:124: warning: missing title after \d
efgroup Esys_Certify
```

For example, here is the page for 
doc/html/group__esys__context.html

```
Esys_ActivateCredential
Enhanced System API » Esys TPM Commands
```

But if I look at the source file, there is content provided for doxygen that is ignored for some reason:

"api/Esys_ActivateCredential.c"
```
/** One-Call function for TPM2_ActivateCredential
 *
 * This function invokes the TPM2_ActivateCredential command in a one-call
 * variant. This means the function will block until the TPM response is
 * available. All input parameters are const. The memory for non-simple output
 * parameters is allocated by the function implementation.
 *
 * @param[in,out] esysContext The ESYS_CONTEXT.
 . . . 
 */    
```

Issue #881, opened 2018-03-21
ESAPI: Set a meaningful default TCTI


AndreasFuchsSIT:
Current code uses the simulator on localhost as default tcti. This is of course not meaningful in any sense.

We need to set a meaningful default behaviour, maybe even probe for a list of tctis on the system.

dantpm:
I think this would be useful if it was documented somewhere.

dantpm:
I can extend the doxygen doc, but I need to find out how to run doxygen.  The following two methods fail (produce a empty document with just headers but no content in the ./html subdirectory:

```
doxygen doc/doxygen.dox
cd doc
doxygen doxygen.dox
```

AndreasFuchsSIT:
@dantpm `./configure --enable-doxygen-doc && make doxygen-doc`

dantpm:
Running 
`./configure --enable-doxygen-doc && make doxygen-doc`
Does not 

Does not work for me. It generates empty files with just the title, but no text from the source file comment blocks.  I filed this issue:
https://github.com/tpm2-software/tpm2-tss/issues/912

I'll close it if there's another step required or the above is incorrect.

PR #929, opened 2018-04-05
Document Default TCTI Behavior

### LINKED ISSUE #928: Document default TCTI behavior
Defaults were implemented for the TCTI to use with the fix to #881 , but not documented.

There is separate documentation in the tools repository, added with
tpm2-software/tpm2-tools#964
but they are not documented here.

Andreas suggested adding to Esys_Initialize() and README.md (see #881)

danintel:
Document the search algorithm used to find the default TCTI when not specified.

Fixes #928.

### DIFF
--- a/src/tss2-esys/esys_context.c
+++ b/src/tss2-esys/esys_context.c
@@ -38,6 +38,12 @@
  *
  * Initialize an ESYS_CONTEXT that holds all the state and metadata information
  * during an interaction with the TPM.
+ * If not specified, load a TCTI in this order:
+ *       Library libtss2-tcti-default.so (link to the preferred TCTI)
+ *       Library libtss2-tcti-tabrmd.so (tabrmd)
+ *       Device /dev/tpmrm0 (kernel resident resource manager)
+ *       Device /dev/tpm0 (hardware TPM)
+ *       TCP socket localhost:2321 (TPM simulator)
  * @param esys_context [OUT] The ESYS_CONTEXT.
  * @param tcti [IN] The TCTI context used to connect to the TPM (may be NULL).
  * @param abiVersion [INOUT] The abi version to check and the abi version


README.md
```diff
@@ -28,6 +28,14 @@ The former should be used for direct access to the TPM through the Linux kernel
 The latter implements the protocol exposed by the Microsoft software TPM2 simulator.
 * The [TCG TSS 2.0 Overview and Common Structures Specification](https://trustedcomputinggroup.org/wp-content/uploads/TSS_Overview_Common_Structures_Version-0.9_Revision-03_Review_030918.pdf) forms the basis for all implementations in this project.
 
+# TCTI Defaults
+If the TCTI is not specified when calling this library, the TCTI defaults to the first found in this order:
+* Library libtss2-tcti-default.so (link to the preferred TCTI)
+* Library libtss2-tcti-tabrmd.so (tabrmd)
+* Device /dev/tpmrm0 (kernel resident resource manager)
+* Device /dev/tpm0 (hardware TPM)
+* TCP socket localhost:2321 (TPM simulator)
```
tstruk: I don't think this description is not quite accurate. What do you mean by "calling this library"?
TSS SAPI exposes two init functions for TCTI:
Tss2_Tcti_Device_Init() and Tss2_Tcti_Mssim_Init()
The user of the SAPI needs to explicitly call the correct function depending on which TPM target they want to talk to.  
On top of that the ESAPI exposes its own Esys_Initialize() function, which takes TSS2_TCTI_CONTEXT *
and if this is NULL it tries to do some magic to figure out what TCTI to use.
In that case it will try to load whatever was configured via --with-tctidefaultmodule or call
Tss2_Tcti_Device_Init() and then Tss2_Tcti_Mssim_Init()

dantpm:
Changing the README was suggested in #881, but I just removed it, so the only change left is the block comment.

Doxygen builds don't work by the way (#912).

Issue #946, opened 2018-04-08
EOF in 'read_all'


flihp:
The `read_all` function from the I/O module doesn't handle `EOF` gracefully. It's so graceless in fact that if a read from the provided `fd` does return `EOF` then we end up in an infinite loop. This bug is currently making the device TCTI unusable.

martinezjavier:
@flihp right, I mentioned this bug in https://github.com/tpm2-software/tpm2-tss/pull/890#issuecomment-376460273 but didn't file an issue nor propose a fix, sorry about that.

flihp:
thanks for the link. I remembered your comment but couldn't find it when I repro'd the issue yesterday.

flihp:
@martinezjavier: you may also find #947 interesting / relevant. I backtracked to the EOF issue you mentioned while trying to figure out why the device TCTI was failing on account of the behavior described there. I guess assuming the behavior described in `man 2 read` would hold for the device driver was a bad idea :unamused:

flihp:
this was resolved in #948

PR #948, opened 2018-04-08
Add EOF handling to 'read_all' function in the io module, and TCTIs.

flihp:
This fixes a possible infinite loop in the TCTIs if / when the other end
of the IPC mechanism shuts down (for whatever reason). We add tests for
this scenario to the unit test for the io module.

Additionally the device and mssim TCTIs now handle this situation and
unit tests added for this scenario too.


flihp:
this resolves #946

flihp:
just fixed up a few empty comments

src/util/io.c
```diff
@@ -59,6 +59,10 @@ read_all (
                          fd, errno, strerror (errno));
             return recvd_total;
         }
+        if (recvd == 0) {
+            LOG_WARNING ("read on fd %d returned 0 / EOF", fd);
```
martinezjavier: I would add in the warning output the number of bytes that were attempt to read since that may be relevant information for debugging, something like:

```
LOG_WARNING ("Attempted to read %zd bytes from fd %d, but EOF returned", size, fd);
```

src/util/io.c
```diff
@@ -59,6 +59,10 @@ read_all (
                          fd, errno, strerror (errno));
             return recvd_total;
```
martinezjavier: Not related to this patch, but `recvd` be returned instead to propagate the error code to the caller?

src/util/io.c
```diff
@@ -59,6 +59,10 @@ read_all (
...
         }
+        if (recvd == 0) {
+            LOG_WARNING ("read on fd %d returned 0 / EOF", fd);
+            return recvd_total;
```
martinezjavier: Here I would also return `recvd` for consistency, even that for EOF will be semantically the same than `recvd_total`.

src/util/io.c
flihp: done

src/util/io.c
flihp: not related, we can take this up in another issue / PR

src/util/io.c
flihp: I've added a comment to the function to clarify this. For now I'm just trying to keep the behavior consistent. Happy to revisit the behavior of this function though. I figured this would be the case anyways when async I/O gets implemented.

flihp:
Just rebased to account for changes to master since this PR was created. Also addressed feedback from @martinezjavier

Issue #947, opened 2018-04-08
device TCTI must read all data in one read


flihp:
The device TCTI currently makes a bad assumption about reading data from the device driver: that we can read the TPM2 response header in one `read` operation (the first 10 bytes) to get the size of the whole thing, and then read the rest of the response in a second read. This is a bad assumption since the driver will close the file descriptor if we don't read the full response in one `read` operation.

This makes supporting the TCTI `receive` function difficult so we're gonna have to come up with a work-around.

martinezjavier:
@flihp is expected that the driver will close the file descriptor if the whole response isn't read in one `read` operation? Should the driver be fixed instead?

AndreasFuchsSIT:
@martinezjavier This is the particular behavior of the TPM driver on Linux. The TPM itself allows the platform to wait after having read the header before reading the rest of the response.

I do know that several Kernel drivers behave this way, e.g. under the /sysfs.

I don't know enough about the Kernel as a whole whether this is a kernel convention or anything.
If you want to tackle the kernel driver, it could also use a poll() implementation... ;-)

martinezjavier:
@AndreasFuchsSIT yeah, that's what I thought but wasn't sure if it was a constraint of the TPM. I may take a look to the driver but I probably won't have time to do it before next week.

In general I prefer us to avoid doing workarounds, following was another example of an issue we had and thought first about doing a workaround in tpm2-tss but but at the end fixed the kernel implementation:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=095531f891e627e408606f2da4008d3d53e6748a

AndreasFuchsSIT:
@martinezjavier Sounds great...

Just a random thought: Does the TPM-device-driver support `stat()` ? That would also be a way to determine the response's size, whilst keeping the current semantic of one-time-reads.

martinezjavier:
@AndreasFuchsSIT I don't think that the `struct stat .st_size` makes sense for a character device since isn't really a file. But I don't see why reading the available buffer in two `read` syscalls as @flihp implemented shouldn't be supported by a driver.

flihp:
I haven't seen any disagreement over whether or not we'd like to see the kernel support these things. It's just a matter of having the time to spend on it.

martinezjavier:
Agreed, and even if the driver is eventually fixed we need your patches so it works with kernel versions that don't implements the new behavior.

PR #949, opened 2018-04-09
tcti: Replace 2 read calls in the 'receive' function with a single read.

flihp:
The linux kernel driver closes connections that don't read the full
response buffer	in one read. This requires that the device TCTI
do exactly this to keep our FDs closed prematurely. It also means we
can't read the response header to know the size of the response ahead
of reading the response body.

If the caller queries for the size of the response before providing us
with the buffer (by providing a NULL buffer) then we just tell them to
give us a 4k buffer to be safe. We do not however require that the
buffer be this large to allow "expert" applications that may want to
precalculate the size of their response buffers in advance.

This commit also updates the unit tests to account for this change.

### DIFF
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -110,7 +110,25 @@ tcti_device_transmit (
     tcti_common->state = TCTI_STATE_RECEIVE;
     return TSS2_RC_SUCCESS;
 }
-
+/*
+ * This receive function deviates from the spec a bit. Calling this function
+ * with a NULL 'tctiContext' parameter *should* result in the required size for
+ * the response buffer being returned to the caller. We would typically do this
+ * by reading the response's header and then returning the size to the caller.
+ * We can't do that on account of the TPM2 kernel driver closing any connection
+ * that doesn't read the whole response buffer in one 'read' call.
+ *
+ * Instead, if the caller queries the size, we return 4k just to be on the
+ * safe side. We do *not* however verify that the provided buffer is large
+ * enough to hold the full response (we can't). If the caller provides us with
+ * a buffer less than 4k we'll read as much of the response as we can given
+ * the size of the buffer. If we get enough of the response to read the size
+ * field, we check to see if the buffer was large enough to get the full
+ * response. If the response header claims it's larger than the provided
+ * buffer we print a warning. This allows "expert applications" to
+ * precalculate the required response buffer size for whatever commands they
+ * may send.
+ */
 TSS2_RC
 tcti_device_receive (
     TSS2_TCTI_CONTEXT *tctiContext,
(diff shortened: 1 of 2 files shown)


flihp:
this resolves #947

src/tss2-tcti/tcti-device.c
```diff
@@ -136,63 +154,59 @@ tcti_device_receive (
...
+        rc = TSS2_TCTI_RC_GENERAL_FAILURE;
+        goto out;
+    }
+    rc = header_unmarshal (response_buffer, &tcti_common->header);
+    if (rc != TSS2_RC_SUCCESS) {
+        goto out;
+    }
+    if ((size_t)size != tcti_common->header.size) {
+        LOG_INFO ("TPM2 header size disagrees with number of bytes read from "
+                  "fd %d. Header says %u but we read %zu bytes.",
```
AndreasFuchsSIT: IMHO this should be LOG_WARNING at least; i.e. we want people to see this on default log-level settings... 
Maybe also rephrase to mention "... read from TPM device node using fd $d ..." or something similar.

src/tss2-tcti/tcti-device.c
```diff
@@ -136,63 +154,59 @@ tcti_device_receive (
...
+                  tcti_dev->fd, tcti_common->header.size, *response_size);
+    }
+    if (*response_size < tcti_common->header.size) {
+        LOG_WARNING ("TPM2 response header size is larger than the provided "
+                     "buffer: future use of this TCTI will likely fail.");
+    }
```
AndreasFuchsSIT: We should `rc = TSS2_TCTI_RC_GENERAL_FAILURE;` in this case...

src/tss2-tcti/tcti-device.c
```diff
@@ -110,7 +110,25 @@ tcti_device_transmit (
...
+ * safe side. We do *not* however verify that the provided buffer is large
+ * enough to hold the full response (we can't). If the caller provides us with
+ * a buffer less than 4k we'll read as much of the response as we can given
+ * the size of the buffer. If we get enough of the response to read the size
+ * field, we check to see if the buffer was large enough to get the full
+ * response. If the response header claims it's larger than the provided
+ * buffer we print a warning. This allows "expert applications" to
+ * precalculate the required response buffer size for whatever commands they
+ * may send.
+ */
```
martinezjavier: As mentioned in the issue tracking this, should we fix the driver instead? Is this just an implementation detail of the tpm driver or a constraint of the TPM? If the former, I really think we should change the driver instead of having workarounds in user-space.

martinezjavier: @flihp @tstruk @AndreasFuchsSIT ^

src/tss2-tcti/tcti-device.c
flihp: Promoted this message to a warning as requested but I left the message as-is. Having the fd # is sufficient data.

src/tss2-tcti/tcti-device.c
flihp: done

src/tss2-tcti/tcti-device.c
flihp: It would be great if interaction with the kernel driver could be made a bit more friendly to user space. Having the time to hack up a patch for the kernel is another issue. For now, userspace needs this workaround.

flihp:
whoops, forgot to fixup test cases to account for the RC change requested by @AndreasFuchsSIT

src/tss2-tcti/tcti-device.c
martinezjavier: @flihp fair enough.

flihp:
Review comments have been addressed IMHO. I'm going to push this to unblock downstream work that requires the device TCTI be functional.

PR #955, opened 2018-04-10
[RFC] Enable meson build

AndreasFuchsTPM:
This is a very first attempt at building things with meson. I have to say, the speedup on repeated compiles is very nice...

For usage examples please see .travis.yml

Please review, and tell me what we are lacking...

Known issues:
- Missing: Creating and installing *.pc files
- Fixing up man page installation


AndreasFuchsSIT:
Issue:
- In meson-options.txt use type:combo for maxloglevel instead of type:string

.travis.yml
```diff
@@ -60,11 +65,26 @@ install:
...
+# Trusty's ninja is too old
+  - wget https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-linux.zip
+  - sha256sum ninja-linux.zip | grep -q d2fea9ff33b3ef353161ed906f260d565ca55b8ca0568fa07b1d2cab90a84a07
+  - unzip ninja-linux.zip
 
 before_script:
   - ./bootstrap
 
 script:
+# if we are building with meson, only the next 5 lines are executed
```
dantpm: This comment does not seem to be accurate--it is more than 5 lines that are executed for Meson builds.

tstruk:
@AndreasFuchsSIT Is this supposed to be used for travis builds only? Autotools also makes reasonable good job on repeated build in development environment, and I think we have no choice, but to use autotools for distro packaging.

martinezjavier:
@tstruk many (most?) open source projects are migrating from autootols to meson, so I think that @AndreasFuchsSIT effort makes a lot of sense.

I haven't worked with meson yet but have only heard good things about it. Can't say the same about autootols. I always struggle with it, even after being developing on Linux systems for more than 10 years...

You are correct thought, that sometimes we don't have a choice for distro packaging. For fast moving distros this isn't a problem, but it's for enterprise distros where meson support isn't available.

I think that an option is to have both build systems in place for some time and eventually drop autools support. @AndreasFuchsSIT it's that your plan or the goal is to drop autootols as soon as your RFC branch is ready for merging?

AndreasFuchsSIT:
@tstruk It's intended as actual build system for everyone.
I've seen Gnome as well as systemd move over to meson and completely ditch autotools...

@martinezjavier Exactly, I wanted to keep introduce it as an alternative for now. That's also why I added meson as a test-matrix into travis...

Being an autotools-fanboy not wanting to migrate (at least for cmake), I can tell you that meson was highly impressive...

I guess it will be pain to have 3 systems now, autotools, VS and meson, but in the log run, meson can also create VS files if I read this correctly (still learning)...

dantpm:
I haven't seen Meson used before but from what I see it's a lot better than Autotools/Automake. X Windows moved from imake templates to Autotools and plans to move to Meson.

AndreasFuchsSIT:
I think this is now maturing slowly...
Not contains:
- Installation of headers
- Documentation in INSTALL.md
- Support for ninja dist in travis build

It would be great, if folks would test this and also point out any meson-options.txt entries I'm missing.

flihp:
IMHO this is less about "if" and more about "when". The build system requirements I was working off when I chose autotools in the beginning:
1) friendly to upstream packagers (adoption by distros is #1 success metric IMHO)
2) flexible parallel test harness support
3) consistency across the projects
and that's about it.

Personally I don't want to support multiple build systems. VS solution files are a necessary evil and if we can get rid of them as we migrate to meson I'll be very happy. Autotools should go away on the same schedule.

We should also consider timing for this change. Probably something we should consider between minor releases of the 2.x code base (when that starts happening).

tstruk:
Closing this for now. We can come back to meson after 2.0 is out.

AndreasFuchsTPM:
I intend to start working on this again, so re-opening

PR #961, opened 2018-04-11
Fix Microsoft Visual Studio Release Builds

### LINKED ISSUE #918: Support for Windows release builds
Builds done through Visual Studio with clang currently fail if we provide any optimization options beyond `-O0` (aka no optimizations). This needs some root-cause analysis and a resolution.

dantpm:
Microsoft Visual Studio gives internal error C1001 when __builtin_expect() is
used for Release builds (with optimization).  The default Debug builds are OK.
This fix disables the macro, which is an optimization suggestion only for the
compiler, for Microsoft compilers.

Fixes #918.

### DIFF
--- a/src/util/log.c
+++ b/src/util/log.c
@@ -6,8 +6,14 @@
 #define LOGMODULE log
 #include "log.h"
 
+#if !defined(_MSC_VER) || defined(__INTEL_COMPILER)
 #define likely(x)       __builtin_expect(!!(x), 1)
 #define unlikely(x)     __builtin_expect(!!(x), 0)
+#else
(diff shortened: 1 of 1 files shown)


src/util/log.c
```diff
@@ -6,8 +6,14 @@
 #define LOGMODULE log
 #include "log.h"
 
+#if !defined(_MSC_VER) || defined(__INTEL_COMPILER)
```
tstruk: Will this also cover Visual studio using clang?

dantpm: Yes.  We only use clang. @flihp said that's because the MS Visual Studio C compiler does not support C99, so we have to use clang.

PR #963, opened 2018-04-13
build: Install udev rule to set ownership and permissions for TPM cha…

martinezjavier:
…rdevs

Currently the udev rule is installed by tpm2-abrmd, but that doesn't seem
correct since users may want to access the TPM directly without using the
user-space resource manager.

Also, conceptually it should be set by the lower layers of the stack that
access the TPM character devices.

The change in this patch was taken from the tpm2-abrmd project.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -84,6 +84,44 @@ Then run the configure script, which generates the makefiles:
 $ ./configure
 ```
 
+### Custom `./configure` Options
+In many cases you'll need to provide the `./configure` script with additional
+information about your environment. Typically you'll either be telling the
+script about some location to install a component, or you'll be instructing
+the script to enable some additional feature or function. We'll cover each
+in turn.
+
+Invoking the configure script with the `--help` option will display
+all supported options.
+
+The default values for GNU installation directories are documented here:
+https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+
+### udev Rules
+The typical operation for the `tpm2-abrmd` is for it to communicate directly
+with the Linux TPM driver using `libtcti-device` from the TPM2.0-TSS project.
+This requires that the user account that's running the `tpm2-abrmd` have both
+read and write access to the TPM device node `/dev/tpm[0-9]`. But users could
+also access the TPM directly so the udev rule is installed by `tpm2-tss`.
+
+#### `--with-udevrulesdir`
+This requires that `udev` be instructed to set the owner and group for this
+device node when its created. We provide such a udev rule that is installed to
+`${libdir}/udev/rules.d`. If your distro stores these rules elsewhere you will
+need to tell the build about this location.
+
+Using Debian as an example we can instruct the build to install the udev
+rules in the right location with the following configure option:
+```
+--with-udevrulesdir=/etc/udev/rules.d
+```
+
+#### `--with-udevrulesprefix`
+It is common for Linux distros to prefix udev rules files with a numeric
+string (e.g. "70-"). This allows for the rules to be applied in a predictable
+order. This option allows for the name of the installed udev rules file to
+have a string prepended to the file name when it is installed.
+
 ## Compiling the Libraries
 Then compile the code using make:
 ```
@@ -102,7 +140,22 @@ won't need to do much more than provide an alternative --prefix option at
 configure time, and maybe DESTDIR at install time if you're packaging for a
 distro.
 
-**NOTE**: It may be necessary to run ldconfig (as root) to update the run-time
+# Post-install
+
+## udev
+Once you have this udev rule installed in the right place for your distro
+you'll need to instruct udev to reload its rules and apply the new rule.
+Typically this can be accomplished with the following command:
+```
+$ sudo udevadm control --reload-rules && sudo udevadm trigger
+```
+
+If this doesn't work on your distro please consult your distro's
+documentation for UDEVADM(8).
+
+## ldconfig
+
+It may be necessary to run ldconfig (as root) to update the run-time
 bindings before executing a program that links against libsapi or a TCTI
 library:
 ```
--- a/Makefile.am
+++ b/Makefile.am
@@ -56,6 +56,11 @@ include src_vars.mk
 include Makefile-test.am
 
 ### Distribution files ###
+# Add udev rule
+if WITH_UDEVRULESPREFIX
+udevrules_DATA   = dist/tpm-udev.rules
+endif
+
 # Adding user and developer information
 EXTRA_DIST += \
     CHANGELOG.md \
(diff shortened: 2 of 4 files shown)


martinezjavier:
@AndreasFuchsSIT sorry for missing that, tested locally so I didn't notice it.

So I disagree, it's not about what the library cares but the users of the library. Let's say that you build an application that uses the tpm2-tss libraries (or a shell script that uses the tpm2-tools), under what user do you expect it to run?

Obviously you don't want it to run it as root, so should either use an unprivileged user that has access to TPM or a user that belongs to a group that has access to it.

By having the udev rule in tpm2-tss, then we can tell users to either use the `tss` user or add their users to the `tss` group. If we don't, what do we suggest to them? Install the tpm2-abrmd even when isn't used is not a correct answer.

In fact, we already ship the TPM udev rule in the tpm2-tss package in Fedora, I'm proposing this changes upstream because I think that makes sense, but if not I'll just keep maintaining my patches in the tpm2 packages.

AndreasFuchsSIT:
You know what ? I'd actually want to see this in 50-udev-default.rules since it's not different from video or printer or whatnot...

I'll contradict myself here for a moment; abrmd is probably not correct, since abrmd will at some point disappear, once all distros' LTSes carry kernerls with tpmrm enabled.
tpm2-tools makes no sense, because libraries may carry this stuff as well...

So maybe you are right, that tpm2-tss is the least of all evils... :-/ But it just does not feel right...

Since we're doing this here now, should we add a rule for /dev/tpmrm[0-9] as well ?
Also, should we give /dev/tpm[0-9] and /dev/tpmrm[0-9] different groups, since the former is for priviledged daemon/root access (since it can clutter and DoS the TPM) and the latter is also for applications (since they cannot DoS the TPM).

And how do we name these groups ?

@martinezjavier What's your thought on the latter ?
@tstruk @flihp @williamcroberts Do you have ideas on the topic ?

martinezjavier:
> You know what ? I'd actually want to see this in 50-udev-default.rules since it's not different from video or printer or whatnot...

@AndreasFuchsSIT I'm all for having it in 50-udev-default.rules, if someone wants to propose it to systemd/udev.

> I'll contradict myself here for a moment; abrmd is probably not correct, since abrmd will at some point disappear, once all distros' LTSes carry kernerls with tpmrm enabled.
> tpm2-tools makes no sense, because libraries may carry this stuff as well...

Agreed, tpm2-tools makes no sense. That's why I suggested to install it with the libraries.

> Since we're doing this here now, should we add a rule for /dev/tpmrm[0-9] as well ?

I already did that, notice that the regex for the udev rule is `"tpmrm[0-9]*|tpm[0-9]*"`.




> Also, should we give /dev/tpm[0-9] and /dev/tpmrm[0-9] different groups, since the former is for priviledged daemon/root access (since it can clutter and DoS the TPM) and the latter is also for applications (since they cannot DoS the TPM).

That's a very good question. I really think that any application that has access to the TPM should be trusted. So I don't think that there's a need to have two different users. As you said in most cases `/dev/tpmrm*` will be used and `/dev/tpm*` will only be used in special cases.

Probably having a single tss user/group for now is enough, anyone that wants more fine grained control could use a LSM based MAC (i.e: SELinux, Apparmor, etc).

AndreasFuchsSIT:
I don't know if we should assume the same level of trust in the application as we do in the daemon.

What do you think about this
```
KERNEL=="tpm[0-9]*", MODE="0660", OWNER="tss"
KERNEL=="tpmrm[0-9]*", MODE="0660", GROUP="tss"
```
i.e. the daemon under user tss gains access to tpm[0-9] directly and applications to tpmrm[0-9]

Or is that bad practice ?

martinezjavier:
I guess you meant?

```
KERNEL=="tpm[0-9]*", MODE="0660", OWNER="tss"
KERNEL=="tpmrm[0-9]*", MODE="0660", OWNER="tss", GROUP="tss"
```

IOW, the tss user can access both the TPM directly and the in-kernel resource manager but the members of the tss group can only access the in-kernel RM, they can't access the TPM directly?

I don't have an argument against that.

AndreasFuchsSIT:
P.S. and anyone else. I thinks this is important so we need good consensus...

williamcroberts:
```
KERNEL=="tpm[0-9]*", MODE="0660", OWNER="tss"
KERNEL=="tpmrm[0-9]*", MODE="0660", OWNER="tss", GROUP="tss"
```
+1 from  me, that seems sane.

tstruk:
@martinezjavier @AndreasFuchsSIT I agree that TSS is a good place for the udev rules. This will make it work for users who only install tss and don't care about tools or RM. The rules also look reasonable. We want to follow the least privilege principle. Having said that should the mode for tpm[0-9]* be MODE="0600"?

INSTALL.md
```diff
@@ -84,6 +84,45 @@ Then run the configure script, which generates the makefiles:
...
+--with-udevrulesdir=/etc/udev/rules.d
+```
+
+#### `--with-udevrulesprefix`
+It is common for Linux distros to prefix udev rules files with a numeric
+string (e.g. "70-"). This allows for the rules to be applied in a predictable
+order. This option allows for the name of the installed udev rules file to
+have a string prepended to the file name when it is installed.
+
+
```
tstruk: Extra blank line

martinezjavier: @tstruk not really, take a look at the file. It uses two lines to separate between sections. But I prefer to first agree on what we should do and then focus on these little details.

martinezjavier:
@AndreasFuchsSIT it seems there's agreement on your proposal.  Let's see what @flihp think and you may want to send a PR to add it to `50-udev-default.rules `.

AndreasFuchsSIT:
I assume @flihp to agree ?
If we don't hear by next week, I'd assume workload overload and merge it given current agreement.

martinezjavier:
@AndreasFuchsSIT Ok, I've updated the PR based on your suggestion.

This is how the permission bits and ownership looks like after using the udev rule:

```
$ ls -lh /dev/tpm*
crw-rw----. 1 tss root  10,   224 Apr 20 14:00 /dev/tpm0
crw-rw----. 1 tss tss  244, 65536 Apr 20 14:00 /dev/tpmrm0
```

flihp:
I like it. Even better if we can get it upstream into `50-udev-default.rules` but I think this is a logical first step. I'll get this merged along with the accompanying patch in the tabrmd.

INSTALL.md
flihp: I'm not seeing two new lines separating any other sections @martinezjavier. Don't doubt your eyes though. Can you point out an example for me?

flihp:
see comment on nit here: https://github.com/tpm2-software/tpm2-tss/pull/963#discussion_r183145948

INSTALL.md
martinezjavier: @flihp I may be blind, I was pretty sure about this and deliberative added that extra blank line ...

Sorry @tstruk for not double checking after your comment. Fixed now.

Issue #952, opened 2018-04-09
TCTI: Remove sockets.h


AndreasFuchsSIT:
tpm2-tss/src/tss2-tcti/sockets.h seems to be a left-over.
 tpm2-tss/src/util/io.h is used instead now...

@flihp can you confirm this ?

flihp:
confirmed and resolved in #966

PR #966, opened 2018-04-17
tss2-tcti: Remove sockets.h

flihp:
This file was left around after the mssim TCTI was refactored and the
socket code moved to the utility library.


flihp:
this resolves #952

PR #970, opened 2018-04-19
Travis: Add scan-build Static Code Analysis

### LINKED ISSUE #969: Add scan-build to Travis builds
Add the clang scan-build static coded analysis to Travis builds.

I will work on this issue.

danintel:
Fixes #969.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,9 +85,19 @@ script:
     if [ "$CC" == "gcc" ]; then
       export CONFIGURE_OPTIONS="--enable-code-coverage";
     fi
-  - ../configure --enable-unit --enable-integration $CONFIGURE_OPTIONS
+  - |
+    if [ "$CC" == "clang" ]; then
+      scan-build ../configure --enable-unit --enable-integration $CONFIGURE_OPTIONS
+    else
+      ../configure --enable-unit --enable-integration $CONFIGURE_OPTIONS
+    fi
   - make -j$(nproc) distcheck
-  - make -j$(nproc) check
+  - |
+    if [ "$CC" == "clang" ]; then
+      scan-build --status-bugs make -j$(nproc) check
+    else
+      make -j$(nproc) check
+    fi
   - cat test-suite.log
   - |
     for LOG in $(ls -1 test/unit/*.log); do
--- a/test/integration/sapi-session-util.c
+++ b/test/integration/sapi-session-util.c
@@ -650,8 +650,15 @@ EncryptDecryptXOR(
     TPM2B_AUTH *authValue)
 {
     TSS2_RC rval = TSS2_RC_SUCCESS;
-    TPM2B_MAX_BUFFER key, mask;
-    int i;
+    TPM2B_MAX_BUFFER key;
+    TPM2B_MAX_BUFFER mask = { .size = 0, .buffer = 0 };
+    UINT16 i;
+    UINT16 size = inputData->size;
+
+    if (size > TPM2_MAX_DIGEST_BUFFER) {
+        LOG_ERROR("Bad value for inputData size: %" PRIu16, size);
+        return TSS2_SYS_RC_GENERAL_FAILURE;
+    }
 
     CopySizedByteBuffer((TPM2B *)&key, (TPM2B *)&session->sessionKey);
     CatSizedByteBuffer((TPM2B *)&key, (TPM2B *)authValue);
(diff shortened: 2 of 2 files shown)


danintel:
This integration first requires a fix to #971 to fix a warning introduced with the recent integration of:
#967 More tpmclient cleanups

danintel:
@AndreasFuchsSIT 
I thought that scan-build required clang too, but that does not appear to be the case.
The documentation states both gcc and clang are supported, although clang produced scan-build.
Also I tried a little experiment which shows that both gcc and clang are supported.

```
$ export CC=gcc           
$ scan-build echo CC is $CC
scan-build: Using '/usr/lib/llvm-3.8/bin/clang' for static analysis
CC is gcc
scan-build: No bugs found.
$ scan-build gcc --version
scan-build: Using '/usr/lib/llvm-3.8/bin/clang' for static analysis
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609
scan-build: No bugs found.
$ scan-build cc --version
scan-build: Using '/usr/lib/llvm-3.8/bin/clang' for static analysis
cc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609
$ scan-build clang --version
scan-build: Using '/usr/lib/llvm-3.8/bin/clang' for static analysis
clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)
scan-build: No bugs found.
```

.travis.yml
```diff
@@ -86,8 +86,8 @@ script:
       export CONFIGURE_OPTIONS="--enable-code-coverage";
     fi
   - ../configure --enable-unit $CONFIGURE_OPTIONS --with-simulatorbin=$(pwd)/../ibmtpm/src/tpm_server
```
AndreasFuchsSIT: Missing scan-build prefix her

danintel:
@AndreasFuchsSIT
1. It makes no sense that make is ran twice, first without the tests, then with the tests.
So I changed it to run scan-build once, with the tests.
Why make is ran twice, with and without tests, I do not know.  There is no comment in the script about why this is done.

2. There is no target check-programs used in the script.
I see this target in file Makefile. What does it do?

How does "make check-programs" benefit scan-build static analysis?

3. Yes, scan-build analysis is implemented with clang, but there is no uncertainty about gcc support.  scan-build documentation states both gcc and clang are supported.  We have seen multiple cases where scan-build finds different errors running gcc and clang.  So it is worth while to run scan-build for both gcc and clang to catch more errors.  I know it is less efficient and a build takes longer, but I think it is worth the extra time.

4. clang is supported with gcc, so the run time should not be faster than clang.
This is another datapoint showing scan-build supports gcc.

.travis.yml
danintel: We are not running static analysis on tpm_server.  We cannot change the simulator source and it has many problems already if we even enable gcc/clang warnings.

AndreasFuchsSIT:
```
If an analyzed project uses an autoconf generated configure script, you will probably need to run configure script through scan-build in order to analyze the project.
```
(source https://clang-analyzer.llvm.org/scan-build.html)

AndreasFuchsSIT:
1. Yes that makes sense: First we run distcheck, then we run tests. Scan-build needs to run only once and the test-config is the right place. So please remove scan-build from distcheck.
2. check-programs does what the name suggests. It compiles all test programs. It was introduced specifically to support static code analysis. Then we do not need to wrap scan-build into the checks. Hence, we first run scan-build on the compilation step (make check-programs) and then we run the test without a static analyzer (make check), so everthing is neatly separated.
3/4. scan-build will allways use clang for the static analysis. The compilation is done independently from this by gcc or clang after the analysis phase. Hence, there is no advantage of running scan-build in both configurations (gcc and clang) except for burning energy that could've been bitcoins...
Hence we only run scan-build on one of he configurations.
And since scan-build is so tightly coupled with clang, please use the clang-configuration to run scan-build.

I hope this makes sense now.

danintel:
I disabled scan-build for gcc.

It's probably not that important, but .... There is nothing in the words "check" and "check-programs" that gives a clue about how they function differently.  Also "check-programs" is not in file .travis.yml. However,  "make check" is in file .travis.yml.  Looking in the generated Makefile I can see how they are different. I see
   `check-programs --> $(check_PROGRAMS)`
and 
`  check --> check-am --> $(check-PROGRAMS) (and check-TESTS)`

So "make check" (used in .travis.yml) is a superset of "make check-programs", so we are OK.

danintel:
Removed another scan-build warning message from another recently integrated change.

danintel:
I removed all scan-build warnings and it's now ready to integrate.

One trick was that I had to install/use clang-5.0 and scan-build-5.0 (instead of the default clang 3.8) to get the same warnings as Travis CI.

tstruk:
@danintel could you please rebase and change --with-simulatorbin to --enable-integration

tstruk:
@danintel Thanks for rebasing. The PRs are integrated in order they are submitted unless changes are requested.

PR #976, opened 2018-04-20
tcti-mssim: Replace liburiparser with custom key/value pair parser.

flihp:
This removes our use of liburiparser with a custom key / value pair
parser. The parser is implemented in the key-value-parse module under
the util directory. The parser is designed to extract a series of key
/ value pairs where each key and value are separated by the '='
character and each pair is separated by the ',' character. The parser
is composed of three parts:

1) The 'parse_key_value' function is used to extract key /value pairs
separated by the '=' character. This function modifies the string by
using the 'strok_r' function to extract the key and value. A reference
to each is then returned to the caller using the key_value_t structure.

2) The 'parse_key_value_string' function is used to extract each key /
value pair parsing the provided string on the ',' character using the
same method as the 'parse_key_value' function. Each string containing a
key / value pair is then passed to the 'parse_key_value' function to
extract the individual components.

3) Finally the 'parse_key_value_string' function is passed a function
pointer conforming to the 'KeyValueFunc' type. This function is invoked
for each key / value pair extracted and is passed a reference to the
key_value_t structure from the previous step. It is also passed a void
pointer to some caller provided data structure which allows the callback
to do something with the key / value pairs. For the purposes of the
mssim TCTI we simply identify valid / invalid keys and store their
values in a data structure.

Once the parser is done extracting data from the provided string the
mssim TCTI will either return an error indicating the configuration
string was invalid, or continue initializing the TCTI based on the
provided data. If a NULL configuration string is provided defaults are
used.

Additionally this commit:
1) Removes the check for liburiparser from the autoconf file.
2) Updates the .travis.yml file to no longer install liburiparser as a
build dependency.
3) Adds unit tests for the functions described above.
4) Updates existing tests and test infrastructure to use configuration
strings in the new format.
5) Updates the man page for the Tss2_Tcti_Mssim_Init fucntion to
describe the new configuration string format.
6) Updates the default configuration string to use 'localhost' instead
of a raw IPv4 address to make the library friendly to platforms that
only support IPv6.


flihp:
branch for required changes to tabrmd: https://github.com/flihp/tpm2-abrmd/tree/missim-conf-key-values

flihp:
Branch for required changes to tools: https://github.com/flihp/tpm2-tools/tree/mssim-conf-key-values. Looks like @williamcroberts is working through some build failures in the tools repo currently so we'll hold this PR up till the tools is 'green' again.

test/integration/context-util.c
```diff
@@ -49,14 +50,14 @@ tcti_device_init(char const *device_path)
...
     TSS2_RC rc;
     TSS2_TCTI_CONTEXT *tcti_ctx;
-    char conf_str[256] = { 0 };
+    char conf_str[TCTI_MSSIM_CONF_MAX] = { 0 };
 
-    snprintf(conf_str, 256, "tcp://%s:%" PRIu16, address, port);
+    snprintf(conf_str, TCTI_MSSIM_CONF_MAX, "host=%s,port=%" PRIu16, host, port);
     rc = Tss2_Tcti_Mssim_Init(NULL, &size, conf_str);
     if (rc != TSS2_RC_SUCCESS) {
         fprintf(stderr, "Faled to get allocation size for tcti context: "
```
danintel: s/Faled/Faled/

src/util/key-value-parse.c
```diff
@@ -0,0 +1,113 @@
...
+#include "util/key-value-parse.h"
+#define LOGMODULE tcti
+#include "util/log.h"
+
+/*
+ * Parse the provided string containing a key / value pair separated by the
+ * '=' character.
+ * NOTE: The 'kv_str' parameter is not 'const' and this function will modify
+ * it as part of the parsing process. The key_value structure will be updated
+ * with referenvces pointing to the appropriate location in the key_value_str
```
danintel: s/referenvces/references/

test/integration/context-util.c
flihp: Please limit your review to code introduced / changed in the PR.

src/util/key-value-parse.c
flihp: fixed

flihp:
@AndreasFuchsSIT : esys default TCTI code and test cases updated

src/tss2-tcti/tcti-mssim.c
```diff
@@ -562,17 +524,41 @@ Tss2_Tcti_Mssim_Init (
...
+    rc = socket_connect (mssim_conf.host,
+                         mssim_conf.port,
+                         &tcti_mssim->tpm_sock);
     if (rc != TSS2_RC_SUCCESS) {
-        return rc;
+        goto fail_out;
     }
 
-    rc = socket_connect (hostname, port + 1, &tcti_mssim->platform_sock);
+    rc = socket_connect (mssim_conf.host,
```
tstruk: We should probably initialize
tcti_mssim->tpm_sock = -1 and 
tcti_mssim->platform_sock = -1
here before calling the first socket_connect.

src/util/key-value-parse.c
```diff
@@ -0,0 +1,113 @@
...
+
+/*
+ * Parse the provided string containing a key / value pair separated by the
+ * '=' character.
+ * NOTE: The 'kv_str' parameter is not 'const' and this function will modify
+ * it as part of the parsing process. The key_value structure will be updated
+ * with references pointing to the appropriate location in the key_value_str
+ * parameter.
+ */
+bool
```
tstruk: I think parse_key_value should be static.

src/tss2-tcti/tcti-mssim.c
flihp: Same response I gave to @danintel yesterday (https://github.com/tpm2-software/tpm2-tss/pull/976#discussion_r183529819) applies: please limit your review to code in the patch. I don't disagree with with the *issue* you're bringing up, just that this isn't the right place to do it. IMHO the right thing to do here is to create an issue to track the work, bonus points if you assign it to yourself and knock out a patch :wink:

PR #979, opened 2018-04-23
tcti: Remove TSS2_TCTI_VERSION type, replace with simple uint32_t.

flihp:
This is based on feedback from TCG TSS2 working group. The 'magic'
field in the INFO structure isn't necessary or useful. This allows us to
remove the VERSION type all together and just use a uint32_t in its
place. The macros in the header are updated as well.


flihp:
TCTI spec out for public review is here: https://trustedcomputinggroup.org/wp-content/uploads/TSS_TCTI_v1.0_r12_PUBLIC_REVIEW.pdf

PR #980, opened 2018-04-24
Tpmclient cleanup

tstruk:


tstruk:
@danintel please see #984. Is this what you requested?

PR #985, opened 2018-04-26
build: Remove --with-simulatorbin, replace with --enable-integration.

flihp:
Bill recently integrated the tools integration tests into the test
harness and his approach assumes that the `tpm_server` binary is on the
PATH. This simplifies a few thing and generally it's just cleaner than
requiring the caller to pass the full path to the executable at
configure time.

This patch removes the --with-simulatorbin configure option replacing it
with the --enable-integration option instead. This option takes no
parameters but it does check that the 'tpm_server' binary is on the
callers path. Additionally this patch consolidates the multiple AC_IF
checks for the definition of the associated variable. The corresponding
option to the int-log-compiler.sh script is removed as well.


flihp:
Didn't think of that, though you mentioned it when we spoke about this the other day. Will have a fix in a bit.

PR #989, opened 2018-04-26
Tests: ESAPI: Add test for Esys_GetPollHandles

AndreasFuchsTPM:


test/unit/esys-getpollhandles.c
```diff
@@ -0,0 +1,156 @@
...
+static void
+test_GetPollHandles(void **state)
+{
+    TSS2_RC r;
+    ESYS_CONTEXT *ectx = (ESYS_CONTEXT *) * state;
+
+    TSS2_TCTI_POLL_HANDLE *handles;
+    size_t count;
+
+    r = Esys_GetPollHandles(ectx, &handles, &count);
```
tstruk: Shouldn't that be, first:
Esys_GetPollHandles(ectx, NULL, &count);
and then
Esys_GetPollHandles(ectx, &handles, &count);
It's strange that the assert_int_equal(count, 2); even passes if you don't call it with the NULL handles first.

AndreasFuchsSIT: No, the two invocations are only needed for TCTI and SAPI.
ESAPI will allways allocate memory for you, i.e. parameter type is TSS2_TCTI_POLL_HANDLE **handles.

AndreasFuchsSIT: No, the two invocations are only needed for TCTI and SAPI.
ESAPI will allways allocate memory for you, i.e. parameter type is TSS2_TCTI_POLL_HANDLE **handles.

PR #994, opened 2018-04-27
build: Add 'Release' builds to .appveyor.yml

flihp:
Dan fixed the release builds a while back but the AppVeyor yml file was
never updated to get these on the build matrix.


danintel:
For the record, the release builds fix was #961, which fixed #918

PR #997, opened 2018-04-30
Changelog cleanup

flihp:
This is preparation for the upcoming CHANGELOG update for the 2.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ the 'common directory.
 ### Fixed
 - Wrong return type for Tss2_Sys_Finalize (API break).
 
+## [1.4.0] - 2018-03-02
+### Added
+- Attached Component commands from the last public review spec.
+### Fixed
+- Essential files missing from release tarballs are now included.
+- Version string generation has been moved from configure.ac to the
+bootstrap script. It is now stored in a file named `VERSION` that is
+shipped in the release tarball.
+- We've stopped shipping the built man page for InitSocketTcti.3 and now
+ship the source.
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,17 @@ the 'common directory.
...
 
+## [1.4.0] - 2018-03-02
+### Added
+- Attached Component commands from the last public review spec.
+### Fixed
+- Essential files missing from release tarballs are now included.
+- Version string generation has been moved from configure.ac to the
+bootstrap script. It is now stored in a file named `VERSION` that is
+shipped in the release tarball.
+- We've stopped shipping the built man page for InitSocketTcti.3 and now
```
tstruk: Prebuilt man page for InitSocketTcti.3 replaced with the source.

CHANGELOG.md
```diff
@@ -16,6 +16,17 @@ the 'common directory.
 ### Fixed
 - Wrong return type for Tss2_Sys_Finalize (API break).
 
+## [1.4.0] - 2018-03-02
+### Added
+- Attached Component commands from the last public review spec.
```
tstruk: Attached Component commands from the last public review spec (rev 1.46)

flihp:
I don't disagree with your proposed improvements / clarifications but this is the changelog entry that was in the 1.4.0 release tarball. I'm not sure changing these is really the right thing to do since it will cause future changelogs to disagree with past changelogs. In fact I'd prefer to leave these as is if you're not too invested in these changes.

tstruk:
If this is how it is in 1.4.0 then sure, we can merge it as is.

Issue #917, opened 2018-04-02
enable / disable mssim TCTI build


flihp:
I've had a request for a configure flag to selectively disable building the mssim TCTI. Something like `--disable-tcti-mssim` using the `AC_ARG_ENABLE` macro. The reason for this is pretty simple: there are very few reasons to install the mssim TCTI on anything but test / development systems.

PR #1006, opened 2018-05-02
Build: Make TCTIs optional

AndreasFuchsSIT:
Fixes: #917

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -90,6 +90,20 @@ AC_ARG_WITH([tctidefaultconfig],
                                 ["The default TCTIs configuration string"])],
             [])
 
+AC_ARG_ENABLE([tcti-device],
+            [AS_HELP_STRING([--enable-tcti-device],
+                            [build the tcti-device module (default is yes)])],
+            [enable_tcti_device=$enableval],
+            [enable_tcti_device=yes])
+AM_CONDITIONAL([ENABLE_TCTI_DEVICE], [test "x$enable_tcti_device" != xno])
+
+AC_ARG_ENABLE([tcti-mssim],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -90,6 +90,20 @@ AC_ARG_WITH([tctidefaultconfig],
...
+            [enable_tcti_device=yes])
+AM_CONDITIONAL([ENABLE_TCTI_DEVICE], [test "x$enable_tcti_device" != xno])
+
+AC_ARG_ENABLE([tcti-mssim],
+            [AS_HELP_STRING([--enable-tcti-mssim],
+                            [build the tcti-mssim module (default is yes)])],
+            [enable_tcti_mssim=$enableval],
+            [enable_tcti_mssim=yes])
+AM_CONDITIONAL([ENABLE_TCTI_MSSIM], [test "x$enable_tcti_mssim" != xno])
+
```
tstruk: We should check if at least one TCTI is enabled.

AndreasFuchsSIT: I intentionally did not do so.

Reason being a case, where someone wants a SAPI in a firmware environment with their own (non-public) TCTI.
Thus, we want to compile without any of our "official" TCTIs.

AndreasFuchsSIT:
@danintel Where and how should these options be mentioned in INSTALL.md ? IMHO INSTALL.md should not be a copy of `./configure --help`. So what do you recommend ?

danintel:
@AndreasFuchsSIT OK--leave it in just` ./configure --help`

configure.ac
AndreasFuchsSIT: added it to the end, so it is easily spotted.

PR #1012, opened 2018-05-10
esys_crypto: fix defects reported by Coverity Scan

tstruk:
Fix two potential null ptr dereference issues reported by ceverity.

CID 1468838:  Null pointer dereferences  (REVERSE_INULL)
Null-checking "partyUInfo" suggests that it may be null,
but it has already been dereferenced on all paths leading to the check.

CID 1468837:  Null pointer dereferences  (REVERSE_INULL)
Null-checking "partyVInfo" suggests that it may be null,
but it has already been referenced on all paths leading to the check.

### DIFF
--- a/src/tss2-esys/esys_crypto.c
+++ b/src/tss2-esys/esys_crypto.c
@@ -949,8 +949,13 @@ iesys_cryptogcry_KDFe(TPM2_ALG_ID hashAlg,
 
     LOG_DEBUG("IESYS KDFe hashAlg: %i label: %s bitLength: %i",
               hashAlg, label, bit_size);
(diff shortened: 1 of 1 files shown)


tstruk:
@AndreasFuchsSIT @JuergenReppSIT This is a quick fix for the coverity issue, but while I'm looking at it, shouldn't all the parameters be mandatory? According to the spec all are needed for KDFa. Should we return an error if any of them is null instead of just omitting it?

tstruk:
@williamcroberts changed

williamcroberts:
> 
> @williamcroberts <https://github.com/williamcroberts>  changed
LGTM
>

PR #1013, opened 2018-05-11
build: Add checks for 'netstat' @ configure and integration test time.

flihp:
I just ran a full build and integration test run on a fresh build system
installing only things that the configure script tested for. Everything
built fine but the test harness hung. This happened because the netstat
utility was missing. I hadn't installed net-tools (Debian).

This commit adds an explicit check for this tool @ configure time. The
integration test wrapper has also been instrumented to ensure that
netstat is on the PATH as well. This causes tests to fail early instead
of timing out because all checks for simulator port conflicts fail which
cause retries and backoff timeouts.

The INSTALL documentation is also updated to include mention of the need
for netstat / the net-tools package.


danintel:
Fixes a problem with #764

PR #1018, opened 2018-05-21
Fix API

tstruk:
Fix invalid order of parameters in Tss2_Sys_PolicyTemplate(), Tss2_Sys_AC_GetCapability(),  Tss2_Sys_AC_Send() and Tss2_Sys_Policy_AC_SendSelect().


flihp:
Good catch @tstruk, having to do a 3.0 release so soon after 2.0 wouldn't have been a good look for us. Merging this now. Glad I forgot to do RC2 on Friday 😉 Will have that done in a bit.

Issue #1026, opened 2018-05-25
Wrong value of TPM2_NUM_PCR_BANKS


tstruk:
Setting TPM2_NUM_PCR_BANKS to 3 worked fine when SHA512 was disabled.
When enabled the Tss2_Sys_GetCapability() fails to numarshal TPML_PCR_SELECTION.
This should be set to 4 or ideally derived from the number of enabled hashes. 
See related discussion in tools: https://github.com/tpm2-software/tpm2-tools/issues/1021

AndreasFuchsSIT:
See my comments on #1027 

VMWare vTPM is one of the unusual TPMs with a bunch of active banks.

But apparently the TCG did not update the TSS spec in order to accomodate for this one TPM that behaves different from all others.

flihp:
We just took this issue back to the TSS WG and we all agreed that the additional overhead of supporting extra PCR banks is a small price to pay for compatibility. We'll be working to hopefully convince VMWare and other vTPM providers to support a small number of banks (2 is ideal) to keep hashing overhead down. The overhead of having 4+ hashes carried out for each extend is non negligible for systems like IMA where an extend operation may be carried out every time a file is read from disk.

So the TSS WG has actually proposed that we increase the value of `TPM2_NUM_PCR_BANKS` from 4 to 16 just to future proof implementations. Embedded implementations that are counting bytes of stack space carefully can themselves tailor this value to their specific environments.

Based on this, if we agree that 16 PCR banks is an acceptable limit we should re-open #1027 and update the patch from 8 to 16. Thoughts?

AndreasFuchsSIT:
Agreed.
Let's however mention the deviation from the linked spec revision in README.md
Preferably prominently in its own section.

PR #1027, opened 2018-05-26
update TPM2_NUM_PCR_BANKS

tstruk:
TPM2_NUM_PCR_BANKS needs to be equal to the number of supported
hash algorithms. Currently 8 hash algorithms are defined by the spec:
SHA1, SHA256, SHA384, SHA512, SM3_256, SHA3_256, SHA3_384, SHA3_512.

### DIFF
--- a/include/tss2/tss2_tpm2_types.h
+++ b/include/tss2/tss2_tpm2_types.h
@@ -41,7 +41,7 @@
 #define TPM2_MAX_SESSION_NUM   3    /* this is the current maximum value */
 
 /* TPM constants for buffer sizes */
-#define TPM2_NUM_PCR_BANKS      3
+#define TPM2_NUM_PCR_BANKS      16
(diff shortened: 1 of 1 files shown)


flihp:
'cannot' is a very strong word. I'd rather be able to support these edge cases than claim compliance with the current draft spec. This issue will likely turn into spec work upstream to resolve the issue definitively. Till that's done I'd favor a work-around even if it means deviation (that will be thoroughly documented).

flihp:
Based on #1026 and discussion in TCG TSS2 working group we've all agreed to update the draft spec our implementation is based on to get us better support with implementations enabling a lot of PCR banks. The number we came up with to "future proof" our implementation is `16`. This will cause a bit of bloat in stack space but highly embedded applications can tweak this value to suit their environment.

AndreasFuchsSIT:
And please write a few words to README.md

flihp:
Still missing data in the README but I'll add that and push directly.

Issue #736, opened 2018-01-29
coveralls authorization redirect loop


flihp:
I've been working to reauthorize the various services that we use for CI, test and code coverage metrics since the migration of the repo last week. It looks like I've got everything except for coveralls working. Unfortunately I think I've run across a bug in coveralls which I've reported here: https://github.com/lemurheavy/coveralls-public/issues/1066

If anyone has experience here I'd love the help. As of now I can't even log into my coveralls account due to the repo reauthorization workflow loop I've stumbled across. As reported in that issue, I've found other reports on github of this issue but the only work-around I've found is to stop using coveralls and use codecov.io instead. If we don't get a response to this issue in a few days that's likely the direction we'll have to go.

PeterHuewe:
Example codecov.io output:
https://codecov.io/gh/PeterHuewe/tpm2-tss
The travis integration is quite easy

flihp:
Just backed out the coveralls stuff from .travis.yml. Should get builds going again. Codecov.io comes next.

PeterHuewe:
When do we plan to add codecov.io ?
Unfortunately it's hard to register the support for these third party apps without having the access rights to the repo/organization.

tstruk:
@flihp did you request codecov.io addition into tpm2-software already?

flihp:
I was playing around with tpm2-abrmd and the code coverage while I was on vacation and it looks like they've fixed their authentication redirect loop bug. I'm tempted to see if I can revert 847234899fa6522bd96376f0e2e7ad906b283ef1 and get it working again. But I want to check first: have we effectively lost confidence in coveralls and still want to move over to codecov.io?

PeterHuewe:
Both, both is good?

Am 6. Juli 2018 20:57:07 MESZ schrieb Philip Tricca <notifications@github.com>:
>I was playing around with tpm2-abrmd and the code coverage while I was
>on vacation and it looks like they've fixed their authentication
>redirect loop bug. I'm tempted to see if I can revert
>847234899fa6522bd96376f0e2e7ad906b283ef1 and get it working again. But
>I want to check first: have we effectively lost confidence in coveralls
>and still want to move over to codecov.io?

-- 
Sent from my mobile

AndreasFuchsSIT:
codecov.io should be fast and easy:
https://github.com/tpm2-software/tpm2-tss-engine/blob/master/.travis.yml#L98
plus
[...]

PR #1047, opened 2018-06-06
build: Remove use of coveralls

flihp:
Coveralls hasn't been able to recover from our transition from the Intel
01org github org to the new tpm2-software org. Coveralls has started
returning errors for this repo and since I haven't even been able
to login to the coveralls.io website for the last few months debugging
this is impossible.

This commit leaves all of the mechanics of collecting code coverage
metrics intact. It only disables use of the coveralls webservice for
maintaining history and providing an interface to the data. This is a
short term solution to get our builds passing again.


flihp:
related to #736, should get our builds "green" again

flihp:
Good catch. I'm having trouble pushing to github right now but will update as soon as I can.

Issue #1035, opened 2018-06-04
ESYS: Make default tcti configurable


AndreasFuchsSIT:

AndreasFuchsSIT:
Add support for an env-varialble TSS2_TCTI=module:conf for Esys_Initialize()

AndreasFuchsSIT:
This will hopefully resolve with https://github.com/tpm2-software/tpm2-tss/wiki/TCTI-loader-library and version 2.3

AndreasFuchsSIT:
This is a non-issue now, since all tools and similar use the ldr and pass their own env variable to it.

PR #1049, opened 2018-06-07
[RFC] ESYS: Change behavior of default tcti

AndreasFuchsSIT:
Now the ESYS first checks for en environment variable TSS2_TCTI=<tctilibrary>[:config].
Then it will use the application provided tcti. Then it will initialize the standard tctis.

See this as an RFC to be merged if people like the behavior.

Fixes: #1035

### DIFF
--- a/src/tss2-esys/esys_context.c
+++ b/src/tss2-esys/esys_context.c
@@ -79,11 +79,9 @@ Esys_Initialize(ESYS_CONTEXT ** esys_context, TSS2_TCTI_CONTEXT * tcti,
     /* Store the application provided tcti to be return on Esys_GetTcti(). */
     (*esys_context)->tcti_app_param = tcti;
 
-    /* If no tcti was provided, initialize the default one. */
-    if (tcti == NULL) {
-        r = get_tcti_default(&tcti);
-        goto_if_error(r, "Initialize default tcti.", cleanup_return);
-    }
+    /* This function will initialize a default tcti if necessary. */
+    r = get_tcti_default(&tcti);
+    goto_if_error(r, "Initializing tcti.", cleanup_return);
 
     /* Initialize the ESAPI */
     r = Tss2_Sys_Initialize((*esys_context)->sys, syssize, tcti, abiVersion);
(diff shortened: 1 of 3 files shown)


flihp:
I think I'm onboard with the general functionality. We should namespace the env var appropriately however and since this is specific to ESYS it should probably be `TSS2_ESYS_TCTI`. This is also something we should delay till after 2.0 release is tagged and signed. It can go into a 2.1 release once we've socialized the functionality a bit with TSS WG etc.

tstruk:
@flihp @williamcroberts do you have any more comments on this one?

joshuagl:
Is there a precedent for the environment overriding the application supplied tcti? It seems like the wrong order to me, I'd expect to use i) application provided tcti ii) tcti from environment iii) try standard tctis

For example the tpm2-tools command-line option always overrides the environment variable.

flihp:
I was just about to reply with the same point. I've not implemented any environment variables for tpm2-abrmd but IMHO command line arguments take precedence override the environment.

Even more damning with this approach:
Lets assume that my application takes command line options and instantiates a TCTI based on this data. If the environment variable is set, then the `Esys_Initialize` function will ignore the TCTI instance passed in and use the data from the environment instead. 

When we're done using the ESYS context and my application calls the `Tss2_Tcti_Finalize` function it will finalize the TCTI it's using. In this case my application will think it's finalizing the TCTI context that I passed in, but it'll be finalizing the one it created instead. In this case my application has no way of knowing that the TCTI it created and passed into the `Initialize` function was ignored and that it still needs to have the `finalize` function called on it. This has the potential to leak resources in very unexpected way.

williamcroberts:
Yeah I agree with @joshuagl

AndreasFuchsSIT:
I guess that makes sense... So only if the provided tcti==NULL will we look at the env variable followed by the default.so files.

AndreasFuchsSIT:
I'll reevaluate. Given that we don't want to override the application choice, the NULL-case can probably be handled with the tss2-tcti-default.so behavior. So I'm closing this PR.

PR #1057, opened 2018-06-12
ESYS: Fix handle returning in Esys_TR_Deserialize

AndreasFuchsSIT:
[updated]
Fix a bug in Esys_TR_Deserialize where no correct ESYS_TR handle was returned to the caller.

### DIFF
--- a/src/tss2-esys/esys_tr.c
+++ b/src/tss2-esys/esys_tr.c
@@ -83,6 +83,7 @@ Esys_TR_Deserialize(ESYS_CONTEXT * esys_context,
     size_t offset = 0;
 
     _ESYS_ASSERT_NON_NULL(esys_context);
+    *esys_handle = esys_context->esys_handle_cnt++;
     r = esys_CreateResourceObject(esys_context, *esys_handle, &esys_object);
     return_if_error(r, "Get resource object");
 


src/tss2-esys/esys_tr.c
```diff
@@ -83,6 +83,7 @@ Esys_TR_Deserialize(ESYS_CONTEXT * esys_context,
     size_t offset = 0;
 
     _ESYS_ASSERT_NON_NULL(esys_context);
+    *esys_handle = esys_context->esys_handle_cnt++;
```
flihp: This looks to be more than just a counting issue as the commit message states. Without this change it looks like the `esys_handle` returned to the caller was never set? It's hard to say since I'm not sure what happens to the `esys_handle` when it's passed to `esys_CreateResourceObject`. The comment block for this function however states that the `esys_handle` parameter is a reference to the `ESYS_TR` object being serialized which implies that it already has an `ESYS_TR` allocated? I don't believe this is actually true but if it is then we're overwriting a valud `ESYS_TR` handle. This does however look like a typo in the documentation that's furthering my confusion.

src/tss2-esys/esys_tr.c
```diff
@@ -221,7 +222,7 @@ Esys_TR_FromTPMPublic_Finish(ESYS_CONTEXT * esys_context, ESYS_TR * object)
         objectHandleNode->rsrc.misc.rsrc_key_pub = *public;
         SAFE_FREE(public);
         SAFE_FREE(name);
-            SAFE_FREE(qualifiedName);
+        SAFE_FREE(qualifiedName);
```
flihp: This change seems to be unrelated to the issue identified in the commit message. Is this just a whitespace fix?

flihp:
Since this PR needs a bit of work and is unrelated to the API I'm planning to delay the merge till after the 2.0 release gets signed (today if all goes well). We can do a bugfix release soon after 2.0 to include this.

src/tss2-esys/esys_tr.c
AndreasFuchsSIT: You are correct. The commit message was wrong. Updated.

src/tss2-esys/esys_tr.c
AndreasFuchsSIT: Yes, I remove this from the commit.

AndreasFuchsSIT:
I addressed the issues.
I'd still like to see this (as last minute fix) in 2.0, since it's a functional bug.

flihp:
Thanks for updating the commit message. This is now much more clear. Merging and cutting another RC in a moment.

Issue #1030, opened 2018-05-30
Tests do no proper cleanup at the end


RSynch:
Currently, a number of tests do not clean up after their test execution (esys-policy-regression.int.c for example does not undefine defined NV space).

Other tests do a cleanup, but not under all circumstances (esys-nv-ram-counter.int.c does undefine its defined NV space, but this is skipped in case an error is returned earlier).

There are more similar issues that leave the TPM in a state where following tests are doomed to fail. When the tests are run against the simulator, this is not a problem since every test starts its own simulator and discards it when finished. If however the tests shall ever run on a HW TPM (even if it is only to prove the TSS is working properly), the tests should be fixed.

The right way to achieve this (fixing each test individually or a more generic approach) remains to be discussed.

tstruk:
Fixed by https://github.com/tpm2-software/tpm2-tss/pull/1060

PR #1060, opened 2018-06-14
test: Undefine NV space and make sure the cleanup is called in any case

RSynch:
Addressing #1030.
This is another occasion where defined NV space is not undefined.
Furthermore, this tests contains a number of possibly unsupported commands. Some of them are checked for, others are not (e.g. PolicyPhysicalPresence is optional in the PC Client Spec, however its RC is not checked for TPM_RC_COMMAND)

This can be seen as a proposal of how the tests in general could behave:
* Have a cleanup label to remove/revert anything that would have lasting impact on the TPM
* Before returning an error, always check for TPM_RC_COMMAND and skip if necessary

Since this would have an impact on a number of tests at least, any suggestions on how to improve this are welcome. I could imagine that the code under the exit label would make sense to be in a macro, for example.

### DIFF
--- a/test/integration/esys-policy-regression.int.c
+++ b/test/integration/esys-policy-regression.int.c
@@ -66,7 +66,7 @@ cmp_policy_digest(ESYS_CONTEXT * esys_context,
 int
 test_invoke_esapi(ESYS_CONTEXT * esys_context)
 {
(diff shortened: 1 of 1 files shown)


RSynch:
Done

Issue #672, opened 2017-12-13
Port simulator TCTI to Windows


flihp:
We removed the windows code from this TCTI some time back as it forced a dependency on C++ libraries. This work needs to be recreated using straight winsock / C only. This should probably be done in tandem with #671

PR #1066, opened 2018-06-20
tcti: Port sumulator TCTI to Windows

### LINKED ISSUE #671: Visual Studio solution for simulator TCTI
This issue covers the creation of a solution file compatible with the free visual studio version capable of producing a dll and lib file for the simulator TCTI

davidjmaria:
Add conditional inclusion directives to separate POSIX and Windows
specific code.

Replace hard coded -1 values with SOCKET_ERROR and INVALID_SOCKET.

Add Visual Studio solution for Simulator TCTI

Add .def file for building the Simulator TCTI dll

Rewrite the TEMP_RETRY macro. Previously this macro was a stement
expression; statement expressions are part of the non-standard gcc
extension statement-exprs.

When compiling on Windows:
 - Initialize winsock
 - Replace errno with WSAGetLastError()
 - Replace "read" with "recv"
 - Replace "write" with "send"
 - Replace "close" with "closesocket"
 - Set _HOST_NAME_MAX to MAX_COMPUTERNAME_LENGTH instead of
 _POSIX_HOST_NAME_MAX
 - Map strtok_r to strtok_s

Turn off parallel build in Appveyor

Fixes #671, #672

Signed-off-by: David Maria <davidjmaria@fb.com>

### DIFF
--- a/src/tss2-mu/tss2-mu.vcxproj
+++ b/src/tss2-mu/tss2-mu.vcxproj
@@ -43,22 +43,22 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141_clang_c2</PlatformToolset>
+    <PlatformToolset>LLVM-vs2014</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v141_clang_c2</PlatformToolset>
+    <PlatformToolset>LLVM-vs2014</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141_clang_c2</PlatformToolset>
+    <PlatformToolset>LLVM-vs2014</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>false</UseDebugLibraries>
-    <PlatformToolset>v141_clang_c2</PlatformToolset>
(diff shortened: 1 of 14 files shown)


src/util/io.h
```diff
@@ -5,20 +5,42 @@
...
-({  int __ret; \
+#ifdef _WIN32
+#define TEMP_RETRY(dest, exp) \
+{   int __ret; \
     do { \
         __ret = exp; \
-    } while (__ret == -1 && errno == EINTR); \
-    __ret; })
+    } while (__ret == SOCKET_ERROR && WSAGetLastError() == WSAEINTR); \
+    dest = __ret; }
```
tstruk: Why do we need to change this? Doesn't Visual Studio support return values from macros?

src/tss2-mu/tss2-mu.vcxproj
```diff
@@ -43,12 +43,12 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v141_clang_c2</PlatformToolset>
```
tstruk: Looks like the first patch changes some of these and patch two changes the test in the same file. Is there a reason why these need to be in separate patches. If not please squash them together.

src/util/io.h
davidjmaria: Error codes made by Windows Sockets are not made available through errno, they have to be retrieved by using WSAGetLastError() instead. More information on this can be found here: https://docs.microsoft.com/en-us/windows/desktop/WinSock/error-codes-errno-h-errno-and-wsagetlasterror-2.

src/tss2-mu/tss2-mu.vcxproj
davidjmaria: The commits are now squashed

src/util/io.h
davidjmaria: Just realized I misunderstood what you were asking about in this comment, sorry about that. I explained this change in the commit message:
"Rewrite the TEMP_RETRY macro. Previously this macro was a statement
expression; statement expressions are part of the non-standard gcc
extension statement-exprs."

I have found that even with the Clang LLVM compiler, visual studio does not support statement expressions. More information about statement expressions can be found here: https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html

src/util/io.c
```diff
@@ -147,8 +183,20 @@ socket_connect (
...
+    const char *h = hostname;
+    WSADATA wsaData;
+    int iResult;
+    iResult = WSAStartup(MAKEWORD(2, 2), &wsaData);
+    if (iResult != 0) {
+        LOG_WARNING("WSAStartup failed: %d", iResult);
+        return TSS2_TCTI_RC_IO_ERROR;
+    }
+#else
+    char host_buff[HOST_NAME_MAX] __attribute__((unused));
```
tstruk: That should char host_buff[_HOST_NAME_MAX]

davidjmaria: Thanks, this is now fixed.

PR #1078, opened 2018-07-05
build: Add esys_crypto files to ESYS VS Solution

davidjmaria:
Add new esys_crypto_gcrypt.c/h files to the Visual Studio solution for ESYS
to fix the current Windows build failures.


davidjmaria:
The commit message has been updated.

PR #1081, opened 2018-07-06
Check for progs

### LINKED ISSUE #1059: Configure: Check for netstat and other progs
We should check for all commandline tools used in the integration tests at configure time it integration tests are enabled.
One that is not checked is `netstat` and I assume `ps`.

This comes after going through some extensive remote debugging that would not have been needed.

flihp:
this resolves #1059


flihp:
I was thinking the same thing while writing this. I remember wrestling with the simulator when I wrote the script and I think the job control code is unnecessary. Also an m4 'for' loop would make the autoconf script even better.

PR #1089, opened 2018-07-11
ESYS: Fix RSA encryption with exponent == 0

JuergenReppSIT:
TPM spec part 2 says: When zero, indicates that the exponent is the default of 2**16 + 1.
This default now will be used in the corresponding crypto functions.


flihp:
LGTM. Verified behavior as described in part 2, section 12.2.3.5, table 186.

Issue #268, opened 2016-08-29
TCTI for TBS on Windows


flihp:
Windows support requires a TCTI for use of the TBS (https://msdn.microsoft.com/en-us/library/windows/desktop/aa446792(v=vs.85).aspx). The TBS is effectively the a resource manager / tpm access broker for Windows.

There is a PR from almost a year ago in the PR queue. @wcarthur1 asked the author for some cleanup that never happened. This must be done, and built as a library (see #267) before we advertise Windows support.

flihp:
superseded by #276

flihp:
PR #82 has an implementation of a TCTI for use with TBS. Anyone interested in Windows support should us this as a starting place.

flihp:
Reopening and tagging 'help wanted'.

tstruk:
At this stage resource manager & access broker for Windows is out of scope for this project.
ABRMD lives in a separate repo https://github.com/tpm2-software/tpm2-abrmd.
Closing the issue.

flihp:
This issue may have been closed a bit prematurely. This isn't so much about implementing any sort of resource management on Windows. Instead it's intended to track the implementation of a TCTI capable of communicating with the existing resource manager in Windows. Microsoft calls this interface / component TBS: https://docs.microsoft.com/en-us/windows/desktop/tbs/about-tbs

This is required if we intend for Windows users to interact with the TPM integrated into their platform. Without it we can only support the software TPM2 simulator and that's not useful for anything other than testing.

If I'm missing something feel free to close this issue again and fill me in on what I'm misunderstanding.

tstruk:
The title was a bit misleading.
In this case I would rename it to something like:
"TCTI for TBS on Windows"

flihp:
works for me

PR #1091, opened 2018-07-11
tcti: Add Windows TBS TCTI

davidjmaria:
Add a TBS TCTI, this allows the tss libraries to interact with physical
TPM devices on Windows.

Add a Visual Studio solution to build the TBS TCTI Library.

Change esys_tcti_default to include the TBS TCTI when building on
Windows.

Fixes: #268 

Signed-off-by: David Maria <davidjmaria@fb.com>


flihp:
Review for this is being held up by #1092.

flihp:
This should be rebased on master to pick up 8507a287916f244bdf3911dcb325841484ba41b5. Without this patch it's not possible to build w/o gcrypt headers installed. This is a minor inconvenience and I've already rebased a local copy for my review.

flihp:
May want to wait till #1098 gets merged to rebase since that flips the switch to build against OpenSSL by default instead of gcrypt.

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,291 @@
...
+    params.includeTpm20 = 1;
+    params.version = TBS_CONTEXT_VERSION_TWO;
+    tbs_rc = Tbsi_Context_Create ((PCTBS_CONTEXT_PARAMS)&params, &(tcti_tbs->hContext));
+    if (tbs_rc != TBS_SUCCESS) {
+        LOG_WARNING ("Failed to create context with TBS error: 0x%x", tbs_rc);
+        return TSS2_TCTI_RC_IO_ERROR;
+    }
+
+    tbs_rc = Tbsi_GetDeviceInfo (sizeof (info), &info);
+    if (tbs_rc != TBS_SUCCESS) {
```
flihp: leaking the context created by `Tbsi_Context_Create` on the error path

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,291 @@
...
+        LOG_WARNING ("Failed to get device information with TBS error: 0x%x", tbs_rc);
+        return TSS2_TCTI_RC_IO_ERROR;
+    }
+    if (info.tpmVersion != TPM_VERSION_20) {
+        LOG_WARNING ("Failed to create context, TPM version is incorrect");
```
flihp: leaking the context created by `Tbsi_Context_Create` on the error path

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,291 @@
...
+        return TSS2_TCTI_RC_IO_ERROR;
+    }
+
+    tcti_tbs->resultBuffer = malloc (sizeof (BYTE) * TBS_RESULT_MAX_BUFFER_SIZE);
+    if (tcti_tbs->resultBuffer == NULL) {
```
flihp: leaking the context created by `Tbsi_Context_Create` on the error path

tstruk:
@davidjmaria you can go ahead and rebase that now. Thanks

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,291 @@
...
+    TSS2_TCTI_SET_LOCALITY (tctiContext) = tcti_tbs_set_locality;
+    TSS2_TCTI_MAKE_STICKY (tctiContext) = tcti_make_sticky_not_implemented;
+    tcti_tbs = tcti_tbs_context_cast (tctiContext);
+    tcti_common = tcti_tbs_down_cast (tcti_tbs);
+    tcti_common->state = TCTI_STATE_TRANSMIT;
+
+    memset (&tcti_common->header, 0, sizeof (tcti_common->header));
+    tcti_common->locality = 0;
+
+    params.includeTpm20 = 1;
```
flihp: Does this leave the `includeTpm12` field uninitialized? I'm reading through the docs for the `Tbsi_Context_Create` function and if both are set (or one uninitialized and set unintentionally) then the create function may return a context that's communicating with a 1.2 TPM. You check below using the `Tbsi_GetDeviceInfo` function so the code will catch the error but if we can only communicate with TPM2 devices then initializing this field to `0` (possibly where the structure is declared) makes sense. Can't hurt to check again via `GetDeviceInfo` as well if we don't trust / want to double check the create function.

davidjmaria:
This is now rebased onto master.

src/tss2-tcti/tcti-tbs.c
davidjmaria: This should be fixed now, `includeTpm12` is now initialized to `0` on the line below this. I plan on keeping the `GetDeviceInfo` check just in case though.

src/tss2-tcti/tcti-tbs.c
davidjmaria: These should be fixed now, I've added calls to `Tbsip_Context_Close` on the error paths. Currently there is no check for the success of `Tbsip_Context_Close`, since `TSS2_TCTI_RC_IO_ERROR` will be returned either way. If you would like to check for the success of `Tbsip_Context_Close` for logging purposes I'll add that as well.

src/tss2-tcti/tcti-tbs.c
davidjmaria: Fixed, view reply to comment on line 261.

src/tss2-tcti/tcti-tbs.c
davidjmaria: Fixed, view reply to comment on line 261.

src/tss2-tcti/tcti-tbs.c
flihp: I'm fine ignoring the response from the `close` function. By this point we've already logged one `WARNING` that will clearly identify the code where the failure occurred.

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,295 @@
...
+    }
+    rc = tcti_common_transmit_checks (tcti_common, command_buffer);
+    if (rc != TSS2_RC_SUCCESS) {
+        return rc;
+    }
+    LOGBLOB_DEBUG (command_buffer,
+                   command_size,
+                   "sending %zu byte command buffer:",
+                   command_size);
+    tbs_rc = Tbsip_Submit_Command (tcti_tbs->hContext,
```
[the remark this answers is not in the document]
davidjmaria: What are your thoughts on making a separate `commandBuffer` to use for this purpose, instead of using one buffer for both the commands and results? Not sure if its worth the extra memory usage, but I think it would improve the readability and might ease the troubleshooting of any future issues.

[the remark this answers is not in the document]
flihp: Any reason we can't pass the caller provided `response_buffer` directly to the `Tbsip_Submit_Command` function call when it's moved to the `tcti_tbs_receive` function? If so then we'll only need a single buffer that will hold the contents of the caller provided `command_buffer` in `tcti_tbs_transmit` and then passed to `Tbsip_Submit_Command` as the `pabCommand` parameter.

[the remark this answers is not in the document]
davidjmaria: I think this could cause issues when a null `response_buffer` is passed into `tcti_tbs_receive`. In this case `tcti_tbs_receive` is expected to return the size of the response, which is stored in `resSize` by `Tbsip_Sumbit_Command`. In this case `Tbsip_Submit_Command` would be called with the `pabResult`  parameter as null since `response_buffer` is null, and the response would be lost.

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,295 @@
...
+    TSS2_TCTI_CONTEXT *tctiContext,
+    size_t command_size,
+    const uint8_t *command_buffer)
+{
+    TSS2_TCTI_TBS_CONTEXT *tcti_tbs = tcti_tbs_context_cast (tctiContext);
+    TSS2_TCTI_COMMON_CONTEXT *tcti_common = tcti_tbs_down_cast (tcti_tbs);
+    TSS2_RC rc = TSS2_RC_SUCCESS;
+    TBS_RESULT tbs_rc;
+    /* Reset resSize */
+    tcti_tbs->resSize = sizeof (BYTE) * TBS_RESULT_MAX_BUFFER_SIZE;
```
tstruk: Is sizeof(BYTE) ever be not equal 1?

src/tss2-tcti/tcti-tbs.h
```diff
@@ -0,0 +1,23 @@
...
+/*
+ * Copyright (c) 2018 Intel Corporation
+ * All rights reserved.
+ */
+#ifndef TCTI_TBS_H
+#define TCTI_TBS_H
+
+#include "tcti-common.h"
+
+#define TBS_RESULT_MAX_BUFFER_SIZE 8192
```
tstruk: The result will never be bigger then 4K. The other 4K will never be used.

src/tss2-tcti/tcti-tbs.c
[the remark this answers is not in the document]
tstruk: Where is the Tbsip_Submit_Command() defined?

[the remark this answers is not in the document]
davidjmaria: `Tbsip_Submit_Command` is part of Microsoft's TPM Base Services, the documentation for `Tbsip_Submit_Command` can be found [here](https://docs.microsoft.com/en-us/windows/desktop/api/tbs/nf-tbs-tbsip_submit_command).

src/tss2-tcti/tcti-tbs.c
davidjmaria: In the Malloc on line 273 I take sizeof (BYTE), so I did the same here to be consistent in the unlikely chance that BYTE was ever not equal to 0.

src/tss2-tcti/tcti-tbs.h
davidjmaria: This should be fixed now, I have changed TBS_RESULT_MAX_BUFFER_SIZE to 4096.

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,295 @@
...
+    return TSS2_TCTI_RC_NOT_IMPLEMENTED;
+}
+
+TSS2_RC
+tcti_tbs_set_locality (
+    TSS2_TCTI_CONTEXT *tctiContext,
+    uint8_t locality)
+{
+    /*
+     * TBS currently only supports locality 0
```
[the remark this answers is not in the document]
davidjmaria: I think it would be best to defer this until after this PR is merged.

src/tss2-tcti/tcti-tbs.c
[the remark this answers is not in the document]
flihp: I hadn't considered that case @davidjmaria. I'm still of the mind that the blocking call needs to be `receive`  to keep us from straying too far from the intended behavior. To allow callers to get the size of the response buffer using the `receive` function will require the TCTI maintain / allocate buffers to hold both the command and response then. This is horribly inefficient but seems to be a necessary side effect of wrapping synchronous function in an API designed for an asynchronous one. We should consider this as a use-case when we next update the TCTI API 😒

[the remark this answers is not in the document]
davidjmaria: The other option would be to have `tcti_tbs_receive` behave similarly to `tcti_device_receive`, which just returns the max result buffer size whenever `response_buffer` is null, instaed of returning the actual size of the response.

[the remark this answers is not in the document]
flihp: Works for me. Like you say, I've taken that route in the past.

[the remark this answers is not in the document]
davidjmaria: @flihp Are you aware of any maximum command size defined in the specs? I haven't found anything, in which case I'm going to have to have the transmit and receive functions manage the command buffer instead of having one large command buffer that is managed in context create and finalize.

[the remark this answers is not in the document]
tstruk: The spec only doesn't define it, but it practice both command and response can't be bigger than 4K. TSS defines both to 4K:
https://github.com/tpm2-software/tpm2-tss/blob/master/include/tss2/tss2_tpm2_types.h#L18
so does the reference TPM implementation from Microsoft:
https://github.com/Microsoft/ms-tpm-20-ref/blob/master/TPMCmd/tpm/include/Implementation.h#L196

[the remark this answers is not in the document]
davidjmaria: Since this was a big change, I added it as a separate commit to make review of it easier. I plan on squashing that commit after the review.

src/tss2-tcti/tcti-tbs.c
```diff
@@ -0,0 +1,295 @@
...
+{
+    TSS2_TCTI_TBS_CONTEXT *tcti_tbs = tcti_tbs_context_cast (tctiContext);
+    TSS2_TCTI_COMMON_CONTEXT *tcti_common = tcti_tbs_down_cast (tcti_tbs);
+    TBS_RESULT tbs_rc;
+
+    if (tcti_tbs == NULL) {
+        return;
+    }
+
+    free (tcti_tbs->resultBuffer);
```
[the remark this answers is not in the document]
davidjmaria: This should be fixed.

flihp:
Please squash this @davidjmaria. Will give it a final "once-over" after that's done.

davidjmaria:
Squashed and rebased.

include/tss2/tss2_tcti_tbs.h
```diff
@@ -0,0 +1,24 @@
+/* SPDX-License-Identifier: BSD-2 */
+/*
+ * Copyright (c) 2015 - 2018, Intel Corporation
```
flihp: @davidjmaria you should talk to your legal department about this copyright statement since you / Facebook originated this code, not Intel.

davidjmaria: I talked to the legal department, I have now added the facebook copyright header but have left the intel copyright header in as well since this code is heavily based off of, and includes code from, the code for the device TCTI.

Issue #1097, opened 2018-07-15
AppVeyor build times are bad.


flihp:
The current configuration mirrors the linux build using GNU TLS / libgcrypt. In most cases I'd say this is the right thing to do but downloading and building gcrypt and all of its dependencies increases build times by nearly 4x. This is way too much time. We need to either mirror pre-built binaries or hopefully something easier.

flihp:
@davidjmaria : Interested in your input since the proposed solution in #1098 may have an impact on your work.

PR #1098, opened 2018-07-15
build: Use OpenSSL library for crypto backend to tss2-esys.

flihp:
For now this is for efficiency since the Windows CI stuff doesn't
include pre-built gcrypt libraries. IMHO it makes more sense as the
default as well due to the popularity of OpenSSL.

AppVeyor build images include prebuilt OpenSSL 1.0.2 libraries & headers
at C:\OpenSSL-(Win32|Win64). The final esys library is linked against
libeay32.lib. The code works against version 1.1.0 / libcrypto.lib but
using the default on the appveyor image is the right place to start.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,14 +3,6 @@ pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
 image: Visual Studio 2017
-before_build:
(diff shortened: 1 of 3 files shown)


flihp:
This resolves #1097

PR #1103, opened 2018-07-17
TEST ESYS: Enable tests for physical TPM.

JuergenReppSIT:
* To enable tests with a physical TPM the configure options:
     --with-ptpm=[device]
     --with-ptpmtests=[test cases]

   were added. To compile the integration tests --enable-integration has
   to be used. The test cases are a comma separated list of:
   mandatory, optional and destructive. The default is mandatory.
   To avoid parallel usage of the TPM the tests should be executed with:
   make check-ptpm
   or with:
   make check -j 1.
   The compiled integration tests for a simulator test can be used for
   tests with a physical TPM without re-compilation.

* A second int_log compiler script was added to execute these tests.

* Split policy regression tests into two parts. Optional commands are
  moved to the test policy-regression-opt.

* The test policy-ticket will return success if the PolicyTicket command
  is not available, but all other commands in this tests were
  successful.

* Exponent 0 will be used in the ESAPI integration tests.
  Exponent 65537 for RSA keys is optional in TPM spec while exponent
  is mandatory:
  A TPM compatible with this specification and supporting RSA shall
  support two primes and an exponent of zero. Support for other values
  is optional. (Spec Part2 12.2.3.5 TPMS_RSA_PARMS).

* The check whether optional commands are available are adapted for the
  usage with a resource manager TPM.

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -33,8 +33,13 @@ TESTS_LDADD = $(noinst_LTLIBRARIES) $(lib_LTLIBRARIES) $(LIBCRYPTO_LIBS) $(libut
 
 # test harness configuration
 TEST_EXTENSIONS = .int
+if TESTPTPM
+INT_LOG_COMPILER = $(srcdir)/script/int-log-compiler-ptpm.sh
+INT_LOG_FLAGS = --ptpm=$(PTPM)
+else
 INT_LOG_COMPILER = $(srcdir)/script/int-log-compiler.sh
-EXTRA_DIST += $(INT_LOG_COMPILER)
+endif
+EXTRA_DIST += $(srcdir)/script/int-log-compiler.sh $(srcdir)/script/int-log-compiler-ptpm.sh
 AM_TESTS_ENVIRONMENT = PATH="$(PATH)"
 
 check-programs: $(check_PROGRAMS)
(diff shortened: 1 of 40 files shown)


tstruk:
Just some comments on the commit message:

* "Exponent 65537 for RSA keys is optional in TPM spec while exponent
is mandatory" - should be: "... while exponent of zero is mandatory"
* There is nothing about exponent in Part 2 12.2.3.4 TPMS_ASYM_PARMS. Did you mean 12.2.3.5 TPMS_RSA_PARMS ?

Makefile-test.am
```diff
@@ -99,6 +104,10 @@ endif #UNIT
 
 if ENABLE_INTEGRATION
 noinst_LTLIBRARIES += test/integration/libtest_utils.la
+
+if TESTPTPM
```
tstruk: instead of doing if else endif just do:
if !TESTTPM
endif

JuergenReppSIT: Yes that's true I meant 12.2.3.5 TPMS_RSA_PARMS. Sorry.

JuergenReppSIT:
Related to exponent 65537: Yes that's true I meant 12.2.3.5 TPMS_RSA_PARMS. Sorry.

tstruk:
Could you amend it please

JuergenReppSIT:
The description is amended

tstruk:
You have changed the PR description, but the commit message still refers to 12.2.3.4. You need to do it by git commit --amend

PR #1104, opened 2018-07-18
ESYS: Fix the bignum conversion in the OSSL backend.

JuergenReppSIT:
There is an error in the conversion if the first >= 8 bits of
the bignumber are zero. The computed offset has to be added to the
pointer.


src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -38,7 +38,7 @@ int BN_bn2binpad(const BIGNUM *bn, unsigned char *bin, int bin_length)
     int len_bn = BN_num_bytes(bn);
     int offset = bin_length - len_bn;
     memset(bin,0,offset);
-    BN_bn2bin(bn, bin);
+    BN_bn2bin(bn, bin+offset);
```
tstruk: Just a nitpick - there should be a space on both sides of the '+'
There is also no Signed-by-off in the commit message.

Issue #1105, opened 2018-07-18
Libgcrypt init is missing


tstruk:
Esapi integ tests with gcrypt leave a trace in the system log:
esys-make-credential-session.int[22878]: Libgcrypt warning: missing initialization - please fix the application

tstruk:
This is fixed by https://github.com/tpm2-software/tpm2-tss/pull/1107

PR #1107, opened 2018-07-19
ESYS: Add initialization function for crypto backend.

JuergenReppSIT:
Adressing #1105

 * For the gcrypt backend the function gcry_check_version will be called.
   A version mismatch will be checked and the internal tables of libgrypt
   will be initialized.
 * For the OpenSSL backend the function OpenSSL_add_all_algorithms will
   be called.


src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -38,7 +38,7 @@ int BN_bn2binpad(const BIGNUM *bn, unsigned char *bin, int bin_length)
     int len_bn = BN_num_bytes(bn);
     int offset = bin_length - len_bn;
     memset(bin,0,offset);
-    BN_bn2bin(bn, bin);
+    BN_bn2bin(bn, bin+offset);
```
tstruk: You need to rebase it. This commit is not part of the PR.

PR #1108, opened 2018-07-19
tcti: add support for partial reads

tstruk:
This enables partial reads in tcti-device.
The feature has dependecy on the same being enable in the driver
so it is disabled by default. It can be enabled by a new configure
time option --enable-tcti-partial-reads=yes


tstruk:
This requires this driver change to be enabled:
https://patchwork.kernel.org/patch/10534823/

tstruk:
@AndreasFuchsSIT, yes, this can be closed until we agree on the driver behavior.
Currently the driver ignores O_NONBLOCK flag.

PR #1113, opened 2018-07-30
ESYS: Type checks for ESAPI input parameters added.

JuergenReppSIT:
If possible a type check for every input parameter is performed for
every ESAPI (async) call.

### DIFF
--- a/src/tss2-esys/tpm2_type_check.c
+++ b/src/tss2-esys/tpm2_type_check.c
@@ -1,0 +1,3768 @@
+/* SPDX-License-Identifier: BSD-2 */
+/*******************************************************************************
+ * Copyright 2017-2018, Fraunhofer SIT sponsored by Infineon Technologies AG
+ * All rights reserved.
+ ******************************************************************************/
+#include <inttypes.h>
+
+#include "tss2_esys.h"
+#include "tpm2_type_check.h"
+#include "esys_iutil.h"
+#define LOGMODULE esys
+#include "util/log.h"
+
+/*** Table 5 - Definition of Types for Documentation ClarityTable 5 - Definition of Types for Documentation Clarity ***/
+
+/**
+ * Check, if a variable is actually of type TPM2_ALG_ID.
+ * @param[in] in variable to check.
+ * @retval TSS2_RC_SUCCESS on success.
+ * @retval TSS2_ESYS_RC_BAD_VALUE for type error.
+ */
+TSS2_RC
+iesys_TPM2_ALG_ID_check(
+    const TPM2_ALG_ID in)
+{
+    if ((in == TPM2_ALG_ERROR) ||
+        (in == TPM2_ALG_RSA) ||
+        (in == TPM2_ALG_SHA) ||
+        (in == TPM2_ALG_SHA1) ||
+        (in == TPM2_ALG_HMAC) ||
+        (in == TPM2_ALG_AES) ||
+        (in == TPM2_ALG_MGF1) ||
+        (in == TPM2_ALG_KEYEDHASH) ||
+        (in == TPM2_ALG_XOR) ||
+        (in == TPM2_ALG_SHA256) ||
+        (in == TPM2_ALG_SHA384) ||
+        (in == TPM2_ALG_SHA512) ||
+        (in == TPM2_ALG_NULL) ||
+        (in == TPM2_ALG_SM3_256) ||
+        (in == TPM2_ALG_SM4) ||
+        (in == TPM2_ALG_RSASSA) ||
+        (in == TPM2_ALG_RSAES) ||
+        (in == TPM2_ALG_RSAPSS) ||
+        (in == TPM2_ALG_OAEP) ||
+        (in == TPM2_ALG_ECDSA) ||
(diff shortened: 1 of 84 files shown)


src/tss2-esys/tpm2_type_check.c
```diff
@@ -0,0 +1,3768 @@
...
+ */
+TSS2_RC
+iesys_TPMT_HA_check(const TPMT_HA *in)
+{
+    TSS2_RC ret;
+
+    if (in == NULL)
+        return TSS2_RC_SUCCESS;
+
+    ret = iesys_TPMI_ALG_HASH_check ((in == NULL)? 0 : in->hashAlg);
```
tstruk: The check (in == NULL) ? is redundant here.

src/tss2-esys/tpm2_type_check.c
```diff
@@ -0,0 +1,3768 @@
...
+ */
+TSS2_RC
+iesys_TPMS_ALGORITHM_DESCRIPTION_check(const TPMS_ALGORITHM_DESCRIPTION *in)
+{
+    TSS2_RC ret;
+
+    if (in == NULL)
+        return TSS2_RC_SUCCESS;
+
+    ret = iesys_TPM2_ALG_ID_check ((in == NULL)? 0 : in->alg);
```
tstruk: the check in == NULL is not needed

src/tss2-esys/tpm2_type_check.c
```diff
@@ -0,0 +1,3768 @@
...
+/**
+ * Check, if a variable is actually of type TPMI_ALG_SIG_SCHEME.
+ * @param[in] in variable to check.
+ * @retval TSS2_RC_SUCCESS on success.
+ * @retval TSS2_ESYS_RC_BAD_VALUE for type error.
+ */
+TSS2_RC
+iesys_TPMI_ALG_SIG_SCHEME_check(const TPMI_ALG_SIG_SCHEME in)
+{
+    if ((in == TPM2_ALG_RSASSA) ||
```
tstruk: In some cases we use switch() and is some if(). Any reason for that?

src/tss2-esys/tpm2_type_check.c
```diff
@@ -0,0 +1,3768 @@
...
+ * Check, if a variable is actually of type TPM2_CC.
+ * @param[in] in variable to check.
+ * @retval TSS2_RC_SUCCESS on success.
+ * @retval TSS2_ESYS_RC_BAD_VALUE for type error.
+ */
+TSS2_RC
+iesys_TPM2_CC_check(
+    const TPM2_CC in)
+{
+    if ((in == TPM2_CC_NV_UndefineSpaceSpecial) ||
```
tstruk: Why not just
if (in < TPM2_CC_NV_UndefineSpaceSpecial || in >= TPM2_CC_LAST) return TSS2_ESYS_RC_BAD_VALUE;
else return TSS2_RC_SUCCESS;

src/tss2-esys/tpm2_type_check.c
```diff
@@ -0,0 +1,3768 @@
...
+ */
+TSS2_RC
+iesys_TPMS_CLOCK_INFO_check(const TPMS_CLOCK_INFO *in)
+{
+    TSS2_RC ret;
+
+    if (in == NULL)
+        return TSS2_RC_SUCCESS;
+
+    ret = iesys_TPMI_YES_NO_check ((in == NULL)? 0 : in->safe);
```
tstruk: The check (in == NULL) is not needed.
Same for all the iesys_*_check() below

JuergenReppSIT:
It's not  required by the spec. So for sanity checks it might be sufficient to check that there
are no NULL pointers used for TPML, TPMS, and TPMT parameters. For TPM2B the
TSS Spec (3.8.2.2) states:
>If any command parameter (after the handles) is a TPM2B:
>   If the TPM2B parameter is NULL, the implementation marshals a TPM2B with a size of 0.

SAPI checks TPML, TPMS, and TPMT parameters in all cases but does not marshal
a TPM2B with size 0 if TPM2B parameters are NULL in all cases. Perhaps it would be sufficient to
add this in the macro TPM2B_MARSHAL instead of returning  TSS2_MU_RC_BAD_REFERENCE?
In the cases where ESAPI parameters are used before SAPI is called, ESAPI also has NULL pointer
checks.

tstruk:
The Marshaling spec in 3.3.2 states that for TPM2B, when the 'src' is NULL then TSS2_MU_RC_BAD_REFERENCE error code should be returned. If we add the check to TPM2B_MARSHAL we will break this behavior. This check should be done in the Sys layer, and as far as I remember it does this in most cases, like here for example:
https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-sys/api/Tss2_Sys_Certify.c#L40
We probably need to review if this is done in all the cases. We should also keep in mind that in some of the commands the params are not optional and in case of NULL it should return an error instead of marshaling 0.

JuergenReppSIT:
I think the intention of 3.8.2.2 was not to differentiate between optional and not optional commands. So in every case the NULL point should be checked.

PR #1119, opened 2018-08-02
build: doxygen causes make distcheck to fail

tstruk:
Number of issues with doxygen encountered on make distcheck.
This change fixes the problems.


m4/ax_prog_doxygen.m4
```diff
@@ -547,7 +547,7 @@ doxygen-doc: doxygen-run \$(DX_PS_GOAL) \$(DX_PDF_GOAL)
 ]m4_foreach([DX_i], [DX_loop],
 [[\$(DX_DOCDIR]DX_i[)/\$(PACKAGE).tag: \$(DX_CONFIG]DX_i[) \$(pkginclude_HEADERS)
 	\$(A""M_V_at)rm -rf \$(DX_DOCDIR]DX_i[)
-	\$(A""M_V_at)mkdir \$(DX_DOCDIR]DX_i[)
+	\$(A""M_V_at)mkdir -p \$(DX_DOCDIR]DX_i[)
```
williamcroberts: Shouldn't this have gone upstream?

tstruk: The latest upstream version is different to what we have.

PR #1121, opened 2018-08-09
tcti-device: change device tcti to nonblocking mode

tstruk:
Change device tcti to nonblocking mode of operation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -70,6 +70,15 @@ AC_ARG_ENABLE([esapi],
     [enable_esapi="yes"])
 AM_CONDITIONAL(ESAPI, test "x$enable_esapi" = "xyes")
 
+AC_ARG_ENABLE([tcti-device-async],
+    AS_HELP_STRING([--enable-tcti-device-async],
+	           [Enable asynchronus operation on TCTI device
+		    (note: This needs to be supported by the kernel driver). default is no]),
+    [enable_tcti_device_async=$enableval],
+    [enable_tcti_device_async=no])
+AS_IF([test "x$enable_tcti_device_async" = "xyes"],
+	AC_DEFINE([TCTI_ASYNC],[1]))
+
 AC_ARG_WITH([crypto],
             [AS_HELP_STRING([--with-crypto={gcrypt,ossl}],
                             [sets the ESAPI crypto backend (default is gcrypt)])],
@@ -305,4 +314,5 @@ AC_MSG_RESULT([
     debug:              $enable_debug
     maxloglevel:        $with_maxloglevel
     doxygen:            $DX_FLAG_doc $enable_doxygen_doc
+    tcti-device-async:  $enable_tcti_device_async
 ])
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -64,13 +64,22 @@ AC_ARG_ENABLE([esapi],
             [enable_esapi=$enableval],
             [enable_esapi=yes])
 
-
```
flihp: whitespace. This is good cleanup but not related to this PR / feature.

configure.ac
```diff
@@ -64,13 +64,22 @@ AC_ARG_ENABLE([esapi],
...
     [enable_esapi="yes"])
 AM_CONDITIONAL(ESAPI, test "x$enable_esapi" = "xyes")
 
+AC_ARG_ENABLE([tcti-device-async],
+    AS_HELP_STRING([--enable-tcti-device-async],
+	           [Enable asynchronus operation on TCTI device
+		    (note: This needs to be supported by the kernel driver). default is no]),
+    [enable_tcti_device_async=$enableval],
+    [enable_tcti_device_async=no])
+AM_CONDITIONAL(TCTI_ASYNC, test "x$enable_tcti_device_async" = "xyes")
```
flihp: I don't see a place in the makefile where you use this variable. Unless I missed it, this isn't necessary.

src/tss2-tcti/tcti-device.c
```diff
@@ -118,7 +119,7 @@ tcti_device_receive (
     TSS2_TCTI_DEVICE_CONTEXT *tcti_dev = tcti_device_context_cast (tctiContext);
     TSS2_TCTI_COMMON_CONTEXT *tcti_common = tcti_device_down_cast (tcti_dev);
     TSS2_RC rc = TSS2_RC_SUCCESS;
-    ssize_t  size;
+    ssize_t size = 0;
```
flihp: This good cleanup but it doesn't belong in this commit.

src/tss2-tcti/tcti-device.c
```diff
@@ -251,6 +294,7 @@ Tss2_Tcti_Device_Init (
     TSS2_TCTI_DEVICE_CONTEXT *tcti_dev;
     TSS2_TCTI_COMMON_CONTEXT *tcti_common;
     const char *dev_path = conf != NULL ? conf : TCTI_DEVICE_DEFAULT;
+    mode_t mode = O_RDWR;
```
flihp: This isn't a mode, it's a flag since it's used in the second parameter: `int open(const char *pathname, int flags, mode_t mode);`. The reason you're not seeing a type error is that the `flags` paramter is an `int` and `mode_t` maps back to `int`. So this variable should be an `int` and it should be called `flags`.

configure.ac
tstruk: Removed

src/tss2-tcti/tcti-device.c
tstruk: The size needs to be initialized because the compiler complains otherwise

src/tss2-tcti/tcti-device.c
tstruk: Fixed

tstruk:
@flihp Thanks for your feedback. I have reworked the code accordingly, updated the commit message, and added unit tests to cover the poll() interface. Please have a look.

PR #1137, opened 2018-08-22
ESAPI: Add missing defines in ESAPI header file.

JuergenReppSIT:
* Two missing defines were added (ESAPI Spec: 5.2 Typedef ESYS_TR)
* Two defines not included in the spec were deleted.


tstruk:
Just a general comment. We need to agree on the preferred prefixes in the commit message.
Looking at the 4 PR we have "ESYS:", "ESAPI:", and "esapi:"
looking at the git log there is also "esys:". Looks like ESAPI is the mos common one.
Any preference?

tstruk:
Sure, I will update the CONTRIBUTING.md with this recommendation.

PR #1139, opened 2018-08-22
Esys cleanup crypto backends

JuergenReppSIT:
* Unused functions were removed.
 * The order of error checks was unified to enable equal unit tests for both backends.
 * Macros for returning errors were used instead of if blocks.
 * Several formatting fixes.
 * An unit test for the crypto backends, to check error cases, which are not covered by
   the integration tests, was added.

Signed-off-by: Juergen Repp Juergen.Repp@sit.fraunhofer.de


JuergenReppSIT:
yes it was just for coverage.

PR #1143, opened 2018-08-29
util: C99 compatible logging

LukasJaegerSIT:
tpm2-tss is supposed to be ISO-C99 compatible. So far, util/log.c
used the function `strncasecmp`, which is part of the POSIX- and
BSD-standard but not of the ISO-C99-standard.

Therefore, a new function, `caseInsensitiveStrncmp` is introduced. It
creates lowercase-variants of the strings to compare and calls
the C99-compatible standard function `strncmp` on them. This
mimicks the behavior of `strncasecmp` but remains compatible with
C99.

* Introduced caseInsensitiveStrncmp
* Replaced all calls of strncasecompare with it
* Adjusted names and indentations

### DIFF
--- a/src/util/log.c
+++ b/src/util/log.c
@@ -2,6 +2,7 @@
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
+#include <ctype.h>
 
 #define LOGMODULE log
 #include "log.h"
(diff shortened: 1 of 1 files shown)


src/util/log.c
```diff
@@ -15,6 +16,34 @@
 #define unlikely(x)     (x)
 #endif
 
+static int caseInsensitiveStrncmp(char const * const string1,
```
AndreasFuchsSIT: Could you please add some documentation for this function, esp regarding the reason for us to reimplement strcasecmp ?
Thanks

src/util/log.c
```diff
@@ -15,6 +16,34 @@
 #define unlikely(x)     (x)
 #endif
 
+static int caseInsensitiveStrncmp(char const * const string1, 
+        char const * const string2, 
+        size_t n){
+    if ((string1 == NULL) && (string2 == NULL)){
```
AndreasFuchsSIT: space between ) and {
Also below...

src/util/log.c
```diff
@@ -15,6 +16,34 @@
 #define unlikely(x)     (x)
 #endif
 
+static int caseInsensitiveStrncmp(char const * const string1, 
+        char const * const string2, 
+        size_t n){
```
AndreasFuchsSIT: linefeed between ) and {

src/util/log.c
AndreasFuchsSIT: The second const is without meaning.
Also, current convetion is to use `const char *string1`

LukasJaegerSIT:
@AndreasFuchsSIT Done.

LukasJaegerSIT:
Done.

src/util/log.c
```diff
@@ -15,6 +16,56 @@
...
+    if ((string1 == NULL) && (string2 != NULL)) {
+        return -1;
+    }
+    if ((string1 != NULL) && (string2 == NULL)) {
+        return 1;
+    }
+
+    size_t string1_size = strlen (string1);
+    size_t string2_size = strlen (string2);
+    char lower_string1 [string1_size];
```
tstruk: If you would declare these as unsigned char there would be no need for casting at all.

src/util/log.c
```diff
@@ -15,6 +16,56 @@
...
+ * @param string1 The first of the two strings to compare
+ * @param string2 The second of the two strings to compare
+ * @param n The maximum number of bytes to compare
+ * @return 0 if both strings are equal (case insensitive),
+ *  an integer greater than zero if string1 is greater than
+ *  string 2 and an integer smaller than zero if string1 is
+ *  smaller than string2
+ *
+ */
+static int caseInsensitiveStrncmp(const char *string1,
```
tstruk: Could you rename it to something that doesn't use camel case please.
See the recommendation in https://github.com/tpm2-software/tpm2-tss/blob/master/doc/coding_standard_c.md#naming-variables-functions-and-other-stuff

src/util/log.c
LukasJaegerSIT: You mean the two arrays `lower_string1` and  `2`? They need to be char, otherwise `strncmp` won't work on them (signedness warnings that result in errors). The casting is also not due to signedness but due to `toLower` taking integer values as input and output and those lines cause size warnings if I don't cast here.

src/util/log.c
LukasJaegerSIT: Augh....yes, I should have seen that right away. Sorry. The latest commit should contain the necessary changes.

src/util/log.c
tstruk: Yes, but now there are 4 casts, 2 for each call to tolower(). With unsigned char only two would be needed. If we change strncmp() to memcmp(), then no cast is needed:

    unsigned char lower_string1 [string1_size];
    unsigned char lower_string2 [string2_size];

    for (size_t i = 0; i < string1_size; i++)
        lower_string1[i] = tolower(string1[i]);

    for (size_t i = 0; i < string2_size; i++)
        lower_string2[i] = tolower(string2[i]);

    return memcmp (lower_string1, lower_string2, n);

LukasJaegerSIT: Ah...now I get it. It's done and the commits are squashed.

PR #1145, opened 2018-08-29
ESYS: Fix leak of dlopen handle and Makefile

AndreasFuchsSIT:
- The handle of dlopen used to leak. Save this handle and dlclose on Esys_Finish().
- ESYSFLAGS was actually never set. Probably leftover from an earlier version.


flihp:
😅

AndreasFuchsSIT:
@flihp Any hints ? I don't know VS...

AndreasFuchsSIT:
Ok, I got it fixed by wrapping it in #ifdef NO_DL, similar to esys_tcti_default.c.
I guess, windows LoadLibrary Support is a todo still.

PR #1159, opened 2018-09-25
Fix OpenSSL crypto backend

AndreasFuchsSIT:
- Fixes #1157 
- Adds OpenSSL (in two versions) to tests.

@flihp @tstruk Hope to get this in before 2.2 (since the code-change is small)

PR #1160, opened 2018-09-27
Build: Fix autotools warning on _LDFLAGS

AndreasFuchsSIT:
autotools does not like variables to end on _LDFLAGS and spits out
warnings. So we rename the variable.

Addresses issue posted in #1159 review.


AndreasFuchsSIT:
I'll merge this since we have no review according to rules...

PR #1161, opened 2018-09-29
test: Add integration test for StirRandom

PeterHuewe:
Add a simple test for StirRandom to increase code coverage.
It does only check whether the command itself works and returns success
for suitable input data.

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>


AndreasFuchsSIT:
The Appveyor failure seems to be a misconfigured server.
Since we don't run the tests on Appveyor, this patch in unrelated and I'll merge it.

PR #1163, opened 2018-10-02
RFC ESAPI helper functions

joshuagl:
I'm working on porting the tpm2-tools to the ESAPI and have ended up making a couple of changes to the TSS, per this RFC PR.

The first patch was implemented because I found myself carrying a similar, though less complete, function to `iesys_handle_to_tpm_handle()` in the tpm2-tools library code. I suspect if this change is acceptable there will need to be some modifications, it's currently a path of least resistance patch to expose existing functionality.

The second patch was implemented in order to maintain the _tpm2_evictcontrol_'s ability to detect when a passed handle is persistent: https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_evictcontrol.c#L138
Perhaps there's a more appropriate way to do that in an ESYS world?

### DIFF
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -3229,6 +3229,12 @@ Esys_Free(
     void *__ptr);
 
 TSS2_RC
+esys_tr_get_tpm_handle(
+    ESYS_CONTEXT *esysContext,
+    ESYS_TR handle,
+    TPM2_HANDLE *tpmHandle);
+
+TSS2_RC
 esys_handle_to_tpm_handle(
     ESYS_TR handle,
     TPM2_HANDLE *tpmHandle);
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -3228,6 +3228,11 @@ void
 Esys_Free(
     void *__ptr);
 
+TSS2_RC
+esys_handle_to_tpm_handle(
+    ESYS_TR handle,
+    TPM2_HANDLE *tpmHandle);
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/tss2-esys.def
+++ b/lib/tss2-esys.def
@@ -341,6 +341,7 @@ EXPORTS
     Esys_TR_GetName
     Esys_TR_Serialize
     Esys_TR_SetAuth
+    esys_handle_to_tpm_handle
     Esys_TestParms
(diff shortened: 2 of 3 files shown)


joshuagl:
AppVeyor failure doesn't seem to be related to this change:

> Build FAILED.
> "C:\projects\tpm2-tss\tpm2-tss.sln" (default target) (1) ->
> "C:\projects\tpm2-tss\src\tss2-mu\tss2-mu.vcxproj" (default target) (2) ->
> (PlatformPrepareForBuild target) -> 
>   C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\Microsoft.Cpp.Platform.targets(67,5): error MSB8020: The build tools for LLVM-vs2014 (Platform Toolset = 'LLVM-vs2014') cannot be found. To build using the LLVM-vs2014 build tools, please install LLVM-vs2014 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [C:\projects\tpm2-tss\src\tss2-mu\tss2-mu.vcxproj]

joshuagl:
@AndreasFuchsSIT I'd appreciate your thoughts on the suitability of these changes.

joshuagl:
Thanks for the feedback @AndreasFuchsSIT. For the first patch that's relatively easy copy/paste, I can live with that.

I'm not even sure it's possible to achieve the functionality of the second patch outside of the TSS? If we only want to implement spec, is it worth trying to add this functionality to the spec for ESAPI? Or are there simpler/cleaner/better ways to go about achieving this?

Thanks

AndreasFuchsSIT:
@joshuagl For many cases, you can actually call Esys_TR_GetName, since the name will be the marshalled handle in case of sessions and NV indices.
I'd love to actually know the use cases for these functions.

Best way to get it into spec:
- write up 5-10 lines of use cases
- write up the function specification (similar to existing ESAPI functions)
- Post this stuff somewhere in a Wiki to discuss outside TCG first
- Send it to TCG TSSwg for adoption in next spec.
- Wait 6-12 months for the next TSSwg specs (for this reason public first)

joshuagl:
Thanks @AndreasFuchsSIT, I was missing that detail on Esys_TR_GetName - with that knowledge I can withdraw the second patch of this PR.

PR #1165, opened 2018-10-08
Fix for Tpm2b marshaling code

tstruk:
Fixes an issue found during fuzzing

### DIFF
--- a/src/tss2-mu/tpm2b-types.c
+++ b/src/tss2-mu/tpm2b-types.c
@@ -121,6 +121,12 @@ TSS2_RC Tss2_MU_##type##_Unmarshal(uint8_t const buffer[], size_t buffer_size, \
              (size_t)size); \
         return TSS2_MU_RC_INSUFFICIENT_BUFFER; \
     } \
+    if ((sizeof(*dest) - 2) < size) { \
+        LOG_ERROR("The dest field size of %zu is too small to unmarshal %d bytes", \
+                  sizeof(*dest), size); \
+        return TSS2_MU_RC_INSUFFICIENT_BUFFER; \
+    } \
+\
     if (dest != NULL) { \
         dest->size = size; \
         memcpy(((TPM2B *)dest)->buffer, &buffer[local_offset], size); \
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
If there is any kind of padding, I think sizeof(*dest) could be larger than sizeof(dest->size) + sizeof(dest->buffer).

But I understand that the buffer-field is not allways called buffer, but something else. Maybe we would add a parameter to the makros for the buffername or similar ?

tstruk:
The TPM2B are always size UINT16 and buffer BYTE[]
There is no padding in any of them so sizeof(*dest) will never be larger than sizeof(dest->size) + sizeof(dest->buffer). Or more specifically - sizeof(*dest) - 2 will always be equal to sizeof(dest->buffer_name).
As you said the buffer itself is called differently in different structures, but I don't see a need for an extra param.

tstruk:
To make sure I just checked that with pahole, and for instance:
        /* typedef TPM2B_DATA */ struct {
                /* typedef UINT16 -> uint16_t -> __uint16_t */ short unsigned int size;  /* 2 */
                /* typedef BYTE -> uint8_t -> __uint8_t */ unsigned char      buffer[64]; /* 64 */
        } /* 66 */

/* typedef TPM2B_SENSITIVE_CREATE */ struct {
        /* typedef UINT16 -> uint16_t -> __uint16_t */ short unsigned int size; /* 2 */
        /* typedef TPMS_SENSITIVE_CREATE */ struct {
               /* typedef TPM2B_AUTH -> TPM2B_DIGEST */ struct {
                       /* typedef UINT16 -> uint16_t -> __uint16_t */ short unsigned int size; /* 2 */
                       /* typedef BYTE -> uint8_t -> __uint8_t */ unsigned char  buffer[64]; /* 64 */
                                 } userAuth; /* 66 */
                                 /* typedef TPM2B_SENSITIVE_DATA */ struct {
                                        /* typedef UINT16 -> uint16_t -> __uint16_t */ short unsigned int size; /* 2 */
                                        /* typedef BYTE -> uint8_t -> __uint8_t */ unsigned char  buffer[256]; /* 256 */
                                } data; /* 258 */
                   } sensitive; /* 324 */
   }; /* 326 */

That's for gcc and clang.

tstruk:
Actually TPM2B_PUBLIC does have padding. Given that and that this is compiler specific I think adding an extra param 'buf_name' and using it in the sizeof() keeps us on the safe site. A new PR on the way.

PR #1168, opened 2018-10-09
Tpm2b fix

tstruk:


AndreasFuchsSIT:
@tstruk 2 ideas:
- Should we start putting input-sanitization into #ifdefs in sapi so firmware implementors can disable them ?
- Should we also add this size-check for the marshalling functions ?

AndreasFuchsSIT:
I was thinking about code-size. Since SAPI is good for microcontrollers with limited RAM, this might be interesting...

tstruk:
To be spec compliant we still need to keep most of the checks and return the response codes as described by the spec in the behavior section of the (un)marshaling for the different TPM2B types.

PR #1172, opened 2018-10-22
test: fix tpm2b unit tests failing on big endian

tstruk:
tpm2b unit test fails on big endian machines

### DIFF
--- a/test/unit/TPM2B-marshal.c
+++ b/test/unit/TPM2B-marshal.c
@@ -210,17 +210,17 @@ tpm2b_unmarshal_success(void **state)
     assert_int_equal (rc, TSS2_RC_SUCCESS);
     assert_int_equal (dgst.size, 4);
     ptr = (uint32_t *)dgst.buffer;
-    assert_int_equal (*ptr, value);
+    assert_int_equal (le32toh(*ptr), value);
     assert_int_equal (offset, 6);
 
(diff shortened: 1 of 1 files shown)


test/unit/TPM2B-marshal.c
```diff
@@ -210,17 +210,17 @@ tpm2b_unmarshal_success(void **state)
     assert_int_equal (rc, TSS2_RC_SUCCESS);
     assert_int_equal (dgst.size, 4);
     ptr = (uint32_t *)dgst.buffer;
-    assert_int_equal (*ptr, value);
+    assert_int_equal (le32toh(*ptr), value);
```
AndreasFuchsSIT: I was irritated: this made no sense, since the digest buffer shall be the same endianess as the original input buffer...

The actual problem however is that the input-byte-buffer in https://github.com/tpm2-software/tpm2-tss/blob/d8d9b90f4f29313f6807c63e208b1a140f371c9c/test/unit/TPM2B-marshal.c#L198 is little-endian encoded.
Thus, we should change the original buffer to be-encoding and use be32toh here instead.

tstruk: @AndreasFuchsSIT yes, we can change this to assert_mem_equal(), but this is not relevant to the big endian issue. Since the buffers are 4 and 8 bytes only then ether way should work.

AndreasFuchsSIT:
Ok, I'll merge it then for fixing the bug, but open an new issue about the upcasting...

PR #1173, opened 2018-10-23
switch default ESAPI crypto backend to openSSL

tstruk:
Switch the default ESAPI crypto backend to openSSL.
This should also fix the problem with forcing applications to
link against libgcrypt.


AndreasFuchsSIT:
We should also add Requires.private and Libs.private to all .pc-files I guess...

tstruk:
I will create a new issue for that.

tstruk:
Issue created https://github.com/tpm2-software/tpm2-tss/issues/1174

jacmet:
Please consider this commit for the 2.1.x / 2.2.x branches as well. The hardcoded -lgcrypt makes building with only openssl impossible

PR #1176, opened 2018-10-30
Partial reads

tstruk:
This adds partial reads support for TCTI-device

### DIFF
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -17,7 +17,7 @@
 
 #include "tss2_tcti.h"
 #include "tss2_tcti_device.h"
-
+#include "tss2_mu.h"
 #include "tcti-common.h"
 #include "tcti-device.h"
 #include "util/io.h"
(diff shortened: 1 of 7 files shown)


src/tss2-tcti/tcti-device.c
```diff
@@ -144,19 +149,84 @@ tcti_device_receive (
...
+        LOG_DEBUG("Partial read - reading response size");
+        fds.fd = tcti_dev->fd;
+        fds.events = POLLIN;
+
+        rc_poll = poll(&fds, nfds, timeout);
+        if (rc_poll < 0) {
+            LOG_ERROR ("Failed to poll for response from fd %d, got errno %d: %s",
+                       tcti_dev->fd, errno, strerror (errno));
+        }
+        if (rc_poll < 0) {
```
AndreasFuchsSIT: This looks like duplicate code...

tstruk: Yes, you are right. I'll remove the duplicate.

tstruk:
@AndreasFuchsSIT in order to test it change this patch is required on the driver side:
https://patchwork.kernel.org/patch/10689297/
Looks like it will be included in the 4.21 kernel.

AndreasFuchsSIT:
@tstruk Then I propose to wait for 4.21 so we know for sure we're compatible with that kernel release...
Ok ?

tstruk:
I'll merge it after the driver patch is merged into the kernel tree.
We don't need to check until 4.21 is out. This is disabled by default anyway and should be enabled by people who will package it.

src/tss2-tcti/tcti-device.c
tstruk: Duplicate code removed.

Issue #1177, opened 2018-10-31
use gnulib for linker version script


flihp:
just ran across this: https://github.com/tpm2-software/tpm2-tss/pull/647

Any reason we're not adding a dependency on gnulib (the upstream for this script) instead of copying it? The license allows this but mirroring this script in this repo seems like a great way for us to get bug reports that we end up having to debug bugs that have already been fixed upstream.

PR #1180, opened 2018-11-02
m4: remove carried macros V7

williamcroberts:
* update bootstrap to allow options to be passed to autoreconf
* update CI script to pass gnulib ld script location to autoreconf via -I and rm ld script
* update CI script to install autoconf-archive in a default search path for autoconf.

**v4**
* Add gnulib package to .travis.yml

**v5**
* add gnulib dependency and bootstrap options to INSTALL.md

**v6**
* Rebase on top of master

**v7**
  * Cherry pick https://github.com/tpm2-software/tpm2-tss/pull/1199/commits/9e747ea603ff41a0e4fd553885b62740d63661e8 - cp the newer doxygen m4 macro

**Travis**
Flunks the patch. Not sure why it's getting a missing separator error in the makefile, I cannot repro locally.

Fixes: #1177


tstruk:
Also there are some local modifications on m4/ax_prog_doxygen.m4 because it caused some issues for make distcheck See: https://github.com/tpm2-software/tpm2-tss/pull/1119
I will need to check if the latest version form autoconf-archive works fine

.travis.yml
```diff
@@ -39,7 +40,7 @@ addons:
       name: "01org/TPM2.0-TSS"
       description: Build submitted via Travis-CI
     notification_email: philip.b.tricca@intel.com
-    build_command_prepend: "./bootstrap && ./configure"
+    build_command_prepend: "./bootstrap -I $(GNULIB_M4) && ./configure"
```
tstruk: Should be -I {GNULIB_M4}

.travis.yml
```diff
@@ -87,7 +88,7 @@ install:
   - make -C ibmtpm/src -j$(nproc)
 
 before_script:
-  - ./bootstrap
+  - ./bootstrap -I $(GNULIB_M4)
```
tstruk: ditto

diabonas:
@williamcroberts @tstruk This seems indeed to be a problem with m4/ax_prog_doxygen.m4: the "missing separator" occurs while trying to run make distcheck, apparently because the version of ax_prog_doxygen.m4 included in the autoconf-archive package used by Travis is broken. Using a newer version instead fixes the CI build, see #1199.

williamcroberts:
@tstruk I got it through configure but now doxygen is having issues:
```
/home/travis/build/tpm2-software/tpm2-tss/README.md:81: warning: Illegal command ref as part of a <h1> tag
  CC       src/tss2-mu/tpms-types.lo
/bin/bash: doc/doxygen/tpm2-tss.tag: No such file or directory
make: *** [doc/doxygen/tpm2-tss.tag] Error 1
make: *** Waiting for unfinished jobs....
The command "if [ "$CC" == "clang" ]; then
  scan-build --status-bugs make -j$(nproc) check
```

Not sure if it's the side-effect of killing mkdir -p patch.

diabonas:
@williamcroberts Yeah, that is a problem with the current doxygen configuration that was circumvented by the `mkdir` patch in `ax_prog_doxygen.m4`. #1205 should fix the issue, so after it gets merged a rebase of this PR should cause all checks to pass.

diabonas:
@williamcroberts It seems that 7c1346a2490d1fc9c730c04448eeb08059eec3f9 got lost in the rebase, so `ld-version-script.m4` is missing because `gnulib` didn't get installed. With that commit included, the build passes, see #1199.

tstruk:
The v8 has been merged so v7 can be closed

PR #1189, opened 2018-11-05
build: Fix breakage in generated configure script when CFLAGS is set.

flihp:
This fixes a bug introduced in 3980bad87fe18ad9d32914e9d21dba145eba973f.
That patch references the documentation for AC_PROG_CC but it misread
the recommended workaround. The leading colon (aka `:`) in the
documentation is significant.

The `:` is a shell 'builtin' command that is equivalent to invoking the
'true' command. By placing the conditional substitution of the CFLAGS
after this, the result of the substitution will be ignored and the
script won't fail. Without this the contents of the CFLAGS variable set
in the environment will be interpreted as a command and since they're
not commands the configure script will abort.

Signed-off-by: Philip Tricca <flihp@twobit.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,7 +31,7 @@ AC_INIT([tpm2-tss],
         [],
         [https://github.com/tpm2-software/tpm2-tss])
 AC_CONFIG_MACRO_DIR([m4])
-${CFLAGS=""}
+: ${CFLAGS=""}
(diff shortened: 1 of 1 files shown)


williamcroberts:
Interesting and good know. We likely have this same issue in other projects.

PR #1190, opened 2018-11-05
Remove SPDX from LICENSE itself

PeterHuewe:
It seems to confuse some tools if the SPDX appears in the License file - e.g. Github which did not show the LICENSE in the right tab.

Signed-off-by: Peter Huewe <peter.huewe@gmx.de>


flihp:
yeah that's kinda silly

PR #1191, opened 2018-11-07
Dockerfile: Apply best practices

jhachenbergerSIT:
This 4 commits change the Dockerfile to keep it more in-line with the [best practices for Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/). They should not introduce any functional change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 FROM ubuntu:18.04
-RUN apt-get update
-RUN apt-get install -y \
+RUN apt-get update && apt-get install -y \
     autoconf \
     autoconf-archive \
     automake \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:18.04
-RUN apt-get update
-RUN apt-get install -y \
+
+RUN apt update
```
AndreasFuchsSIT: From `man apt`:

```
SCRIPT USAGE AND DIFFERENCES FROM OTHER APT TOOLS
       The apt(8) commandline is designed as an end-user tool and it may change behavior between versions. While it tries not to break backward compatibility this is not guaranteed either if a change seems
       beneficial for interactive use.

       All features of apt(8) are available in dedicated APT tools like apt-get(8) and apt-cache(8) as well.  apt(8) just changes the default value of some options (see apt.conf(5) and specifically the
       Binary scope). So you should prefer using these commands (potentially with some additional options enabled) in your scripts as they keep backward compatibility as much as possible.
```

jhachenbergerSIT: Good point, I removed the commit from the pull request

PR #1184, opened 2018-11-03
configure: add check for doxygen

tstruk:
Fixes: #1179

Signed-off-by: Tadeusz Struk <tadeusz.struk@intel.com>

PR #1199, opened 2018-11-12
m4: remove carried macros V8

diabonas:
This rehash of #1180 fixes the Travis CI build by using `ax_prog_doxygen.m4` from a newer version of autoconf-archive because the one included with Travis seems to be broken: `make distcheck` fails with a "missing separator" error. The same strategy is already used for [`ax_code_coverage.m4`](https://github.com/tpm2-software/tpm2-tss/blob/ad8140ed4a73b6244ebd5dc539aee89d9a67c511/.travis.yml#L63), so this requires no additional effort.


.travis.yml
```diff
@@ -61,6 +63,7 @@ install:
   - sha256sum autoconf-archive-2017.09.28.tar.xz | grep -q 5c9fb5845b38b28982a3ef12836f76b35f46799ef4a2e46b48e2bd3c6182fa01 || travis_terminate 1
   - tar xJf autoconf-archive-2017.09.28.tar.xz
   - cp autoconf-archive-2017.09.28/m4/ax_code_coverage.m4 m4/
+  - cp autoconf-archive-2017.09.28/m4/ax_prog_doxygen.m4 m4/
```
tstruk: This still doesn't solve the doxygen problem. Currently travis doesn't generate anything because doxygen is not listed as a dependency in .Travis.yml. If you add this:

https://github.com/tpm2-software/tpm2-tss/pull/1184/commits/0b46de97ff4497e7a1c0995aa832e3905515d78a

it will try to generate the docs, but it will fail with: 

/bin/sh: doc/doxygen/tpm2-tss.tag: No such file or directory
make[1]: *** [Makefile:13655: doc/doxygen/tpm2-tss.tag] Error 1
make[1]: Leaving directory '/tpm2-tss'
make: *** [Makefile:13253: dist] Error 2

When I compare the new file and the one that we currently have it looks like it is missing the line that we have modified locally:

$ diff autoconf-archive-2017.09.28/m4/ax_prog_doxygen.m4 m4/ax_prog_doxygen.m4
549a550
> 	\$(A""M_V_at)mkdir -p \$(DX_DOCDIR]DX_i[)

in this commit:

https://github.com/tpm2-software/tpm2-tss/commit/bb3fef3dd69916d659a4fc40a74dbf3a21d93d3c

tstruk: @williamcroberts @diabonas could you try to rebase this on top of https://github.com/tpm2-software/tpm2-tss/pull/1184 please

diabonas: @tstruk Ah, I see! Sadly, the rebased version exhibits the error described by you above, so it seems that the local modification to `ax_prog_doxygen.m4`  is necessary for `make distcheck` to work. I will close this pull request as my added commit does not help with the Doxygen problem - instead the solution would be to modify the last commit of @williamcroberts' PR #1180 to leave the modified `ax_prog_doxygen.m4` in the `m4` directory.

diabonas: @tstruk I think the reason that the upstream `ax_prog_doxygen.m4` does not work directly is because the current Doxygen configuration is inconsistent between `Doxyfile.in` and `configure.ac`. I have opened #1205 to fix the configuration. If #1205 gets merged, I can reopen this PR as well because `make distcheck` will then work with the upstream `ax_prog_doxygen.m4`.

.travis.yml
```diff
@@ -35,6 +35,7 @@ addons:
     - lcov
     - libssl-dev
     - uthash-dev
+    - gnulib
```
tstruk: This supposed to be in the previous commit.

tstruk:
Just a minor comment about in which commit the gnulib supposed to be added, but I don't think anybody is ever going to bisect using travis so we should be fine. Thanks

PR #1205, opened 2018-11-14
doxygen: change output directory

diabonas:
Currently the Doxygen output directory is set to two different values by separate files: [In `configure.ac`, it is specified as `doc/doxygen`](https://github.com/tpm2-software/tpm2-tss/blob/b9d1d2b64cccea8f30726642ee063c0c25ebc253/configure.ac#L289), while [in `Doxyfile.in`, it is set simply as `doc`](https://github.com/tpm2-software/tpm2-tss/blob/b9d1d2b64cccea8f30726642ee063c0c25ebc253/Doxyfile.in#L66). Doxygen honours the latter declaration, so the only thing that ends up in `doc/doxygen` is [a timestamp file (`tpm2-tss.tag`) autogenerated by `ax_prog_doxygen.m4`](https://github.com/tpm2-software/tpm2-tss/blob/b9d1d2b64cccea8f30726642ee063c0c25ebc253/m4/ax_prog_doxygen.m4#L552).

This constellation seems a little off and causes problems with the autoconf-archive file `ax_prog_doxygen.m4`, which [needed to be modified](https://github.com/tpm2-software/tpm2-tss/pull/1119) to support this configuration.

This pull request uses a unified directory `doxygen-doc` as the Doxygen output directory. The reason for using a new top-level directory instead of one of the already existing `doc` or `doc/doxygen` is as follows:

- We cannot use `doc` because [`ax_prog_doxygen.m4` removes the whole Doxygen directory before building the documentation](https://github.com/tpm2-software/tpm2-tss/blob/b9d1d2b64cccea8f30726642ee063c0c25ebc253/m4/ax_prog_doxygen.m4#L549). As there are other files residing in `doc`, we do not want this.
- `doc/doxygen` would seem to be the most natural choice, but [Doxygen does not create nested directories if they do not exist](https://www.stack.nl/~dimitri/doxygen/manual/starting.html#step2):
    > If the output directory does not exist, doxygen will try to create it for you (but it will *not* try to create a whole path recursively, like `mkdir -p` does).

    This would cause issues with `make distcheck` when the `doc` directory does not yet exist.

Therefore I opted for a new top-level folder `doxygen-doc` (the default name chosen by `ax_prog_doxygen.m4`). I also reverted `ax_prog_doxygen.m4` to the current upstream version, as the [custom `mkdir`](https://github.com/tpm2-software/tpm2-tss/blob/b9d1d2b64cccea8f30726642ee063c0c25ebc253/m4/ax_prog_doxygen.m4#L550) added by #1119 is not necessary any more with this PR.

Lastly, I fixed another little inconsistency between `Doxygen.in` and `configure.ac`, where RTF output was enabled in the former, but disabled in the latter.


tstruk:
I will merge it in to unlock @williamcroberts , but it looks like this still doesn't fix the distcheck:
https://travis-ci.org/tstruk/tpm2-tss/jobs/456103495#L6790

tstruk:
or we can build and install a more recent version of doxygen from .travis.yml similarly to what we do with openSSL. I would prefer to solve it that way instead of adding workarounds into Makefiles. If you could prepare a PR for it that would be great. Also include a separate patch that removes the workaround from travis.yml. Thanks:
    --- a/.travis.yml
    +++ b/.travis.yml
    @@ -127,7 +127,7 @@ script:
     # make distcheck
       - mkdir ./distcheck_build
       - pushd ./distcheck_build
    -  - ../configure --disable-doxygen-doc --enable-unit --enable-integration CFLAGS=-I${PWD}/../osslinstall/usr/local/include LDFLAGS=-L${PWD}/../osslinstall/usr/local/lib
    +  - ../configure --enable-unit --enable-integration CFLAGS=-I${PWD}/../osslinstall/usr/local/include LDFLAGS=-L${PWD}/../osslinstall/usr/local/lib
       - make -j$(nproc) distcheck
       - popd

PR #1207, opened 2018-11-15
ESYS: Fix build with LibreSSL

casantos:
RAND_bytes() is declared in rand.h. Also, LibreSSL does not provide OAEP
macros EVP_PKEY_CTX_set0_rsa_oaep_label and EVP_PKEY_CTX_set_rsa_oaep_md
so use them conditionally.

Signed-off-by: Carlos Santos <casantos@datacom.com.br>

### DIFF
--- a/src/tss2-esys/esys_crypto_ossl.c
+++ b/src/tss2-esys/esys_crypto_ossl.c
@@ -10,6 +10,7 @@
 #include <openssl/aes.h>
 #include <openssl/rsa.h>
 #include <openssl/engine.h>
+#include <openssl/rand.h>
 #include <stdio.h>
 
 #include "tss2_esys.h"
@@ -660,6 +661,8 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
                    "Could not set RSA passing.", cleanup);
     }
 
+/* LibreSSL does not provide these OAEP functions */
+#ifdef EVP_PKEY_CTX_set0_rsa_oaep_label
     if (1 != EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, label, strlen(label)+1)) {
         goto_error(r, TSS2_ESYS_RC_GENERAL_FAILURE,
                    "Could not set RSA label.", cleanup);
@@ -669,6 +672,9 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
         goto_error(r, TSS2_ESYS_RC_GENERAL_FAILURE,
                    "Could not set hash algorithm.", cleanup);
     }
+#else
+    (void) label;
+#endif
 
     /* Determine out size */
     if (1 != EVP_PKEY_encrypt(ctx, NULL, out_size, in_buffer, in_size)) {


src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -660,6 +661,8 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
                    "Could not set RSA passing.", cleanup);
     }
 
+/* LibreSSL does not provide these OAEP functions */
+#ifdef EVP_PKEY_CTX_set0_rsa_oaep_label
```
tstruk: Should we check if both EVP_PKEY_CTX_set_rsa_oaep_md and EVP_PKEY_CTX_set0_rsa_oaep_label are defined instead of wrapping the two calls in one check? Something like:
#if defined(EVP_PKEY_CTX_set_rsa_oaep_md) && defined(EVP_PKEY_CTX_set0_rsa_oaep_label)

JuergenReppSIT: Also together with (void)label;
 goto_error(r, TSS2_ESYS_RC_NOT_IMPLEMENTED,
                   "RSA OAEP not implemented.", cleanup);
 should be added, because the secret encryption in StartAuthSession will not be possible. The integration tests with secret encryption will fail. But instead of "0x2c4 TPM_RC_VALUE" the message "RSA OAEP not implemented" would be  displayed in the log of the test.

casantos: OK, I will update the change to reflect these comments. Thanks.

casantos:
Libressl is a fork of OpenSSL as of 1.0.1g. Those function were introduced in OpenSSL version 1.1.0, so they don't exist on LibreSSL.

AndreasFuchsSIT:
Hmmm.... Ok, we're only testing against OpenSSL 1.0.2stable and 1.1.0stable branch...

How was this functionality achieved in 1.0.1g then ?

AndreasFuchsSIT:
@casantos I dug into the code a little and found the following:

LibreSSL does not support OAEP-mode with labels at all, even though the internal OAEP-padding-function includes the parameters already: https://github.com/libressl/libressl/blob/ba04546060b835e279c0e8a26261e73a1964927c/src/crypto/rsa/rsa_eay.c#L197
(see the last 2 parameters, that represent the label)

Further, the internal OAEP-padding-function does not support variable hash algs, but staticly uses SHA1, see OpenSSL in comparison: https://github.com/openssl/openssl/blob/32f803d88ec3df7f95dfbf840c271f7438ce3357/crypto/rsa/rsa_pmeth.c#L302 (the last 2 parameters are the digest alg for the label)

Are you involved with the LibreSSL community ? Could LibreSSL add support for OAEP-Labels ? Then we'd happily support LibreSSL as well...

casantos:
> Are you involved with the LibreSSL community ? Could LibreSSL add support for OAEP-Labels ? Then we'd happily support LibreSSL as well...

I'm, not involved with the LibreSSL community. I'm working on porting tpm2-tss to Buildroot.

PR #1208, opened 2018-11-16
 travis: update doxygen to version 1.8.14 

diabonas:


.travis.yml
```diff
@@ -119,19 +119,13 @@ script:
...
       ../configure --enable-unit --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS CFLAGS=-I${PWD}/../osslinstall/usr/local/include LDFLAGS=-L${PWD}/../osslinstall/usr/local/lib
     fi
+  - make -j$(nproc) distcheck
   - |
     if [ "$CC" == "clang" ]; then
       scan-build --status-bugs make -j$(nproc) check
     else
       make -j$(nproc) check
     fi
-  - |
```
tstruk: Well, we still want to do this as a separate step in a new subdir, just remove the --disable-doxygen-doc from configure, please

.travis.yml
```diff
@@ -66,6 +65,10 @@ install:
   - pip install --user cpp-coveralls
 # IBM-TPM
   - wget https://download.01.org/tpm2/ibmtpm974.tar.gz
+# doxygen
+  - wget http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.14.linux.bin.tar.gz
```
tstruk: The comment says "Compiled using Ubuntu 14.04 with kernel 3.13.0 and gcc 4.8.2" that sould be ok for trusty, but we need to remember to update this as well when we update the travis dist. Could you at least add a comment here please.

diabonas: According to the CI log, the precompiled binary does not work for some reason, so I will build doxygen from source instead.

.travis.yml
diabonas: Will do!

diabonas:
@tstruk This PR should be fit to merge now, [`make distcheck` works with doxygen](https://travis-ci.org/tpm2-software/tpm2-tss/jobs/456194121#L6931) and all tests have passed successfully.

PR #1209, opened 2018-11-17
configure: fix check for doxygen

### LINKED ISSUE #1204: --disable-doxygen-doc still requires doxygen
We now check for the existence of doxygen when we try to build the documentation and fail if it's not available (a very good thing). Unfortunately this check still happens even if the caller supplies the `--disable-doxygen-doc` option to the configure script. I would expect that if the caller explicitly disables the doxygen docs (possibly because they can't install it) that we shouldn't fail the configure script if doxygen isn't installed.

diabonas:
Fix #1204: The doxygen feature is controlled by the variable <code>$*enable*_doxygen_doc</code>, which is why the current check does not work.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -297,7 +297,7 @@ DX_PS_FEATURE(OFF)
 DX_INIT_DOXYGEN($PACKAGE_NAME, [Doxyfile], [doxygen-doc])
 AM_CONDITIONAL(DOXYMAN, [test $DX_FLAG_man -eq 1])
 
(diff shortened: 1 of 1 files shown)


tstruk:
eval prints the result and clutters the output from configure. I think the 
[test "x$enable-doxygen-doc" != xno]
option worked fine.

diabonas:
@tstruk Ah right, I didn't see that. OK, rolling back to the previous version of the PR..

PR #1211, opened 2018-11-20
TPM Types: Fix tss2_tpm_types.h

JuergenReppSIT:
* Missing type casts are added as defined in "TCG TSS 2.0 Overview and Common Structures Specification".
*  Wrong type name _ID_OBJECT is fixed.
* The type The type TPM2_NT is added.


AndreasFuchsSIT:
The problem is  that currently a comparison of a TPMI_DH_CONTEXT variable against TPM2_TRANSIENT_FIRST will result in a compiler warning about signedness. So the current code cannot stay as is.

There are 2 solutions, append a bunch of U and UL declarations to the defines or do the type casts.
IMHO the latter is nicer because it will give warning if the defines are errorous and it is the same as the spec.

PR #1226, opened 2018-12-13
MU: Fix marshaling of TPM2Bs with sub types.

JuergenReppSIT:
* The offset was not computed correctly if a NULL buffer was passed.
  The TSS MU spec states: If the 'buffer' parameter is NULL the
  implementation shall not write any marshaled data but the 'offset'
  parameter shall be updated as as though it had.
* The unnecessary test of NULL buffer and not NULL offset was removed
  and the corresponding unit test was adapted.
* Also a unit test which compares the result of NULL buffer and
  not NULL buffer marshaling was added.


tstruk:
Looks like something is wrong with doxygen in travis, which causes the build to fail.

PR #1229, opened 2018-12-14
travis: update to Xenial VM

diabonas:
Travis now provides a Xenial image, see https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures and https://github.com/tpm2-software/tpm2-abrmd/pull/561/. This allows us to use the packaged versions of cmocka and Doxygen instead of building them from source, which should make the CI a bit more robust.

The updated version of Clang in the image reports some additional bugs with `scan-build`, which need to be addressed so that the build passes.

Somewhat unrelated, I also removed an invalid `CLEANFILES` directive in the Makefile related to Doxygen. This does not lead to errors, but adds noise by generating an extra warning
```
rm: cannot remove 'doxygen-doc': Is a directory
make: [Makefile:14043: clean-generic] Error 1 (ignored)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ script:
 # build with no tests enabled
   - mkdir ./build-no-tests
   - pushd ./build-no-tests
-  - ../configure --disable-doxygen-doc CFLAGS=-I${PWD}/../osslinstall/usr/local/include LDFLAGS=-L${PWD}/../osslinstall/usr/local/lib
+  - ../configure CFLAGS=-I${PWD}/../osslinstall/usr/local/include LDFLAGS=-L${PWD}/../osslinstall/usr/local/lib
   - make -j$(nproc)
   - popd
 # build with all tests enabled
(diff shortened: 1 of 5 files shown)


tstruk:
Ok, let's disable doxygen first to make the build passing using the old image. Then we can fix the warnings, switch to the new image and enable doxygen again.

tstruk:
@diabonas do you want to fix the warnings in this PR?

diabonas:
@tstruk I will have a look later today, unless someone is faster than me. There are essentially two problems: one is in [`test/integration/esys-create-session-auth.int.c`](https://github.com/tpm2-software/tpm2-tss/blob/6996bc1fd2c9f40ac7f1fbdecc00f8f313974eb1/test/integration/esys-create-session-auth.int.c#L220), where the analyser complains that `sessionAttributes` is uninitialised because it doesn't see the `memset`. This one should be easy to fix.

The other one is more problematic: there is a "division by zero" warning when using `HASH_ADD_INT`, which is not under our control because it comes from [uthash](https://github.com/troydhanson/uthash/blob/49428699aa376f6668d2aedcfa9929480132b66e/src/uthash.h#L497). Currently I am not sure what to do about this one, except [excluding this code from analysis altogether](https://clang-analyzer.llvm.org/faq.html#exclude_code). What do you think?

diabonas:
The build failure doesn't seem to be my fault: apparently AppVeyor simply failed to start [one of the jobs](https://ci.appveyor.com/project/tstruk/tpm2-tss/builds/21008620/job/1k5qbakn8tcbm7n9), its log is completely empty. The other jobs and the Travis build ran successfully.

tstruk:
@diabonas since these warnings are in integration tests only and it looks they are false positives in uthash I think that should be fine. Thanks for fixing these. BTW any idea why clang-check doesn't see memset?

diabonas:
@tstruk No idea, I cannot even reproduce that warning on my own machine with the same version of Clang, so guess it is a really weird Clang bug? For the record, the warning originally occurred in the first version of this PR for [Travis job #1777.7, l. 8075](https://travis-ci.org/tpm2-software/tpm2-tss/jobs/467803879#L8075).

PR #1237, opened 2019-01-04
Fix number of NULL ptr issues in ESYS

tstruk:
esys: Check for NULL auth in iesys_compute_session_value
esys: Fix NULL ptr auth in Esys_TR_SetAuth
esys: handle NULL sequenceHandleNode
test: esys: Add test for Esys_TR_SetAuth with NULL auth

### DIFF
--- a/src/tss2-esys/esys_tr.c
+++ b/src/tss2-esys/esys_tr.c
@@ -356,7 +356,8 @@ Esys_TR_Close(ESYS_CONTEXT * esys_context, ESYS_TR * object)
  * every Esys_TR_Deserialize.
  * @param esys_context [in,out] The ESYS_CONTEXT.
  * @param esys_handle [in,out] The ESYS_TR for which to set the auth value.
- * @param authValue [in] The auth value to set for the ESYS_TR.
+ * @param authValue [in] The auth value to set for the ESYS_TR or NULL to zero
+ *        the auth.
  * @retval TSS2_RC_SUCCESS on Success.
  * @retval TSS2_ESYS_RC_BAD_REFERENCE if the esysContext is NULL.
  * @retval TSS2_ESYS_RC_BAD_TR if the ESYS_TR object is unknown to the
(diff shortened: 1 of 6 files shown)


src/tss2-esys/api/Esys_HMAC_Start.c
```diff
@@ -25,12 +25,12 @@ static void store_input_parameters (
     esysContext->in.HMAC_Start.handle = handle;
     esysContext->in.HMAC_Start.hashAlg = hashAlg;
     if (auth == NULL) {
-        esysContext->in.HMAC_Start.auth = NULL;
```
williamcroberts: Commit message: sed -i s/esey/esys

I wish github allowed comments on the commit message itself.

src/tss2-esys/esys_tr.c
```diff
@@ -369,6 +369,7 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
     RSRC_NODE_T *esys_object;
     TSS2_RC r;
     _ESYS_ASSERT_NON_NULL(esys_context);
+    _ESYS_ASSERT_NON_NULL(authValue);
```
williamcroberts: I'm not 100% sure this is the right fix. IIRC, authValue NULL implies 0 length auth buffer.

williamcroberts: Commit message body lacking justification

tstruk: Right, just checked the spec and if auth is NULL the auth len should be set to zero. Patch updated.

src/tss2-esys/esys_tr.c
```diff
@@ -369,10 +369,16 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
     RSRC_NODE_T *esys_object;
     TSS2_RC r;
     _ESYS_ASSERT_NON_NULL(esys_context);
+
```
williamcroberts: whitespace change: Newline added.....

PR #1238, opened 2019-01-07
build: update to ax_code_coverage.m4 version 2019.01.06

diabonas:
When using this project in combination with the freshly released Autoconf Archive 2019.01.06, the build fails with a "missing separator" error in the makefile. This is because upstream made a [breaking change](http://git.savannah.nongnu.org/cgit/autoconf-archive.git/commit/m4/ax_code_coverage.m4?id=b5a2d69f7a7e0133733673586231b88464a98d58) by removing `@CODE_COVERAGE_RULES@` and replacing it with `include $(top_srcdir)/aminclude_static.am`, see the [documentation](https://www.gnu.org/software/autoconf-archive/ax_code_coverage.html) for reference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 AUTHORS
 tags
 aclocal.m4
+aminclude_static.am
 autom4te.cache/
 [Bb]uild/
 [Dd]ebug/
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,11 +47,11 @@ addons:
 
 install:
 # Autoconf archive
-  - wget https://download.01.org/tpm2/autoconf-archive-2017.09.28.tar.xz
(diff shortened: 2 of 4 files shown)


AndreasFuchsSIT:
Is there a way to change the Makefile.am such that it's compatible with the old and the new version of ax_code_coverage ? Like an ifdef or something ?

I'd even be ok with setting this from configure.ac

Rational is that even though we want to support the new autotools for when distros switch, it's a pitty if we don't support 1 year old distros and LTSes anymore without downloading this.

diabonas:
Somewhat ugly, but we can add a test for the existence of the macro `_AX_CODE_COVERAGE_RULES` which was newly introduced in version 2019.01.06, see the updated PR. I successfully tested this with version 2018.03.13 and 2019.01.06.

AndreasFuchsSIT:
I like this.
@tstruk what do you think ?

joshuagl:
FWIW, I like this solution too.

PR #1243, opened 2019-01-09
build: add doxygen.dox to distribution

diabonas:
I just tried building 2.2.0-rc0 with a source tarball generated by `make distcheck`. The build process itself works fine, but `make install` fails with the following errors:
```
/usr/bin/install: cannot stat './doxygen-doc/man/ESYS_TR.3': No such file or directory
/usr/bin/install: cannot stat './doxygen-doc/man/ESYS_TR_defines.3': No such file or directory
...
/usr/bin/install: cannot stat './doxygen-doc/man/Esys_HierarchyControl.3': No such file or directory
make[1]: *** [Makefile:13304: install-man3] Error 1
```
The problem is that `doxygen-doc`, the rule from [`ax_prog_doxygen.m4`](https://www.gnu.org/software/autoconf-archive/ax_prog_doxygen.html) that is [used to generate the documentation](https://github.com/tpm2-software/tpm2-tss/blob/acd54f261ad963bc682bbb7d5cd491cae5f093c7/Makefile.am#L140), completely [removes the man pages directory](https://github.com/autoconf-archive/autoconf-archive/blob/1999013d84b42db111fce9182296cc6452af1497/m4/ax_prog_doxygen.m4#L549) before generating the documentation. This wouldn't be a problem if the man pages were successfully regenerated by doxygen afterwards, but the release tarball is missing the necessary [`doxygen.dox`](https://github.com/tpm2-software/tpm2-tss/blob/master/doc/doxygen.dox) file to do this.

The solution is simply to add `doxygen.dox` to the distribution. Unfortunately there does not seem to be a way to stop the regeneration process, so including the autogenerated man pages in the distribution is useless, which is why I removed them. Furthermore, manually setting `DX_CLEANFILES` is not necessary as this is already done by `@DX_RULES@`.

The reason why this issue first occurs during the current release is PR #1205: it changed the doxygen directory for `ax_prog_doxygen` to match the directory specified in the `Doxyfile`. While this itself is correct, it means that now the actual doxygen directory is deleted when generating the documentation, previously it was just an empty "dummy directory" that was getting deleted.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,7 +23,6 @@ noinst_PROGRAMS =
 
 # ax_doxygen
 @DX_RULES@
-DX_CLEANFILES = doxygen-doc
 MOSTLYCLEANFILES += $(DX_CLEANFILES)
 
 if DOXYMAN
@@ -138,7 +137,6 @@ DOXYMAN3 = \
     doxygen-doc/man/Esys_VerifySignature.3 \
     doxygen-doc/man/Esys_ZGen_2Phase.3
 $(DOXYMAN3): doxygen-doc
-EXTRA_DIST += $(DOXYMAN3)
 else #DOXYMAN
 DOXYMAN3 =
 endif #DOXYMAN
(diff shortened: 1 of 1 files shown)


tstruk:
@diabonas yes, I will add this to 2.2.0 and create a new rc soon

PR #1246, opened 2019-01-12
INSTALL: Add missing tests dependencies

ArchangeGabriel:
Had to add them when building for Arch Linux.


ArchangeGabriel:
I see that net-tools was actually recently replaced with ss, feel free to edit my commit and remove that part.

AndreasFuchsSIT:
Merged #1247

PR #1254, opened 2019-01-24
Add ld-version-script.m4 back

joshuagl:
From the Gnulib website:

> Gnulib takes a different approach. Its components are intended to be shared at the source level, rather than being a library that gets built, installed, and linked against. Thus, there is no distribution tarball; the idea is to copy files from Gnulib into your own source tree.

and thus several distributions (Arch Linux, Clear Linux, openSUSE, etc) don't package it.
Therefore add back `ld-version-script.m4` and update collateral to reflect this.


AndreasFuchsSIT:
+1 from me...
@tstruk Add to 2.2 ?

PR #1256, opened 2019-01-25
build: fix build on systems with autoconf-archive < 2019.01.06

### LINKED ISSUE #1255: Build broken on systems with autoconf-archive < 2019.01.06
Since #1238 landed I'm seeing:

> $ ./bootstrap -I /usr/share/gnulib/m4/
> Generating file lists: src_vars.mk
> automake: error: cannot open < aminclude_static.am: No such file or directory
> autoreconf: automake failed with exit status: 1

joshuagl:
Since #1238 landed I'm seeing:

    $ ./bootstrap -I /usr/share/gnulib/m4/
    automake: error: cannot open < aminclude_static.am: No such file or directory
    autoreconf: automake failed with exit status: 1


diabonas:
Huh, weird, I didn't  get this error on my system when testing with version 2018.03.13 - but maybe the system-wide installation 2019.01.06 interfered with my locally installed older version? Sorry about that...

The problem with `-include` seems to be that the file is not parsed by Automake any more, so the macros are not expanded properly, causing failures on systems with version 2019.01.06. I guess the easiest solution is to make sure that `aminclude_static.am` exists by using
```
AM_COND_IF([AUTOCONF_CODE_COVERAGE_2019_01_06],[],[AX_ADD_AM_MACRO_STATIC([])])
```
in `configure.ac`, however I am unable to properly test this on my own machine.

joshuagl:
Thanks for bringing some automake knowledge to the issue!

> AM_COND_IF([AUTOCONF_CODE_COVERAGE_2019_01_06],[],[AX_ADD_AM_MACRO_STATIC([])])
> ```

It seems that `AM_COND_IF` doesn't like the if-true branch to be empty, but my local testing suggest we can just call `AX_ADD_AM_MACRO_STATIC([])` unconditionally.

Let's see what the Travis matrix makes of it.

diabonas:
OK, that should be fine as well as multiple calls to `AX_ADD_AM_MACRO_STATIC` just append to the file.

flihp:
Same issue broke tpm2-abrmd autobuilder and my local stuff too. Patches here work for me.

PR #1261, opened 2019-01-30
sys: add mechanism to disable weak crypto

tstruk:
Core Infrastructure best practices require usage only strong
cryptographic algorithms. To comply with this requirement we
need to have a way to disable usage of algorithms that are
considered weak. The diable mechanism will validate
the TPM2_PUBLIC template of an object and will not allow
to create or load object which the public template will
contain references to the not allowed algorithms.
Additionally it will validate the nameAlg of inPublic and
NV_PUBLIC, hashAlg of the different signing schemes, and
hashAlg of PCR selection for relevant commands.

Algorithms that will be disabled by this are as follows:
 - RSA with key size < 2048 bits
 - Symmetric ciphers with key size < 128 bits
 - SHA1

The new configure flag is called --disable-weakcrypto

### DIFF
--- a/src/tss2-sys/api/Tss2_Sys_Create.c
+++ b/src/tss2-sys/api/Tss2_Sys_Create.c
@@ -22,6 +22,10 @@ TSS2_RC Tss2_Sys_Create_Prepare(
     if (!ctx || !creationPCR)
         return TSS2_SYS_RC_BAD_REFERENCE;
 
+    rval = ValidateTPML_PCR_SELECTION(creationPCR);
+    if (rval)
+        return rval;
+
     rval = CommonPreparePrologue(ctx, TPM2_CC_Create);
     if (rval)
         return rval;
@@ -55,6 +59,11 @@ TSS2_RC Tss2_Sys_Create_Prepare(
                                       &ctx->nextData);
 
     } else {
+
+        rval = ValidatePublicTemplate(inPublic);
+
+        if (rval)
+            return rval;
 
         rval = Tss2_MU_TPM2B_PUBLIC_Marshal(inPublic, ctx->cmdBuffer,
                                             ctx->maxCmdSize,
--- a/src/tss2-sys/sysapi_util.c
+++ b/src/tss2-sys/sysapi_util.c
@@ -6,11 +6,12 @@
  ***********************************************************************/
 
 #include <string.h>
-#include <stdbool.h>
 
 #include "tss2_mu.h"
 #include "sysapi_util.h"
 #include "util/tss2_endian.h"
+#define LOGMODULE sys
+#include "util/log.h"
 
 void InitSysContextFields(_TSS2_SYS_CONTEXT_BLOB *ctx)
 {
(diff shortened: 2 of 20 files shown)


tstruk:
@AndreasFuchsSIT Thanks for pointing out the other commands and structs that need to be checked. PR updated and all should be covered now.

src/tss2-sys/api/Tss2_Sys_Create.c
```diff
@@ -56,6 +60,11 @@ TSS2_RC Tss2_Sys_Create_Prepare(
 
     } else {
 
+        rval = ValidatePublicTemplate(inPublic);
+
```
williamcroberts: extra space here and in other locations.

Does this actually comply with TCG?

AndreasFuchsSIT:
@williamcroberts Nope, it breaks TCG, but it assembles compliance with NIST and thus CII.

I'll look at it next week, I guess we have time since the next release is far out...

DomiNic5787:
On the one hand I think it is a good idea to have this feature. But on the other hand, as @AndreasFuchsSIT mentioned, this would break TCG compliance.
So for me it is ok, but only as an optional feature (NEVER default!).

tstruk:
@DomiNic5787 I agree, the default is not to disable anything. It only disables the algorithms when one explicitly passes --disable-weakcrypto to the configure script.

williamcroberts:
The problem is, some distro will flip this on and everyone will be like WTF why is my client application failing and can I work around it simply. I would make this an ENV variable... getenv() is stdlib and then if it's set in the environment disable weak crypto. Thus folks can turn it on when they want/need it.

AndreasFuchsSIT:
I understand the concern. We might want to add a HUGE WARNING to configure for this.

I dislike having too many ENVs to begin with.

But I also wonder if an ENV would fulfill the requirements we're trying to solve; see https://bestpractices.coreinfrastructure.org/de/projects/2332#security
To, "configure" means `./configure` intuitively, but that is probably arguable.

williamcroberts:
Perhaps use configure to set the default (ie allow or disallow weak crypto) and an env variable to override. You'll want to provide a run-time way to have apps run in a portable fashion, even when apps do something stupid.

tstruk:
@williamcroberts Once this is disabled at configure time there will be no way to enable it back in runtion with env variable. We could possibly disable it based on env variable only if it is enabled at configure/build time, but in this case it will defeat the purpose of being compliant with Core Infrastructure best practices.

AndreasFuchsSIT:
As mentioned, I don't want to have more env stuff.
I see the use case in distribution policies; i.e. in 5 years all dists will just want to be able to flip the switch and have all application "secured". Or gov may want to do the same thing. And this is just the preparation for this day X.
Thus, having override ENV for this would defeat the purpose of the use case.

@williamcroberts Would you be very unhappy with the merger ? Otherwise, I'd hit the button.

src/tss2-sys/sysapi_util.c
```diff
@@ -304,3 +303,111 @@ int GetNumResponseHandles(TPM2_CC commandCode)
 {
     return GetNumHandles(commandCode, 0);
 }
+
+#ifdef DISABLE_WEAK_CRYPTO
+bool IsAlgorithWeak(TPM2_ALG_ID algorithm, TPM2_KEY_SIZE key_size)
```
webmeister: Shouldn't this be called `IsAlgorithmWeak`?

src/tss2-sys/sysapi_util.c
```diff
@@ -304,3 +303,111 @@ int GetNumResponseHandles(TPM2_CC commandCode)
...
+{
+    switch (algorithm) {
+        case TPM2_ALG_RSA:
+            if (key_size <= 2048)
+                return true;
+        break;
+        case TPM2_ALG_AES:
+            if (key_size <= 128)
```
webmeister: Why is this considered weak? The smallest possible key size for AES is 128 bits, and that is generally still considered secure, isn't it? There are no known attacks that significantly reduce the strength of AES (certainly not below the NIST limit of 112 bits).

src/tss2-sys/sysapi_util.c
```diff
@@ -304,3 +303,111 @@ int GetNumResponseHandles(TPM2_CC commandCode)
 {
     return GetNumHandles(commandCode, 0);
 }
+
+#ifdef DISABLE_WEAK_CRYPTO
+bool IsAlgorithWeak(TPM2_ALG_ID algorithm, TPM2_KEY_SIZE key_size)
+{
+    switch (algorithm) {
+        case TPM2_ALG_RSA:
+            if (key_size <= 2048)
```
webmeister: Shouldn't this be `key_size < 2048`? RSA-2048 is fine according to NIST to which CII refers.

src/tss2-sys/sysapi_util.c
tstruk: Yes, symmetric key of 128 provides effective security strength of 128 bits and is safe to use.
This should only disable anything that is 112 and smaller. This will make most of the ciphers (AES, SM4 and Camellia) usable with all supported key sizes (128, 192, and 256), so this check will probably be redundant, but the the spec also defines a generic TPM2_ALG_SYMCIPHER ALG_ID and the algorithm registry [1] also calls out Triple DES I have updated the check to cover all ciphers and bail out on key < 128. Thanks for pointing this out.
[1] https://trustedcomputinggroup.org/wp-content/uploads/TCG-_Algorithm_Registry_Rev_1.27_FinalPublication.pdf

src/tss2-sys/sysapi_util.c
tstruk: Correct, condition updated. Thanks

src/tss2-sys/sysapi_util.c
tstruk: Right, fixed.

tstruk:
> Are sure these are all the important places? Except for the hash commands, you only changed commands dealing with creating objects, not with using them. So if I already have an insecure key within the TPM (maybe because it sat there for the last five years), I can still use it for example to create insecure signatures.
> 
> If you wanted to completely disable an algorithm, wouldn't it be better to deal with that at the marshalling layer, where you could simply reject TPM2_ALG_SHA1 instead of encoding into any command stream, no matter for what it is going to be used?
>

Correct, but it will not be possible to disable using an insecure keys created beforehand as many commands only take key handles (TPMI_DH_OBJECT) to perform operations. This assumes a clean platform with a fresh install of TSS, which will protect the application from creating weak keys and using weak algorithms.
 
> Also, should the TPM's GetCapability responses be modified to exclude algorithms that will be rejected anyway? That would allow applications to detect whether SHA1 is still usable with a TPM.

No, we don't want to mask out the capabilities of the TPM itself. We just want to make sure that they are not used.

webmeister:
> which will protect the application from creating weak keys and using weak algorithms.

I haven't looked into that in much detail, but I'd assume that with the current implementation you can still create a key with no restrictions on which algorithms it may be used with (scheme=ALG_NULL for Create), and then specify a weak algorithm during usage (e.g. SHA1 for Sign).

As you already pointed out, this cannot be detected as easily when the parameters are used the other way around (a key with a weak scheme already exists within the TPM, Sign then just uses ALG_NULL). But you could detect it when looking at the response before returning it to the caller, since it needs to contain the algorithm identifiers somewhere.

>> Also, should the TPM's GetCapability responses be modified to exclude algorithms that will be rejected anyway? That would allow applications to detect whether SHA1 is still usable with a TPM.
>
> No, we don't want to mask out the capabilities of the TPM itself. We just want to make sure that they are not used.

But that makes it harder for applications to correctly deal with the TPM. For example, I've got an application that simply asks the TPM for all hash algorithms that it supports, then tries to hash some data with all those algorithms. The intention is to have something that works with every TPM, which is why there is no fixed list of algorithms. But now this application has to learn that an algorithm can be "supported, but unavailable" when it runs against this specific TSS implementation, but not when it uses any other way to talk to the TPM. And as far as I can see, there is no way for the application to learn that information other than actually trying to use the algorithm.

tstruk:
> I haven't looked into that in much detail, but I'd assume that with the current implementation you can still create a key with no restrictions on which algorithms it may be used with (scheme=ALG_NULL for Create), and then specify a weak algorithm during usage (e.g. SHA1 for Sign).
> 
> As you already pointed out, this cannot be detected as easily when the parameters are used the other way around (a key with a weak scheme already exists within the TPM, Sign then just uses ALG_NULL). But you could detect it when looking at the response before returning it to the caller, since it needs to contain the algorithm identifiers somewhere.

As far as I remember for objects with the sign attribute set, the specification requires a valid signing scheme for the key type. If you think that some checks are missing please provide a specific examples and we can add the checks later.

> > No, we don't want to mask out the capabilities of the TPM itself. We just want to make sure that they are not used.
> 
> But that makes it harder for applications to correctly deal with the TPM. For example, I've got an application that simply asks the TPM for all hash algorithms that it supports, then tries to hash some data with all those algorithms. The intention is to have something that works with every TPM, which is why there is no fixed list of algorithms. But now this application has to learn that an algorithm can be "supported, but unavailable" when it runs against this specific TSS implementation, but not when it uses any other way to talk to the TPM. And as far as I can see, there is no way for the application to learn that information other than actually trying to use the algorithm.

Or the application can simply follow the security guidance and best practices, for instance: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-175b.pdf

tstruk:
@webmeister @williamcroberts thanks for your comments on this PR. I will wait for couple of more days and if there will be no objections I will merge this in. Thanks

PR #1266, opened 2019-02-08
Esys memory leak fixes

williamcroberts:


src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -689,16 +702,13 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
...
 
-    return TSS2_RC_SUCCESS;
+    r = TSS2_RC_SUCCESS;
 
  cleanup:
     OSSL_FREE(ctx, EVP_PKEY_CTX);
-    OSSL_FREE(rsa_key, RSA);
     OSSL_FREE(evp_rsa_key, EVP_PKEY);
+    OSSL_FREE(rsa_key, RSA);
     OSSL_FREE(bne, BN);
```
tstruk: We will leak the BIGNUM *n on openssl >= 1.1.0

williamcroberts: I don't think that statement is correct.

On version 1.1.0g:
in file crypto/rsa/rsa_lib.c RSA_free() --> BN_clear_free(r->n); which free's n.

I checked against versions:
- 1.0.0
- 1.0.1
- 1.0.2
- 1.0.h
all have the same behavior.

Considering that their was a segfault in addition to the memory leaking indicates this error path was never exercised including that conditional block, which was incorrect and will cause an additional segfault when left in place.

tstruk: Right, this memory management in different versions on OpenSSL gets tricky.

PR #1267, opened 2019-02-09
esys: dont orphan rsa->n in iesys_cryptossl_pk_encrypt

williamcroberts:
in OSSL versions less than 1.1.0, rsa is non-opaque and
the code assigning to rsa->n assumed n was not initialized
and thus orphaned the pointer. Rather than duplicate memory,
just use the BN *ret argument to BN_bin2bn() to specify
setting that value.

### DIFF
--- a/src/tss2-esys/esys_crypto_ossl.c
+++ b/src/tss2-esys/esys_crypto_ossl.c
@@ -576,9 +576,7 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
     BIGNUM* bne = NULL;
     int padding;
     char *label_copy = NULL;
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L
-    BIGNUM *n = NULL;
-#endif
+
     if (!(hashAlg = get_ossl_hash_md(pub_tpm_key->publicArea.nameAlg))) {
         LOG_ERROR("Unsupported hash algorithm (%"PRIu16")",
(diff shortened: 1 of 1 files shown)


src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -631,10 +629,14 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
...
-                           NULL);
+                           rsa_key->n)) {
+        goto_error(r, TSS2_ESYS_RC_GENERAL_FAILURE,
+                   "Could not create rsa n.", cleanup);
+    }
 #else
+    BIGNUM *n = NULL;
     if (!(n = BN_bin2bn(pub_tpm_key->publicArea.unique.rsa.buffer,
                         pub_tpm_key->publicArea.unique.rsa.size,
                         NULL))) {
```
williamcroberts: FYI I thought about using RSA_get0_key() to get a pointer to N and then passing that to BN_bin2bn() to help further reduce the delta between OSSL versions, but RSA_get0_key() returns a const pointer, they don't want you doing things like what Initially thought.

tstruk: We could probably use RSA_set_method() and provide our own esys_rsa_keygen() to make it look the same on both ossl < 1.1.0 and ossl >=1.1.0.

williamcroberts: That's kind of how we did it in tools... just added the missing functions

PR #1271, opened 2019-02-15
Update session flags

tstruk:
Update session enc/dec flags depending on what is supported by the command.

### DIFF
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -552,6 +552,85 @@ iesys_gen_caller_nonces(ESYS_CONTEXT * esys_context)
     }
     return TSS2_RC_SUCCESS;
 }
+
+/** Update session attributes.
+ *
+ * In case where command does not support param encryption/decryption
+ * store the original session attributes and update them accordingly.
+ * Return true is command support param encryption.
+ *
+ * @retval TRUE if command support param encryption
+ * @retval FLASE if command does not support param encryption
+ */
+static bool
+iesys_update_session_flags(ESYS_CONTEXT * esys_context,
+                           IESYS_SESSION *rsrc_session)
+{
+    TSS2_RC r = TSS2_RC_SUCCESS;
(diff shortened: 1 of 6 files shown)


src/tss2-esys/esys_iutil.c
```diff
@@ -552,6 +552,85 @@ iesys_gen_caller_nonces(ESYS_CONTEXT * esys_context)
...
 }
+
+/** Update session attributes.
+ *
+ * In case where command does not support param encryption/decryption
+ * store the original session attributes and update them accordingly.
+ * Return true is command support param encryption.
+ *
+ * @retval TRUE if command support param encryption
+ * @retval FLASE if command does not support param encryption
```
williamcroberts: I was questioning this, and my tests show this version failing.

Why did you do this? The purpose was if flag set and command doesn't support it, quietly turn the flag down.

src/tss2-esys/esys_iutil.c
```diff
@@ -552,6 +552,85 @@ iesys_gen_caller_nonces(ESYS_CONTEXT * esys_context)
...
+ */
+static bool
```
williamcroberts: return type: void

src/tss2-esys/esys_iutil.c
```diff
@@ -552,6 +552,85 @@ iesys_gen_caller_nonces(ESYS_CONTEXT * esys_context)
...
+iesys_update_session_flags(ESYS_CONTEXT * esys_context,
+                           IESYS_SESSION *rsrc_session)
+{
+    TSS2_RC r = TSS2_RC_SUCCESS;
+    size_t param_size;
+    const uint8_t *param_buffer;
+    bool encrypt = true;
```
williamcroberts: no need for this AFAICT

tstruk: We do need this. It's no good if we unset the flags, but still encrypt the parameter.

PR #1276, opened 2019-02-19
Make case_insensitive_strncmp more robust

PeterHuewe:
Valgrind complaints that "conditional jump or move depends on uninitialized value(s)” in memcmp of `case_insensitive_strncmp`,
 if `n` is larger than `strlen` of `string1` or `string2`. Although the result of memcmp will be correct, it still accesses unitialized memory in this case (which might even crash). Using strncmp fixes the problem, but operates on (regular) chars.

Moreover manpage of `tolower` indicates that its parameter should be casted to unsigned char (as char might be signed and thus create problem. Even though the return value is int, reference also assigns it to a regular char.

Signed-off-by:Peter Huewe <peter.huewe@infineon.com>

### DIFF
--- a/src/util/log.c
+++ b/src/util/log.c
@@ -54,21 +54,21 @@ case_insensitive_strncmp(const char *string1,
 
     size_t string1_size = strlen (string1);
     size_t string2_size = strlen (string2);
-    unsigned char lower_string1 [string1_size + 1];
-    unsigned char lower_string2 [string2_size + 1];
+    char lower_string1 [string1_size + 1];
+    char lower_string2 [string2_size + 1];
     size_t i;
 
     for (i = 0; i < string1_size; i++)
-        lower_string1[i] = tolower(string1[i]);
+        lower_string1[i] = tolower((unsigned char) string1[i]);
 
     lower_string1[i] = '\0';
 
     for (i = 0; i < string2_size; i++)
-        lower_string2[i] = tolower(string2[i]);
+        lower_string2[i] = tolower((unsigned char) string2[i]);
 
     lower_string2[i] = '\0';
 
-    return memcmp (lower_string1, lower_string2, n);
+    return strncmp (lower_string1, lower_string2, n);
 }
 
 log_level


AndreasFuchsSIT:
Agreed, it should be `strncmp` not `memcmp`.
I'm curious about the `char = int` assignment here and loss of precision warnings in static analysis or newer compilers.
Pending this, I'd approve.

PeterHuewe:
The patch does not change any `char = int` assignments.

PeterHuewe:
Maybe it just makes sense to rewrite it completely, not using strncmp at all.
Since we are already touching each byte in the two tolower loops....

AndreasFuchsSIT:
it turned `unsigned char = int` into `char = int`
but scanbuild on travis is happy, thus so am I... :-)

a rewrite could work. We don't even need cmp functionality, only eq. We can even `return FALSE` if `strlen1 != strlen2` of both are larger than `n`.

tstruk:
We can remove it completely and convert the whole "module" string to lower case before comparing the strings. Something like:
```
@@ -157,14 +102,23 @@ getLogLevel(const char *module, log_level logdefault)
 {
     log_level loglevel = logdefault;
     char *envlevel = getenv("TSS2_LOG");
-    char *i = envlevel;
+    char *i = envlevel, module_lower[strlen(module) + 1];
+    size_t x;
+
     if (envlevel == NULL)
         return loglevel;
+
+    for (x = 0; x < strlen(module); x++)
+        module_lower[x] = tolower((unsigned char) module[x]);
+
+    module_lower[x] = '\0';
+
     while ((i = strchr(i, '+')) != NULL) {
         if ((envlevel <= i - strlen("all") &&
-            case_insensitive_strncmp(i - 3, "all", 3) == 0) ||
-            (envlevel <= i - strlen(module) &&
-             case_insensitive_strncmp(i - strlen(module), module, strlen(module)) == 0)) {
+            strncmp(i - 3, "all", 3) == 0) ||
+            (envlevel <= i - strlen(module_lower) &&
+             strncmp(i - strlen(module_lower), module_lower,
+                    strlen(module_lower)) == 0)) {
             log_level tmp = log_stringlevel(i+1);
             if (tmp != LOGLEVEL_UNDEFINED)
                 loglevel = tmp;
```

PeterHuewe:
Superseeded by #1277

PR #1279, opened 2019-02-19
mu: TPM2B: Validate requested size to marshall

pdxjohnny:
For TPM2B types with both size and buffer, validate that the size given
for the buffer to be marshaled is not more than the max size of the
buffer to be marshaled.


src/tss2-mu/tpm2b-types.c
```diff
@@ -47,6 +47,13 @@ TSS2_RC Tss2_MU_##type##_Marshal(type const *src, uint8_t buffer[], \
              local_offset, \
              sizeof(src->size) + src->size); \
         return TSS2_MU_RC_INSUFFICIENT_BUFFER; \
+    } else if ((sizeof(type) - sizeof(src->size)) < src->size) { \
```
pdxjohnny: This `sizeof(type)` should maybe be the `sizeof(src->buf_name)` like in the unmarshals:

https://github.com/tpm2-software/tpm2-tss/blob/edbcfc688278e96c83d9c20c479bfdc5b739be62/src/tss2-mu/tpm2b-types.c#L137

I tried doing `sizeof(((TPM2B *)src)->buffer)` but that was always `1`, maybe the unmarshal way is more correct than what I have here?

src/tss2-mu/tpm2b-types.c
```diff
@@ -47,6 +47,13 @@ TSS2_RC Tss2_MU_##type##_Marshal(type const *src, uint8_t buffer[], \
...
+        LOG_WARNING(\
+             "size: %u for buffer of " #type " is larger than max length" \
+             " of buffer: %zu", \
+             src->size, \
+             (sizeof(type) - sizeof(src->size))); \
+        return TSS2_MU_RC_BAD_REFERENCE; \
```
tstruk: *_RC_BAD_REFERENCE is usually for NULL ptrs. Here TSS2_MU_RC_BAD_SIZE error code will work better.

src/tss2-mu/tpm2b-types.c
AndreasFuchsSIT: TPM2B is a different type, thus sizeof resolved to that.
To make it correct, the macro would need a second parameter similar to the unmarshalling, unfortunately.

tstruk: (sizeof(type) - sizeof(src->size)) gives exactly what the sizeof(dest->buf_name) gives in unmarshal.

PR #1286, opened 2019-02-20
Session flags

tstruk:
The enc flag needs to be check earlier, before bailing out on TSS2_SYS_RC_NO_ENCRYPT_PARAM error.


williamcroberts:
from git am:
```
Applying: esys: check the enc param flags before bailing out
Applying: esys: formating - replace tabs with spaces in iesys_check_response()
```

PR #1287, opened 2019-02-20
Upload coverage metrics for selected builds only

tstruk:
There is no need to upload 12 sets of metrics from a single build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ dist: xenial
 
 env:
   matrix:
-  - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_0_2-stable
-  - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_1_0-stable
+  - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_0_2-stable COVERALLS=send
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
I think it's only 6, since we only upload from GCC.

With this patch, we will lose a lot of coverage in the stats compared to what we really have, because this commit will throw away all the ifdef'ed code.

I'd suggest at least uploading lines 1, 2 and 4.

tstruk:
@AndreasFuchsSIT That is true, but will it make any difference? Say the build against ossl has 85% coverage and against gcrypt has 85%, and the overlap of these two covers 90%. This doesn't mean that if we upload the metrics from the two we will get 90%. The last 85% will simply overwrite the first 85%. To get the max coverage we will need to have everything enabled at the same time.

AndreasFuchsSIT:
Looks at least somewhat more intelligent that this, because both `esys_crypto_gcrypt.c` as well as `esys_crypto_ossl.c` are marked "100%" on https://coveralls.io/github/tpm2-software/tpm2-tss?branch=master

And inside https://coveralls.io/builds/21753468/source?filename=src/tss2-esys/esys_crypto_ossl.c#L643 both, the `#if` as well as the `#else` are visited.

tstruk:
Updated to upload 1, 2 and 4.

PR #1288, opened 2019-02-21
tcti: fix null ptr dereference in tcti_tbs_receive

tstruk:
If response_size is NULL the function will dereference a NULL ptr before checking.

### DIFF
--- a/src/tss2-tcti/tcti-tbs.c
+++ b/src/tss2-tcti/tcti-tbs.c
@@ -111,7 +111,7 @@ tcti_tbs_receive (
     TSS2_TCTI_COMMON_CONTEXT *tcti_common = tcti_tbs_down_cast (tcti_tbs);
     TSS2_RC rc = TSS2_RC_SUCCESS;
     TBS_RESULT tbs_rc;
-    int original_size = *response_size;
+    int original_size;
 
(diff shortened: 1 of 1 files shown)


src/tss2-tcti/tcti-tbs.c
```diff
@@ -325,4 +327,4 @@ const TSS2_TCTI_INFO*
 Tss2_Tcti_Info (void)
 {
     return &tss2_tcti_info;
-}
\ No newline at end of file
+}
```
williamcroberts: Nit: superfluous white space change, split to another patch.

tstruk: There is no change. This is just the git way of noting that the file does not ends with '\n'
It looks different in gitub GUI. The git diff in terminal looks like this:
```
@@ -325,4 +327,4 @@ const TSS2_TCTI_INFO*
 Tss2_Tcti_Info (void)
 {
     return &tss2_tcti_info;
-}
\ No newline at end of file
+}
```
See https://mrkiffie.com/2016/newlines-and-version-control/

williamcroberts: Freaking UI's

PR #1296, opened 2019-03-01
Fuzzing

tstruk:
Some improvements & fixes for fuzzing

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10,8 +10,8 @@ AC_INIT([tpm2-tss],
         [https://github.com/tpm2-software/tpm2-tss])
 AC_CONFIG_MACRO_DIR([m4])
 : ${CFLAGS=""}
-AC_PROG_CXX
-AC_PROG_CC
+AC_PROG_CXX([clang++ g++])
+AC_PROG_CC([clang gcc])
 LT_INIT()
 AM_INIT_AUTOMAKE([foreign
                   subdir-objects])
--- a/configure.ac
+++ b/configure.ac
@@ -228,7 +228,9 @@ AM_CONDITIONAL([ENABLE_FUZZING],[test "x$with_fuzzing" != "xnone"])
 AS_IF([test "x$with_fuzzing" != "xnone"],
       [ADD_COMPILER_FLAG([-fsanitize=fuzzer-no-link], [required])
        AS_IF([test "x$enable_tcti_fuzzng" = "xno"],
-             AC_MSG_ERROR([Fuzz tests can not be enabled without the TCTI_FUZZING module]))])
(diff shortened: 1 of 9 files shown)


configure.ac
```diff
@@ -228,7 +228,9 @@ AM_CONDITIONAL([ENABLE_FUZZING],[test "x$with_fuzzing" != "xnone"])
 AS_IF([test "x$with_fuzzing" != "xnone"],
       [ADD_COMPILER_FLAG([-fsanitize=fuzzer-no-link], [required])
        AS_IF([test "x$enable_tcti_fuzzng" = "xno"],
-             AC_MSG_ERROR([Fuzz tests can not be enabled without the TCTI_FUZZING module]))])
+             AC_MSG_ERROR([Fuzz tests can not be enabled without the TCTI_FUZZING module]))
+       AS_IF([test "x$GEN_FUZZ" != "x1"],
+             AC_MSG_ERROR([Fuzz tests can not be enabled without "GEN_FUZZ=1" variable]))])
```
pdxjohnny: `GEN_FUZZ` env variable is only used in bootstrap to see if it needs to generate the cpp and makefile

pdxjohnny:
I am now seeing that I enabled travis for fuzzing wrong. It didn't set `GEN_FUZZ` in front of bootstrap and so they were never generated. Hopefully this fixes that and they work as intended. I'll review the travis log to confirm once it finishes, Thanks!

configure.ac
tstruk: The thing it that you can run ./bootstrap without setting GEN_FUZZ and then run ./configure --enable_tcti_fuzzng=yes and thing don't work. This check is a best attempt to ensure that if you run configure --enable_tcti_fuzzng=yes the fuzz.cpp files are actually generated.

tstruk:
Yes, they are generated - see https://api.travis-ci.org/v3/job/500575432/log.txt

pdxjohnny:
Sweet! This looks great to me then

configure.ac
dilyanpalauzov: Calling
```
GEN_FUZZ=1 ./bootstrap
./configure --with-fuzzing=libfuzzer --enable-tcti-fuzzing --disable-tcti-device --disable-tcti-mssim
```
should be valid, but fails, as during ./configure the variable is not set.  Rather than checking whether the GEN_FUZZ variable is set to one, ./configure shall validate, that the file $(top_buiddir)/test/fuzz/Tss2_Sys_Clear_Complete.fuzz.cpp exists, e.g. with `AC_CHECK_FILE()`.

I do not plan to provide a PR.

PR #1309, opened 2019-03-07
configure.ac: preserve default behaviour of CFLAGS

dilyanpalauzov:
In all Autoconf projects is expected that:

• calling “./configure” implies passing -O2 -g to the compiler, while
• calling “CFLAGS=–flto ./configure” implies passing -flto to the compiler
but omiting “-O2 -g”.

This patch restores the expected behaviour.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,7 +9,6 @@ AC_INIT([tpm2-tss],
         [],
         [https://github.com/tpm2-software/tpm2-tss])
 AC_CONFIG_MACRO_DIR([m4])
-: ${CFLAGS=""}
 AC_PROG_CXX([clang++ g++])
 AC_PROG_CC([clang gcc])
 LT_INIT()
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
@tstruk Would you have a look, since it was your original patch ?
@dilyanpalauzov the origin of this is https://github.com/tpm2-software/tpm2-tss/commit/3980bad87fe18ad9d32914e9d21dba145eba973f or the note under AC_PROG_CC at https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/C-Compiler.html

So I guess it's about `./configure "CFLAGS=-Og -ggdb"` vs `./configure --enable-debug` right ?

dilyanpalauzov:
The rationale for introducing “: ${CFLAGS=""}” here was connected with abolishing CONFIG_SITE and having mutually exclusive optimized versus debug builds.

* Users still will use CONFIG_SITE.
* Sometimes it is necessary to have an optimized build that has debug information inside.  E.g. the only way I can run Gnome Evolution under valgrind on my computer is to have it compiled with -O3, otherwise it is very, very, very, very slow.  -O3 -flto makes it just very, very slow.
* Optimizations can trigger errors, so one needs debug information in optimized (-D_FORTIFY_SOURCE=2) builds to debug them.
* I expect that “CFLAGS=-flto ./configure” will omit -g

I do not see the benefit of these two extra lines in configure.ac.

AndreasFuchsSIT:
I'm convinced.
@tstruk ?

PR #1317, opened 2019-03-13
Fix upcasting issues in Tss2_Sys_GetRspAuths.c

LukasJaegerSIT:
Upcasting in Tss2_Sys_GetRspAuths.c causes memory alignment
errors on some platforms. Therefore, their occurance is
replaced with a more secure implementation using memcpy.

Signed-off-by: Lukas Jäger <lukas.jaeger@sit.fraunhofer.de>

### DIFF
--- a/src/tss2-sys/api/Tss2_Sys_GetRspAuths.c
+++ b/src/tss2-sys/api/Tss2_Sys_GetRspAuths.c
@@ -9,6 +9,7 @@
 #include "tss2_tpm2_types.h"
 #include "tss2_mu.h"
 #include "sysapi_util.h"
+#include <string.h>
 
 TSS2_RC Tss2_Sys_GetRspAuths(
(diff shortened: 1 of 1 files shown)


src/tss2-sys/api/Tss2_Sys_GetRspAuths.c
```diff
@@ -42,8 +43,10 @@ TSS2_RC Tss2_Sys_GetRspAuths(
         if (offset_tmp > ctx->rsp_header.responseSize)
             return TSS2_SYS_RC_MALFORMED_RESPONSE;
 
-        offset_tmp += sizeof(UINT16) +
-            BE_TO_HOST_16(*(UINT16 *)(ctx->cmdBuffer + offset_tmp));
+				UINT16 tmp;
```
AndreasFuchsSIT: Tab indentations here and following.

PR #1322, opened 2019-03-17
deheader: remove double #include’d headers

dilyanpalauzov:
and fix unbalanced #ifdef .. #endif

Found by http://www.catb.org/~esr/deheader/ .


src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -22,7 +22,6 @@
 #define LOGMODULE esys_crypto
 #include "util/log.h"
 #include "util/aux_util.h"
-#include "esys_crypto_ossl.h"
```
AndreasFuchsSIT: IMHO, a XXX.c file should always include the XXX.h file. But probably as very first...

PR #1323, opened 2019-03-18
Configure: further tweaks

dilyanpalauzov:
* print in the summary the default value for tctidefaultmodule, if --with-tctidefaultmodule was not supplied
* replace "x$with_cryptor" → "x$with_crypto"

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -85,17 +85,14 @@ AC_ARG_WITH([crypto],
 AM_CONDITIONAL(ESYS_OSSL, test "x$with_crypto" = "xossl")
 AM_CONDITIONAL(ESYS_GCRYPT, test "x$with_crypto" = "xgcrypt")
 
-AS_IF([test "x$with_crypto" != "xgcrypt"],
-    AS_IF([test "x$with_crypto" != "xossl"],
-          AC_MSG_ERROR([Bad value for --with-crypto $with_crypto])))
+AS_IF([test "x$with_crypto" != xgcrypt -a "x$with_crypto" != xossl],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -116,7 +113,8 @@ AC_ARG_WITH([tctidefaultmodule],
 [The default TCTI module for ESAPI. (Default: libtss2-tcti-default.so)])],
             [AC_DEFINE_UNQUOTED([ESYS_TCTI_DEFAULT_MODULE],
                                 ["$with_tctidefaultmodule"],
-                                ["The default TCTI library file"])])
+                                ["The default TCTI library file"])],
+	    [with_tctidefaultmodule=libtss2-tcti-default.so])
```
AndreasFuchsSIT: This change will change the logic of ESYS.

If with_tctidefaultmodule is unset then ESYS will use libtss2-tcti-default.so but also attempt to use fallback options if libtss2-tcti-default.so is not present.
If with_tctidefaultmodule is set then ESYS will _only_ use that library without the fallback options.

So this change must not happen.

dilyanpalauzov: ./configure --help prints `--with-tctidefaultmodule The default TCTI module for ESAPI. (Default: libtss2-tcti-default.so)`.

If ESYS does fallbacks, when ESYS_TCTI_DEFAULT_MODULE is unset, then the --help output is misleading.

In any case, this change does not affect anyhow the value of ESYS_TCTI_DEFAULT_MODULE in config.h, the purpose is to have at the end of ./configure’s output behind `tctidefaultmodule:` text, that corresponds to "(Default: libtss2-tcti-default.so)".

Issue #1320, opened 2019-03-15
ESYS: "openssl" is not an openssl-engine


AndreasFuchsSIT:
For following code is incorrect:
https://github.com/tpm2-software/tpm2-tss/blob/1d57398857b6302acbec7a601b49cb46745a827e/src/tss2-esys/esys_crypto_ossl.c#L33

I think it should be "default" but that needs to be confirm for PR.

AndreasFuchsSIT:
Ok, the engine should just be NULL statically

PR #1330, opened 2019-03-20
Fix OpenSSL issue and bump version

AndreasFuchsSIT:
Instead of attempting to load the non-existent engine "openssl" which
returned NULL we are now using the engine NULL aka default directly.
This removes some side-effects with openssl's s_client to crash when
using the tpm2-tss-engine project.

This change is save to do, since even if tpm2-tss-engine is set as the
default engine, the keys are not tpm2-tss-engine keys. The tpm2-tss-engine
will forward operations to OpenSSL's own software implementation for
such standard key objects.

Also bumping version to 2.3.0-dev so downstream project can require >=2.3.0

Fixes: #1320 

Should be backported to 2.2.X

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,7 +4,7 @@
 # All rights reserved.
 
 AC_INIT([tpm2-tss],
-        [2.2.0],
+        [2.3.0-dev],
         [https://github.com/tpm2-software/tpm2-tss/issues],
         [],
         [https://github.com/tpm2-software/tpm2-tss])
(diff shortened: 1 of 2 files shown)


tstruk:
Also do we still need this include:
#include <openssl/engine.h>

AndreasFuchsSIT:
We should keep ENGINE_load_builtin_engines() since I don't know whether it's needed, or if that may e.g. enable crypto accelerators.

Issue #1332, opened 2019-03-20
ESYS: Session attribute automation mixup


AndreasFuchsSIT:
The automated clearing of session attributes in ESYS mixes up encrypt and decrypt somewhere.
If only one of these flags is set, neither flag remains set, e.g. in GetRadnom.

tstruk:
I noticed something similar in the resubmit logic, but I thought I fixed it.

PR #1338, opened 2019-03-21
ESYS: Fix session attribute automation problems

AndreasFuchsSIT:
Add fix and testcase.

Fixes: #1332

### DIFF
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -652,9 +652,17 @@ iesys_encrypt_param(ESYS_CONTEXT * esys_context,
             continue;
         IESYS_SESSION *rsrc_session = &session->rsrc.misc.rsrc_session;
         TPMT_SYM_DEF *symDef = &rsrc_session->symmetric;
-        if (rsrc_session->sessionAttributes & TPMA_SESSION_ENCRYPT) {
+        if (rsrc_session->sessionAttributes & TPMA_SESSION_ENCRYPT)
             return_if_notnull(encryptNonce, "More than one encrypt session",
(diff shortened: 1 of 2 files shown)


src/tss2-esys/esys_iutil.c
```diff
@@ -611,13 +604,16 @@ iesys_restore_session_flags(ESYS_CONTEXT *esys_context)
...
     LOG_DEBUG("Restoring session attribs");
 
-    if (esys_context->enc_session) {
-        LOG_DEBUG("Orig Session Attrs 0x%x, altered Attrs x%x",
-                  esys_context->enc_session->origSessionAttributes,
-                  esys_context->enc_session->sessionAttributes);
+    for (int i = 0; i < 3; i++) {
+        RSRC_NODE_T *session = esys_context->session_tab[i];
+        IESYS_SESSION *rsrc_session = &session->rsrc.misc.rsrc_session;
+        if (session == NULL)
```
tstruk: Looks like we first dereference the session pointer and then check if it is NULL afterwards?

AndreasFuchsSIT: Totally true of course...
Monday then, or you just rewrite my patch...

tstruk: Monday is fine

AndreasFuchsSIT: Fixed

AndreasFuchsSIT:
Should be all done and ready now...

tstruk:
Just noticed that we have now two session flag test files:
test/integration/esys-session-attributes.int.c
test/integration/esys-auto-session-flags.int.c
Probably makes sense to merge them?

PR #1339, opened 2019-03-21
Makefile.am: compile test utilities only on “make test”

dilyanpalauzov:
In particular do not compile on simple $(make) `test/helper/tpm_startup`, `test/helper/tpm_transientempty` and `test/integration/libtest_utils.a`.

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -6,7 +6,8 @@
 TESTS_CFLAGS = $(AM_CFLAGS) $(LIBCRYPTO_CFLAGS) -I$(srcdir)/src/tss2-mu \
     -I$(srcdir)/src/tss2-sys -I$(srcdir)/src/tss2-esys \
     -Wno-unused-parameter -Wno-missing-field-initializers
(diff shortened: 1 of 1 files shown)


diabonas:
This change causes `test/helper/tpm_startup` and `test/helper/tpm_transientempty` to be executed as additional tests, which is undesired as these are only helper programs for [`int-log-compiler.sh`](https://github.com/tpm2-software/tpm2-tss/blob/9ef74e99f6dadb51870386cf7837997fae85d98a/script/int-log-compiler.sh#L231) and [`int-log-compiler-ptpm.sh`](https://github.com/tpm2-software/tpm2-tss/blob/9ef74e99f6dadb51870386cf7837997fae85d98a/script/int-log-compiler-ptpm.sh#L88). If they are executed as standalone tests, the test suite will fail, see the [CI build log](https://travis-ci.org/tpm2-software/tpm2-tss/jobs/509507583#L5963).

tstruk:
This commit causes build fail

dilyanpalauzov:
This is fixed on https://github.com/tpm2-software/tpm2-tss/pull/1343 .

PR #1340, opened 2019-03-21
Makefile.am: remove nodist_ before pkgconfig_DATA

dilyanpalauzov:
From the Automake manual: “Files listed with the ‘_DATA’ primary are not automatically part of the tarball built with ‘make dist’”.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -182,8 +182,8 @@ CLEANFILES += AUTHORS
 
 # pkg-config setup. pc-file declarations happen in the corresponding modules
 pkgconfigdir          = $(libdir)/pkgconfig
-nodist_pkgconfig_DATA =
-CLEANFILES += $(nodist_pkgconfig_DATA)
+pkgconfig_DATA =
+CLEANFILES += $(pkgconfig_DATA)
 
 %.pc : %.pc.in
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
Won't `./configure` on the tar-ball recreate the `.pc` files from the `.pc.in` files ?
Thus, isn't correct that the `.pc.in` files are dist but the `.pc` files are `nodist_` ?

dilyanpalauzov:
The .pc.in files are templates, containing placeholders, e.g. for CFLAGS or prefix.  These must be in the .tar file and are put in the .tar file by EXTRA_DIST.

The .pc files are created from the .pc.in files.  Usually by configure.ac:AC_CONFIG_FILES([x.pc]) and therefore by ./configure.  In case of tpm2-tss the .pc files are created from .pc.in by the Makefile, not by ./configure.  This is however irrelevant.

To sum up: `nodist_pkgconfig_DATA` and `pkgconfig_DATA` is the same, according to the Automake manual, here the logic is not changed anyhow.

AndreasFuchsSIT:
OK, I got it now...

AndreasFuchsSIT:
Please rebase to (now fixed) master, so CI can run through...

PR #1344, opened 2019-03-22
Revert "Makefile.am: compile test utilities only on “make test”"

AndreasFuchsSIT:
Reverts tpm2-software/tpm2-tss#1339 since it makes the build to fail, because startup becomes a test instead of helper for tests


AndreasFuchsSIT:
Closing in favor of #1343

PR #1347, opened 2019-03-22
script/gen_fuzz.py: don’t include test/fuzz/*.fuzz.cpp files on “make dist”

dilyanpalauzov:
Before this change, when `GEN_FUZZ=1 ./bootstrap` has filled Makefile-fuzz-generated.am, the aforementioned files were genetared, and included on “make dist”.

Likewise, test/fuzz/main-sapi.cpp was sometimes considered by “make dist” and test/fuzz/tcti/tss2_tcti_fuzzing.h was always forgotten.

### DIFF
--- a/Makefile-fuzz.am
+++ b/Makefile-fuzz.am
@@ -7,6 +7,7 @@ FUZZ_LOG_COMPILER = $(srcdir)/script/fuzz-log-compiler.sh
 INCLUDE_DIRS += -I$(srcdir)/test/fuzz/tcti
 TESTS_LIBADD = $(lib_LTLIBRARIES) $(libtss2_mu) $(libtss2_sys) $(libutil)
 
+EXTRA_DIST += test/fuzz/main-sapi.cpp
 # tcti library used for fuzzing
 if ENABLE_TCTI_FUZZING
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -12,7 +12,7 @@ AM_LDFLAGS      = $(EXTRA_LDFLAGS) $(CODE_COVERAGE_LIBS)
 # Initialize empty variables to be extended throughout
 lib_LTLIBRARIES =
 noinst_LTLIBRARIES =
-EXTRA_DIST =
+EXTRA_DIST = test/fuzz/main-sapi.cpp
```
tstruk: This should probably go to Makefile-fuzz.am to keep the test and fuzz separate from the rest of the code. As the comment above says this only
initializes empty variables...

dilyanpalauzov: I think it will be simpler, if there is eventually one EXTRA_DIST on a single place including all relevant files, rather than splitting it over different places.

tstruk: I'm not sure about that. Currently the appropriate files are appended to EXTRA_DIST "by component", which is nice and logical.
In some cases this might be conditional. I would prefer this to be in Makefile-fuzz.am.

AndreasFuchsSIT: I agree. I also prefer the `+=` convention where all settings of a component are bundled together.

PR #1352, opened 2019-03-25
Automake: a .h file needs to be included only once in _SOURCES or _HEADERS for the whole project

dilyanpalauzov:
The sole purpose of mentioning a .h file in Makefile*.am is either:
• for installing and distributing it, being in a _HEADERS primary, or
• only for distributing it, when it is in a _SOURCE primary.

The dependencies between .c and .h files (when a .h file is changed, recompile the .c file and relink) are anyway generated by $CC -M.

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -235,14 +235,14 @@ test_unit_tcti_device_CFLAGS  = $(CMOCKA_CFLAGS) $(TESTS_CFLAGS)
 test_unit_tcti_device_LDADD   = $(CMOCKA_LIBS) $(libtss2_mu) $(libutil)
 test_unit_tcti_device_LDFLAGS = -Wl,--wrap=read -Wl,--wrap=write, -Wl,--wrap=poll
 test_unit_tcti_device_SOURCES = test/unit/tcti-device.c \
-    src/tss2-tcti/tcti-common.c src/tss2-tcti/tcti-common.h \
+    src/tss2-tcti/tcti-common.c \
     src/tss2-tcti/tcti-device.c src/tss2-tcti/tcti-device.h
 
 test_unit_tcti_mssim_CFLAGS  = $(CMOCKA_CFLAGS) $(TESTS_CFLAGS)
 test_unit_tcti_mssim_LDADD   = $(CMOCKA_LIBS) $(libtss2_mu) $(libutil)
 test_unit_tcti_mssim_LDFLAGS = -Wl,--wrap=connect,--wrap=read,--wrap=select,--wrap=write
 test_unit_tcti_mssim_SOURCES = test/unit/tcti-mssim.c \
-    src/tss2-tcti/tcti-common.c src/tss2-tcti/tcti-common.h \
+    src/tss2-tcti/tcti-common.c \
     src/tss2-tcti/tcti-mssim.c src/tss2-tcti/tcti-mssim.h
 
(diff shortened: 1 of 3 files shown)


Makefile-test.am
```diff
@@ -235,14 +235,14 @@ test_unit_tcti_device_CFLAGS  = $(CMOCKA_CFLAGS) $(TESTS_CFLAGS)
...
+    src/tss2-tcti/tcti-common.c \
     src/tss2-tcti/tcti-device.c src/tss2-tcti/tcti-device.h
 
 test_unit_tcti_mssim_CFLAGS  = $(CMOCKA_CFLAGS) $(TESTS_CFLAGS)
 test_unit_tcti_mssim_LDADD   = $(CMOCKA_LIBS) $(libtss2_mu) $(libutil)
 test_unit_tcti_mssim_LDFLAGS = -Wl,--wrap=connect,--wrap=read,--wrap=select,--wrap=write
 test_unit_tcti_mssim_SOURCES = test/unit/tcti-mssim.c \
-    src/tss2-tcti/tcti-common.c src/tss2-tcti/tcti-common.h \
+    src/tss2-tcti/tcti-common.c \
     src/tss2-tcti/tcti-mssim.c src/tss2-tcti/tcti-mssim.h
```
tstruk: In this case should the src/tss2-tcti/tcti-device.h and src/tss2-tcti/tcti-mssim.h also be removed?

dilyanpalauzov: If they are removed completely from Makefile*.am then “make dist” will not include them in the tarball.

tstruk: They are listed in
src_tss2_tcti_libtss2_tcti_device_la_SOURCES and
src_tss2_tcti_libtss2_tcti_mssim_la_SOURCES
so if should be ok to remove them form test_unit_tcti_*_SOURCES, right?

dilyanpalauzov: The proposed change removes from Makefile.am:src_tss2_tcti_libtss2_tcti_device_la_SOURCES “src/tss2-tcti/tcti-device.h” and removes from Makefile.am:src_tss2_tcti_libtss2_tcti_mssim_la_SOURCES “src/tss2-tcti/tcti-mssim.h”.

After this change src/tss2-tcti/tcti-device.h will be mentioned only in Makefile-test.am:test_unit_tcti_device_SOURCES and src/tss2-tcti/tcti-mssim.h will be mentioned only in Makefile-test.am:test_unit_tcti_mssim_SOURCES.

tstruk: Ok, sorry, I see it now.
And looks like src/tss2-tcti/tcti-common.h will be listed in test_fuzz_tcti_libtss2_tcti_fuzzing_la_SOURCES in Makefile-fuzz.am?
Wouldn't it make more sense if we remove it from Makefile-fuzz.am and leave it in Makefile-test.am?

dilyanpalauzov: src/tss2-tcti/tcti-common.h will be included over Makefile-fuzz.am:test_fuzz_tcti_libtss2_tcti_fuzzing_la_SOURCES.

From my perspective it will not make any difference if the .h files will be referenced in Makefile-test.am instead of Makefile-fuzz.am.

tstruk: I think it make more sense to include them via one Makefile instead of two unrelated Makefiles.
Let's list them all in Makefile-test.am.

PR #1355, opened 2019-03-27
Tpmclient: refactor NV tests

flihp:
This is some refactoring that I ended up doing while attempting to liberate a different test. Removing the test I was working on caused breakage in these tests so I went ahead and cleaned these up too.

### DIFF
--- a/test/integration/sapi-nv-policy-locality.int.c
+++ b/test/integration/sapi-nv-policy-locality.int.c
@@ -1,0 +1,306 @@
+/*
+ * SPDX-License-Identifier: BSD-2
+ * Copyright (c) 2019, Intel Corporation
+ */
+
+#include <assert.h>
+#include <inttypes.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "tss2_sys.h"
+
+#include "context-util.h"
+#include "sapi-util.h"
+#include "util/aux_util.h"
+#define LOGMODULE test
+#include "util/log.h"
+
+#define NV_INDEX 0x01800003
+#define NV_SIZE 96
+
+#define TPM2B_SIZE_MAX(type) (sizeof (type) - 2)
+
+/*
+ * This test creates an NV index governed by a policy and then performs
+ * several operations on the NV region to exercise this policy. The NV
+ * region created is modeled after the TXT AUX region as defined by the
+ * Intel TXT software developers guide:
+ * https://www.intel.com/content/dam/www/public/us/en/documents/guides/intel-txt-software-development-guide.pdf
+ * Read is controlled by authValue and is unrestricted since authValue is
+ * set to emptyBuffer.
+ * Write is controlled by policy that allows writes from locality 3 and 4.
+ */
+/*
+ * This function creates a policy session asserting that the locality is
+ * either 3 or 4. If this policy is used when executing a command and the
+ * policy is not satisfied (locality is not 3 or 4) then the command will
+ * fail.
+ */
+static TSS2_RC
+create_policy_session (TSS2_SYS_CONTEXT *sys_ctx,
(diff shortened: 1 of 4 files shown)


test/integration/sapi-nv-policy-locality.int.c
```diff
@@ -0,0 +1,311 @@
...
+static TSS2_RC
+setup_nv (TSS2_SYS_CONTEXT *sys_ctx)
+{
+    TSS2_RC rc;
+    TPMI_SH_AUTH_SESSION auth_handle = 0;
+    TPM2B_DIGEST policy_hash = { .size = TPM2B_SIZE_MAX (policy_hash), };
+    TPM2B_AUTH nv_auth = { 0, };
+    TSS2L_SYS_AUTH_RESPONSE auth_rsp = { 0, };
+    TPM2B_NV_PUBLIC public_info = {
+        .size = sizeof (TPMI_RH_NV_INDEX) + sizeof (TPMI_ALG_HASH) +
```
jeffpferreira: .size = 0 for consistency. See #1353

test/integration/sapi-nv-readwrite.int.c
```diff
@@ -0,0 +1,191 @@
...
+{
+    TSS2_RC rc;
+    TPMI_SH_AUTH_SESSION auth_handle;
+    TPM2B_DIGEST  policy_hash = {
+        .size = TPM2B_SIZE_MAX (policy_hash),
+    };
+    TPM2B_AUTH  nv_auth = { .size = 0, };
+    TSS2L_SYS_AUTH_RESPONSE auth_rsp;
+    TPM2B_NV_PUBLIC public_info = {
+        .size = sizeof (TPMI_RH_NV_INDEX) + sizeof (TPMI_ALG_HASH) +
```
jeffpferreira: .size = 0 for consistency. See #1353

test/integration/sapi-nv-policy-locality.int.c
flihp: Good catch. Forgot tss2-sys does these calcs for us now. This code is ancient & predates that change.

flihp:
Commit updated accounting for feedback from @jeffpferreira: the size fields of TPM2B structures are only initialized to non-zero values when appropriate.

flihp:
Also worth noting that the decrease in test coverage (0.01%) reported by codecov is probably related to #1360.

test/integration/sapi-nv-readwrite.int.c
```diff
@@ -0,0 +1,185 @@
...
+}
+int
+test_invoke (TSS2_SYS_CONTEXT *sys_ctx)
+{
+    TSS2_RC rc;
+
+    rc = setup_nv (sys_ctx, INDEX_LCP_OWN);
+    return_if_error (rc, "setup_nv for INDEX_LCP_OWN");
+    rc = setup_nv (sys_ctx, INDEX_LCP_SUP);
+    return_if_error (rc, "setup_nv for INDEX_LCP_SUP");
```
tstruk: If the second setup_nv() fails, or in case of any other failure along the way, we wont clean the INDEX_LCP_OWN NV index. Instead of return_if_error let's use "goto error" and call teardown_nv() for both indexes always.

flihp: Sounds like the requirement here is to clean up the NV regions created in the case of a test failure? I didn't do exactly what you suggest but I think the outcome is the same.

tstruk: That looks good. Thanks @flihp 
Can we do the same in sapi-nv-policy-locality.int.c (next commit)

flihp: Just pushed changes to the policy-locality test @tstruk, rebased as well.

Issue #1354, opened 2019-03-27
Test breaks PCR allocations


AndreasFuchsSIT:
The integration tests break the PCR allocations here:
https://github.com/tpm2-software/tpm2-tss/blob/35c121f6814beef3dbc369e60a2ca189eb381911/test/integration/esys-pcr-basic.int.c#L122

They need to be reset.

PR #1357, opened 2019-03-28
Restore original PCR allocations in test and detect TPM state changes

AndreasFuchsSIT:
Fixes a PCR allocation test.
Also adds some code to detect TPM state changes during tests, such as NV, objects, sessions and theoretically PCR allocations.
PCR allocation changes only take effect after reboot, so they are not really discovered now.

Future TODO: We need to figure out, how to make the tpm-simulator boot before a third call to dumpstate in the script.

Fixes: #1354

### DIFF
--- a/script/int-log-compiler-ptpm.sh
+++ b/script/int-log-compiler-ptpm.sh
@@ -92,6 +92,18 @@ if [ $? -ne 0 ]; then
     break
 fi
 
+TPMSTATE_FILE1=${TEST_BIN}_state1
+TPMSTATE_FILE2=${TEST_BIN}_state2
+
+env TPM20TEST_TCTI_NAME="device" \
+    TPM20TEST_DEVICE_FILE=${PTPM} \
+    G_MESSAGES_DEBUG=all ./test/helper/tpm_dumpstate>$TPMSTATE_FILE1
+if [ $? -ne 0 ]; then
+    echo "Error during dumpstate"
+    ret=99
+    break
+fi
+
 echo "Execute the test script"
 env TPM20TEST_TCTI_NAME="device" \
     TPM20TEST_DEVICE_FILE=${PTPM} \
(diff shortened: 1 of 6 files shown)


script/int-log-compiler-ptpm.sh
```diff
@@ -92,6 +92,18 @@ if [ $? -ne 0 ]; then
     break
 fi
 
+TPMSTATE_FILE1=${TEST_BIN}_state1
+TPMSTATE_FILE2=${TEST_BIN}_state2
+
+env TPM20TEST_TCTI_NAME="device" \
+    TPM20TEST_DEVICE_FILE=${PTPM} \
+    G_MESSAGES_DEBUG=all ./test/helper/tpm_dumpstate>$TPMSTATE_FILE1
+if [ $? -ne 0 ]; then
```
tstruk: You need to add the test/integration/*.int_state[2|1] files to the $CLEANFILES otherwise make distcheck will complain.

script/int-log-compiler.sh
```diff
@@ -253,16 +266,52 @@ env TPM20TEST_TCTI_NAME="socket" \
...
-    G_MESSAGES_DEBUG=all ./test/helper/tpm_transientempty
+    G_MESSAGES_DEBUG=all ./test/helper/tpm_dumpstate>$TPMSTATE_FILE2
 if [ $? -ne 0 ]; then
-    echo "TPM transient area not empty or generally failed after test"
+    echo "Error during dumpstate"
     ret=99
     break
 fi
 
+if [ "$(cat $TPMSTATE_FILE1)" != "$(cat $TPMSTATE_FILE2)" ]; then
```
tstruk: Not something that prevents merging it, but would using diff be more appropriate here?
if [ "$(diff $TPMSTATE_FILE1 $TPMSTATE_FILE2 -ne 0) ];
Probably something to improve.

script/int-log-compiler.sh
```diff
@@ -253,16 +266,52 @@ env TPM20TEST_TCTI_NAME="socket" \
...
+    TPM20TEST_SOCKET_ADDRESS="127.0.0.1" \
+    TPM20TEST_SOCKET_PORT="${SIM_PORT_DATA}" \
+    G_MESSAGES_DEBUG=all ./test/helper/tpm_dumpstate>$TPMSTATE_FILE2
+if [ $? -ne 0 ]; then
+    echo "Error during dumpstate"
+    ret=99
+    break
+fi
+
+if [ "$(cat $TPMSTATE_FILE1)" != "$(cat $TPMSTATE_FILE2)" ]; then
```
tstruk: ditto

script/int-log-compiler.sh
AndreasFuchsSIT: Honestly, I tried it, but for some reason did not work, so I went the lazy route known to work... ;-)

PR #1369, opened 2019-04-04
Makefile.am: set CODE_COVERAGE_RULES before rules are substituted

tstruk:


flihp:
This is a noop according to the `check-code-coverage` make target. Both this branch and master report the same:
Overall coverage rate:
  lines......: 74.2% (13601 of 18342 lines)
  functions..: 79.2% (921 of 1163 functions)

tstruk:
yes @flihp, just trying.

PR #1370, opened 2019-04-05
Spec 1.1

tstruk:
Spec 1.1 changes

### DIFF
--- a/test/unit/sys-execute.c
+++ b/test/unit/sys-execute.c
@@ -1,0 +1,164 @@
+/* SPDX-License-Identifier: BSD-2 */
+/*******************************************************************************
+ * Copyright 2019, Intel Corporation
+ *
+ * All rights reserved.
+ ******************************************************************************/
+
+#include <stdarg.h>
+#include <inttypes.h>
+#include <string.h>
+#include <stdlib.h>
+#include <setjmp.h>
+#include <cmocka.h>
+
+#include "tss2_sys.h"
+#include "sysapi_util.h"
+#include "tss2-tcti/tcti-common.h"
+
+
+#define LOGMODULE test
+#include "util/log.h"
+
+/**
+ * Test calls Tss2_Sys_Execute() many times after receiving TPM2_RC_RETRY
+ */
+
+const uint8_t ok_response[] = {
+    0x80, 0x01,                 /* TPM_ST_NO_SESSION */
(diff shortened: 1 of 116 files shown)


AndreasFuchsSIT:
I think 265d0ebdebb9b4ac86453243b406d39c9f86c828 should need a unit test...

tstruk:
added unit test for https://github.com/tpm2-software/tpm2-tss/commit/265d0ebdebb9b4ac86453243b406d39c9f86c828

test/unit/sys-execute.c
```diff
@@ -0,0 +1,167 @@
...
+    TSS2_RC r;
+
+    size_ctx = Tss2_Sys_GetContextSize(0);
+    sys_ctx = calloc (1, size_ctx);
+    assert_non_null (sys_ctx);
+    _tcti_v1_ctx = (TSS2_TCTI_CONTEXT_COMMON_V1 *) &_tcti_ctx;
+    _tcti_v1_ctx->version = 1;
+    _tcti_v1_ctx->transmit = tcti_transmit;
+    _tcti_v1_ctx->receive = tcti_receive;
+    _tcti_ctx.state = TCTI_STATE_TRANSMIT;
```
AndreasFuchsSIT: This variable is never used.

tstruk: Do you mean _tcti_ctx? We need to set the internal TCTI state to transmit otherwise the Execute() will fail with TSS2_TCTI_RC_BAD_SEQUENCE https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-tcti/tcti-common.c#L50
Also this is the storage space for the whole TCTI context.
Both the tcti_ctx * and _tcti_v1_ctx * point to &_tcti_ctx, which is TSS2_TCTI_COMMON_CONTEXT.

AndreasFuchsSIT: `.state` is never used. And tcti_common_transmit_checks is not called from tcti_transmit function in this file.

I guess none of the extra fields from TSS2_TCTI_COMMON_CONTEXT are called, so using TSS2_TCTI_CONTEXT_COMMON_V1 would even be sufficient for _tcti_ctx...

tstruk: Ok, you are right, TSS2_TCTI_CONTEXT_COMMON_V1 is sufficient to setup the test. Code updated.

PR #1376, opened 2019-04-16
SYS & ESYS symbol maps

tstruk:
Add symbol maps for SYS & ESYS


src/tss2-esys/tss2-esys.vcxproj
```diff
@@ -251,10 +251,10 @@
     <ClCompile Include="api\Esys_Vendor_TCG_Test.c" />
```
AndreasFuchsSIT: SYS and ESYS are structured differently.
in SYS, all TCG-Spec functions are under api/
in ESYS, all TPM-based functions are under api/

So this change screws over the structure of ESYS.
Of course we are screwed anyways with having a difference here to begin with...

Makefile-test.am
```diff
@@ -57,7 +57,6 @@ if UNIT
 TESTS_UNIT  = \
     test/unit/CommonPreparePrologue \
     test/unit/CopyCommandHeader \
-    test/unit/GetNumHandles \
```
AndreasFuchsSIT: This is an endless argument between devs I guess.
@flihp removed all statics, because unit tests. I did not care too much... ;-)

But I guess removing a unit test in favor of statics is not too good.
I guess we can remove the functions from the headers (I'm all for that) but then declare them locally in the corresponding test ?

src/tss2-esys/tss2-esys.vcxproj
tstruk: For the sake of consistency I think we should keep TCG defined Esys_* functions in api/ also.
Esys_Free was easy to move since it is in a separate .c file. It is more work for TR or context related functions so I left it there.

Makefile-test.am
tstruk: To declare private functions locally in test will not work if a function is static.
In the case of GetNumHandles we have a good coverage as it is called for every TPM command.

tstruk:
Patches re-arranged and intermediate versions tested.

src/tss2-esys/tss2-esys.vcxproj
AndreasFuchsSIT: I'd argue to either move all or none in one go... ;-)

Makefile-test.am
AndreasFuchsSIT: Let's await the email outcome...

tstruk: Here is the coverage:
https://codecov.io/gh/tpm2-software/tpm2-tss/pull/1376/src/src/tss2-sys/sysapi_util.c?before=src/tss2-sys/sysapi_util.c#D1-286

AndreasFuchsSIT: I know, but unit tests fail earlier, if you know what I mean... ;-)
OTOH, non-statics may lead to collisions...
So I'm undecided, but I guess we want to be consistent...

src/tss2-esys/tss2-esys.vcxproj
AndreasFuchsSIT: How about this now ? Prospone this until we move esys-context.c and esys-tr.c as well ?

tstruk: That's fine I can drop this one from the PR. We can move them all later.

PR #1378, opened 2019-04-16
Improve logging quality

PeterHuewe:
This PR makes not exported functions static, removes unused code and adds a simple test for codecoverage.

Signed-off-by: Peter Huewe <peter.huewe@infineon.com>

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -60,6 +60,7 @@ TESTS_UNIT  = \
     test/unit/GetNumHandles \
     test/unit/io \
     test/unit/key-value-parse \
+    test/unit/log \
     test/unit/tcti-device \
     test/unit/tcti-mssim \
     test/unit/UINT8-marshal \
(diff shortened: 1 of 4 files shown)


AndreasFuchsSIT:
Would it make sense to turn the test into an exe plus shell script and check the output of the program ?

If course I like the added coverage either way...

PeterHuewe:
> Would it make sense to turn the test into an exe plus shell script and check the output of the program ?
Maybe - or wrap vfprintf.

One Question: 
What should getLogLevel in case of an unknown loglevel in the TSS2_LOG  environment variable?
Should it return LOGLEVEL_UNDEFINED (what I would expect) or the default loglevel (what it does).

I updated the testcases once more to get a little bit more coverage.

AndreasFuchsSIT:
I guess you could also get rid of the logdefault parameter and instead add some code at the callers to set the loglevel to logdefault if getLogLevel returns LOGLEVEL_UNDEFINED.

AndreasFuchsSIT:
I guess I'll merge this for now and if you want to move the logdefault assignement out of getLogLevel and rework it to LOGLEVEL_UNDEFINED you can do so in a subsequent PR.

Issue #1377, opened 2019-04-16
Exclude Tests from CodeCoverage


PeterHuewe:
CodeCoverage should not include the testcases itself, but only the source to be tested.
This gives a more realistic view on the covered code.

AndreasFuchsSIT:
Hmmm... we set the
https://github.com/tpm2-software/tpm2-tss/blob/e341d5dbc011a07c50d1d1015d23d2922ee3073b/Makefile.am#L29
but it happens after the
https://github.com/tpm2-software/tpm2-tss/blob/e341d5dbc011a07c50d1d1015d23d2922ee3073b/Makefile.am#L27

Or is it rather codecov.io ? and we need a `.codecov.yml` file to declare tests and library independently
- https://docs.codecov.io/docs/commit-status#section-excluding-tests-example-

There we should also set a threshold of `1%` i guess
- https://docs.codecov.io/docs/commit-status#section-threshold

I've tried to come up with a `.codecov.yml`
```
coverage:
  status:
    project:
      default: false
      tests:
        target: 90%
        paths: "test/"
      tpm2-tss:
        threshold: 1%
        paths: "!test/"  # remove all files in "tests/"
comment:
  behavior: new
```
What I don't know is how this will effect the badge or if we need a new badge-URL ?
Also, I have now declared the coverage-status rules for the project (on master I presume), but will this also affect PRs or not ?

I'll open a PR so we can have a look...

@tstruk @PeterHuewe What do you think, any updates / ideas ?

PeterHuewe:
See tpm2-pkcs11  - there the problem is fixed.
https://github.com/tpm2-software/tpm2-pkcs11/blob/master/codecov.yml

williamcroberts:
> See tpm2-pkcs11 - there the problem is fixed.
> https://github.com/tpm2-software/tpm2-pkcs11/blob/master/codecov.yml
> 

We need to fix that threshold stuff Andreas brought up. Ive been looking at doing that. Sick of
It flunking CI for 0.1% negative change in coverage. We're not at a maturity level to be able
To enforce something that strict in the project yet

AndreasFuchsSIT:
I doupt that it makes sense ever...

The one issue I have now is following:
- If we just ignore the tests directory, we have no means of checking if a certain test file exits preemtively via code coverage. codecov.io also recommends checking the tests directory for 100% code coverage, which seems logical.
[...]

PR #1379, opened 2019-04-16
CI: configure codecov

AndreasFuchsSIT:
This is a testing-PR mostly for checking the settings of the codecov config file.

Fixes: #1377

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,11 @@
+ignore:
+  - "test"
+
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1%
+
+comment:
+  behavior: new


tstruk:
This doesn't help. Perhaps something like this might help:
https://docs.codecov.io/docs/ignoring-paths

tstruk:
Didn't know that the "Fixes: ..." tag can also be in the PR description. Usually I have added them to commit messages.

AndreasFuchsSIT:
If you click on the green checkmark, it will actually give you tpm2-tss and tests split out...
The question is of course, if there are then separate badges for each of these components and how the heck we get the bot to post them as well... :-(

Yes, I allways had the Fixes in PRs, never in commits...
Interesting, isn't it ? ;-)

PR #1397, opened 2019-05-04
esys: don't stringify ESYS_TCTI_DEFAULT_CONFIG

diabonas:
The macro is already quoted, so using stringification will lead to a configuration string like `"/dev/tpmrm0"`, including the double quotes. Just use the macro directly instead like `ESYS_TCTI_DEFAULT_MODULE`.

An alternative solution would be to keep the stringification and remove the double quotes from the [definition of `ESYS_TCTI_DEFAULT_MODULE` in `configure.ac`](https://github.com/tpm2-software/tpm2-tss/blob/2f3bff12ad2527668c84a684711b8ae478442adc/configure.ac#L110). However this wouldn't be consistent with the other strings in `config.h` (including the ones automatically defined by Autoconf) which all have quotes.

### DIFF
--- a/src/tss2-esys/esys_tcti_default.c
+++ b/src/tss2-esys/esys_tcti_default.c
@@ -23,9 +23,6 @@
 #define LOGMODULE esys
 #include "util/log.h"
 
-#define _STR(A) #A
-#define _XSTR(A) _STR(A)
-
 #define ARRAY_SIZE(X) (sizeof(X)/sizeof(X[0]))
 
 struct {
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
To be included in next bugfix release ?

diabonas:
Makes sense I think, I guess nobody is using the `--with-tctidefaultconfig` feature at the moment because with this bug, nobody could actually set a valid configuration (at least for the known TCTI libraries).

PR #1404, opened 2019-05-09
build: Use default OutDir for all VS projects.

flihp:
The previous configuration was a misguided attempt to get all build
output in a central location. Turns out VS doesn't like this since I
rolled the intermediate output directory into the same location. This
causes a race condition since there's no guarantee that the intermediate
build output (object files) will have unique names.

The right way to resolve this is to just remove the configuration and
use the VS defaults.


AndreasFuchsSIT:
@flihp Also to be cherry-picked for 2.2.3 ?

flihp:
If you're willing. Getting the VS build files into the release tarball was important (thanks for that BTW). But with me spending so much time in VS these days it's likely that smaller fixes like this will continue to trickle into the PR queue. Eventually you're gonna have to cut me off :smile: When you do so is at your discretion.

PR #1405, opened 2019-05-10
Makefile-test.am: remove target check-programs

dilyanpalauzov:
This reverts completely commit 24fb1306e, partially reverted with commit a4e8c0ce.

The initial motivation for introducing the check-programs target was, that parallel builds have not worked.  If Makefile is correct, then parallel builds will run rigth.


AndreasFuchsSIT:
make check on ptpm still requires `-j1`, so the motivation is still valid.

PR #913, opened 2018-03-30
Tpm transientempty no tests

flihp:
this resolves #907

PR #1406, opened 2019-05-10
Improvements to Travis CI jobs configuration

diabonas:
- tpm2-tss is built first without tests enabled in order to prevent test-specific code from causing errors if disabled (#913). Improve this test to respect `WITH_TCTI_PARTIAL` and `WITH_TCTI_ASYNC` for better configuration options coverage.
- The fuzz target is tested using a [separate build matrix entry](https://github.com/tpm2-software/tpm2-tss/blob/0e778bf959b20d8f38b5c4f440cfd6c6d55c48b7/.travis.yml#L14). This is redundant because it starts a job for both GCC and Clang (and fuzzing can be used only with Clang) and runs the integration tests already included in [another job](https://github.com/tpm2-software/tpm2-tss/blob/0e778bf959b20d8f38b5c4f440cfd6c6d55c48b7/.travis.yml#L13) before doing the fuzz target test. Move this test to a separate job in `matrix.include` instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
   matrix:
   - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_0_2-stable
   - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_1_0-stable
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
LGTM... I'll await CI result

PR #1407, opened 2019-05-12
esys_iutil.c: Update session attribute in cmdauths structure for policy sessions

idesai:
TPM commands like tpm2_encryptdecrypt can clear the continuesession flag in the
sessionAttributes. However, when it is intended to continue the policy session
inorder to be able to flush the session at a later time in order to:
1. Chain further TPM operations that are intended to modify policy digest
2. Run additional TPM commands using the session with satisfied policy

Signed-off-by: Imran Desai <imran.desai@intel.com>

### DIFF
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -1299,11 +1299,12 @@ iesys_gen_auths(ESYS_CONTEXT * esys_context,
                 auths->auths[auths->count].sessionHandle = session->rsrc.handle;
                 if (objects[session_idx] == NULL) {
                     auths->auths[auths->count].hmac.size = 0;
-                    auths->count += 1;
                 } else {
                     auths->auths[auths->count].hmac = objects[session_idx]->auth;
-                    auths->count += 1;
                 }
+                auths->auths[auths->count].sessionAttributes =
+                    session->rsrc.misc.rsrc_session.sessionAttributes;
+                auths->count += 1;
                 continue;
             }
         }
--- a/test/integration/esys-policy-password.int.c
+++ b/test/integration/esys-policy-password.int.c
@@ -280,6 +280,9 @@ test_esys_policy_password(ESYS_CONTEXT * esys_context)
     r = Esys_FlushContext(esys_context, sessionTrial);
     goto_if_error(r, "Flushing context", error);
 
+    r = Esys_FlushContext(esys_context, policySession);
+    goto_if_error(r, "Flushing context", error);
+
     return EXIT_SUCCESS;
 
  error:


williamcroberts:
@idesai I wish I understood your commit message better, can you elaborate? Shouldn't we be setting the ocntinuesession bit? When I queried the session created by ESAPI it appeared to have that bit set, was it not propagating? Somehow just code change in the tools caused the new failures, which makes me think the proper fix is in the tools, no?

idesai:
@williamcroberts, you are correct the continuesession bit still persists in the ESAPI resource node structure. However, It is lost in transition from ESAPI to SAPI @ https://github.com/tpm2-software/tpm2-tss/blob/0e778bf959b20d8f38b5c4f440cfd6c6d55c48b7/src/tss2-esys/esys_iutil.c#L1283.
I may have to git bisect to check when the tpm2-tool started to fail...my guess would be when tpm2-tools switched over to esapi; Unless esapi_util.c changed at some point.

idesai:
@AndreasFuchsSIT, your thoughts on this PR?

williamcroberts:
Interesting storm of events where my changes in the tool fixed a session handling bug but uncovered a session handling bug in ESAPI. Fun times.

williamcroberts:
@flihp any love on this to get it merged? It blocks the tools at the moment.

AndreasFuchsSIT:
I'm still trying to figure out the patch.
If I see this correctly, then the sessionAttributes are actually set, but just inside the compute_hmac function (not that that was an intuitive name)...
See
https://github.com/tpm2-software/tpm2-tss/blob/a601f8e1f9538226a1f73ff5709770979bcfa4b5/src/tss2-esys/esys_iutil.c#L1223-L1224

So I don't see the added value from this change.

Could you provide me with a detailed bug description on what happens in which case ?

AndreasFuchsSIT:
Ok, so the only difference I see is that the old code used `auths->auths[session_idx]` to call `compute_hmac` where the sessionAttributes are then set. The new code sets the sessionAttributes on `auths->auths[auths->count]`.

This should be the same though unless we have some weirdness around session types and orders.  I'll try to reproduce this tomorrow on my dev-machine, but this needs some more investigation, I guess.

williamcroberts:
@AndreasFuchsSIT this bug is reproduced with master tools and the policysecret.sh test code. Imran has a more concise test case he can share on this thread. Somehow, this patch solves our problem.

idesai:
> I'm still trying to figure out the patch.
> If I see this correctly, then the sessionAttributes are actually set, but just inside the compute_hmac function (not that that was an intuitive name)...
> See
> 
> https://github.com/tpm2-software/tpm2-tss/blob/a601f8e1f9538226a1f73ff5709770979bcfa4b5/src/tss2-esys/esys_iutil.c#L1223-L1224
> 
> So I don't see the added value from this change.
> 
> Could you provide me with a detailed bug description on what happens in which case ?

@AndreasFuchsSIT, I have a summary of the findings in the tools issue tracker here: https://github.com/tpm2-software/tpm2-tools/issues/1463

idesai:
> @AndreasFuchsSIT this bug is reproduced with master tools and the policysecret.sh test code. Imran has a more concise test case he can share on this thread. Somehow, this patch solves our problem.

I used the policypassword integration test script, but here it is more concisely
tpm2_startauthsession -S session.ctx
tpm2_policypassword -S session.ctx -o policy.password
tpm2_flushcontext -S session.ctx
rm session.ctx
tpm2_createprimary -a o -o primary_key.ctx
tpm2_create -g sha256 -G aes -u key.pub -r key.priv -C primary_key.ctx \
  -L policy.password -p testpswd
tpm2_load -C primary_key.ctx -u key.pub -r key.priv -o key.ctx
tpm2_encryptdecrypt -c key.ctx -o encrypted.txt -i plain.txt -p testpswd
tpm2_startauthsession -S session.ctx --policy-session 
tpm2_policypassword -S session.ctx 
tpm2_encryptdecrypt -c key.ctx -i encrypted.txt -o decrypted.txt -p session:session.ctx+testpass -D 
tpm2_flushcontext -S session.ctx 
rm -f session.ctx

williamcroberts:
@idesai why does tpm2-tss/test/integration/esys-policy-password.int.c pass? If you add a save/restore session ctx in that test does it trigger the bug?

AndreasFuchsSIT:
Ok, now I understand it.
This bug was specific to the handling of PolicyPassword sessions.
@idesai Your fix needs to go into this line:
https://github.com/tpm2-software/tpm2-tss/blob/4a3bcf7e8594e6d2c8788c6f28da15767f55af75/src/tss2-esys/esys_iutil.c#L1307
right before the continue.

Your unit-test will not help. We'll need to setup a new unit-test that uses PolicyPassword for this to be triggered.

On a side-note (@williamcroberts ): The use of PolicyPassword is highly discouraged, because it means that passwords will be transferred to the TPM in **plain text**. You should rather use PolicyAuthValue instead. I'd even go as far as saying PolicyPassword is dangerous and should not have a tpm2_tool available.

idesai:
> Ok, now I understand it.
> This bug was specific to the handling of PolicyPassword sessions.
> @idesai Your fix needs to go into this line:
> 
> https://github.com/tpm2-software/tpm2-tss/blob/4a3bcf7e8594e6d2c8788c6f28da15767f55af75/src/tss2-esys/esys_iutil.c#L1307
> 
> 
> right before the continue.
> Your integration-test will not help. We'll need to setup a new integration-test that uses PolicyPassword for this to be triggered.
> 
> On a side-note (@williamcroberts ): The use of PolicyPassword is highly discouraged, because it means that passwords will be transferred to the TPM in **plain text**. You should rather use PolicyAuthValue instead. I'd even go as far as saying PolicyPassword is dangerous and should not have a tpm2_tool available.

@AndreasFuchsSIT, relocating the code to line 1307 right before continue causes both esys-policy-password.int and tpm2-tools to fail.

AndreasFuchsSIT:
@idesai well, because auths->count is incremented before, so you need to put it to auths->count - 1 or move the increment as well... ;-)

idesai:
> @flihp any love on this to get it merged? It blocks the tools at the moment.

@AndreasFuchsSIT @flihp, changes are addressed, need this merged for tools to be unblocked.
Thanks.

williamcroberts:
@AndreasFuchsSIT yes that is a problem with policypassword. My goal is to enable tpm features, not make policy decisions on behalf of the user. For instance if the API is bad, the driver could block it, you could remove it from ESAPI, etc. However, we should make this clear in the manual page.

Issue #1410, opened 2019-05-15
Use some more AX_ makros


AndreasFuchsSIT:
Some candidates that may make live easier. Need to be evaluated though:
- [AX_IS_RELEASE(dash-version)](https://www.gnu.org/software/autoconf-archive/ax_is_release.html) since we are now using -dev as version suffix
- [AX_CHECK_ENABLE_DEBUG](https://www.gnu.org/software/autoconf-archive/ax_check_enable_debug.html)
- [AX_ADD_FORTIFY_SOURCE](https://www.gnu.org/software/autoconf-archive/ax_add_fortify_source.html#ax_add_fortify_source)
- [AX_GENERATE_CHANGELOG](https://www.gnu.org/software/autoconf-archive/ax_generate_changelog.html#ax_generate_changelog)
- ~~[AX_LIB_GCRYPT](https://www.gnu.org/software/autoconf-archive/ax_lib_gcrypt.html#ax_lib_gcrypt) Instead of the current manual check~~ this is for an --with-gcrypt option
- [git.mk](https://github.com/behdad/git.mk)

Issue #1356, opened 2019-03-27
bootstrap warnings


tstruk:
$ ./bootstrap
Makefile-fuzz.am:35: warning: variable 'FUZZ_LDADD' is defined but no program or
Makefile-fuzz.am:35: library has 'FUZZ' as canonical name (possible typo)
Makefile.am:161:   'Makefile-fuzz.am' included from here

2raziel3:
I am getting the same warning, any update ?

AndreasFuchsSIT:
That's non-critical. We just have to rename the variable to FUZZLDADD or something similar.
We are falsely matching Autotools naming convention of _LDADD suffixes.
Good first issue...

2raziel3:
@AndreasFuchsSIT thanks for the info

PR #1412, opened 2019-05-15
Use more automagic

AndreasFuchsSIT:
A bunch of build-fixes.

I'd like to get some feedback from packagers and as many as possible.
I just saw that the tools also already use the ax_enable_debug stuff, so that should be fine.

Fixes: #1410 
Fixes: #1356


Makefile.am
```diff
@@ -17,6 +19,27 @@ DISTCLEANFILES =
...
+    $(GITIGNORE_MAINTAINERCLEANFILES_MAKEFILE_IN) \
+    $(GITIGNORE_MAINTAINERCLEANFILES_M4_LIBTOOL) \
+    src_vars.mk \
+    Makefile-fuzz-generated.am \
+    aminclude_static.am \
+    m4/ax_ac_append_to_file.m4 \
+    m4/ax_ac_print_to_file.m4 \
+    m4/ax_add_am_macro_static.m4 \
+    m4/ax_am_macros_static.m4 \
+    m4/ax_check_compile_flag.m4 \
```
diabonas: On my system, the otherwise untracked `m4/ax_check_gnu_make.m4` needs to be added to `MAINTAINERCLEANFILES`.

AndreasFuchsSIT: I'll add that as well...

PR #1416, opened 2019-05-16
Remove trailing whitespaces and empty lines and add to CI

AndreasFuchsSIT:
Inspired by @diabonas https://github.com/tpm2-software/tpm2-tools/pull/1502

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,8 @@ before_script:
   - ./bootstrap
 
 script:
+# Check for whitespace errors
+  - git diff --check $(git hash-object -t tree /dev/null)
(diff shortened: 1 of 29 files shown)


.travis.yml
```diff
@@ -114,6 +114,8 @@ script:
     else
       make -j$(nproc) check
     fi
+# Check for whitespace errors
+  - git diff --check $(git hash-object -t tree /dev/null)
```
flihp: What do you think about moving this check all the way up to the top of the `script` section? Our ci loop is getting pretty long and failing early is good in these situations.

AndreasFuchsSIT: Done

Issue #1174, opened 2018-10-24
Add appropriate Requires.private and Libs.private to pkgconfig .pc files


tstruk:
Add Requires.private and Libs.private to pkgconfig .pc files where required.

Issue #1398, opened 2019-05-07
On “pkg-config --static --libs tss2-esys”


dilyanpalauzov:
When a program is dynamically linked with libtss2-esys, during the linking the output of `pkg-config --libs tss2-esys` shall be considered.  It returns currently `-L/usr/local/lib -ltss2-esys -ltss2-sys -ltss2-mu`.

When a program is statically linked wit libtss2-esys, during the linking the output of `pkc-config --libs --static tss2-esys` shall be considered.  It returns currently `-L/usr/local/lib -ltss2-esys -ltss2-sys -ltss2-mu`.  Missing are -lssl/-lgcrypt/-lcrypt.

To my understanding, `pkg-config --libs tss2-esys` should just return what is necessary to link, in order to use the program.  If the implementation of libtss2-esys.so changes later, so that it does not depend on libtss2-mu, programs linked wit libtss2-esys should not be recompiled, in order to stop loading libtss2-mu (which is added as DT_NEEDED).

Unfortunately, the way libtool is currently implemented, is to write in the libtss2-esys.la file ` -ldl -L/usr/local/lib64 -lssl -lcrypto /usr/local/lib/libtss2-sys.la /usr/local/lib/libtss2-tcti-device.la /u
sr/local/lib/libtss2-tcti-mssim.la /usr/local/lib/libtss2-mu.la` and when libtool links with -ltss2-esys, to add libcrypto to DT_NEEDED of the resulting binary.  The implication is, that once esys is compiled with OpenSSL and a program links with tss2-esys, the program will get explicit DT_NEEDED for libssl .  If libtss2-esys is then recompiled to use gcrypt, when the program is loaded, it will load both libssl (because it is in the DT_NEEDED of the program) and libgcrypt (because it is in the DT_NEEDED of libtss2-esys).  The program should just list the libraries as DT_NEEDED it directly depends on.

Ignoring libtool, `pkg-config --libs tss2-esys` shall return `-Llibdir -ltss2-esys` and `pkg-config --libs --static tss2-esys` shall return `-L/usr/local/lib -ltss2-esys -ltss2-sys -ltss2-mu` and whatever crypto library is used.

To utilize the --static, pkg-config foresees Libs.private and Requires.private.

[...]

Issue #1365, opened 2019-03-29
Linking libtss2-{mu, sys, tcti-mssim and tcti-device}.so without -lgcrypt on --with-crypto=gcrypt


dilyanpalauzov:
configure.ac contains:
```
AS_IF([test "x$enable_esapi" != xno -a "x$with_crypto" = xgcrypt],
      [AC_CHECK_LIB([gcrypt], [gcry_mac_open], [], [AC_MSG_ERROR([Missing required library: gcrypt.])])])
```
The `[]` mean that on `./configure --with-crypro=gcrypt` puts in the LIBS variable `-lgcrypt` and then everything is linked with libgcrypt.

* Do not link libtss2-mu.so, libtss2-sys.so, libtss2-tcti-mssim.so and libtss2-tcti-device.so with libgrypt, as they don’t use any symbols from there.

* Consider cheching for libgrypt using [AM_PATH_LIBGCRYPT](https://www.gnupg.org/documentation/manuals/gcrypt/Building-sources-using-Automake.html#Building-sources-using-Automake).

PR #1417, opened 2019-05-19
pkg-config: sort out dependencies

diabonas:
The pkg-config files should contain the modules required by the respective library according to the [following guidelines](https://people.freedesktop.org/~dbn/pkg-config-guide.html#writing):

> `Requires` and `Requires.private` define other modules needed by the library. It is usually preferred to use the private variant of `Requires` to avoid exposing unnecessary libraries to the program that is linking with your library. If the program will not be using the symbols of the required library, it should not be linking directly to that library. See the discussion of [overlinking](https://wiki.openmandriva.org/en/Overlinking_issues_in_packaging) for a more thorough explanation.
>
> Since `pkg-config` always exposes the link flags of the `Requires` libraries, these modules will become direct dependencies of the program. On the other hand, libraries from `Requires.private` will only be included when static linking. For this reason, it is usually only appropriate to add modules from the same package in `Requires`.
>
> The `Libs` field contains the link flags necessary to use that library. In addition, `Libs` and `Libs.private` contain link flags for other libraries not supported by `pkg-config`. Similar to the `Requires` field, it is preferred to add link flags for external libraries to the `Libs.private` field so programs do not acquire an additional direct dependency.

The currently specified dependencies in `Requires` are only used internally, so according to these guidelines, they should be in `Requires.private` instead. **Note that this will break downstream programs like tpm2-abrmd and tpm2-tools** because they currently use functions from tss2-mu without explicitly adding this dependency, relying on the dependencies specified in pkg-config. If it is decided that this PR is the right approach, I would open separate pull requests for these projects before merging this PR to add the missing libraries. If you are worried about breaking compatibility like this, it might also be ok to keep these dependencies in `Requires` according to
> For this reason, it is usually only appropriate to add modules from the same package in `Requires`.

above. However keep in mind that the current states lead to unnecessary overlinking for *all* binaries using tss2-sys or tss2-esys.

The other change in this PR is completely backwards-compatible: tss2-sys and tss2-esys are missing some internal dependencies, most notably the cryptographic for the ESAPI. These should be specified in `Requires.private`/`Libs.private` to allow static linking, see also #1398.

Fixes: #1174

### DIFF
--- a/lib/tss2-esys.pc.in
+++ b/lib/tss2-esys.pc.in
@@ -7,6 +7,6 @@ Name: tss2-esys
 Description: TPM2 Enhanced System API library.
 URL: https://github.com/tpm2-software/tpm2-tss
 Version: @VERSION@
-Requires: tss2-mu tss2-sys
+Requires.private: tss2-mu tss2-sys
 Cflags: -I${includedir}
 Libs: -ltss2-esys -L${libdir}
--- a/lib/tss2-esys.pc.in
+++ b/lib/tss2-esys.pc.in
@@ -7,6 +7,7 @@ Name: tss2-esys
 Description: TPM2 Enhanced System API library.
 URL: https://github.com/tpm2-software/tpm2-tss
 Version: @VERSION@
-Requires.private: tss2-mu tss2-sys
+Requires.private: tss2-mu tss2-sys tss2-tcti-device tss2-tcti-mssim
 Cflags: -I${includedir}
 Libs: -ltss2-esys -L${libdir}
+Libs.private: @LIBDL_LDFLAGS@ @LIBSOCKET_LDFLAGS@ @TSS2_ESYS_LDFLAGS_CRYPTO@
--- a/lib/tss2-sys.pc.in
+++ b/lib/tss2-sys.pc.in
@@ -10,3 +10,4 @@ Version: @VERSION@
 Requires.private: tss2-mu
 Cflags: -I${includedir}
 Libs: -ltss2-sys -L${libdir}
+Libs.private: @LIBSOCKET_LDFLAGS@
--- a/lib/tss2-sys.pc.in
+++ b/lib/tss2-sys.pc.in
@@ -7,6 +7,6 @@ Name: tss2-sys
 Description: TPM2 System API library.
 URL: https://github.com/tpm2-software/tpm2-tss
 Version: @VERSION@
-Requires: tss2-mu
+Requires.private: tss2-mu
 Cflags: -I${includedir}
 Libs: -ltss2-sys -L${libdir}
--- a/lib/tss2-tcti-device.pc.in
+++ b/lib/tss2-tcti-device.pc.in
@@ -7,6 +7,6 @@ Name: tss2-tcti-device
 Description: TCTI library for communicating with a TPM device node.
 URL: https://github.com/tpm2-software/tpm2-tss
 Version: @VERSION@
-Requires: tss2-mu
+Requires.private: tss2-mu
 Cflags: -I${includedir}
 Libs: -ltss2-tcti-device -L${libdir}
--- a/lib/tss2-tcti-mssim.pc.in
+++ b/lib/tss2-tcti-mssim.pc.in
@@ -7,6 +7,6 @@ Name: tss2-tcti-mssim
 Description: TCTI library for communicating with the Microsoft TPM2 simulator.
 URL: https://github.com/tpm2-software/tpm2-tss
 Version: @VERSION@
-Requires: tss2-mu
+Requires.private: tss2-mu
 Cflags: -I${includedir}
 Libs: -ltss2-tcti-mssim -L${libdir}


diabonas:
Since this change potentially affects downstream projects, I would also welcome feedback from @flihp @williamcroberts.

dilyanpalauzov:
If there is use case for a program to link only with tss2-esapi or tss2-sapi, but not with tss2-mu, and still be functuinal then the .pc files must allow this scenario.  If there is no such scenario, then there is more than one way to do the right thing.

diabonas:
> If there is use case for a program to link only with tss2-esapi or tss2-sapi, but not with tss2-mu, and still be functional then the .pc files must allow this scenario. If there is no such scenario, then there is more than one way to do the right thing.

@dilyanpalauzov There certainly are valid use cases where you only need tpm2-(e)sys, but not tpm2-mu, e.g. if you only want to [read and write data to and from a NVRAM index](https://github.com/tpm2-software/tpm2-totp/blob/b7f29328dc1e332a44c1f9b477e763f3c9aa70a7/src/libtpm2-totp.c#L533). Marshalling is used quite often, but is by no means compulsory, so I think it really should be in `Requires.private`.

> The question is though: If we break downstreams, do we need to roll a major release ?
> I think no, since this is a bug-fix, not a removed feature. But I'm open to comments.

@AndreasFuchsSIT Good question: IMHO this is clearly a bug that is fixed, see above, so one can argue that it is wrong to rely on the current behaviour. On the other hand, pkg-config files are part of the public API which we are changing in a backwards-incompatible way. In cases where downstream software might depend on an upstream bug, SemVer [seems to suggest](https://github.com/semver/semver/issues/445) trying to gauge how many users might be affected by the change and to error on the side of caution (i.e. do a major release) if in doubt.

diabonas:
I have opened pull requests for [tpm2-abrmd](https://github.com/tpm2-software/tpm2-abrmd/pull/629), [tpm2-tools](https://github.com/tpm2-software/tpm2-tools/pull/1512), [tpm2-tss-engine](https://github.com/tpm2-software/tpm2-tss-engine/pull/115), [tpm2-pkcs11](https://github.com/tpm2-software/tpm2-pkcs11/pull/230) and [tpm2-totp](https://github.com/tpm2-software/tpm2-totp/pull/31). [tpm2-tcti-uefi](https://github.com/tpm2-software/tpm2-tcti-uefi) is not affected since it doesn't use pkg-config.

PeterHuewe:
PR for  [tpm2-pkcs11](https://github.com/tpm2-software/tpm2-pkcs11/pull/230) approved and merged.

AndreasFuchsSIT:
> > The question is though: If we break downstreams, do we need to roll a major release ?
> > I think no, since this is a bug-fix, not a removed feature. But I'm open to comments.
> 
> @AndreasFuchsSIT Good question: IMHO this is clearly a bug that is fixed, see above, so one can argue that it is wrong to rely on the current behaviour. On the other hand, pkg-config files are part of the public API which we are changing in a backwards-incompatible way. In cases where downstream software might depend on an upstream bug, SemVer [seems to suggest](https://github.com/semver/semver/issues/445) trying to gauge how many users might be affected by the change and to error on the side of caution (i.e. do a major release) if in doubt.

IMHO, then we'll probably leave this PR laying around until end of the year release. Until then, all downstreams will have had one minor release cycle include the appropriate bugfix...

diabonas:
Sure, fine by me!

PeterHuewe:
Can we track the progress on the downstreams some how? (i.e. when they have a minor release).

AndreasFuchsSIT:
> Can we track the progress on the downstreams some how? (i.e. when they have a minor release).

No idea. I guess we just have to pay attention... ;-)

tstruk:
@Fixes: https://github.com/tpm2-software/tpm2-tss/issues/1174

dilyanpalauzov:
As a reminder on the overlinking topic, `-lgcrypt` is used when linking any tpm2-tss library `--with-crypto=gcrypt`, cf. https://github.com/tpm2-software/tpm2-tss/issues/1365.

tstruk:
Since all downstream PRs are merged I think we should merge it sooner rather than later. It will be part of the next 2.3 TSS release, but there is nothing that prevents this to be merged now.

diabonas:
@tstruk Sounds good, this helps preventing further problems being introduced downstream by relying on the old behaviour.

AndreasFuchsSIT:
I wanted to postpone it until all downstreams have had a release first, i.e. until tpm2-tss 2.4 release.
Reason being, that I wanted to retain current compilation of downstream stables and distros.
For 2.3 we should add a huge warning in the CHANGELOG and maybe even README.

Or am I wrong with this concern ?

diabonas:
You are certainly right in worrying about breaking downstream releases, so putting a warning into the changelog for the next release before putting it into the release after next seems like an excellent idea. What I think @tstruk is suggesting is to merge the PR now while still keeping it out of the next release. Since the downstream projects already include fixes, this shouldn't break anything and makes sure that no new problems are introduced. It might also keep new software from relying on the old behaviour if they build against tpm2-tss master.

On the other hand the next release is scheduled in about a month's time, so merging it directly after the release should work as well and saves the rebasing for 2.3.0 that would be necessary if you merge it now.

AndreasFuchsSIT:
Thanks for pinning down the pro and con so nicely.

I like the other hand. :+1:

tstruk:
Merging  this to master doesn't break anything since all downstream patches are already merged.
The benefit of doing it early is that the downstream can try it early and make additional adjustments.
Keeping it out just before 2.3 doesn't add any benefit.

AndreasFuchsSIT:
I was intending on merging it just after 2.3
Thus a 3.1 or 3.2 tpm2-tools will still compile against a 2.3 tpm2-tss
Because afaik, none of the downstreams has backported the related changes, and I don't think they will or even should ?

diabonas:
Backporting the changes removes a few surprises for users mixing different versions of the toolchain, but shouldn't be necessary in general as long as we make sure that every downstream has had a release before doing the change in a released version of tpm2-tss.

For tpm2-tools I have opened https://github.com/tpm2-software/tpm2-tools/pull/1534 to get this into the 3.2 release, tpm2-pkcs11 doesn't have any releases yet anyway, we still need to keep track of tpm2-abrmd, tpm2-tss-engine and tpm2-totp.

diabonas:
The rationale for getting this into tpm2-tools 3.2 is that tpm2-tools 4.0 will contain breaking changes, so many distributions will probably hold back updating for quite some time or at least provide long-term support for the 3.x series. This shouldn't prevent them from updating tpm2-tss to the latest version solely due to "trivial" build problems.

flihp:
Is there anything holding this PR up any longer? It has approval from @AndreasFuchsSIT and the comments on the PR make a good case (IMO) for merging the patch. If no one speaks up I'm merging.

flihp:
did some digging and this has been held up so we can get releases of downstreams ready before this gets merged & released. Goal is to have at least one release of all downstreams with fixes released before this gets merged. the tools has already been done, tabrmd is left as are the engine & totp (see linked issues).

flihp:
The tpm2-abrmd release with the relevant fix is done: https://github.com/tpm2-software/tpm2-abrmd/releases/tag/2.2.0

AndreasFuchsSIT:
engine and totp have rcs out...

AndreasFuchsSIT:
Since all downstreams are now done, I'll hit merge...

PR #1418, opened 2019-05-19
ESYS: Fix computation of session name in Esys_StartAuthSession.

JuergenReppSIT:
* The session name was no added to meta data in StartAuthSession.
* This caused an error in HMAC computation for policy sessions.

### DIFF
--- a/src/tss2-esys/api/Esys_StartAuthSession.c
+++ b/src/tss2-esys/api/Esys_StartAuthSession.c
@@ -478,6 +478,14 @@ Esys_StartAuthSession_Finish(
 
         sessionHandleNode->rsrc.misc.rsrc_session.sessionKey.size = authHash_size;
     }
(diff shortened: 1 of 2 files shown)


AndreasFuchsSIT:
@JuergenReppSIT Could you add/trigger a test that triggers this issue, so it won't happen in the future ?

src/tss2-esys/api/Esys_StartAuthSession.c
```diff
@@ -1,4 +1,4 @@
-/* SPDX-License-Identifier: BSD-2-Clause */
+/* SPDX-License-Identifier: BSD-2 */
```
AndreasFuchsSIT: Please revert this change

AndreasFuchsSIT: This is still wrong...

Issue #1419, opened 2019-05-20
tests: fuzzing: OSS-Fuzz build error


pdxjohnny:
```log
Step #1: Step 13/15 : RUN git clone --depth 1   https://github.com/tpm2-software/tpm2-tss $SRC/tpm2-tss/
Step #1:  ---> Running in ffc384e17846
Step #1: [91mCloning into '/src/tpm2-tss'...
Step #1: [0mRemoving intermediate container ffc384e17846
Step #1:  ---> 80a167a9a1e4
Step #1: Step 14/15 : WORKDIR $SRC/tpm2-tss/
Step #1:  ---> Running in 8eecfa3b4a95
Step #1: Removing intermediate container 8eecfa3b4a95
Step #1:  ---> ad39a11a12fe
Step #1: Step 15/15 : COPY build.sh $SRC/
Step #1:  ---> 39feb1ec2397
Step #1: Successfully built 39feb1ec2397
Step #1: Successfully tagged gcr.io/oss-fuzz/tpm2-tss:latest
Finished Step #1
Starting Step #2
Step #2: Already have image: gcr.io/oss-fuzz/tpm2-tss
Step #2: ++ tempfile
Step #2: + SRCMAP=/tmp/file5dLQ1o
Step #2: + echo '{}'
Step #2: ++ find /src -name .git -type d
Step #2: + for DOT_GIT_DIR in '$(find $SRC -name ".git" -type d)'
Step #2: ++ dirname /src/tpm2-tss/.git
Step #2: + GIT_DIR=/src/tpm2-tss
Step #2: + cd /src/tpm2-tss
Step #2: ++ git config --get remote.origin.url
Step #2: + GIT_URL=https://github.com/tpm2-software/tpm2-tss
Step #2: ++ git rev-parse HEAD
Step #2: + GIT_REV=010ec6016dd3957d35951cb2a8ebf95658812b13
Step #2: + jq_inplace /tmp/file5dLQ1o '."/src/tpm2-tss" = { type: "git", url: "https://github.com/tpm2-software/tpm2-tss", rev: "010ec6016dd3957d35951cb2a8ebf95658812b13" }'
Step #2: ++ tempfile
Step #2: + F=/tmp/fileXzTuDr
Step #2: + cat /tmp/file5dLQ1o
[...]

PR #1420, opened 2019-05-20
Dockerfile: use tpm2software/tpm2-tss as base image

diabonas:
Using the [organisation-wide Dockerfile](https://github.com/tpm2-software/tpm2-software-container) as a base helps avoiding duplicate efforts in keeping the image up to date, see https://github.com/tpm2-software/tpm2-tss/issues/1419#issuecomment-493987654. This change depends on https://github.com/tpm2-software/tpm2-software-container/pull/7 to configure a more recent clang++ for the fuzzing target to work properly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,48 +1,4 @@
-FROM ubuntu:18.04 AS base
-RUN apt-get update && apt-get install -y \
-    autoconf \
-    autoconf-archive \
-    automake \
-    build-essential \
-    doxygen \
-    gcc \
-    git \
-    libssl-dev \
-    libtool \
-    pkg-config \
-    wget
-
-# OpenSSL
-ARG openssl_name=openssl-1.1.0h
-WORKDIR /tmp
-RUN wget --quiet --show-progress --progress=dot:giga https://www.openssl.org/source/$openssl_name.tar.gz \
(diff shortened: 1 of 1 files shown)


diabonas:
The tpm2software/tpm2-tss Docker image is used by [tpm2-tools](https://github.com/tpm2-software/tpm2-tools/blob/5b96ac3983407fa273b4455a091a399ee15f6b48/.ci/travis.run#L7) and [tpm2-pkcs11](https://github.com/tpm2-software/tpm2-pkcs11/blob/8c9f31fbe73d5d0497b3b17c0f6e7666078bb715/.ci/travis.run#L12), also to [build tpm2-tss](https://github.com/tpm2-software/tpm2-tools/blob/5b96ac3983407fa273b4455a091a399ee15f6b48/.ci/download-deps.sh#L10), so I'd say this is pretty well-tested. For the fuzzing target @pdxjohnny is the expert, everything builds fine on my system using the updated Dockerfile.

I am not aware of any direct users of this Dockerfile. Given that it is currently broken due to the outdated version of Autoconf Archive shipped by the base Ubuntu image I would guess nobody uses it on a regular basis, or else they probably would have complained.

AndreasFuchsSIT:
Point taken, you got me good there... ;-)

PR #1422, opened 2019-05-21
build: use unified TSS2_ESYS_{C,LD}FLAGS_CRYPTO for crypto backend

### LINKED ISSUE #1366: Makefile-test.am: inconsistent usage of $(esyscryCFLAGS) and $(esyscryLDLAGS)
Makefile-test.am contains:
```
if ESYS_OSSL
esyscryCFLAGS = -DOSSL $(LIBSSL_CFLAGS) $(LIBCRYPTO_CFLAGS)
esyscryLDFLAGS = $(LIBSSL_LIBS) $(LIBCRYPTO_LIBS) -ldl
else
if ESYS_GCRYPT
esyscryCFLAGS =
esyscryLDFLAGS = -lgcrypt -ldl
endif
endif
```
For compiling and linking a program using only one of esyscryCFLAGS and esyscryLDFLAGS is suspicious (why shall a program link with -lssl, if the preprocessor cannot find the appropriate headers).

test/unit/esys-context-null uses $(esyscryLDFLAGS) but does not use $(esyscryCFLAGS).

test/integration/esys-policy-ticket.int uses $(esyscryCFLAGS) but not $(esyscryLDFLAGS).

If this is intended, write comments in Makefile-test.am.

diabonas:
Split out the backwards-compatible change a89ce2afae195aa58a0579da7d7019a004ff1edd from #1417  in order to prevent it from getting stale. In theory we could also merge 1a7be9bbaaad53ab770b9a5afe2c252e6a1f149f now because it is backwards-compatible as well, but I think I would prefer correcting the pkg-config files in one go.

- `configure.ac`: Introduce `TSS2_ESYS_CFLAGS_CRYPTO` and `TSS2_ESYS_LDFLAGS_CRYPTO` reflecting the chosen ESAPI cryptographic backend.
- `Makefile.am`: Use the newly introduced variables instead of relying on the `ESYS_OSSL`/`ESYS_GCRYPT` conditionals.
- `Makefile-test.am`: following suggestions from @dilyanpalauzov, remove the `esyscryCFLAGS`/`esyscryLDFLAGS`/`esyscrySRC` and replace them with the newly introduced variables. Also fix #1366 (a missing CFLAGS specification) in the process.

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -43,18 +43,6 @@ test_helper_tpm_dumpstate_LDFLAGS = $(TESTS_LDFLAGS)
 test_helper_tpm_dumpstate_LDADD = $(TESTS_LDADD)
 endif #ENABLE_INTEGRATION
 
-if ESYS_OSSL
-esyscryCFLAGS = -DOSSL $(LIBCRYPTO_CFLAGS)
-esyscryLDFLAGS = $(LIBCRYPTO_LIBS) -ldl
-esyscrySRC = src/tss2-esys/esys_crypto_ossl.c
-else
-if ESYS_GCRYPT
-esyscryCFLAGS =
-esyscryLDFLAGS = -lgcrypt -ldl
-esyscrySRC = src/tss2-esys/esys_crypto_gcrypt.c
-endif
-endif
-
 if UNIT
 TESTS_UNIT  = \
     test/unit/CommonPreparePrologue \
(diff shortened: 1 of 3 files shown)


dilyanpalauzov:
The common pattern is to have _CFAGS and _LDFLAGS at the end of the variable names.  This would allow replacing `PKG_CHECK_MODULES([LIBCRYPTO], [libcrypto])` with `PKG_CHECK_MODULES([TSS2_ESYS_CRYPTO], [libcrypto])` and deleting `TSS2_ESYS_CFLAGS_CRYPTO="$LIBCRYPTO_CFLAGS"`, `TSS2_ESYS_LDFLAGS_CRYPTO="$LIBCRYPTO_LIBS"`.

AndreasFuchsSIT:
Great point. Will you send a PR ?

dilyanpalauzov:
I will not send a PR.

PR #1426, opened 2019-05-22
makefile.am: conditionally append map file

tstruk:
Apped the map files for SYS and ESYS to LDFLAGS only
if HAVE_LD_VERSION_SCRIPT is true.


tstruk:
@AndreasFuchsSIT according to https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html that is on platforms that don’t use GNU LD.

AndreasFuchsSIT:
Should we add
```
else
libfoo_la_LDFLAGS += -export-symbols-regex '^libfoo_.*'
endif
```
then for those, as your link recommends ?

tstruk:
Probably. For now I just wanted to make the sys and esys consistent with the rest.

Issue #1424, opened 2019-05-21
`maintainer-clean` target removes files needed by configure


flihp:
Going by documentation from GNU:
> If the maintainer built it (for instance, a .info file), then maintainer-clean should delete it. However maintainer-clean should not delete anything that needs to exist in order to run ‘./configure && make’. 

Currently running `maintainer-clean` will remove several files installed by `autoreconf` / `aclocal` (in the `bootstrap` script). This runs afoul of the final bit of the quote above which I take to mean we should be able to run `maintainer-clean` followed by `configure` and then `make` without error.

Issue #1348, opened 2019-03-22
Files ignored by “make dist”


dilyanpalauzov:
“make dist” ignores:
```
lib/tss2-esys.def
lib/tss2-mu.def
lib/tss2-tcti-mssim.def
lib/tss2-sys.def
lib/tss2-tcti-tbs.def
.gitattributes
.appveyor.yml
doc/TSS block diagram.png
Dockerfile
.dockerignore
bootstrap
src/tss2-sys/tss2-sys.vcxproj
src/tss2-tcti/tcti-tbs.c
src/tss2-tcti/tss2-tcti-tbs.vcxproj.filters
src/tss2-tcti/tss2-tcti-tbs.vcxproj
src/tss2-tcti/tss2-tcti-mssim.sln
src/tss2-tcti/tcti-tbs.h
src/tss2-tcti/tss2-tcti-mssim.vcxproj.filters
src/tss2-tcti/tss2-tcti-mssim.vcxproj
src/tss2-tcti/tss2-tcti-tbs.sln
src/tss2-esys/tss2-esys.vcxproj.filters
src/tss2-esys/tss2-esys.vcxproj
src/tss2-esys/tss2-esys.sln
src/tss2-mu/tss2-mu.vcxproj
MAINTAINERS
include/tss2/tss2_tcti_tbs.h
script/gen_fuzz.py
script/fuzz-log-compiler.sh
tpm2-tss.sln
.gitignore
.travis.yml
```
Consider tweaking the tests to first call “make dist”, unpack and call the tests on the unpacked code.

flihp:
Thanks for reporting this @dilyanpalauzov. Related PRs are showing up in the thread for the IMHO high-priority items from your list.

AndreasFuchsSIT:
bootstrap or .travis or .gitignore have no place in the dist tarball. I think this list should be stripped at places...

diabonas:
I like the idea of trying to build from a `make dist` release tarball, since there has been at least one instance (#1243) where `make distcheck` didn't catch the distribution not containing all necessary files.

AndreasFuchsSIT:
Sounds neat. PRs welcome.
Should maybe be a env-flag so it can run in parallel to the other tests ?

diabonas:
> Sounds neat. PRs welcome.
> Should maybe be a env-flag so it can run in parallel to the other tests ?

[...]

PR #1427, opened 2019-05-22
Fixes to maintainers clean and dist tar

AndreasFuchsSIT:
- Include block diagram in dist ball
- fix maintainerclean to not kill ./configure
- Add vcs.filters to dist as the final missing files
- Restrict -Werror to non-release versions

Fixes: #1424 
Fixes: #1348


configure.ac
```diff
@@ -252,7 +252,8 @@ AS_IF([test "x$enable_defaultflags" = "xyes"],
       ADD_COMPILER_FLAG([-Wall])
       ADD_COMPILER_FLAG([-Wextra])
       ADD_COMPILER_FLAG([-Wformat-security])
-      ADD_COMPILER_FLAG([-Werror])
+      AS_IF([test "x$ax_is_release" = "xno"],
```
tstruk: Is this really what we want? With the policy == "dash-version" the $ax_is_release will be "no" for X.Y.Z-dev, but it will be "yes" for RCs (X.Y.Z_rcN) and for the final release, which will case the -Werror flag not be added to them.

AndreasFuchsSIT: Hmmm...
What I want is that -Werror is only added to non-releases.
So yes, that's what I want...

Rational being that, if a distro upgrades its compiler, we don't want newly introduce warning to make our releases start failing, due to -Werror.

tstruk: Makes sense. I should read the commit message.

PR #1428, opened 2019-05-23
Esys: Require software random values

AndreasFuchsSIT:
During pk_encrypt we will generate random values for the key as well as the OAEP padding. We need to ensure that we use a CPU-based random function.

### DIFF
--- a/src/tss2-esys/esys_crypto_ossl.c
+++ b/src/tss2-esys/esys_crypto_ossl.c
@@ -545,6 +545,13 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
                            BYTE * out_buffer,
                            size_t * out_size, const char *label)
 {
+    const RAND_METHOD *rand_save = RAND_get_rand_method();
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
Also a backporting candidate unfortunately...

src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -597,6 +604,7 @@ iesys_cryptossl_pk_encrypt(TPM2B_PUBLIC * pub_tpm_key,
     if (1 != RSA_generate_key_ex(rsa_key,
                                  pub_tpm_key->publicArea.parameters.rsaDetail.keyBits,
                                  bne, NULL)) {
+        RAND_set_rand_method(rand_save);
```
tstruk: Don't need that here as this is done in the cleanup block, but we do need this in line 566 before the return.

AndreasFuchsSIT: So true

AndreasFuchsSIT:
Good point... Updated

PR #1433, opened 2019-05-30
added VxWorks support

coolfengyu:
This patch is used to add VxWorks support. With this patch, tpm2-tss could work on VxWorks

### DIFF
--- a/include/tss2/tss2_tcti.h
+++ b/include/tss2/tss2_tcti.h
@@ -43,15 +43,16 @@
 #error Version mismatch among TSS2 header files.
 #endif  /* TSS2_API_VERSION_1_2_1_108 */
 
-#if defined(__linux__) || defined(__unix__) || defined(__APPLE__) || defined (__QNXNTO__)
+#if defined(__linux__) || defined(__unix__) || defined(__APPLE__) || defined (__QNXNTO__) || defined (__VXWORKS__)
+#if defined (__VXWORKS__)
+#include <sys/poll.h>
+#else
 #include <poll.h>
+#endif
 typedef struct pollfd TSS2_TCTI_POLL_HANDLE;
 #elif defined(_WIN32)
 #include <windows.h>
 typedef HANDLE TSS2_TCTI_POLL_HANDLE;
-#elif defined (__VXWORKS__)
-#include <sys/poll.h>
-typedef struct pollfd TSS2_TCTI_POLL_HANDLE;
 #else
 typedef void TSS2_TCTI_POLL_HANDLE;
 #ifndef TSS2_TCTI_SUPPRESS_POLL_WARNINGS
--- a/include/tss2/tss2_tcti.h
+++ b/include/tss2/tss2_tcti.h
@@ -5,6 +5,31 @@
  * Copyright 2015, Andreas Fuchs @ Fraunhofer SIT
  *
  * All rights reserved.
+ *
+ * Copyright (c) 2019, Wind River Systems.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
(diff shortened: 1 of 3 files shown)


src/tss2-esys/esys_tcti_default.c
```diff
@@ -40,6 +64,9 @@ struct {
 #ifdef _WIN32
     { .init = Tss2_Tcti_Tbs_Init, .conf = "",
       .description = "Access to TBS" },
+#elif defined (__VXWORKS__)
+    { .init = Tss2_Tcti_Device_Init, .conf = "/tpm0",
+      .description = "Access to /tpm0" },
```
AndreasFuchsSIT: I assume your intention is to set the default for vxworks in terms of a "distribution"
You should use the `./configure --with-tctidefaultmodule --with-tctidefaultconfig" for this.

include/tss2/tss2_tcti.h
```diff
@@ -5,6 +5,31 @@
  * Copyright 2015, Andreas Fuchs @ Fraunhofer SIT
  *
  * All rights reserved.
+ *
+ * Copyright (c) 2019, Wind River Systems.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
```
AndreasFuchsSIT: You don't need to add the BSD-Clause in again, there is a SPDX-Identifier at the top of the file.
Or do you think that that is not sufficient to reflect your choice of license ?

include/tss2/tss2_tcti.h
```diff
@@ -24,6 +49,9 @@ typedef struct pollfd TSS2_TCTI_POLL_HANDLE;
 #elif defined(_WIN32)
 #include <windows.h>
 typedef HANDLE TSS2_TCTI_POLL_HANDLE;
+#elif defined (__VXWORKS__)
+#include <sys/poll.h>
```
AndreasFuchsSIT: I'm not sure if this shouldn't be done by adding the this directory to the include-dirs ?

Why does vxworks have it at this place anyways ?

src/tss2-tcti/tcti-device.c
```diff
@@ -393,7 +421,11 @@ Tss2_Tcti_Device_Init (
     memset (&tcti_common->header, 0, sizeof (tcti_common->header));
     tcti_common->locality = 3;
 
+#ifdef __VXWORKS__
+    tcti_dev->fd = open (dev_path, O_RDWR, 0666);
```
AndreasFuchsSIT: Since we do not intent to create a file here (no O_CREAT), we should not set a file-access-mode.

include/tss2/tss2_tcti.h
```diff
@@ -24,6 +49,9 @@ typedef struct pollfd TSS2_TCTI_POLL_HANDLE;
...
+typedef struct pollfd TSS2_TCTI_POLL_HANDLE;
```
AndreasFuchsSIT: Since this is the same as for linux, unix, apple and qnx, I'd go ahead and extend that ifdef instead of defining a new one.
Or at least move this elif up there to directly follow the similar ones.

include/tss2/tss2_tcti.h
coolfengyu: Yes, the license needs to be shown here explicit. This is required by Wind River.

src/tss2-tcti/tcti-device.c
coolfengyu: VxWorks will check the flags and mode. Only the fixed value can be accepted by TPM driver.

include/tss2/tss2_tcti.h
coolfengyu: The sys directory includes some header files which are duplicated with other header files in include-dirs. Therefore, sys can't be added to include-dirs.

src/tss2-esys/esys_tcti_default.c
coolfengyu: The build configuration for VxWorks defines NO_DL. If added --with-tctidefaultmodule --with-tctidefaultconfig options, the build will fail.

src/tss2-tcti/tcti-device.c
AndreasFuchsSIT: > Only the fixed value can be accepted by TPM driver.
What do you mean by this ?

Also don't know about 0666 for the mode ? If it's ignored, then 0 would be good enough, right ?

coolfengyu: I updated the code for using macro to replace the magic number.

src/tss2-esys/esys_tcti_default.c
AndreasFuchsSIT: Ok, I looked at the code. What currently happens, if NO_DL is passed is
```
    { .init = Tss2_Tcti_Device_Init, .conf = "/dev/tpmrm0",
      .description = "Access to /dev/tpmrm0" },
    { .init = Tss2_Tcti_Device_Init, .conf = "/dev/tpm0",
.description = "Access to /dev/tpm0" },
```
This should work vor VxWorks, correct ?

include/tss2/tss2_tcti.h
AndreasFuchsSIT: You have 2 `poll.h` files ?
And the function that POSIX.1-2001 defines for `poll.h` you've put into `sys/poll.h` instead ?

src/tss2-tcti/tcti-device.c
```diff
@@ -422,7 +422,7 @@ Tss2_Tcti_Device_Init (
     tcti_common->locality = 3;
 
 #ifdef __VXWORKS__
-    tcti_dev->fd = open (dev_path, O_RDWR, 0666);
+    tcti_dev->fd = open (dev_path, O_RDWR, (S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP));
```
AndreasFuchsSIT: You are missing my point.
If we do not set O_CREAT, then the value that gets passed as `mode` parameter does not do anything, right ?
Thus, I'd put 0 into here.

On a different note: All other posix systems can handle a 3-parameter open as well. So we could get rid of the ifdef here and just use the 3-parameter version for everyone.
Hopefully static code analysis does not dislike this though...

include/tss2/tss2_tcti.h
coolfengyu: Only one poll.h, but there are two errno.h files.

src/tss2-tcti/tcti-device.c
coolfengyu: For file system, mode parameter is only useful when set O_CREAT. For device, it depends on the implementation. TPM driver of VxWorks will check the mode always, but Linux driver doesn’t. Therefore, it works on Linux without mode parameter, but doesn’t work on VxWorks.
The code is more clean if removed the ifdef, but maybe the value of the mode parameter doesn’t satisfy all of the OS. It will introduce other ifdef.

src/tss2-esys/esys_tcti_default.c
coolfengyu: You mean to use /dev/tpmrm0 or /dev/tpm0? It could work but is not a perfect solution.
VxWorks doesn’t support resource manager. /dev/tpmrm0 is nonexistent. ESYS always try to open this device first. It will waste time to open a nonexistent and impact the performance.
VxWorks uses /tpm0 as the device name. If using /dev/tpm0, it needs to change the OS build configuration and impact the compatibility.

Issue #1436, opened 2019-06-04
Support CMAC algorithm


diabonas:
The [TCG Algorithm Registry Family “2.0”, Revision 1.27](https://trustedcomputinggroup.org/wp-content/uploads/TCG-_Algorithm_Registry_Rev_1.27_FinalPublication.pdf) adds `TPM_ALG_CMAC` with value `0x003F`. It needs to be added to [`tss2_tpm2_types.h`](https://github.com/tpm2-software/tpm2-tss/blob/master/include/tss2/tss2_tpm2_types.h) and possibly in other places like [`tpmu-types.c`](https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpmu-types.c) as well.

PR #1437, opened 2019-06-05
types: add CMAC algorithm

diabonas:
The algorithm was added in [revision 1.27 of the TCG Algorithm Registry](https://trustedcomputinggroup.org/wp-content/uploads/TCG-_Algorithm_Registry_Rev_1.27_FinalPublication.pdf). Please advise if there are more places in tpm2-tss where it needs to be added.

Fixes: #1436

### DIFF
--- a/include/tss2/tss2_tpm2_types.h
+++ b/include/tss2/tss2_tpm2_types.h
@@ -100,6 +100,7 @@ typedef UINT16 TPM2_ALG_ID;
 #define TPM2_ALG_ECC                 ((TPM2_ALG_ID) 0x0023)
 #define TPM2_ALG_SYMCIPHER           ((TPM2_ALG_ID) 0x0025)
 #define TPM2_ALG_CAMELLIA            ((TPM2_ALG_ID) 0x0026)
+#define TPM2_ALG_CMAC                ((TPM2_ALG_ID) 0x003F)
 #define TPM2_ALG_CTR                 ((TPM2_ALG_ID) 0x0040)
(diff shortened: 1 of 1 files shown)


diabonas:
I agree it would make more sense if the algorithm was actually used by tpm2-tss. The reason why I would like to see it added nevertheless is that downstream tpm2-tools uses [`tss2_tpm2_types.h`](https://github.com/diabonas/tpm2-tss/blob/master/include/tss2/tss2_tpm2_types.h) to convert the algorithm IDs returned by the TPM to a friendly name, e.g. for `tpm2_getcap`. Since the most recent version 1332 of the [TPM simulator](https://sourceforge.net/projects/ibmswtpm2/) supports CMAC, we need to add it to the list of known algorithms, otherwise an integration test for tpm2-tools fails because the algorithm is reported as "unknown" (https://github.com/tpm2-software/tpm2-tools/issues/1532). Of course I can add the algorithm ID to the corresponding downstream PR https://github.com/tpm2-software/tpm2-tools/pull/1533 instead, but tpm2-tss feels like the correct place to do that.

Do you known when CMAC will be added to the TPM 2.0 specification? It is already in the published revision 1.27 of the TCG Algorithm Registry, so appears to be somewhat official, but is not in the [TCG Trusted Platform Module Library Family 2.0 Level 00 Revision 1.55 – Part 2: Structures](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p55_Part2_Structures_pubrev.pdf#page=43) currently under public review.

AndreasFuchsSIT:
Good reason, point taken. I'll merge... ;-)

PR #1439, opened 2019-06-05
This patch is used to add VxWorks support. 

coolfengyu:
This patch is used to add VxWorks support. With this patch, tpm2-tss could work on VxWorks.

### DIFF
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -2,6 +2,30 @@
 /*
  * Copyright (c) 2015 - 2018 Intel Corporation
  * All rights reserved.
+ * Copyright (c) 2019, Wind River Systems.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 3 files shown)


AndreasFuchsSIT:
Resend of PR #1433
Please do not close and reopen PRs, since this loses all comments and discussions. Please force-push your changes to your branch and the PR will update accordingly automatically.

src/tss2-tcti/tcti-device.c
```diff
@@ -393,7 +421,11 @@ Tss2_Tcti_Device_Init (
     memset (&tcti_common->header, 0, sizeof (tcti_common->header));
     tcti_common->locality = 3;
 
+#ifdef __VXWORKS__
+    tcti_dev->fd = open (dev_path, O_RDWR, (S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP));
```
AndreasFuchsSIT: What's the function of the mode, if O_CREAT is not set ?
IMO, it is being ignored, right ?
If that is the case, then just write `0 /* ignored */` instead of these values.

src/tss2-tcti/tcti-device.c
```diff
@@ -393,7 +421,11 @@ Tss2_Tcti_Device_Init (
     memset (&tcti_common->header, 0, sizeof (tcti_common->header));
     tcti_common->locality = 3;
 
+#ifdef __VXWORKS__
```
AndreasFuchsSIT: I think, instead of having an ifdef, we could just use the 3-parameter version for VxWorks as well as other POSIX systems. We should just add a comment that explains why we use the 3-parameter version.

src/tss2-tcti/tcti-device.c
coolfengyu: For file system, mode parameter is only useful when set O_CREAT. For device, it depends on the implementation. TPM driver of VxWorks will check the mode always, but Linux driver doesn’t. Therefore, it works on Linux without mode parameter, but doesn’t work on VxWorks.
The code is more clean if removed the ifdef, but maybe the value of the mode parameter doesn’t satisfy all of the OS. It will introduce other ifdef.

src/tss2-tcti/tcti-device.c
coolfengyu: except the 3-parameter issue, the flag of open() is different. VxWorks requires O_RDWR without O_NONBLOCK, but other OS require O_RDWR | O_NONBLOCK.

coolfengyu:
Is there any update for this review?

coolfengyu:
Thanks your comment example. I updated the commit. Please review.

PR #1442, opened 2019-06-07
tss2-esys: Support sessions in Esys_TR_FromTPMPublic

williamcroberts:
The getcap command can return a TPM2_HANDLE for the loaded or saved
session however, ESAPI has no way to flush this from the TPM. Modify
Esys_TR_FromTPMPublic() to accept a TPM2_HANDLE representing a session
and create a resource object tagged with IESYSC_SESSION_RSRC.

TODO:
  - [x] add tests
  - [x] add issue tracker

Fixes: https://github.com/tpm2-software/tpm2-tss/issues/1443

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -177,6 +177,7 @@ ESYS_TESTS_INTEGRATION_MANDATORY = \
     test/integration/esys-tpm-tests.int \
     test/integration/esys-tr-fromTpmPublic-key.int \
     test/integration/esys-tr-fromTpmPublic-nv.int \
+    test/integration/esys-tr-fromTpmPublic-session.int \
     test/integration/esys-tr-getName-hierarchy.int \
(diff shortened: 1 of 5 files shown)


AndreasFuchsSIT:
I think it would be better to add a new resource type to the enum for this, since it's a degraded session and practically the only thing we can use it with is FlushContext.

williamcroberts:
> I think it would be better to add a new resource type to the enum for this, since it's a degraded session and practically the only thing we can use it with is FlushContext.

Actually that makes the error code nicely and consitently go to ESYS_TR_BADREFERENCE when used as an actual session handle (with a minor change to the session handle checking code).

williamcroberts:
@AndreasFuchsSIT do you have a release planned with this patch in it?

AndreasFuchsSIT:
@williamcroberts "Mid-July" is the plan, featuring tctildr

PR #1449, opened 2019-06-24
Fix tpm2-abrmd startup failure due to wrong permission of /dev/tpm0

jiazhang0:
The group permission should be also specified.

Signed-off-by: Jia Zhang <zhang.jia@linux.alibaba.com>

### DIFF
--- a/dist/tpm-udev.rules
+++ b/dist/tpm-udev.rules
@@ -1,4 +1,4 @@
 # tpm devices can only be accessed by the tss user but the tss
 # group members can access tpmrm devices
-KERNEL=="tpm[0-9]*", MODE="0660", OWNER="tss"
+KERNEL=="tpm[0-9]*", MODE="0660", OWNER="tss", GROUP="tss"
 KERNEL=="tpmrm[0-9]*", MODE="0660", OWNER="tss", GROUP="tss"


diabonas:
tpm2-abrmd should be [run as the `tss` user](https://github.com/tpm2-software/tpm2-abrmd/blob/master/INSTALL.md#system-user--group) (or root), so if starting it fails with a permission error, something is wrong with your setup (e.g. you might have to specify `--with-systemdsystemunitdir` or `--with-dbuspolicydir` with the correct directory when building tpm2-abrmd).

In general `/dev/tpm[0-9]*` should *not* be accessible by members of the `tss` group: the idea is that you add users to this group that should have access to the TPM through the in-kernel resource manager `/dev/tpmrm[0-9]*`. They shouldn't have direct access to the raw `/dev/tpm[0-9]*` device though in order to prevent them from interfering with the resource manager by exhausting its limited memory resources.

jiazhang0:
@diabonas The story is that the tpm-udev.rules is not installed. So the default permission of /dev/tpm0  is root:root. After installed, the permission of /dev/tpm0 becomes tss:root. Now I see the group tss is not specified in tpm-udev.rules.

jiazhang0:
@diabonas OK. I'm using a 4.9 kernel without tpmrm support. The latter is introduced since 4.14. Looks like we should set the rule of creation of /dev/tpmrm0 before /dev/tpm0?

AndreasFuchsSIT:
So the idea of /dev/tpm0 is that it's a privileged device, because it is singleton open. Thus only the user tss or root may open it; "tss:root".
/dev/tpmrm0 however can be accessed by multiple application simultaniously, thus it also includes group tss.

I guess we should upstream the udev-rules to systemd-udevd at some point, but did not do so yet.

@jiazhang0 Could you explain what you mean my
>   Looks like we should set the rule of creation of /dev/tpmrm0 before /dev/tpm0?

I'll close the PR however, since we really don't want /dev/tpm0 to be group tss.

PR #1454, opened 2019-06-28
Add missing definition of TPMS_TAGGED_POLICY struct

tstruk:
For whatever reson we don't have the definition of the
TPMS_TAGGED_POLICY struct even though its size is used
to calculate the TPM2_MAX_TAGGED_POLICIES value.


flihp:
Just made #1459 to track the overreaction by codecov here. We shouldn't be automatically rejecting PRs for this reason. No clue how to tweak this in codecov though.

PR #1457, opened 2019-07-02
tcti loading & default / NO_DL refactor

flihp:
This is largely in preparation to separate this code into a new library. The NO_DL macro was right in the middle of the interface though. Better to clean that up first, especially since NO_DL wasn't properly integrated into the build. This cleanup removes the need for the NO_DL macro in the source code completely which improves readability to no end.


flihp:
travis builds seem to be failing due to the internet (cloudflare) outage

src/tss2-tcti/tctildr-nodl.c
```diff
@@ -0,0 +1,90 @@
...
+        .init = Tss2_Tcti_Device_Init,
+        .conf = "/dev/tpmrm0",
+        .description = "Access to /dev/tpmrm0",
+    },
+    {
+        .init = Tss2_Tcti_Device_Init,
+        .conf = "/dev/tpm0",
+        .description = "Access to /dev/tpm0",
+    },
+#endif /* TCTI_DEVICE */
```
AndreasFuchsSIT: VxWorks will also need an #ifdef since they export the device under "/tpm0".
See https://github.com/tpm2-software/tpm2-tss/pull/1439/files#diff-e5fc82fbafb52375569e9a8a6c9d603cR67-R69

flihp: > is [9f1bdd1](https://github.com/tpm2-software/tpm2-tss/commit/9f1bdd187c7dbcd1bf4e439f9cf52962483c7468) even needed, if [8d973ab](https://github.com/tpm2-software/tpm2-tss/commit/8d973abcb3fcf446d467861ba918703e5e2152b5) removes the file anyways ?

9f1bdd1 was added to make 8d973ab possible. Couldn't get it to build without that change. If you'd like me to squash them together I'm not against the idea.

flihp: > VxWorks will also need an #ifdef since they export the device under "/tpm0".
> See https://github.com/tpm2-software/tpm2-tss/pull/1439/files#diff-e5fc82fbafb52375569e9a8a6c9d603cR67-R69

That PR hasn't been merged. Are you asking me to do something specific here?

AndreasFuchsSIT: Well, the other PR won't merge ontop of this one and this one will not merge / revert the other one...
So I don't know what to do.
I can merge the other and you rebase and include here ?
Sorry for the inconvenience...

AndreasFuchsSIT:
Yes, I guess, I'd squash 2 and 3.
The result should be an unaltered 3 with 1 as parent, I think ?

src/tss2-tcti/tctildr-nodl.c
flihp: I hadn't been tracking #1433. Saw it get closed, didn't see it reappear in #1439. Since you're asking for input here it seems to me you have reservations about merging #1439? Would you like me to do a review of this PR?

flihp:
I've got a branch with both permutations:
- https://github.com/flihp/tpm2-tss/tree/linux-no_dl-squash (squash 2nd patch into 3rd)
- https://github.com/flihp/tpm2-tss/tree/linux-no_dl-squash-alt (squash 2nd patch into 1st)

I prefer the later primarily because the 1st patch enables use of the NO_DL macro via the `--enable-nodl` and it doesn't make much sense to allow use of this flag if it doesn't work. By combining these two patches we enable easy use of the feature (exposing it in the configure script) and we fix it so that it's usable.

This feels a bit like bike shedding to me though so if you have strong opinions for the other approach I'm happy to go either way. Let me know which you prefer and I'll update this PR.

src/tss2-tcti/tctildr-nodl.c
AndreasFuchsSIT: Ì just re-posted my review with some "hint"...

AndreasFuchsSIT:
I guess that makes sense.

AndreasFuchsSIT:
@flihp Ok, I merged the VxWorks PR now.
Would you rebase and include the __VxWorks__ #ifdef for "/tpm0" to tctildr_nodl ?

Sorry for the mess.

flihp:
Cleaning up after #1439 will be more work than rebasing. I offered to help but since you merged it you get to clean it up :stuck_out_tongue_winking_eye:

AndreasFuchsSIT:
> Cleaning up after #1439 will be more work than rebasing. I offered to help but since you merged it you get to clean it up stuck_out_tongue_winking_eye

I think I don't get what you mean ?

flihp:
nm, not relevant to this PR any longer

PR #1460, opened 2019-07-12
TCTI: Fix the default locality value of tcti-mssim.

JuergenReppSIT:
* The default value should be 0 instead of 3.
* The integration tests relying on locality 3 (extend pcr 17, pcr event pcr 18) were adapted.
* The locality is reset to 0 at the end of the locality integration test.

### DIFF
--- a/src/tss2-tcti/tcti-mssim.c
+++ b/src/tss2-tcti/tcti-mssim.c
@@ -489,7 +489,7 @@ tcti_mssim_init_context_data (
     TSS2_TCTI_SET_LOCALITY (tcti_common) = tcti_mssim_set_locality;
     TSS2_TCTI_MAKE_STICKY (tcti_common) = tcti_make_sticky_not_implemented;
     tcti_common->state = TCTI_STATE_TRANSMIT;
-    tcti_common->locality = 3;
+    tcti_common->locality = 0;
     memset (&tcti_common->header, 0, sizeof (tcti_common->header));
 }
 /*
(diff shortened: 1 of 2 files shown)


AndreasFuchsSIT:
I guess, I'll wait until the distcheck-fix is merged...

flihp:
This was set to locality 3 by default because user space applications are expected to be in locality 3 and the majority (all?) of our use-cases for the simulator (test harness) are userspace applications. Does this change imply that the downstreams need to update their code to set the locality to 3 explicitly?

AndreasFuchsSIT:
IMHO, the linux tpm-driver maps only locality 0 to user-space, correct ?
Thus, all applications developed for a physical TPM rely on locality 0, or am I mistaken ?

AndreasFuchsSIT:
AFAIK, unless you start TXT, Intel systems will disallow access to any locality besides 0.
AMD systems allow access to localities 0-2 or 0-3 (cannot remember)
4 should be TXT/MLE for PCRs 17 and up aka DRTM, but life is a mess... :-(

@JuergenReppSIT please update the commit msg mentioning that locality 0 is the default locality and the only one exposed to userspace under linux.

AndreasFuchsSIT:
I wonder, how locality 3 ended up here in the first place ?!? Maybe related to the tpmclient-tests author ?

flihp:
Given the relationship between locality 3 and TXT / DRTM PCRs I'd say you're right @AndreasFuchsSIT

flihp:
Thanks for updating the commit message. I did a quick build / test cycle with the tabrmd and tools to hopefully head off any downstream issues. Both worked as expected.

PR #1461, opened 2019-07-12
Fixed several build breaks

LarryOsterman:


flihp:
Please take a look at our contributing guidelines in CONTRIBUTING.md. A few of the obvious problems:
- This PR checks in files that are generated and not version controlled.
- The merge commits shouldn't be in this PR.
- No commit message.

PR #1445, opened 2019-06-11
tss2-rc: add TSS2_RC decoding library

williamcroberts:
This library implements the TCG spec:
  -https://trustedcomputinggroup.org/wp-content/uploads/TCG_TSS_RC_v1p0_r12_pub.pdf

This library allows for the conversion of TSS2_RC's from numerical form
to a standardized human readable text string.

Fixes: #1444

Signed-off-by: William Roberts <william.c.roberts@intel.com>

PR #1470, opened 2019-07-19
libtss2-tctildr

flihp:
This library is intended to replace downstream implementations of the TCTI dynamic loading protocol. Should save us LOCs and maintenance burden downstream. Will also save new projects the wasted time / energy of having to implement this stuff themselves.


src/tss2-tcti/tctildr-dl.c
```diff
@@ -32,25 +35,90 @@ struct {
         .description = "Access libtss2-tcti-default.so",
     },
     {
-        .file = "libtss2-tcti-tabrmd.so",
+        .file = "libtss2-tcti-tabrmd.so.0",
         .description = "Access libtss2-tcti-tabrmd.so",
     },
     {
-        .file = "libtss2-tcti-device.so",
+        .file = "libtss2-tcti-device.so.0",
```
pdxjohnny: I'm sure there's a reason, but why prepend the `.0`?

flihp: I tried to covey this in the commit message but may not have been as clear as possible

> When dynamically loading TCTIs the ESYS default selection and search
> logic tries known TCTI library names with the '.so' suffix. This
> works on systems where the library has been installed from source or
> when development packages are installed. Packages for the TCTI (non
> 'dev' packages) however will provide a library with the library version
> in the suffix (eg '.so.0') causing the existing approach to fail for
> the common case. This commit replaces '.so' suffix with '.so.0' as a
> fix. Both the tools and the tabrmd worked through these issues some
> time ago with help from some of the distro packagers (see 2296d48a from
> tabrmd).

Devs from Suse explained this to me here: https://github.com/tpm2-software/tpm2-abrmd/issues/486

src/tss2-tcti/tctildr.c
```diff
@@ -85,3 +97,365 @@ tcti_from_info(TSS2_TCTI_INFO_FUNC infof,
...
+ * - "tcti-name", "tcti-name:" -> tcti-name copied to 'name', 'conf'
+ *   unchanged
+ * - ":tcti-conf" -> tcti-conf copied to 'conf', 'name' unchanged
+ * - "tcti-name:tcti-conf" - "tcti-name" copied to 'name,', "tcti-conf"
+ *   copied to 'conf'
+ */
+TSS2_RC
+tctildr_conf_parse (const char *name_conf,
+                    char *name,
+                    char *conf)
```
pdxjohnny: My only concern is that since `tctildr_conf_parse` requires the caller to pass already allocated `name` and `conf`, we can't stop them from allocating something which would be to small, we can tell them to, but they may not listen. So I think that this is an appropriate place to add `strn`-ish arguments where we make the caller specify `name_length` and `conf_length`. Granted, this isn't an exported function, but could save some maintainer years down the road some headache if they try to pass in buffers that are too small.

src/tss2-tcti/tctildr-dl.c
flihp: RedHat dev patched this in the tabrmd for me: https://github.com/tpm2-software/tpm2-abrmd/pull/489
There are similar patches in the tools as well.

pdxjohnny: Ah thank you. I failed to read the commit messages

src/tss2-tcti/tctildr.c
pdxjohnny: But this is very take it or leave it, since its an internal function

src/tss2-tcti/tctildr-dl.c
flihp: SHAME! :stuck_out_tongue_closed_eyes:

include/tss2/tss2_tctildr.h
```diff
@@ -0,0 +1,21 @@
+/*
+ * SPDX-License-Identifier: BSD-2
```
AndreasFuchsSIT: The correct SPDX should be `BSD-2-Clause`

Also all others are a little different, but that's minor and I don't care too much...
```
/***********************************************************************;
 * Copyright (c) 2015-2018, Intel Corporation
 *
 * All rights reserved.
 ***********************************************************************/
```
But the la

src/tss2-tcti/tctildr-nodl.c
```diff
@@ -49,37 +49,53 @@
...
     TSS2_TCTI_INIT_FUNC init;
     char *conf;
     char *description;
 } tctis [] = {
 #ifdef _WIN32
     {
-        .name = "libtss2-tcti-tbs.so",
+        .names = {
+            "libtss2-tcti-tbs.so.0",
+            "libtss2-tcti-tbs.so",
```
AndreasFuchsSIT: Isn't this superflous ?
If tbs is a name and the `handle_from_name()` will expand this to the .so.0 and .so files anyways, why not provide only `tbs` as the only option ?

src/tss2-tcti/tctildr.c
flihp: I had a version that took the length but I went the route of less code, more assumptions. Since it's an internal function (as you point out) I figured the assumptions were reasonable. Would prefer to "leave it" as is.

include/tss2/tss2_tctildr.h
flihp: Good catch w/r/t the SPDX identifier. Just pushed fixes for the one you caught and 2 others.

flihp: w/r/t the file header / comment block: I'm trying to cut down on the number of characters here. We should be working to eliminate these blocks of comments completely (they have no meaning according to @williamcroberts  in #1445). Not trying to do that here but I drop those long strings of '*'s when I remembered to. My preference is to leave this as-is.

src/tss2-tcti/tctildr-nodl.c
flihp: This code module is only built when `NO_DL` is defined. `handle_from_hame` is only defined when `NO_DL` is undefined.

flihp:
Something went weird when I rebased this yesterday. Codecov now reports test coverage getting all screwed up with the tctildr.c module.

AndreasFuchsSIT:
Seems like `tctildr_conf_parse` allways returns on `combined_length > PATH_MAX - 1` and no GetInfo and no tctildr_transmit/receive are called.

But if tctildr_conf_parse errors out, I guess it makes sense that nothing else is visited...

What I find more concerning is that the tests should probably fail, if the conf_parse only ever fails. At leasts that's how I read the tests...

include/tss2/tss2_tctildr.h
williamcroberts: Yeah just kill these, no need to introduce these if we don't have to.

pdxjohnny: Just to be clear for when I'm doing the python bindings, we don't need these right (I'm gathering from these comments), but why the shift? (I've always thought it was re-redundant but just curious)

flihp:
Not sure I follow @AndreasFuchsSIT. Is there a specific change you're asking for?

flihp:
W/r/t code coverage it looks like codecov can't cope with multiple unit tests testing a single file. I flipped a few switches in `.travis.yml` to get the codecov script to dump data and it looks like only the metrics for the last unit test to touch a given file are uploaded. Since I've spread testing of `src/tss2-tcti/tctildr.c` over 3 unit tests only the metrics for the last one that's run get uploaded (see https://travis-ci.org/flihp/tpm2-tss/jobs/562678076#L8244 and https://travis-ci.org/flihp/tpm2-tss/jobs/562678076#L8297). Naturally the `check-code-coverage` build target works as expected reporting 98.3% coverage while Codecov reports 31.3%.

I'm out of ideas so combining all 3 of the unit tests that cover various parts of the `tctildr.c` module seems like the only course of action :frowning_face:

AndreasFuchsSIT:
Ok, you just gave the the explaination I was missing; aka codecov is dump... :-(

I was concerned about all the unvisited code and starting to wonder how that came whilst tests not failing, but I guess that's cleared up now...

So I assume I can merge, right ?

flihp:
I'd feel more comfortable having this issue with codecov sorted first. Let me grind this down a bit. Will hopefully have a better coverage report when you come online tomorrow.

flihp:
I've added a single commit combining two (of three) of the unit tests that cover the `src/tss2-tcti/tctildr.c` module. Codecov checks are no longer listed in the checks, though the comment added by codecov has not been updated. The report can be found here: https://codecov.io/gh/tpm2-software/tpm2-tss/pull/1470 specifically the change is relative to https://codecov.io/gh/tpm2-software/tpm2-tss/pull/1470/src/src/tss2-tcti/tctildr.c

I've put off any rebasing till the codecov / coverage issue is resolved. If you're OK with the above, let me know how you want the commits arranged and I'll do the rebase.

AndreasFuchsSIT:
However you like to.
I'd just sQuash 52a78be756787d10c02c0901cec7963dc21387f2 and 79c9e7a6c3f62980ec82c2f42176198649fdb5c8 because it feels like a weird intermediate state, but it's also more a Question than a reQuest...

Just give the the go and I'll hit the merge...

P.S. Sorry for the capital Qs but my firefox does not accept Q in lowercase anymore; all other programs do though... :rofl:

include/tss2/tss2_tctildr.h
williamcroberts: You actually don't need any of it, as the LICENSE file indicates the LICENSE for all files. However, having the SPDX identifier is no big deal, its the copyright lines you don't want to add.

flihp:
Alright. Hopefully to answer the question: I partitioned this work on a few criteria in what I think is a priority order
1. generally change one thing at a time (defining this can be hard)
2. size (smaller is better but must be large enough to be "significant")
3. testability / readability / reviewability (a function of size & complexity)

The reason for keeping 52a78be separate from 79c9e7a is that 52a78be is largely concerned with breaking a piece of logic off of the esys implementation and turning it into a library. Almost nothing / very little about how the code behaves was changed (I hope). This makes testing easier / more localized (renaming function calls, changing interfaces, very few functional changes). My goal was 95% coverage for this stuff too so there's a lot of code in the test directory.

79c9e7a on the other hand is a functional change in the code. Granted it's not a big diff code wise but that's because it's not a big renaming / autotools-fu exercise like 52a78be. Having this and the related test cases separate from a diff that's all file renaming and build stuff made managing the tree easier.

This sort of partitioning is why I pushed build changes / fixes for the `NO_DL` stuff ahead of this work. Once that was refactored out of the build changes for libtss2-tctildr I figured it was easier to review as a separate commit / PR. I figured this would make it easier to review but if it looks awkward from your point of view I may have got that part wrong :confused:

If this answers the question feel free to push the merge button.

P.S. I hope your firefox feels better soon :smirk:

PR #1480, opened 2019-07-26
tctildr: downgrade missing TCTI library error to debug message

diabonas:
If `Esys_Initialize` or `Tss2_TctiLdr_Initialize` attempt to get the default TCTI using [`tctildr_get_default`](https://github.com/tpm2-software/tpm2-tss/blob/d650eda45813424823f57ea0df43544c9995f57e/src/tss2-tcti/tctildr-dl.c#L213), errors of the form
```
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:116:handle_from_name() Failed to load TCTI for name "libtss2-tcti-default.so": libtss2-tcti-libtss2-tcti-default.so.so: cannot open shared object file: No such file or directory
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:116:handle_from_name() Failed to load TCTI for name "libtss2-tcti-tabrmd.so.0": libtss2-tcti-libtss2-tcti-tabrmd.so.0.so: cannot open shared object file: No such file or directory
```
are produced until the correct library is found. This PR downgrades these non-errors to debug messages to hide them from the user by default while keeping important messages like "No standard TCTI could be loaded" visible.

Fixes: #1482

### DIFF
--- a/src/tss2-tcti/tctildr-dl.c
+++ b/src/tss2-tcti/tctildr-dl.c
@@ -113,7 +113,7 @@ handle_from_name(const char *file,
     }
     *handle = dlopen(file_xfrm, RTLD_NOW);
     if (*handle == NULL) {
-        LOG_ERROR("Failed to load TCTI for name \"%s\": %s", file, dlerror());
+        LOG_DEBUG("Failed to load TCTI for name \"%s\": %s", file, dlerror());
         return TSS2_TCTI_RC_NOT_SUPPORTED;
     }
 


src/tss2-tcti/tctildr-dl.c
```diff
@@ -113,7 +113,7 @@ handle_from_name(const char *file,
     }
     *handle = dlopen(file_xfrm, RTLD_NOW);
     if (*handle == NULL) {
-        LOG_ERROR("Failed to load TCTI for name \"%s\": %s", file, dlerror());
+        LOG_WARNING("Failed to load TCTI for name \"%s\": %s", file, dlerror());
```
williamcroberts: Can we downgrade this even further? Sucks that their is no way to get something like errno to check for ENOENT....

diabonas: Done, it is now `LOG_DEBUG` like the other `dlerror()` messages in this function and therefore hidden from the user by default.

williamcroberts:
Fixes: #1482

williamcroberts:
@diabonas 
Can you add
```
Fixes: #1482
```
to your commit message or edit the first message on the PR so Github picks it up and closes it on merge?

diabonas:
Done!

PR #1483, opened 2019-07-29
tcti: fix issues found by coverity scan

tstruk:
Fix two issues found by coverity scan:

** CID 1484591:  Null pointer dereferences  (NULL_RETURNS)
/src/tss2-tcti/tctildr.c: 149 in tctildr_conf_parse()

** CID 1484590:  Memory - illegal accesses  (USE_AFTER_FREE)
/src/tss2-tcti/tctildr.c: 369 in Tss2_TctiLdr_GetInfo()

### DIFF
--- a/src/tss2-tcti/tctildr.c
+++ b/src/tss2-tcti/tctildr.c
@@ -146,7 +146,7 @@ tctildr_conf_parse (const char *name_conf,
         name [name_length] = '\0';
         LOG_DEBUG ("TCTI name: \"%s\"", name);
     }
-    if (conf != NULL && split [1] != '\0') {
+    if (conf != NULL && split && split [1] != '\0') {
         /* conf is more than empty string */
         strcpy (conf, &split [1]);
         LOG_DEBUG ("TCTI conf: \"%s\"", conf);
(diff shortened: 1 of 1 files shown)


src/tss2-tcti/tctildr.c
```diff
@@ -361,12 +361,13 @@ Tss2_TctiLdr_GetInfo (const char *name,
...
     if (rc != TSS2_RC_SUCCESS) {
         free (info_tmp);
+        info_tmp = NULL;
         goto out;
     }
     info_tmp->init = NULL;
 out:
     tctildr_finalize_data (&data);
-    *info = info_tmp;
+    *info = info_tmp ? info_tmp : NULL;
```
AndreasFuchsSIT: This does not make too much sense, does it ?
If NULL then NULL
If not NULL then the value ? ;-)
I think the fix in L364 is sufficient...

tstruk: I agree that this check is redundant, but that's the only way to make the coverity happy.

AndreasFuchsSIT: No, L364 is what fixes coverity. L370 does nothing, IMHO

tstruk: updated

PR #1486, opened 2019-07-30
Install hook

tstruk:
Add make install hook

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -403,7 +403,7 @@ man/man7/%.7 : man/%.7.in $(srcdir)/man/man-postlude.troff
 	$(AM_V_GEN)$(call make_man,$@,$<,$(srcdir)/man/man-postlude.troff)
 
 install-man: install-man3 install-man7
-	$(LN_S) Tss2_TctiLdr_Initialize.3 $(DESTDIR)$(man3dir)/Tss2_TctiLdr_Initialize_Ex.3
+	$(LN_S) -f Tss2_TctiLdr_Initialize.3 $(DESTDIR)$(man3dir)/Tss2_TctiLdr_Initialize_Ex.3
 
 EXTRA_DIST += dist/tpm-udev.rules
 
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -415,6 +415,14 @@ uninstall-local:
 	-rm $(DESTDIR)$(udevrulesdir)/$(udevrulesprefix)tpm-udev.rules
 endif
 
+install-exec-hook:
+	cd $(DESTDIR)$(libdir) && \
+		$(LN_S) libtss2-tcti-device.so libtss2-tcti-default.so
```
diabonas: It would be desirable to use `$(LN_S) -f`, see https://github.com/tpm2-software/tpm2-tss-engine/pull/63:
> Additionally I noticed that you cannot `make install` twice to the same directory because creating the symlink [...] a second time fails. Adding `--force` overwrites the existing symlink (exactly like the standard Makefile installation procedure overwrites existing files), making updating an existing installation possible.

This also concerns installation of the [`Tss2_TctiLdr_Initialize_Ex` man page in l. 406](https://github.com/tpm2-software/tpm2-tss/blob/8506c7d155ac16497dd378b61227223767b6cdc9/Makefile.am#L406).

Makefile.am
```diff
@@ -415,6 +415,13 @@ uninstall-local:
...
+
+uninstall-hook:
+	unlink $(DESTDIR)$(libdir)/libtss2-tcti-default.so
```
diabonas: Note that despite the name, `unlink` will remove `libtss2-tcti-default.so` regardless of whether it is a symbolic link or not. In tpm2-tss-engine there is an additional sanity check in the form of [`[ -L file ] && rm -f file`](https://github.com/tpm2-software/tpm2-tss-engine/blob/d79620378b681ecc2b6b63745e61a95ebc9b804e/Makefile.am#L92) to avoid deleting real files. In general I think using `rm -f` instead of `unlink` is the more standard way to do this (but both are POSIX, so should be portable).

tstruk: "[ -L file ] && rm -f file" produces uninstall-hook Error if the file/link doesn't exist. I will just do rm -f file.

diabonas: Ah yes, if the file doesn't exist, that would bail out because the last command to be executed is `[ -L file ]`, which exits with a non-zero exit status. You can do
```
cd $(DESTDIR)$(libdir) && [ -L libtss2-tcti-default.so ] && rm -f libtss2-tcti-default.so || true
```
to continue uninstalling even if the file is not found.

PR #1489, opened 2019-07-31
test: Skip sapi-policy-authorizeNV if Encrypt/Decrypt command is missing

webmeister:
Signed-off-by: Alexander Steffen <Alexander.Steffen@infineon.com>

### DIFF
--- a/test/integration/sapi-policy-authorizeNV.int.c
+++ b/test/integration/sapi-policy-authorizeNV.int.c
@@ -20,6 +20,7 @@
 
 #define LOGMODULE test
 #include "util/log.h"
+#include "test-esapi.h"
 #include "test.h"
 #include "sapi-util.h"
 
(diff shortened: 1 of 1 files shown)


test/integration/sapi-policy-authorizeNV.int.c
```diff
@@ -228,6 +229,30 @@ test_invoke (TSS2_SYS_CONTEXT *sapi_context)
...
+    if (rc == TPM2_RC_COMMAND_CODE) {
+        LOG_WARNING("Encrypt/Decrypt not supported by TPM");
+        rc = Tss2_Sys_NV_UndefineSpace(sapi_context,
+                                       TPM2_RH_OWNER,
+                                       nv_index,
+                                       &cmd_auth,
+                                       &rsp_auth);
+        if (rc != TSS2_RC_SUCCESS) {
+            LOG_ERROR("Tss2_Sys_NV_UndefineSpace failed: 0x%"PRIx32, rc);
+            return 99; /* fatal error */
```
tstruk: I wouldn't think this is an hard error? Perhaps return EXIT_FAILURE would be better here and below. It will also be be more consistent with the error codes returned elsewhere in this file. Later we should also change all the exit(1); to return EXIT_FAILURE;

webmeister: This is what for example the sapi-encrypt-decrypt tests do in the same place, I just tried to be consistent with that. Should all of those be changed as well?

tstruk: Let's keep these consistent within the same file at least.  The sapi-encrypt-decrypt test can be changed later.

PR #1492, opened 2019-08-14
Fixed two TCG spec violations in the TSS library.

LordOfDorks:
Each description is listing the corresponding violation of the Rev 1.38 specs.

### DIFF
--- a/test/integration/esys-commit.int.c
+++ b/test/integration/esys-commit.int.c
@@ -13,6 +13,7 @@
 #include "tss2_esys.h"
 
 #include "esys_iutil.h"
+#include "test-esapi.h"
 #define LOGMODULE test
 #include "util/log.h"
 #include "util/aux_util.h"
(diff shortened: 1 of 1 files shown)


tstruk:
@LordOfDorks thanks for the patch.
The fact the we don't support the latest addition to the spec doesn't mean that we violate it ;)
We strive for both API and ABI backwards compatibility within the same major number releases, as described in [1]. I'm not sure if for this change we will need to bump the major number to 3.x.y

[1] https://github.com/tpm2-software/tpm2-tss/blob/master/RELEASE.md#version-numbers

AndreasFuchsSIT:
Having though about it: The ReadClock update to 1.38 can be handled as a two-stage approach. We can make the API/ABI compatible change now (setting the bits) so it is usable in the async scheme and from Esys.
We delay the API/ABI incompatible change for 1 release and give some deprecation warning in the next ?

AndreasFuchsSIT:
Regarding the hmac-key calculation: Could you provide a test-case / seQuence that demonstrates the bug? Then we could add this test case to the integration test suite.

AndreasFuchsSIT:
@LordOfDorks you just added two unrelated commits to this PR. Also they contain no commit message and not code comments.
Would you mind structuring this a bit into separate PRs with commit messages and more code comments ? Thanks!

PR #1494, opened 2019-08-17
esys: Fix HMAC generation for policy sessions

jeffpferreira:
Policy sessions which do not include TPM2_PolicyAuthValue should not include
authValue in their HMAC computation.

Signed-off-by: Jeffrey Ferreira <jeffpferreira@gmail.com>

### DIFF
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -1019,11 +1019,14 @@ iesys_compute_session_value(RSRC_NODE_T * session,
     /* Then if we are a bound session, the auth value is not appended to the end
        of the session value for HMAC computation. The size of the key will not be
        increased.*/
-    if (iesys_is_object_bound(name, auth_value,
-                              session) &&
-        /* type_policy_session set to POLICY_AUTH by command PolicyAuthValue */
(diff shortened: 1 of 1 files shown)


jeffpferreira:
According to Part 1, 19.6.5 HMAC Computation, authValue Note 1: 

"For policy sessions, the authValue is not included in the HMAC calculation unless the policy session included TPM2_PolicyAuthValue() and it was not superseded by TPM2_PolicyPassword()."

Also in Part 1, 19.6.15 No HMAC Authorization:

"For a policy session, two situations will result in authValue being an Empty Buffer:
1) the authValue of the authorized object is an EmptyAuth, or
2) the policy does not use the authValue of the object (that is, the evaluated policy does not contain TPM2_PolicyAuthValue())"

jeffpferreira:
Does anyone have any thoughts on this?

AndreasFuchsSIT:
Sorry, busy times here...

What you provide seems to make sense... I'm a bit confused that we had a bug in this so far.
Could you maybe add a test to the test-suit that fails without but succeeds with your patch ?

That would help to understand the concrete issue and prevent this bug in the future.

PR #1496, opened 2019-08-22
Add the enable-asan configure option

dlrobertson:
Add the `enable-asan` configure option to allow building with `-fsanitize=address`.
I also fixed a simple leak in the unit tests detected by `LeakSanitizer`. There were a
few more memory errors that were detected with `enable-asan`, but I'd need a bit
more time to figure out the cause of them.

Comments and critiques are welcomed! :smile:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -201,6 +201,29 @@ AS_IF([test "x$enable_integration" = "xyes"],
        AC_SUBST([ENABLE_INTEGRATION], [$enable_integration])])
 AM_CONDITIONAL([ENABLE_INTEGRATION],[test "x$enable_integration" = "xyes"])
 #
+# sanitizer compiler flags
+#
+AC_ARG_WITH([sanitizer],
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -201,6 +201,18 @@ AS_IF([test "x$enable_integration" = "xyes"],
        AC_SUBST([ENABLE_INTEGRATION], [$enable_integration])])
 AM_CONDITIONAL([ENABLE_INTEGRATION],[test "x$enable_integration" = "xyes"])
 #
+# asan compiler flags
+#
+AC_ARG_ENABLE([asan],
```
dlrobertson: Note: Might be better as a `with` configure argument so that a user could choose `(address, memory, thread, etc.)

AndreasFuchsSIT:
Seems to uncover some issues, so I like it.
I guess we want the fixes in the codebase first, before adding the scanners, right ?

PR #1501, opened 2019-08-26
Fix typos in TPM2_PT_ type names

tstruk:
Some TPM2_PT_ types have invalid names TPM2_PT_TPM2_ (doube TPM2).
Fix the names, but keep the old macros, just mark the old macros as deprecated.


AndreasFuchsSIT:
Hmmm... the problem with `GCC warn` is that it's not affected by `-Wno-deprecation`
I guess we need a third option...

tstruk:
Superseeded by https://github.com/tpm2-software/tpm2-tss/pull/1503

PR #1503, opened 2019-08-28
types: fix misspelling of some macros

### LINKED ISSUE #1500: tss2_tpm2_types.h incorrectly defines some TPM2_PT_FOO values as TPM2_PT_TPM2_FOO
```
$ git grep -o 'TPM2_PT_TPM2_[A-Z_]*'
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_HR_TRANSIENT_MIN
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_HR_PERSISTENT_MIN
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_HR_NV_INDEX
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_HR_TRANSIENT_AVAIL
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_HR_PERSISTENT
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_HR_PERSISTENT_AVAIL
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_PCR_FIRST
include/tss2/tss2_tpm2_types.h:TPM2_PT_TPM2_PCR_LAST
```

All of the `TPM2_PT_TPM2_` prefixes here should be just `TPM2_PT_`. The string "TPM2_PT_TPM2" does not appear within the [TCG TSS 2.0 Overview and Common Structures Specification](https://trustedcomputinggroup.org/wp-content/uploads/TSS_Overview_Common_Structures_Version-0.9_Revision-03_Review_030918.pdf).

Related: tpm2-software/tpm2-tools#1689.

mdempsky:
It seems like these were generated by a scraper tool that goofed up on
these definitions.

Note that this change breaks backwards compatibility with downstream
applications that depended on the misspelled names. Those applications
shouldn't be relying on standard headers providing non-standard
definitions, and tpm2-tss as an implementation of a standard shouldn't
be providing non-standard behavior either.

If applications need to use these definitions and need to build
against an old version of tpm2-tss that lacks the definitions, they
should add their own compatibility macros. For example, see:
https://github.com/tpm2-software/tpm2-tools/pull/1719

Fixes #1500.

Signed-off-by: Matthew Dempsky <matthew@dempsky.org>

### DIFF
--- a/include/tss2/tss2_tpm2_types.h
+++ b/include/tss2/tss2_tpm2_types.h
@@ -482,8 +482,8 @@ typedef UINT32 TPM2_PT;
 #define TPM2_PT_FIRMWARE_VERSION_1       ((TPM2_PT) (TPM2_PT_FIXED + 11)) /* the most significant 32 bits of a TPM vendor-specific value indicating the version number of the firmware. See 10.12.2 and 10.12.8. */
 #define TPM2_PT_FIRMWARE_VERSION_2       ((TPM2_PT) (TPM2_PT_FIXED + 12)) /* the least significant 32 bits of a TPM vendor-specific value indicating the version number of the firmware. See 10.12.2 and 10.12.8. */
 #define TPM2_PT_INPUT_BUFFER             ((TPM2_PT) (TPM2_PT_FIXED + 13)) /* the maximum size of a parameter typically a TPM2B_MAX_BUFFER */
-#define TPM2_PT_TPM2_HR_TRANSIENT_MIN    ((TPM2_PT) (TPM2_PT_FIXED + 14)) /* the minimum number of transient objects that can be held in TPM RAM. NOTE This minimum shall be no less than the minimum value required by the platforms-pecific specification to which the TPM is built. */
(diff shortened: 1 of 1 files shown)


mdempsky:
@tstruk Ping.

tstruk:
@mdempsky wouldn't it make more sense to have this compatibility layer https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_getcap.c#L20 in TSS instead of in every downstream project that used the misspelled types?

mdempsky:
@tstruk I don't think so.

tpm2-tss is just one particular TSS implementation, and applications should want to be portable to any conforming TSS implementation. Other TSS implementations won't have these compatibility macros, so applications will need to provide compatibility themselves anyway.

tstruk:
@mdempsky but other TSS implementations don't have these silly typoes ;)

mdempsky:
> but other TSS implementations don't have these silly typoes 

That's my point.

PR #1505, opened 2019-08-29
RFC: add sphinx build for Read the Docs

joholl:
For easier access, the doxygen documentation could be hosted (and referenced by `README.md`).  Readthedocs is a really cool site but it does not support doxygen (only [sphinx](https://www.sphinx-doc.org)). Howver, you can create an empty sphinx project, make it generate the doxygen doc and include the html files.

I have an example hosted at [tpm2-tss.readthedocs.io](https://tpm2-tss.readthedocs.io/en/latest/). The build is completely automatic and is triggered by pushing a commit to a specific branch.

What do you think?

Note: Unfortunately the doxygen doc cannot be generated by make (readthedocs does not have autoconf). Instead, I had to convert `Doxyfile.in` to `Doxyfile` using `sed` and then call doxygen manually (see [here](https://github.com/tpm2-software/tpm2-tss/compare/master...joholl:enable_readthedocs?expand=1#diff-5820bd443376e6f878b54152642ea8aaR26)). Currently the doxygen version is `master`. The readthedocs version is `latest` since `master` cannot be set to a custom branch. We might change that later.


joholl:
Travis is failing at one build because `apt-get install` failed. Not a bug.

tstruk:
We probably need to add index.rst or README.rst that would point to https://tpm2-tss.readthedocs.io/en/latest/

joholl:
I'd rather suggest hosting only the doxygen doc on Read the Docs. In the end, [tpm2-tss.readthedocs.io](https://tpm2-tss.readthedocs.io) will point to doxygen's `index.html` directly.

As I said, we probably want to have Read the Docs build the documentation for master which I cannot do at the moment since I am working on a feature branch. However, you can set the root page to whatever branch you wish (including master and latest which is an arbitrary branch).

joholl:
At the moment an empty `index.rst` is created which does not come into account because the `index.html` by doxygen is used.

joholl:
It's been a week now. My suggestion of going forward with this is first merging this PR and then setting up the Read the Docs account.

If you want, we can use my Read the Docs account and switch to master once you merged. Alternatively I can delete my account and release the now occupied domain *tpm2-tss.readthedocs.io* for you.

tstruk:
> If you want, we can use my Read the Docs account and switch to master once you merged. Alternatively I can delete my account and release the now occupied domain _tpm2-tss.readthedocs.io_ for you.

I would prefer the later. Thanks

joholl:
Done. If you have problems with setting up the account just let me know.

tstruk:
The new tpm2-tss.readthedocs.io created and looks good. Thanks

PR #1506, opened 2019-08-29
travis: exclude gcc run with SCANBUILD=yes

williamcroberts:
Rather than overriding CC with clang when compiler is gcc, just
exclude it from the build matrix. This will reduce the build
count by one.

Details can be found here:
  - https://docs.travis-ci.com/user/build-matrix/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ cache: ccache
 
 env:
   matrix:
-  - SCANBUILD=yes CC=clang WITH_TCTI_ASYNC=yes WITH_TCTI_PARTIAL=yes WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_1_0-stable
   - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_0_2-stable
   - WITH_CRYPTO=ossl OPENSSL_BRANCH=OpenSSL_1_1_0-stable
   - WITH_CRYPTO=gcrypt OPENSSL_BRANCH=NONE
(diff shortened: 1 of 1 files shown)


williamcroberts:
Didn't work... hrmmm

williamcroberts:
There's issues in the travis file:
travis lint
```
[x] in matrix.include section: unexpected key script, dropping
[x] in addons.coverity_scan.project section: unexpected key description, dropping
[x] in addons section: unexpected key apt, dropping
```
Ill start by fixing these

williamcroberts:
> There's issues in the travis file:
> travis lint
> 
> ```
> [x] in matrix.include section: unexpected key script, dropping
> [x] in addons.coverity_scan.project section: unexpected key description, dropping
> [x] in addons section: unexpected key apt, dropping
> ```
> 
> Ill start by fixing these

Looks like the linter is pretty much worthless:
https://github.com/travis-ci/travis.rb/issues/376

williamcroberts:
I'm playing with it over in the tools, it looks like exclude on env vars has issues with multiple lines:
https://travis-ci.org/tpm2-software/tpm2-tools/builds/578510247

If I do something like:
```yaml
env:
  - FOO=BAR
  - DOCKER_TAG=ubuntu-16.04 ENABLE_COVERAGE=true
  - DOCKER_TAG=ubuntu-18.04

# only one build should report coverage stats
matrix:
  exclude:
  - env: FOO=BAR
```
It works.... but If I try to exclude on ENABLE_COVERAGE=true, it doesn't work...

Either I need ENABLE_COVERAGE=false OR move each var assignment to its own line.

williamcroberts:
Ahh the whole env string needs to match exactly. I think flipping the logic is the better way here, generate all the general purpose builds with the matrix and then enable the specific feature via the include directive.

williamcroberts:
Hooray, thins looks right. 14 builds. 2 compilers X 6 env = 12 + 2 include builds = 14.

williamcroberts:
@tstruk and love on this?

PR #1510, opened 2019-09-04
ci: convert to using docker builds

williamcroberts:
This will likely be broken and in need of care. This should be a good
starting point at increasing distro covereage. An additional benefit,
if we go to a common build ci system we can coalesce the CI project into
its own repo and have each project just set variables in .travis.yml or
travis itself.

To get code coverage right, this will depends on:
  -https://github.com/tpm2-software/tpm2-tss/pull/1506

Increase test coverage by switching to a docker based build runtime
where we can test different distros like Fedora, Ubuntu and OpenSuse.
This will also drop the need on the OpenSSL wget, as the mix of distros
covered provides testing on both 1.0.2 and 1.1.0 flavors currently being
packaged by some of the distros.


williamcroberts:
The real question here is how we want to handle the matrix, are all the combos really that interesting? Might be best to look at trimming this down.

williamcroberts:
todo list:
  - [ ] update commit message
  - [ ] fix fedora-30 with gcc (missing libasan and libusan)
  - [ ] ensure codecov coverage details are uploaded from one build

williamcroberts:
Suggestion for builds:

### via matrix
- 4 distros with vanilla clang and gcc builds (8 builds)

### via include
- 4 distros with gcrypt and 1 compiler (4 builds)
- 1 distro and 1 compiler with the 3 settings on WITH_TCTI_ASYNC, WITH_TCTI_PARTIAL (3 builds)
- 1 distro and compiler for scan-build (1 build)
- 1 distro for codecov (1 build)
- 1 distro for fuzz (1 build)
 
That gets us to 18 builds, aprox half of what is currently being built and should get us near to the existing ci times we had before.

tstruk:
It looks like the FUZZ build doesn't do the fuzzing:
https://travis-ci.org/tpm2-software/tpm2-tss/jobs/581411066#L443

something is still missing.

williamcroberts:
> It looks like the FUZZ build doesn't do the fuzzing:
> https://travis-ci.org/tpm2-software/tpm2-tss/jobs/581411066#L443
> 
> something is still missing.

I think it's that the CXX and GEN_FUZZ env variables appear to be missing in the docker.env file and thus not getting propagated.

williamcroberts:
working now:
https://travis-ci.org/tpm2-software/tpm2-tss/jobs/585666448

.ci/coverity.run
```diff
@@ -0,0 +1,35 @@
...
+
+echo "Collecting Coverity data for submission"
+rm -fr README
+export AUTHOR="$(git log -1 $TRAVIS_COMMIT --pretty="%aN")"
+echo "Name: $AUTHOR" >> README
+echo "Email: william.c.roberts@intel.com" >> README
+echo "Project: 01org/tpm2.0-tools" >> README
+echo "Build-Version: $TRAVIS_COMMIT" >> README
+echo "Description: $TRAVIS_REPO_SLUG $TRAVIS_BRANCH" >> README
+echo "Submitted-by: tpm2-tools CI" >> README
```
tstruk: This section needs updates

PR #1512, opened 2019-09-09
Public is not updated when CreateLoaded() completes

LordOfDorks:
This results in Esys_TR_GetName() returning the name of the template and not the actual public that was created.

### DIFF
--- a/src/tss2-esys/api/Esys_CreateLoaded.c
+++ b/src/tss2-esys/api/Esys_CreateLoaded.c
@@ -392,6 +392,8 @@ Esys_CreateLoaded_Finish(
             "in Public name not equal name in response", error_cleanup);
 
     /* Update the meta data of the ESYS_TR object */
+    objectHandleNode->rsrc.rsrcType = IESYSC_KEY_RSRC;
+    objectHandleNode->rsrc.misc.rsrc_key_pub = *loutPublic;
     objectHandleNode->rsrc.name = name;
     objectHandleNode->auth = esysContext->in.CreateLoaded.inSensitive->sensitive.userAuth;
     if (outPublic != NULL)


src/tss2-esys/api/Esys_CreateLoaded.c
```diff
@@ -392,6 +392,8 @@ Esys_CreateLoaded_Finish(
             "in Public name not equal name in response", error_cleanup);
 
     /* Update the meta data of the ESYS_TR object */
+    objectHandleNode->rsrc.rsrcType = IESYSC_KEY_RSRC;
+    objectHandleNode->rsrc.misc.rsrc_key_pub = *loutPublic;
```
tstruk: Isn't it done in line 321 and 323 above already?
What kind of key are you trying to create that is failing?

williamcroberts: The only thing I can tell is that size in the resource object TPM2B is left at 0 after the unmarshal. However, calls to check name uses the TPM2B_PUBLIC from loutpublic, which has size. I thought perhaps subsequent calls to Esys_TR_GetName() would be wrong, but it doesn't make use of size. So not really sure what the issue is either.

tstruk: > The only thing I can tell is that size in the resource object TPM2B is left at 0 after the unmarshal.

The MU library sets also the size in all TPM2B structs when it unmarshals them.
Otherwise how the user would know if anything was returned by the TPM?

williamcroberts: > > The only thing I can tell is that size in the resource object TPM2B is left at 0 after the unmarshal.
> 
> The MU library sets also the size in all TPM2B structs when it unmarshals them.
> Otherwise how the user would know if anything was returned by the TPM?

The TPM2_PUBLIC in loutpublic has size set, but the TPM2B_PUBLIC in the resource node, size is 0.

williamcroberts: It's the wrong name... the template is missing a lot of the parameters (like unique) returned in the loutpublic variable. So if you call Esys_TR_GetName() it doesn't match ReadPublic. Test and patch forthcoming. 

has: 000b6e59ef657bca3b9624088244e6f67ab28d7843a8f53df58e37c67b0df7152f44
got: 000bd657c3a10b77897d20dd7e38ef2718349dae19795ed4e3db6d8a629ea62a5c4f

williamcroberts: See: https://github.com/tpm2-software/tpm2-tss/pull/1517

tstruk:
Thanks, but this is superseded by #1517.

PR #1521, opened 2019-09-19
tcti-device: getPollHandles should allow num_handles query

jeffpferreira:
Signed-off-by: Jeffrey Ferreira <jeffpferreira@gmail.com>


jeffpferreira:
See [TCTI API Specification](https://trustedcomputinggroup.org/resource/tss-tcti-specification/), Section 3.2.5.5 which states "In order to query the number of handles that a TCTI module needs to have monitored, the application may pass NULL for handles; in this case, it returns the number of handles."

PR #1527, opened 2019-09-25
esys: add ECC parameter padding for Esys_LoadExternal

### LINKED ISSUE #1518: Implement ECC input parameter padding
I noticed that for some TPMs and some ECC keys, LoadExternal fails. I'll give an example for ECC P256  keys. The public key is a point on the curve which consists of two coordinates *x* and *y* which are 256 bit (or 32 byte) each. The private key is a scalar *a* with 256 bit (32 byte) as well.

The problem arises when one of these parameters *x*, *y* or *a* are small enough to have one or more leading zero bytes. In this case, the user can transmit a lower number of bytes to the TPM (e.g. 31 instead of 32 bytes). I discovered this because the tool `tpm_loadexternal` was failing on me in some cases if fed with openssl generated keys.

Unfortunately, padding is not very well defined in the spec. There is only [this](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-1-Architecture-01.38.pdf#%5B%7B%22num%22%3A1706%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C300%2C0%5D) paragraph which basically states:

> When  ECC  parameters  are  input  parameters  to  the  TPM,  the  padding  is  optional

#### Loading an unpadded public key together with its private key 
If one coordinate *x* or *y* is unpadded and the corresponding private key is given, everything will work fine. (Tested for both simulator and SLM9670)

#### Loading an unpadded public key **without private key**
If one coordinate *x* or *y* is unpadded and the corresponding private key is **not** given, the TPM **MAY** return `TPM_RC_KEY`. This is documented in the [latest errata](https://trustedcomputinggroup.org/wp-content/uploads/TPM2.0-Library-Spec-v1.38-Errata_v1.8_for_publication_fixed_references.pdf#%5B%7B%22num%22%3A107%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C69%2C246%2C0%5D).

#### Loading an public key with an unpadded private key
I have not tested this. This is the most undocumented case.

#### Suggested solution
I suggest adding ECC input parameter padding for both public key coordinates as well as the private key to the TSS. That means if a ECC key parameter for a given curve is smaller than expected, leading zeros should be added. Personally, I would add that feature to the SAPI.

joholl:
When ECC parameters (private scalar, public x and y coordinates) are not padded, problems might arise. E.g. if one of the x, y coordinates do not have the expected size (due to leading zero bytes) and the private part is not given, the TPM might return `TPM_RC_KEY`. To avoid any unexpected behaviour, generally add padding to ECC parameters. Fixes #1518.

Could someone please review if I got the keysizes in `ecc_get_keysize()` right?

### DIFF
--- a/src/tss2-esys/api/Esys_LoadExternal.c
+++ b/src/tss2-esys/api/Esys_LoadExternal.c
@@ -148,6 +148,10 @@ Esys_LoadExternal_Async(
     TPMI_RH_HIERARCHY hierarchy)
 {
     TSS2_RC r;
+    TPM2B_SENSITIVE inPrivate_padded;
+    TPM2B_PUBLIC inPublic_padded;
+    size_t keysize;
+
     LOG_TRACE("context=%p, inPrivate=%p, inPublic=%p,"
               "hierarchy=%"PRIx32 "",
               esysContext, inPrivate, inPublic, hierarchy);
@@ -162,6 +166,23 @@ Esys_LoadExternal_Async(
     if (r != TSS2_RC_SUCCESS)
         return r;
     esysContext->state = _ESYS_STATE_INTERNALERROR;
+
+    /* Pad ECC parameters */
+    if (inPublic && inPublic->publicArea.type == TPM2_ALG_ECC) {
+        r = ecc_get_keysize(inPublic->publicArea.parameters.eccDetail.curveID, &keysize);
+        if (r != TSS2_RC_SUCCESS)
+            return r;
+
+        inPublic_padded = *inPublic;
+        ecc_public_add_padding(&inPublic_padded, keysize);
+        inPublic = (const TPM2B_PUBLIC*) &inPublic_padded;
+
+        if (inPrivate) {
+            inPrivate_padded = *inPrivate;
+            ecc_sensitive_add_padding(&inPrivate_padded, keysize);
+            inPrivate = (const TPM2B_SENSITIVE*) &inPrivate_padded;
+        }
+    }
 
     /* Check input parameters */
     r = check_session_feasibility(shandle1, shandle2, shandle3, 0);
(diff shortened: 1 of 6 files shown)


src/tss2-esys/api/Esys_LoadExternal.c
```diff
@@ -163,6 +167,23 @@ Esys_LoadExternal_Async(
...
 
+    /* Pad ECC parameters */
+    if (inPublic && inPublic->publicArea.type == TPM2_ALG_ECC) {
+        r = ecc_get_keysize(inPublic->publicArea.parameters.eccDetail.curveID, &keysize);
+        if (r != TSS2_RC_SUCCESS)
+            return r;
+
+        inPublic_padded = *inPublic;
+        ecc_public_add_padding(&inPublic_padded, keysize);
+        inPublic = (const TPM2B_PUBLIC*) &inPublic_padded;
```
tstruk: This will mess up the HMAC sessions. What if someone would want to precompute the auth of a session, where he sends a command and one of the parameters is the object handle of the just loaded ECC key. 
The name computed for the the original inPublic.public will not be valid as soon as the Esys_LoadExternal_Async() returns. What will happen if one would then bound a new session to this object? Again - in my opinion the padding should be done by the user of the API.

joholl: You're totally right. I did not notice this issue until now. Padding the public portion of the key changes its name from name<sub>unpadded</sub> to name<sub>padded</sub>. This might be a problem if the user assumes name<sub>unpadded</sub> because it is used to calculate authHMAC for the HMAC session.

Sorry for the inconvenience.

Issue #1522, opened 2019-09-23
Clarity between the ESAPI API Spec and the ESAPI implemented API.


joe-cryptosecurellc:
Background: The ESAPI spec (v0.90 Rev 4) defines 'hierarchy' types as ESYS_TR for Hash and Sequence Finish commands. The implementation instead uses TPMI_RH_HIERARCY as the type for this command. The README for ESAPI explains that the API is a 1-to-1 mapping of the commands defined in Pt3 of the TPM2 specification.

Analysis: It is difficult to map 1-to-1 all commands defined in Pt.3 of the TPM specification for ESAPI, since many arguments in ESAPI API must use ESYS_TR types for handles and auth types.

Conclusion: If it is required for the ESAPI Spec and API implementation to be consistent, then should the API implementation follow the ESAPI spec's definition, or should the ESAPI Spec change to make the parameters all TPM-types? If it is not required, then the description of the ESAPI implementation could change to explain that it is a 1-to-1 mapping of only the parameters defined in Pt.3 of the TPM2 Specification (since authorization handles must be ESYS_TR types for ESAPI).

tstruk:
It is messed up alright. Not only Esys_Hash() and Esys_SequenceComplete(), but also Esys_LoadExternal() is also screwed the same way. These need to be converted to take ESYS_TR, as in the spec, and then convert the ESYS handle to TPM handle internally using iesys_handle_to_tpm_handle() function. This change would need a major version bump though.

PR #1531, opened 2019-09-26
esys: change the hierarchy type according to the spec

tstruk:
There is a discrepancy between ESYS specification v0.90 rev4 and the
ESYS implementation, where the specification defines hierarchy
as ESYS_TR type and the implementation uses TPMI_RH_HIERARCHY type.
The affected functions include: Esys_Hash(), Esys_HierarchyControl(),
Esys_LoadExternal(), and Esys_SequenceComplete() plus their _Async()
version. This change makes the spec and the implementation consistent.

### DIFF
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -421,7 +421,7 @@ Esys_LoadExternal(
     ESYS_TR shandle3,
     const TPM2B_SENSITIVE *inPrivate,
     const TPM2B_PUBLIC *inPublic,
-    TPMI_RH_HIERARCHY hierarchy,
+    ESYS_TR hierarchy,
     ESYS_TR *objectHandle);
 
 TSS2_RC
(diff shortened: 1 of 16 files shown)


tstruk:
@idesai this can be certainly put on hold for now. This PR is to start the discussion on what the best solution would be.

tstruk:
@idesai just thinking that to keep the options interface compatible you could actually support both types TPMI_RH_HIERARCHY and ESYS_TR and remap to the correct value before invoking the ESYS function. That one possible solution.

idesai:
> @idesai just thinking that to keep the options interface compatible you could actually support both types TPMI_RH_HIERARCHY and ESYS_TR and remap to the correct value before invoking the ESYS function. That one possible solution.

what do the handle values look like for ESYS_TR? Is it fixed?

tstruk:
> what do the handle values look like for ESYS_TR? Is it fixed?
Yes, see https://github.com/tpm2-software/tpm2-tss/blob/master/include/tss2/tss2_esys.h#L54

AndreasFuchsSIT:
That's a big one... :-(

I guess the tools would not mind the command line options, but basically the tools would reuire an #ifdef for tss2_esys < 3.0 and tss2_esys >= 3.0

Seems like we're now guaranteed to break ABI...
You could add a commit to this PR to set the version to 3.0-dev in configure.ac... ;-)

tstruk:
Tools can handle this change as follows: https://github.com/tpm2-software/tpm2-tools/pull/1776

tstruk:
Fixes: #1522

PR #1532, opened 2019-10-04
Add C++ linkage to tss2_rc.h

s98felix:
Without the change, using Tss2_RC_Decode() or Tss2_RC_SetHandler() from
C++ code results in undefined reference error during linkage.


tstruk:
Please add your Signed-off-by: to the commit message. Thanks

Issue #1533, opened 2019-10-04
ESAPI should expose the API to retrieve cpHash


idesai:
The function [tpm_calc_phash](https://github.com/tpm2-software/tpm2-tss/blob/7da2391547c08157ce3046f8049aa527cfb105fc/test/integration/sapi-session-util.c#L340) or [iesys_compute_cp_hashtab](https://github.com/tpm2-software/tpm2-tss/blob/70e9fae7ef535e7cf27a72ddbc818dfefcbdbdbb/src/tss2-esys/esys_iutil.c#L211) should be exposed through the ESAPI api interface.
This is required TPM commands that need cpHash like PolicySigned/ PolicySecret/ etc.

tstruk:
What about Tss2_Sys_GetCpBuffer()?

idesai:
> What about Tss2_Sys_GetCpBuffer()?
How do I access this function from ESAPI?

idesai:
> What about Tss2_Sys_GetCpBuffer()?
> How do I access this function from ESAPI?

Something similar has been done to retrieve session nonceTPM [here](https://github.com/tpm2-software/tpm2-tss/commit/2a67f9e4b80d4875e4e70c9a6c899a749c57c608).

AndreasFuchsSIT:
Yes, so I guess a TPM-less SYS_CONTEXT would allow for a seuence of Tss2_Sys_<command>_Prepare(); Tss2_Sys_GetCpBuffer().

If we want to make it systematic, I guess I'd like to introduce a library next to tss2-esys.so that exposes the cpHash functions. Tss2_util_<command>_cpHash(hashAlg, obj1, obj2, param1, param2) or similar I guess ?

PR #1536, opened 2019-10-09
esys: add a new function to retrieve sys context from esys

tstruk:
In some cases it can be necessary to call the underlining
sys function before invoking esys function. One example
would be to retrieve cpHash buffer of a command that needs
to be fed to Esys_PolicySigned or Esys_PolicySecret calls.
This can be done by using the Tss2_Sys_GetCpBuffer()
sys call as long as the user can access the sys context.
That's one possible (and simple) solution.
See issue: #1533

### DIFF
--- a/doc/doxygen.dox
+++ b/doc/doxygen.dox
@@ -23,6 +23,8 @@ Both the synchronous and asynchronous API are exposed through a single library: 
  \fn TSS2_RC Esys_GetTcti(ESYS_CONTEXT * esys_context, TSS2_TCTI_CONTEXT ** tcti)
  \fn TSS2_RC Esys_GetPollHandles(ESYS_CONTEXT * esys_context, TSS2_TCTI_POLL_HANDLE ** handles, size_t * count)
  \fn TSS2_RC Esys_SetTimeout(ESYS_CONTEXT *esys_context, int32_t timeout)
+ \fn TSS2_RC Esys_GetSysContext(ESYS_CONTEXT *esys_context, TSS2_SYS_CONTEXT **sys_context)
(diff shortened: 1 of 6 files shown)


tstruk:
> I still prefer the tcti-less syscontext approach.
> Reason is that this change will bind ESYS to SYS forever. We will never be able to ditch SYS and only provide ESYS without breaking API; nor will anybody else...

That is true, but will there ever be ESYS without SYS? Isn't it already bind forever by all the diagrams like in [1], where ESYS box is always above SYS box.

[1] Esys spec - https://trustedcomputinggroup.org/wp-content/uploads/TSS_TSS-2.0-Enhanced-System-API_V0.9_R03_Public-Review-1.pdf

AndreasFuchsSIT:
> > I still prefer the tcti-less syscontext approach.
> > Reason is that this change will bind ESYS to SYS forever. We will never be able to ditch SYS and only provide ESYS without breaking API; nor will anybody else...
> 
> That is true, but will there ever be ESYS without SYS? Isn't it already bind forever by all the diagrams like in [1], where ESYS box is always above SYS box.

That diagram is just informative... This change would make this binding normative.

tstruk:
@AndreasFuchsSIT if we ever implement SYS-less ESYS then we will need to add a bunch of new functions like the mentioned Esys_getCpHash etc, and after we have that we could change Esys_GetSysContext() to always return NULL sys context, indicating that this ESYS is SYS-less.

tstruk:
@idesai do you have any feedback on this?

AndreasFuchsSIT:
LGTM

PR #1544, opened 2019-11-01
test/integration: move TPM simulator from ibmswtpm2 to ms-tpm-20-ref

diabonas:
To continue the discussion from https://github.com/tpm2-software/tpm2-tss/pull/1542#issuecomment-548825653, here are the necessary changes to use [ms-tpm-20-ref](https://github.com/microsoft/ms-tpm-20-ref) instead of [ibmswtpm2](https://sourceforge.net/projects/ibmswtpm2/) for the TPM simulator. If we decide to move forward with this approach, the first commit should be moved to the Dockerfile in [tpm2-software-container](https://github.com/tpm2-software/tpm2-software-container) instead of building the simulator in each tpm2-tss CI run.

### DIFF
--- a/.ci/docker-prelude.sh
+++ b/.ci/docker-prelude.sh
@@ -13,7 +13,7 @@ source $TRAVIS_BUILD_DIR/.ci/download-deps.sh
 get_deps "$WORKSPACE"
 
 export LD_LIBRARY_PATH=/usr/local/lib/
-export PATH=/root/.local/bin/:/ibmtpm974/src:$PATH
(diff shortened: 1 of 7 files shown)


AndreasFuchsSIT:
Working on alternative in #1542

AndreasFuchsSIT:
@diabonas Would you mind putting the changes to misc.m4 into a separate PR ? That seams like a good addition if people want to use their own different simulator...

diabonas:
@AndreasFuchsSIT Sure, I branched out this commit in #1545.

PR #1548, opened 2019-11-11
[RFC][WIP] ESYS: Iniaitial Implementation of AC commands

williamcroberts:
Implement the Attached Component Commands for ESAPI.

TODO:
  - is the interface sane?
  - How to get the sendObject as an ESYS_TR, is it via:
    - Esys_TR_FromPublic() or are they hard coded addresses?
  - Testing, is it possible?

### DIFF
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -9,7 +9,7 @@
 #include "tss2_tcti.h"
 #include "tss2_sys.h"
 
-#ifdef __cplusplus
+#ifdef _cplusplus
 extern "C" {
 #endif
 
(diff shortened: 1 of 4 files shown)


AndreasFuchsSIT:
`TPMI_RH_AC ac` should probably be an ESYS_TR following the scheme from
https://github.com/tpm2-software/tpm2-tss/blob/30afabfc81496bc8a225e67b4d090ba1e169a22a/include/tss2/tss2_esys.h#L54-L59

Does that make sense ?

williamcroberts:
> `TPMI_RH_AC ac` should probably be an ESYS_TR following the scheme from
> https://github.com/tpm2-software/tpm2-tss/blob/30afabfc81496bc8a225e67b4d090ba1e169a22a/include/tss2/tss2_esys.h#L54-L59
> 
> Does that make sense ?

I wasn't 100% sure if they were constants, so if they are, then yes it makes sense. I need to go look at the actual spec to find the constant values.

include/tss2/tss2_esys.h
```diff
@@ -9,7 +9,7 @@
 #include "tss2_tcti.h"
 #include "tss2_sys.h"
 
-#ifdef __cplusplus
+#ifdef _cplusplus
```
tstruk: This should be double "__"
https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html

include/tss2/tss2_esys.h
```diff
@@ -3221,19 +3221,103 @@ Esys_Vendor_TCG_Test_Finish(
...
+    TPM2B_NAME *authHandleName,
+    TPM2B_NAME *acName,
+    TPMI_YES_NO includeObject);
+
 /*
  * TPM 2.0 ESAPI Helper Functions
  */
 void
 Esys_Free(
-    void *__ptr);
```
tstruk: What does this have todo with AC commands?

AndreasFuchsSIT:
Is this PR postponed until the next version of the TPM library spec ?
Then I'd close it for now

AndreasFuchsSIT:
Ok, closing for now.
Since it was RFC only anyways...

PR #1559, opened 2019-11-21
pkg-config: remove TCTI loaders from ESYS dependencies

diabonas:
Since commit b258ab9618ded645777917a742b917a3ba72f107 the ESYS library is not linked to any TCTI libraries any more, they are dlopen()'ed if available instead. Removing them from the dependencies also fixes an error when using pkg-config to discover tss2-esys if tpm2-tss was built with `--disable-tcti-mssim`.

Fixes: https://github.com/tpm2-software/tpm2-tools/issues/1851

### DIFF
--- a/lib/tss2-esys.pc.in
+++ b/lib/tss2-esys.pc.in
@@ -7,7 +7,7 @@ Name: tss2-esys
 Description: TPM2 Enhanced System API library.
 URL: https://github.com/tpm2-software/tpm2-tss
 Version: @VERSION@
-Requires.private: tss2-mu tss2-sys tss2-tcti-device tss2-tcti-mssim
+Requires.private: tss2-mu tss2-sys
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
But don't we have to add tss2_tctildr to Reuires.private ?

diabonas:
A valid question! Apparently tss2-esys [statically pulls in the tss2-tctildr sources](https://github.com/tpm2-software/tpm2-tss/blob/1c99117f4d102fd465655ed86aca58909a13c10d/Makefile.am#L357-L366), so there is no link-time dependency on `libtss2-tctildr.so`. This is also confirmed by `ldd libtss2-esys.so`, so I don't think it needs to be in `Requires.private`.

AndreasFuchsSIT:
That surprising, but I guess it's ok...

Issue #1568, opened 2019-12-03
Tss2_TctiLdr_Initialize_Ex returns TSS2_RC_SUCCESS on memory allocation failure


Pagten:
Tss2_TctiLdr_Initialize and Tss2_TctiLdr_Initialize_Ex return TSS2_RC_SUCCESS when the calloc of a TSS2_TCTILDR_CONTEXT struct fails.

Pagten:
See #1569

PR #1569, opened 2019-12-03
Return proper error code on memory allocation failure

Pagten:
See issue #1568


tstruk:
Looks like test/unit/tctildr test needs to be updated

PR #1576, opened 2019-12-14
tcti: try to open /dev/tpmrm0 before /dev/tpm0 during tcti-device init

### LINKED ISSUE #1564: tcti device default value vs tctildr load preference
The tctildr load preference is "first `/dev/tpm0` then `/dev/tpmrm0`.

https://github.com/tpm2-software/tpm2-tss/blob/d5cca2a51434c0d0fe7cf4237b161d498baed02d/src/tss2-tcti/tctildr-dl.c#L41-L50

However, when passing a `NULL` conf string to `Tss2_Tcti_Device_Init()`, the tcti-device chooses `/dev/tpm0` by default.

https://github.com/tpm2-software/tpm2-tss/blob/d5cca2a51434c0d0fe7cf4237b161d498baed02d/src/tss2-tcti/tcti-device.c#L62

For the tools, this means not specifying a tcti will result in using `/dev/tpmrm0` (assuming tabrmd is not installed) but passing `--tcti device`, will result in using `/dev/tpm0`. This is not what I would expect.

What do you think about having tcti-device try `/dev/tpmrm0` first and fallback to `/dev/tpm0` if a `NULL` conf string is passed to `Tss2_Tcti_Device_Init()`?

joholl:
The tctildr library tries to open `/dev/tpmrm0` before `/dev/tpm0` if no tcti was specified. Previously, if the tcti `device` was specified, the tcti-device module would try to open `/dev/tpm0` only.

To streamline the loading behavior, have tcti-device try to open `/dev/tpmrm0` and fallback to `/dev/tpm0`.

Fixes #1564.

### DIFF
--- a/test/unit/tcti-device.c
+++ b/test/unit/tcti-device.c
@@ -9,6 +9,7 @@
 #endif
 
 #include <inttypes.h>
+#include <stdarg.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <string.h>
@@ -38,6 +39,21 @@ static uint8_t tpm2_buf [BUF_SIZE] = {
     0xca, 0xfe, 0xba, 0xbe,
     0xfe, 0xef
 };
+int
+__real_open(const char *pathname, int flags, ...);
+/* wrap function for open required to test init */
+int
+__wrap_open(const char *pathname, int flags, ...)
+{
+    const char* pathname_prefix_dev = "/dev";
+    if (strncmp(pathname, pathname_prefix_dev, strlen(pathname_prefix_dev)) == 0) {
+        return mock_type (int);
+    } else {
+        /* only mock opening of device files as the open() syscall is needed
+           for code coverage reports as well */
+        return __real_open(pathname, flags);
+    }
+}
 /**
  * When passed all NULL values ensure that we get back the expected RC
  * indicating bad values.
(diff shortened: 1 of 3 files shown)


joholl:
Unfortunately I'm not sure why the test fails. On first glance it looks like a cmocka bug. All tests in the test group `tcti-device` are passing and `test-suite.log` clearly says `[  PASSED  ] 14 test(s)`. Nevertheless, running the binary `/workspace/tpm2-tss/build/test/unit/.libs/tcti-device` results in an exit code `255`.

Locally, my build succeeds but I can reproduce the problem with the `ubuntu-18.04` docker image.

Steps to reproduce (basically copied from the [travis log](https://travis-ci.org/tpm2-software/tpm2-tss/jobs/630346511?utm_medium=notification&utm_source=github_status)):

```bash
# setup (see travis log)
git clone --depth=50 https://github.com/tpm2-software/tpm2-tss.git tpm2-software/tpm2-tss
cd tpm2-software/tpm2-tss
git fetch origin +refs/pull/1576/merge:
git checkout -qf FETCH_HEAD

export coverity_scan_run_condition='"$CC" = gcc'
export DOCKER_TAG=ubuntu-18.04
export TRAVIS_COMPILER=gcc
export CC=${CC:-gcc}
export CC_FOR_BUILD=${CC_FOR_BUILD:-gcc}

# ./.ci/travis.run (edited to bash into container)
ci_env=$(bash <(curl -s https://codecov.io/env))
build="build-$RANDOM"

docker run $ci_env --env-file .ci/docker.env \
    -v "$(pwd):/workspace/tpm2-tss" \
    --name $build -it "tpm2software/tpm2-tss:$DOCKER_TAG" \
    /bin/bash -c '/workspace/tpm2-tss/.ci/docker.run || /bin/bash'
```

diabonas:
Apparently `open()` is used for collecting code coverage, see the blog post ["Unit testing C code with CMocka", Section "Issues with gcov and system calls mocking"](https://blog.microjoe.org/2017/unit-tests-c-cmocka-coverage-cmake.html#issues-with-gcov-and-system-calls-mocking). The solution is to mock only opening the device files necessary for the unit tests and pass through the real system call otherwise:
```diff
diff --git a/test/unit/tcti-device.c b/test/unit/tcti-device.c
index 0d093e2c..997337ef 100644
--- a/test/unit/tcti-device.c
+++ b/test/unit/tcti-device.c
@@ -39,11 +39,20 @@ static uint8_t tpm2_buf [BUF_SIZE] = {
     0xca, 0xfe, 0xba, 0xbe,
     0xfe, 0xef
 };
+int
+__real_open(const char *pathname, int flags, ...);
 /* wrap function for open required to test init */
 int
 __wrap_open(const char *pathname, int flags, ...)
 {
-    return mock_type (int);
+    const char* pathname_prefix_dev = "/dev";
+    if (strncmp(pathname, pathname_prefix_dev, strlen(pathname_prefix_dev)) == 0) {
+        return mock_type (int);
+    } else {
+        /* only mock opening of device files as the open() syscall is needed
+           for code coverage reports as well */
+        return __real_open(pathname, flags);
+    }
 }
 /**
  * When passed all NULL values ensure that we get back the expected RC
```
With this change CI passes, see https://github.com/diabonas/tpm2-tss/commit/e71a5a3481b4636ec8aa7d8ddc5b8277aa58e7af.

tstruk:
@joholl could you update the PR as suggested by @diabonas please

joholl:
Many thanks, @diabonas, you're absolutely right!

test/unit/tcti-device.c
```diff
@@ -63,6 +79,63 @@ tcti_device_init_size_test (void **state)
...
+    size_t tcti_size = 0;
+    TSS2_RC ret = TSS2_RC_SUCCESS;
+    TSS2_TCTI_CONTEXT *ctx = NULL;
+
+    ret = Tss2_Tcti_Device_Init (NULL, &tcti_size, NULL);
+    assert_true (ret == TSS2_RC_SUCCESS);
+    ctx = calloc (1, tcti_size);
+    assert_non_null (ctx);
+    errno = EACCES; /* Permission denied */
+    will_return (__wrap_open, -1);
```
tstruk: Looks like one will_return is redundant here.

joholl: This is needed here. Passing a `NULL` conf will result in trying to open the specified default files one after another until `open()` succeeds. See the [call to open()](https://github.com/tpm2-software/tpm2-tss/blob/051160c39db6215a8931887ce677cc2040e9b035/src/tss2-tcti/tcti-device.c#L459) inside of the for loop.

These default files are:

- `/dev/tpm0`
- `/dev/tpmrm0`

If both attempts fail (thus returning `-1` twice) an `TSS2_TCTI_RC_IO_ERROR` is returned by `Tss2_Tcti_Device_Init()`.

PR #1577, opened 2019-12-18
tpmclient: Refactor SimpleHmacOrPolicyTest into integration test

flihp:
This test uses an NV region to test both HMAC and POLICY sessions. In
the current tpmclient code only the policy path is taken. This commit
refactors the SimpleHmacOrPolicyTest function, enables the HMAC code
path (gets us a bit more code coverage), moves some code to the
sapi-util test library, and moves the test code to a new new
integration test module: sapi-hmac-auth.int.


test/integration/sapi-hmac-auth.int.c
```diff
@@ -0,0 +1,431 @@
...
+        LOG_ERROR ("compute_command_hmac failed with RC: 0x%x", rc);
+        return rc;
+    }
+
+    /*
+     * And now read the data back.
+     * If the command is successful, the command
+     * HMAC was correct.
+     */
+    rc = Tss2_Sys_NV_Read (sys_ctx,
```
tstruk: This will call Tss2_Sys_NV_Read_Prepare() again. Shouldn't this rather be
Tss2_Sys_Execute() followed by Tss2_Sys_NV_Read_Complete()?
All the auths will probably be the same, only the params will be written twice.

flihp: Are you waiting for me to make this change? Not clear since you've marked the PR 'approved'. I don't disagree with your analysis but the outcome is effectively the same.

PR #1582, opened 2019-12-20
FAPI: make directories suitable for packaging

### LINKED ISSUE #1393: tss user / group creation
https://github.com/tpm2-software/tpm2-abrmd/issues/613 brings up relevant points about assumptions made about the existence of the `tss` user account / group. Since tpm2-tss took over the udev rules we now have the `tss` user and group hard coded: https://github.com/tpm2-software/tpm2-tss/commit/0075f88466d791b2b58c60adafed87ec814441df

I figure we can make the name of the user / group a configure time option without much trouble. I don't see a viable way to guarantee that the user / group exist at install time without changing the default to 'root'. Anyone overriding this default explicitly via configure script would be taking on responsibility for ensuring they exist (likely in a package script). There was also mention of using some systemd mechanism: https://www.freedesktop.org/software/systemd/man/sysusers.d.html

Any strong opinions on this one @AndreasFuchsSIT @tstruk @williamcroberts?

diabonas:
- Don't use `chown tss:tss` in `make install`, this does not work when packaging because the necessary `tss` user and groups are not created yet. Instead fully rely on [tmpfiles.d](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html) to create the directories with the correct owner and access rights.
- To create the `tss` system user and group, install a [sysusers.d](https://www.freedesktop.org/software/systemd/man/sysusers.d.html) configuration file. This also fixes #1393.
- According to the [Filesystem Hierarchy Standard](http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05.html), applications must not create directories directly under `/var`, the correct directory for the key store is under `/var/lib/tpm2-tss`.

Additional smaller fixes:
- `./configure --runstatedir` is not yet available in the latest Autoconf 2.69, it is only added in the unreleased version 2.70. Add a workaround `--with-runstatedir` for distributions using the official, unpatched Autoconf.
- Name the tmpfiles.d configuration file as prescribed by the specification.


AndreasFuchsSIT:
The reason for the install hooks and creating the tpmfiles directory on make install was to be able to directly use fapi after make install.
With only installing the tmpfiles.d IIRC the directories would only be created upon reboot.

Can you confirm this ?

diabonas:
> With only installing the tmpfiles.d IIRC the directories would only be created upon reboot.

To create the directories directly, it is possible to run `systemd-tmpfiles --create`. Most distributions do this automatically when installing packages so that the software can be used directly.

AndreasFuchsSIT:
Should we trigger this `systemd-tmpfiles --create` then from the install-exec-hook: ?

I'm also concerned for non-packaged installs... ;-)

diabonas:
I'm not a big fan of running things requiring root access during `make install`, but I guess attempting to run `systemd-sysusers` and `systemd-tmpfiles` should be fine - worst case it fails and the user/directories are only created after a reboot. See the last commit in the update PR.

diabonas:
While playing around with the implementation, I noticed that you need to specify `--sysconfdir` during `./configure`, otherwise `fapi-config.json` can't be found. The reason is that the default value for `sysconfdir` is `${prefix}/etc`, which gets copied without expansion to the `SYSCONFDIR` macro.

I added another commit that expands the prefix to a full path before defining the macro so that it works with `sysconfdir` containing references to other variables.

tstruk:
I think with this one in we can tag 2.4.0-rc0

PR #1583, opened 2019-12-20
 FAPI: fix pkg-config file

diabonas:
Fix the pkg-config file to include all required dependencies in <code>Requires<em>.private</em></code>, cf. #1417, and improve the linking a bit.


tstruk:
Looks fine, and it should also fix this: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19639

PR #1604, opened 2020-01-22
mu: Remove use of VLAs for Marshalling TPML types

josephlr:
All of the `Tss2_MU_*_Marshal()` functions have the property that
`buffer` can be NULL, `offset` can be NULL, but both cannot be
NULL. Some Marshal functions check this directly (returning
`TSS2_MU_RC_BAD_REFERENCE` on error), but most do this by composing
existing Marshaling functions, and calling them with the same
`buffer`/`offset` arguments as the parent function.

The TMPL Marshal function macro does things differently. It creates a local
VLA `local_buffer[buffer_size]` and uses that as the buffer pointer if
a NULL buffer is given. This is unnecessary, as this pointer is only used:
  - for debug logging
  - as an argument to other Marshaling functions (which will correctly handle a NULL buffer).

Note that the VLA in the existing code is of length `buffer_size` (the
length of the _entire_ buffer, _not_ the length of the data being
unmarshaled). This can potentially result in a very large stack
allocation or stack overflow.

*EDIT*: I don't think this change needs more tests, as this condition is already checked in `test/unit/TPML-marshal.c`, but I can add some if they are necessary.

Signed-off-by: Joe Richey <joerichey@google.com>

### DIFF
--- a/src/tss2-mu/tpml-types.c
+++ b/src/tss2-mu/tpml-types.c
@@ -29,8 +29,6 @@ TSS2_RC Tss2_MU_##type##_Marshal(type const *src, uint8_t buffer[], \
     size_t  local_offset = 0; \
     UINT32 i, count = 0; \
     TSS2_RC ret = TSS2_RC_SUCCESS; \
-    uint8_t *buf_ptr = buffer; \
-    uint8_t local_buffer[buffer_size]; \
 \
     if (offset != NULL) { \
         LOG_TRACE("offset non-NULL, initial value: %zu", *offset); \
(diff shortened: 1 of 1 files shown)


tstruk:
The reason why we need to use the the local buffer is to satisfy the TCG MU spec [1], which states:
"The offset parameter will be updated by the function to the index of the next byte after the marshaled data in the buffer byte array. If the buffer parameter is NULL the implementation will not write any marshaled data but the offset parameter MUST be updated as though it had."
so in case when the user calls a marshal function with a NULL buffer, but a valid offset we need to use the local buffer and unmarshal the given type, and all TPML struct members in order to update the offset. We also need to use on the stack allocation, because of other TCG spec requirement, which states that the System API shell not allocate memory. See "2.2 SAPI High Level Design Requirements" in [2]

[1] https://trustedcomputinggroup.org/wp-content/uploads/TSS_Marshaling_Unmarshaling_API_v1_r05_21jan20.pdf
[2] https://trustedcomputinggroup.org/wp-content/uploads/TSS_SAPI_v1p1_r29_pub_20190806.pdf

josephlr:
> The reason why we need to use the the local buffer is to satisfy the TCG MU spec [1], which states:
> "The offset parameter will be updated by the function to the index of the next byte after the marshaled data in the buffer byte array. If the buffer parameter is NULL the implementation will not write any marshaled data but the offset parameter MUST be updated as though it had."

I understand the requirements of the MU spec. The point I was trying to make in my PR description is that all this `local_buffer` stuff is unnecessary.

 Even after this change, if a user calls the Marshal functions with a NULL buffer and valid offset, the offset will be update correctly. As I noted above, there's already a test for this with TPML types, and that test is still passing with this change.

tstruk:
Looks like you're right. This seams work because all the sub-members types of the TPML_* structs are const size, even the TPML_DIGEST, which contains TPM2B_DIGEST. I will add a test with TPML_DIGEST type to cover that as well before merging.

PR #1618, opened 2020-01-30
[WIP] ci: add FreeBSD build with Cirrus CI

tstruk:


tstruk:
Merging this with UT enabled only so that all checks for PRs would pass.

PR #1621, opened 2020-01-31
Fapi fixes

AndreasFuchsSIT:
One more.
Now mostly style things and some renaming.

2 Bugs fixed with PolicyAuthorize and PolicyOr, where a slightly different behaviour was wanted.

Things are starting to slow down here now.


src/tss2-fapi/ifapi_policy_callbacks.c
```diff
@@ -771,7 +771,7 @@ search_policy(
         policy_object = calloc(sizeof(struct POLICY_LIST), 1);
         return_if_null(policy_object, "Out of memory.", TSS2_FAPI_RC_MEMORY);
 
-        policy_object->path = context->fsearch.current_path;
+        policy_object->path = strdup(context->fsearch.current_path);
```
tstruk: Check if NULL after alloc maybe?

AndreasFuchsSIT: yes

AndreasFuchsSIT:
Ok, I fixed the commit, which not only missed the OOM-check but also triggerd some leak check after fixing.

Also extended this PR with 1 more commit that targets chaining of PolicyAuthorize commands

PR #1627, opened 2020-02-12
Fixes #1611. Adapted Makefile.am to create FAPI folders without systemd.

### LINKED ISSUE #1611: "make install" does not create FAPI folders in systems without systemd (Makefile.am)
Installing the TSS with `make install` does not create FAPI folders in systems with no *systemd*. The problem is the `Makefile.am` which relies on *systemd* to create all FAPI folders in the system. In Docker, for example, there is no *systemd*, and as a result FAPI folders are not created. Additionally, folder permissions are not set, and the "tss" user and group are not created. This leads to errors in program code which uses the FAPI, e.g. when calling `Fapi_Provision(...)`.

As a solution, I suggest to augment the `Makefile.am` by doing the following, in case the *systemd* call fails:

1. create "tss" user and group
2. create FAPI folders
3. set owner and group of FAPI folders to "tss" user
4. set permissions accordingly

These steps should do the same as *systemd*. Here is a diff of the original and my proposed new `Makefile.am`:

```diff
diff --git a/Makefile.am b/Makefile.am
index aca6de27..edc0d6e1 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -521,6 +521,28 @@ endif #DOXYMAN
 
 endif #FAPI
 
+### Helper Functions ###
+define make_parent_dir
+    if [ ! -d $(dir $1) ]; then mkdir -p $(dir $1); fi
+endef
+
+define make_tss_user_and_group
+    (groupadd -r tss || true) && \
+    (useradd -r -g tss tss)
+endef
+
+define make_tss_dir
+    ($(call make_parent_dir,$1)) && \
+    (chown tss:tss "$1") && \
+    (chmod 775 "$1")
+endef
+
+define make_fapi_dirs
+    ($(call make_tss_dir,$(DESTDIR)$(runstatedir)/tpm2-tss/eventlog/) || true) && \
+    ($(call make_tss_dir,$(DESTDIR)$(localstatedir)/lib/tpm2-tss/system/keystore/))
+endef
+
+
 ### Man Pages
 man3_MANS = \
     man/man3/Tss2_Tcti_Device_Init.3 \
@@ -557,7 +579,9 @@ endif
 
 # Create tss user and FAPI directories directly after installation (vs. after a reboot)
 install-exec-hook:
-	systemd-sysusers && systemd-tmpfiles --create || true
+	(systemd-sysusers && systemd-tmpfiles --create) || \
+	(($(call make_tss_user_and_group) || true) && \
+	($(call make_fapi_dirs) || true))
 
 uninstall-hook:
 	cd $(DESTDIR)$(man3dir) && \
@@ -585,11 +609,6 @@ CLEANFILES += \
     $(man3_MANS) \
     $(man7_MANS)
 
-### Helper Functions ###
-define make_parent_dir
-    if [ ! -d $(dir $1) ]; then mkdir -p $(dir $1); fi
-endef
-
 # function to transform man .in files to man pages
 # $1: target
 # $2: .in file
```

MichaelEckelSIT:
Adapted `Makefile.am` to create 'tss' user and group as well as all FAPI
directories manually as a fallback solution, in case systemd is not available on the
system, such as in Docker (by default).

Signed-off-by: Michael Eckel <michael.eckel@sit.fraunhofer.de>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -522,6 +522,36 @@ endif #DOXYMAN
 
 endif #FAPI
 
+### Helper Functions ###
+define make_parent_dir
+    if [ ! -d $(dir $1) ]; then mkdir -p $(dir $1); fi
+endef
+
+define make_tss_user_and_group
+    (id -g tss || groupadd -r tss) && \
+    (id -u tss || useradd -r -g tss tss)
+endef
+
+define make_tss_dir
+    ($(call make_parent_dir,$1))
+endef
+
+define set_tss_permissions
+    (chown -R tss:tss "$1") && \
+    (chmod -R 775 "$1")
+endef
+
+define make_fapi_dirs
+    ($(call make_tss_dir,$(DESTDIR)$(runstatedir)/tpm2-tss/eventlog/) || true) && \
+    ($(call make_tss_dir,$(DESTDIR)$(localstatedir)/lib/tpm2-tss/system/keystore/))
+endef
+
+define set_fapi_permissions
+    ($(call set_tss_permissions,$(DESTDIR)$(runstatedir)/tpm2-tss)
+    ($(call set_tss_permissions,$(DESTDIR)$(localstatedir)/lib/tpm2-tss)
+endef
+
+
 ### Man Pages
 man3_MANS = \
     man/man3/Tss2_Tcti_Device_Init.3 \
@@ -558,7 +588,8 @@ endif
 
 # Create tss user and FAPI directories directly after installation (vs. after a reboot)
 install-exec-hook:
-	systemd-sysusers && systemd-tmpfiles --create || true
+	(systemd-sysusers && systemd-tmpfiles --create) || \
+	($(call make_tss_user_and_group) && $(call make_fapi_dirs) && ($call set_fapi_permissions)) || true
 
 uninstall-hook:
 	cd $(DESTDIR)$(man3dir) && \
@@ -585,11 +616,6 @@ EXTRA_DIST += \
 CLEANFILES += \
     $(man3_MANS) \
     $(man7_MANS)
-
-### Helper Functions ###
-define make_parent_dir
-    if [ ! -d $(dir $1) ]; then mkdir -p $(dir $1); fi
-endef
 
 # function to transform man .in files to man pages
 # $1: target


Makefile.am
```diff
@@ -522,6 +522,28 @@ endif #DOXYMAN
 
 endif #FAPI
 
+### Helper Functions ###
+define make_parent_dir
+    if [ ! -d $(dir $1) ]; then mkdir -p $(dir $1); fi
+endef
+
+define make_tss_user_and_group
+    groupadd -r tss && \
```
AndreasFuchsSIT: I think we should change this to not fail if the user already exists:
```
    (id -g tss || groupadd -r tss) && \
    (id -u tss || useradd -r -g tss tss)
```

Makefile.am
```diff
@@ -558,7 +580,9 @@ endif
 
 # Create tss user and FAPI directories directly after installation (vs. after a reboot)
 install-exec-hook:
-	systemd-sysusers && systemd-tmpfiles --create || true
+	(systemd-sysusers && systemd-tmpfiles --create) || \
+	(($(call make_tss_user_and_group) || true) && \
```
AndreasFuchsSIT: If make_tss_user_and_group fails, then make_fapi_dirs will also fail, since the chown fails.

IMHO a single || true on the outest would be better.

Makefile.am
MichaelEckelSIT: I totally agree.

Makefile.am
MichaelEckelSIT: Like this?
```bash
	(systemd-sysusers && systemd-tmpfiles --create) || \
	($(call make_tss_user_and_group) && \
	$(call make_fapi_dirs)) || true
```
This would not trigger FAPI folder creation in case user and group creation fails. Is that intended? Because I think, the folders should be created even if user and group creation fails. This way, at least root user can use FAPI.

AndreasFuchsSIT: Are we sure that we don't intend for this behavior ?
If not, however, I'd split and rearrange things a little:
`make_fapi_dirs && make_tss_user_and_group && chown_fapi_dirs || true`

tstruk:
@MichaelEckelSIT could you update the PR please

tstruk:
@MichaelEckelSIT thanks for the update. Could you squash the two commits into one please.

PR #1632, opened 2020-02-19
WIP: tcti-mssim: Add ability to set platform port

pdxjohnny:
So I was playing around with the simulator hoping to get it to be able to bind to any available port (port 0), to help us avoid the random number dance. But, I went to go grab the newly bound port from stdout, turns out its `write`ing to the tty...

```
write(1</dev/pts/25<char 136:25>>, "Platform server listening on por"..., 40Platform server listening on port 37543
```

Wasn't expecting this. I’m currently trying to run it from a Python subprocess [here](https://github.com/pdxjohnny/tpm2-pytss/blob/38ced89a9f55bef59b3ab55e5414e36aa6748a08/tpm2_pytss/util/simulator.py#L93-L151)

stracing the python that runs `tpm2-simulator` the ports are bound to be the above `write` never happens... werid. Not sure whats up. Or if I'll pursue this further, but thought I'd post the WIP here and then close later in case anyone else has interest, or I have an epiphany.

### DIFF
--- a/man/Tss2_Tcti_Mssim_Init.3.in
+++ b/man/Tss2_Tcti_Mssim_Init.3.in
@@ -67,10 +67,12 @@ The host may be an IPv4 address, an IPv6 address, or a host name. The port
 must be a valid uint16_t in string form. If a NULL
 .I conf
 string is provided by the caller then the default of
-"host=localhost,port=2321" is used. If either
+"host=localhost,port=2321,pport=2322" is used. If either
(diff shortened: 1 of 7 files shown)


tstruk:
@pdxjohnny since we switched to swtmp do you still need this?

pdxjohnny:
No I never was able to get the simulator building on Windows so this is dead

Issue #1573, opened 2019-12-09
ESYS_TestParams: logging expected errors


williamcroberts:
When using TestParams, certain error conditions are expected, but ESAPI always logs them.  I think we would want the "param 1 xxx not supported" errors (tpm:parameter(1):curve not supported) to be moved to something like verbose/debug level?

```
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001e6) 
WARNING:esys:src/tss2-esys/api/Esys_TestParms.c:269:Esys_TestParms_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_TestParms.c:95:Esys_TestParms() Esys Finish ErrorCode (0x000001e6) 
```

tstruk:
but these look like legitimate TPM errors, that should not be silenced, especially when ESAPI call returns an error.

williamcroberts:
> but these look like legitimate TPM errors, that should not be silenced, especially when ESAPI call returns an error.

No that error literally means, the curve is not supported. But you called test params to ask if the curve is supported, the routine can return:
1. yes
2. no
3. error so I couldn't tell you

We log on cases 2 and 3, which we should only log on case 3.

williamcroberts:
Example code,
```C
CK_RV tpm_is_rsa_keysize_supported(tpm_ctx *tctx, CK_ULONG test_size) {

    TPMT_PUBLIC_PARMS input = { 0 };

    input.type = TPM2_ALG_RSA;
    input.parameters.rsaDetail.exponent = 0;
    input.parameters.rsaDetail.scheme.scheme = TPM2_ALG_NULL;
    input.parameters.rsaDetail.symmetric.algorithm = TPM2_ALG_NULL;
    input.parameters.rsaDetail.keyBits = test_size;

    TSS2_RC rval = Esys_TestParms(tctx->esys_ctx,
            ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE, &input);
    if (rval != TSS2_RC_SUCCESS) {
        if ((rval & (TPM2_RC_P | TPM2_RC_1)) == (TPM2_RC_P | TPM2_RC_1)) {
            rval &= ~(TPM2_RC_P | TPM2_RC_1);
[...]

PR #1635, opened 2020-02-21
esys: don't print error msgs on expected errors

tstruk:
Fixes: #1573

### DIFF
--- a/src/util/aux_util.h
+++ b/src/util/aux_util.h
@@ -15,15 +15,27 @@ extern "C" {
 #define TPM2_ERROR_FORMAT "%s%s (0x%08x)"
 #define TPM2_ERROR_TEXT(r) "Error", "Code", r
 
+#define PARAM_ERR (TPM2_RC_P | TPM2_RC_1)
+#define not_expected_param_err(r) ({ \
+    TSS2_RC _r = r; \
+    if ((_r & PARAM_ERR) == PARAM_ERR) { \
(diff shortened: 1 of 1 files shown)


tstruk:
> Is this really something that we want to see applied to every TPM command ?
> I know we (Bill) want it for TestParams.
> But is RC_KEYSIZE maybe an actual error in some cases ?
That's why I'm looking for feedback on what we really want to suppress.
@williamcroberts what do you need for now? Let's start with minimum and then expand as needed.
I will also need to ad an else case with LOG_DEBUG() before it goes in.

tstruk:
> LGTM. Surprised that macro expansion that declares a variable inside an if worked in older gcc version like 5 too.
Statement expressions are supported since gcc3 I think:
https://gcc.gnu.org/onlinedocs/gcc-3.2.3/gcc/Statement-Exprs.html

tstruk:
@williamcroberts any comments on what we want to suppress?

williamcroberts:
> @williamcroberts any comments on what we want to suppress?

I completely misread this patch. This actually appears to be filtering at every TSS/TPM command level, we only care about TestParams AFAICT.

We want to make this only for TestParams.

williamcroberts:
I have a draft PR that addresses this:
https://github.com/tpm2-software/tpm2-tss/pull/1664

tstruk:
Superseded by https://github.com/tpm2-software/tpm2-tss/pull/1664

PR #1637, opened 2020-02-24
Some fixes and a lot of documentation

AndreasFuchsSIT:
This PR is the last before 2.4.0 unless there's any hickups.

Fixes:
- Some unnecessary async vs sync mode code removed
- Several return code fixes
- Unified curl usage
- Added root ca certs for all known vendors

@tstruk We removed the SSL_NO_VERIFY option from curl, please check that this still works with PTT.

After merge we can roll the next rc and go into release.


src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2434,85 +2434,79 @@ write_curl_buffer_cb(void *contents, size_t size, size_t nmemb, void *userp)
...
+        free(curl_buffer.buffer);
+    curl_easy_cleanup(curl);
+out_global_cleanup:
     curl_global_cleanup();
-    free(curl_buffer.buffer);
-    return r;
+out_memory:
+    return ret;
 }
+
```
tstruk: Please pay attention to these empty lines at the EOF because it looks like it upsets the CI:
https://travis-ci.org/tpm2-software/tpm2-tss/jobs/654483762#L2446

PR #1642, opened 2020-02-28
FAPI: Parse certificate data out of encoded format

tstruk:
Intel PTT EK certs on the host website are json encoded
where the cert data in "certificate:" tags. To get the cert
it needs to be parsed out of the json format before converting
to a binary PEM cert. An example can be seen here:
https://ekop.intel.com/ekcertservice/WVEG2rRwkQ7m3RpXlUphgo6Y2HLxl18h6ZZkkOAdnBE%3D

### DIFF
--- a/src/tss2-fapi/ifapi_get_intl_cert.c
+++ b/src/tss2-fapi/ifapi_get_intl_cert.c
@@ -13,6 +13,7 @@
 #include <openssl/buffer.h>
 #include <openssl/evp.h>
 #include <openssl/sha.h>
+#include <json-c/json.h>
 
 #include "fapi_crypto.h"
 #include "ifapi_helpers.h"
@@ -112,7 +113,8 @@ err:
     return NULL;
 }
 
-char *base64_encode(const unsigned char* buffer)
+static char *
+base64_encode(const unsigned char* buffer)
 {
     BIO *bio, *b64;
     BUF_MEM *buffer_pointer;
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/api/Fapi_Provision.c
```diff
@@ -932,15 +932,46 @@ Fapi_Provision_Finish(FAPI_CONTEXT *context)
...
                 r = ifapi_get_intl_ek_certificate(context, &pkey->public, &cert_buffer,
-                                                  &cert_size);
+                                                  &buffer_size);
                 goto_if_error_reset_state(r, "Get certificates", error_cleanup);
 
-                r = ifapi_cert_to_pem(cert_buffer, cert_size, &command->pem_cert,
-                                      NULL, &public);
+                LOGBLOB_DEBUG(cert_buffer, buffer_size, "%s", "Certificate");
+                /* Parse certificate data out of XML */
+                cert_start = strstr((char *)cert_buffer, "\"certificate\":\"");
```
AndreasFuchsSIT: Since this is very intl-specific, this json and b64 decoding should be done inside the get_intl_cert.c

src/tss2-fapi/api/Fapi_Provision.c
```diff
@@ -932,15 +932,46 @@ Fapi_Provision_Finish(FAPI_CONTEXT *context)
...
+                if (cert_start == NULL) {
+                    SAFE_FREE(cert_buffer);
+                    goto_error_reset_state(r, TSS2_FAPI_RC_GENERAL_FAILURE,
+                                           "Invalid EK cert data", error_cleanup);
+                }
+                cert_start += 15;
+                cert_end = strstr((char *)cert_buffer, "\"}");
```
AndreasFuchsSIT: You could also use libjson-c to decode the certs to be more robust against whitespaces added and removed on this stuff in the future.

src/tss2-fapi/ifapi_get_intl_cert.h
```diff
@@ -12,4 +12,11 @@ ifapi_get_intl_ek_certificate(
     unsigned char ** buffer,
     size_t *cert_size);
 
+char *
+base64_encode(const unsigned char* buffer);
```
AndreasFuchsSIT: This should be somewhat namespaced imho. its a way to generic name otherwise.

However, if the JSON/B64 decoding is moved into ifapi_get_intl_cert these two functions can be made static.

tstruk: Both [en|de]code function changed to static

src/tss2-fapi/api/Fapi_Provision.c
tstruk: changed to libjson-c parsing

src/tss2-fapi/api/Fapi_Provision.c
tstruk: moved to get_intl_cert.c

tstruk:
@AndreasFuchsSIT thanks for reviewing. All feedback should be addresses in the updated version.

PR #1647, opened 2020-03-03
Fapi simplification

williamcroberts:
Just some code that:
- simplifies the OSSL cleanup paths in FAPI
- Diffrentiates OSSL signature verification failure from OSSL errors.

### DIFF
--- a/src/tss2-fapi/fapi_crypto.c
+++ b/src/tss2-fapi/fapi_crypto.c
@@ -872,8 +872,12 @@ ecdsa_verify_signature(
     eccKey = EVP_PKEY_get1_EC_KEY(publicKey);
 
     /* Try to verify the signature using ECDSA, note that param 0 is unused */
-    if (!ECDSA_verify(0, digest, digestSize, signature, signatureSize, eccKey)) {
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/fapi_crypto.c
```diff
@@ -876,15 +876,14 @@ ecdsa_verify_signature(
                      "Invalid ECDSA DER signature", error_cleanup);
     }
 
-    eccKey = EVP_PKEY_get1_EC_KEY(publicKey);
+    eccKey = EVP_PKEY_get0_EC_KEY(publicKey);
```
tstruk: Looks like there is no EVP_PKEY_get0_EC_KEY or EVP_PKEY_get0_RSA in OpenSSL 1.0.2 that comes with ubuntu 16

williamcroberts: Oh yeah, easy enough to add compat layer. I'm surprised they have they 1 versions in 1.0.2.

williamcroberts: Probably better just to leave that in place, and when OSSL < 1.1.0 support is no longer needed, we can just update it. Id rather have consistent code, than adding a compat layer. The other patch(s) make sense though

Issue #1641, opened 2020-02-27
Wrong hmac calculation for response when executing TPM2_CC_Clear command


idesai:
If this command is authorized using lockoutAuth, the HMAC in the response shall use the new lockoutAuth value (that is, the Empty Buffer) when computing the response HMAC. 

The fix may be to [conditionally](https://github.com/tpm2-software/tpm2-tss/blob/d8fd8c0ad935fb2022b9967d613c4776a35955e1/src/tss2-sys/api/Tss2_Sys_GetRspAuths.c#L77) retrieve response auths when executing clear command.

tstruk:
The spec only mentions the lockout auth, but I wonder if this is true in all cases, since tpm_clear sets all ownerAuth, endorsementAuth, and lockoutAuth to an Empty Buffer. I have a fix PR for this, could you have a look please
https://github.com/tpm2-software/tpm2-tss/pull/1648

tstruk:
The spec is correct, it only needs to be done for lockoutAuth.

PR #1648, opened 2020-03-03
Tpm_clear auth

tstruk:
After tpm2_clear command is executed it sets all ownerAuth, endorsementAuth, and lockoutAuth to the Empty Buffer and then this is used for a response auth calculation.
This requires to recalculate the esys session auth value after tpm2_clear is executed or the calculated response HMAC value will be invalid.

Fixes: #1641

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -145,6 +145,7 @@ if ESAPI
 ESYS_TESTS_INTEGRATION_DESTRUCTIVE = \
     test/integration/esys-change-eps.int \
     test/integration/esys-clear.int \
+    test/integration/esys-tpm-clear-auth.int \
     test/integration/esys-clear-session.int \
     test/integration/esys-field-upgrade.int \
     test/integration/esys-firmware-read.int \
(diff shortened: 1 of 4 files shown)


tstruk:
@JuergenReppSIT I removed the call to iesys_DeleteAllResourceObjects() from Esys_Clear_Finish(), and changed the test to flush the session obj after each Esys_Clear call. Somehow I thought that the tpm_clear will also remove the session objects (the spec says that it removes all transient objects). I returned skip from the test, because, as the comment stated, the TPM caps were different between before and after the tests and it failed here: https://github.com/tpm2-software/tpm2-tss/blob/master/script/int-log-compiler.sh#L379
even though the test itself returned success. It looks like flushing the session fixes that and I don't need to do that anymore. PR updated. Thanks for reviewing.

JuergenReppSIT:
The PR is OK for me now.

tstruk:
> The PR is OK for me now.

Thank you @JuergenReppSIT

Issue #1650, opened 2020-03-09
fapi: Invalid JSON in default config


pdxjohnny:
https://github.com/tpm2-software/tpm2-tss/blob/9dd5cc404ce79896e70a7e5d9a8cbe013d978100/dist/fapi-config.json.in#L8

The last comma shouldn't be there. It will cause some JSON parsers to throw errors. For example Python.

PR #1651, opened 2020-03-09
FAPI: Ensure default config adheres to JSON spec

pdxjohnny:
Remove a trailing comma from the last line of the default FAPI config
file, making it valid JSON.

Fixes: #1650

### DIFF
--- a/dist/fapi-config.json.in
+++ b/dist/fapi-config.json.in
@@ -5,5 +5,5 @@
      "system_dir": "@localstatedir@/lib/tpm2-tss/system/keystore",
      "tcti": "",
      "system_pcrs" : [],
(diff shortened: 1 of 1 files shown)


pdxjohnny:
@tstruk It would be great if this could make it in 2.4, if it doesn't then we won't be able to parse the defaults from the json file installed in `SYSCONFDIR / tpm2-tss / fapi-config.json` from the Python API.

tstruk:
> @tstruk It would be great if this could make it in 2.4, if it doesn't then we'd have to add extra logic to be able to parse the defaults from the json file installed in `SYSCONFDIR / tpm2-tss / fapi-config.json` from the Python API.

@pdxjohnny will do. Thanks

PR #1652, opened 2020-03-11
test/esys: Check unsupported symmetric  cipher used for Esys_EncryptD…

JuergenReppSIT:
…ecrypt test.

The key type TPM2_ALG_SYMCIPHER might not be supported. In this case the test will be
skipped.

### DIFF
--- a/test/integration/esys-encrypt-decrypt.int.c
+++ b/test/integration/esys-encrypt-decrypt.int.c
@@ -207,6 +207,12 @@ test_esys_encrypt_decrypt(ESYS_CONTEXT * esys_context)
                     &outPrivate2,
                     &outPublic2,
                     &creationData2, &creationHash2, &creationTicket2);
+
+    if (r == 0x2c2) { /*<< tpm:parameter(2):inconsistent attributes */
+        LOG_WARNING("Unsupported symmetric cipher.");
+        failure_return = EXIT_SKIP;
+        goto error;
+    }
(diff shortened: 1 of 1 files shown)


test/integration/esys-encrypt-decrypt.int.c
```diff
@@ -207,6 +207,12 @@ test_esys_encrypt_decrypt(ESYS_CONTEXT * esys_context)
                     &outPrivate2,
                     &outPublic2,
                     &creationData2, &creationHash2, &creationTicket2);
+
+    if (r == 0x2c2) { /*<< tpm:parameter(2):inconsistent attributes */
+        LOG_WARNING("Unsupported symmetric cipher.");
+        failure_return = EXIT_SKIP;
+        goto error;
```
tstruk: All these goto for 'not supported' cases are redundant.

AndreasFuchsSIT: This goto is only for 0x2c2 the goto_if_error triggers for everything else.

AndreasFuchsSIT:
We should probably get this into 2.4.0 and backport to 2.x

test/integration/esys-encrypt-decrypt.int.c
tstruk: for everything else including 0x2c2

tstruk:
> We should probably get this into 2.4.0 and backport to 2.x

I will merge it as soon as the CI passes.

Issue #1534, opened 2019-10-08
Bring Tss2_Sys_ReadClock() upto the Rev 1.38 specs.


tstruk:
In Rev 1.38 both audit or encrypt are allowed.

AndreasFuchsSIT:
This is not fixed yet.
The header of SAPI needs to be extended by auth-arrays...

See:
https://github.com/tpm2-software/tpm2-tss/blob/master/include/tss2/tss2_sys.h#L1788-L1797

tstruk:
@AndreasFuchsSIT right, Thanks for the reminder :+1: 
see https://github.com/tpm2-software/tpm2-tss/pull/1663

PR #1661, opened 2020-03-17
Bump libtss2-sys soname version

tstruk:


diabonas:
Does 1414f73902cd0c64e447522c64c8e439fa2074b7 actually break the documented ABI, i.e. is it necessary to change, or at least recompile, applications using `Tss2_Sys_ReadClock`? To me it looks like this is just an extension of the current ABI, so applications unaware of the fact that audit sessions are now allow will continue working as expected.

If that is the case, a major soname version bump would not be necessary. It would in theory be possible to bump the minor version instead to signal a backwards-compatible change to applications relying on the new behaviour. However library versioning is [somewhat complicated and not really standartised](https://autotools.io/libtool/version.html). As a result, usually only the major version is used in the soname, at least on Linux. I'd argue that applications needing a specific newer (backwards-compatible) version of the library should check that at compile time using pkg-config.

diabonas:
Gotcha, https://github.com/tpm2-software/tpm2-tss/issues/1534#issuecomment-600538804 is certainly a reason for a soname bump.

PR #1664, opened 2020-03-18
Esys_TestParams: silence expected errors

williamcroberts:
The TestParams code is used to test paramters against the TPM to
understand if a specific algorithm and scheme is supported. Thus,
certain errors are expected. The expected errors are:
1. From any layer
2. Are Format 1 errors
3. Are P1 or Parameter 1 errors
4 are one of:
  - RC_CURVE
  - RC_VALUE
  - RC_ASSYMETRIC
  - RC_KEY_SIZE

Fixes: #1573

### DIFF
--- a/src/util/aux_util.h
+++ b/src/util/aux_util.h
@@ -6,6 +6,9 @@
 #ifndef AUX_UTIL_H
 #define AUX_UTIL_H
 
+#include <stdbool.h>
+
+#include "tss2_tpm2_types.h"
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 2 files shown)


src/util/aux_util.h
```diff
@@ -84,6 +89,45 @@ extern "C" {
         r_max = r; \
     }
 
+static inline TSS2_RC
```
tstruk: Since this is only needed in Esys_TestParms.c maybe it should be defined there?

williamcroberts: Thats fine, I struggled with that decision myself. I figured though, that this is generally usable. But I have no strong opinion, so Ill just move it over. We can always move it again if needed.

williamcroberts: Most routines like this seem to be not in the API implementation folders. The only static routines I see are the store_input routines. Which also have a slightly different style in that the return type and function name are on one line rather than down another level...

src/util/aux_util.h
```diff
@@ -6,6 +6,10 @@
 #ifndef AUX_UTIL_H
 #define AUX_UTIL_H
 
+#include <stdbool.h>
+
+#include "tss2_tpm2_types.h"
+
```
williamcroberts: superflous newline...

Issue #1658, opened 2020-03-17
FAPI: Missing return code checking


AndreasFuchsSIT:
We need to get the return value into `r` here:
https://github.com/tpm2-software/tpm2-tss/blob/c94a14e65cd482cb9b3767a49535976cfd66a687/src/tss2-fapi/ifapi_keystore.c#L1412-L1413

Issue #1655, opened 2020-03-13
Wrong sizeof


AndreasFuchsSIT:
Should be sizeof(TPMU_HA).
https://github.com/tpm2-software/tpm2-tss/blob/c990d03b6e2eaad73627c354d1ad827f175b43a2/src/tss2-fapi/tpm_json_serialize.c#L1250
Thanks to coverity

PR #1665, opened 2020-03-19
FAPI: Add session usage to FAPI provisioning.

JuergenReppSIT:
The auth values were not encrypted during provisioning. Session usage
to ensure encryption was added.

* A session with the EK used as session key is created for the
  SRK generation. Thus now the EK has to be generated before the SRK.
* A session with the SRK used as session key is created for the other
  commands executed during provisioning.


tstruk:
@AndreasFuchsSIT I think this will be third one to backport. We need two more PRs for:
https://github.com/tpm2-software/tpm2-tss/issues/1655 and
https://github.com/tpm2-software/tpm2-tss/issues/1658

AndreasFuchsSIT:
Yep, very true...

PR #1666, opened 2020-03-19
CI: reduce code duplication in INT_LOG_COMPILER

tstruk:
Removed most of the code duplicated between
int-log-compiler-mssim.sh and int-log-compiler-swtpm.sh.
Moved the common part to int-log-compiler-common.sh
and split the rest between fint-log-compiler.sh
for FAPI integration test, and int-log-compiler.sh
for the rest. That way we don't generate EKs, certs, etc
for SAPI or ESAPI tests, where none of these are needed.


JuergenReppSIT:
looks good to me

PR #1667, opened 2020-03-20
build: update git.mk

tstruk:
Add few more files to the auto-generated .gitignore file.

### DIFF
--- a/git.mk
+++ b/git.mk
@@ -120,6 +120,8 @@ GITIGNORE_MAINTAINERCLEANFILES_M4_LIBTOOL = \
 			ltsugar.m4 \
 			ltversion.m4 \
 			lt~obsolete.m4 \
+			ax_recursive_eval.m4 \
+			libgcrypt.m4 \
 		; do echo "$$MACRO_DIR/$$x"; done; \
 	 fi`
 #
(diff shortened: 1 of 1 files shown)


git.mk
```diff
@@ -120,6 +120,8 @@ GITIGNORE_MAINTAINERCLEANFILES_M4_LIBTOOL = \
 			ltsugar.m4 \
 			ltversion.m4 \
 			lt~obsolete.m4 \
+			ax_recursive_eval.m4 \
+			libgcrypt.m4 \
```
AndreasFuchsSIT: These should be added here:
https://github.com/tpm2-software/tpm2-tss/blob/9f0584a1e093d9aa0b37b9e3b17359d372b4b43d/Makefile.am#L27-L50

git.mk
```diff
@@ -372,6 +374,8 @@ $(srcdir)/.gitignore: Makefile.am $(top_srcdir)/git.mk $(top_srcdir)/configure.a
 			"*.gcno" \
 			"*.$(OBJEXT)" \
 			$(DEPDIR) \
+			"*.pem" \
+			"*_state[1|2]" \
```
AndreasFuchsSIT: These should be added to CLEANFILES or moved to maybe better moved to the tempdir.

tstruk:
I made these changes because the git.mk doesn't work for me as expected. Every time I run git status I get output cluttered by a bunch of files that the git.mk didn't cover. I think we have all the required files in CLEANFILES or make distcheck would fail. I have to say that I was more happy with our handcrafted .gitignore file. We need to improve the git.mk or switch back to the old way. Why do we care how much git.mk will diverge from the upstream version?

AndreasFuchsSIT:
Well, my idea was that improvements to upstream could be easily merged here if we keep the delta to a minimum.
Since upstream seems dormant though, I guess we can do what we want...
So please, go ahead.

tstruk:
> Well, my idea was that improvements to upstream could be easily merged here if we keep the delta to a minimum.

That's a fair point, but if there will be any substantial changes/improvements upstream we can always merge them and resolve potential conflicts locally. That shouldn't be too hard to do.

PR #1668, opened 2020-03-21
Ci tcti config test

tstruk:
* Only build TCTI unit test if the type is enabled
* Add CI build tests for all possible TCTI configurations


tstruk:
Fixes: #1671

PR #1672, opened 2020-03-23
esys test: Add check whether session can be used by Esys_ReadClock.

JuergenReppSIT:
Use of an authorization session might not be supported by TPMs before revision 1.38

### DIFF
--- a/test/integration/sapi-read-clock.int.c
+++ b/test/integration/sapi-read-clock.int.c
@@ -16,6 +16,7 @@
 
 #define LOGMODULE test
 #include "util/log.h"
+#include "test-esapi.h"
 #include "test.h"
 /*
  * This is an incredibly simple test to create the most simple session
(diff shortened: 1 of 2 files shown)


tstruk:
Thanks @JuergenReppSIT. Do we need similar condition in https://github.com/tpm2-software/tpm2-tss/blob/master/test/integration/sapi-read-clock.int.c#L90

JuergenReppSIT:
@tstrunk Y

> Thanks @JuergenReppSIT. Do we need similar condition in https://github.com/tpm2-software/tpm2-tss/blob/master/test/integration/sapi-read-clock.int.c#L90

You're right. In our testfarm I did only run esys and fapi tests  for physical TPMs,
So I didn't recognize the problem .

tstruk:
> You're right. In our testfarm I did only run esys and fapi tests for physical TPMs,
> So I didn't recognize the problem

Could you add the same condition to sapi test and update this commit please. There is no need for two commits for this. Thanks

test/integration/sapi-read-clock.int.c
```diff
@@ -16,6 +16,7 @@
 
 #define LOGMODULE test
 #include "util/log.h"
+#include "test-esapi.h"
```
tstruk: Is this for EXIT_SKIP only?
I would prefer to use 77 magic number instead of including esapi header here. I will fix that. Thanks

PR #1675, opened 2020-03-23
FAPI: Prevent overwriting ob objects in key store.

JuergenReppSIT:
* During provisioning and object creation it will be checked whether
  the object already exists in key store.
* The test fapi-nv-set-bits which did include provisioning two times was
  modified to check whether second provisioning was recognized.

### DIFF
--- a/src/tss2-fapi/ifapi_keystore.c
+++ b/src/tss2-fapi/ifapi_keystore.c
@@ -627,6 +627,61 @@ cleanup:
     return r;
 }
 
+/**  Check whether the key path for a new object does not exist in key store.
+ *
+ * To prevent overwriting of objects the functions returns an error
+ * if the object is already stored in key store.
+ * The FAPI path will be expanded to absolute path appropriate for
+ * the object to be checked.
+ *
+ * @param[in] keystore The key directories and default profile.
+ * @param[in] path The relative path of the object. For keys the path will
+ *           expanded if possible.
+ * @param[in] object The object to be checked.
+ * @retval TSS2_RC_SUCCESS if the object does not exist and a new object
+ *         can be written.
+ * @retval TSS2_FAPI_RC_PATH_ALREADY_EXISTS: if the object exists in key store.
+ * @retval TSS2_FAPI_RC_MEMORY: if memory could not be allocated to hold the output data.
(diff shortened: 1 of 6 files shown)


src/tss2-fapi/ifapi_keystore.h
```diff
@@ -1,4 +1,3 @@
-/* SPDX-License-Identifier: BSD-2-Clause */
```
tstruk: Don't we need License on this file anymore?

src/tss2-fapi/ifapi_keystore.c
```diff
@@ -627,6 +627,58 @@ ifapi_keystore_store_async(
...
+ * @param[in] object The object to be checked.
+ * @retval TSS2_RC_SUCCESS if the object does not exist.
+ * @retval TSS2_FAPI_RC_PATH_ALREADY_EXISTS: if the object exists.
+ * @retval TSS2_FAPI_RC_MEMORY: if memory could not be allocated to hold the output data.
+ * @retval TSS2_FAPI_RC_BAD_VALUE if an invalid value was passed into
+ *         the function.
+ * @retval TSS2_FAPI_RC_BAD_REFERENCE a invalid null pointer is passed.
+ */
+TSS2_RC
+ifapi_keystore_object_does_not_exist(
```
tstruk: The function name doesn't reflect the behavior or match the comments.
The function checks if object exists already, and so say the comments.
Should the name of the function be ifapi_keystore_object_exists()?

JuergenReppSIT: The function returns TSS2_RC_SUCCESS if the object does not exist in keystore. So I would say the behaviour is OK.  So it was possible to use the standard mechanism for error checking instead of "if (ifapi_keystore_object_exists(..) == TSS2_RC_SUCCESS) error_handling."
Would it be sufficient if the comment is changed and describes exactly this behaviour or would you prefer the example above?

tstruk: It would be good to make that clear in the comment.

tstruk:
I'm guessing this one needs to be cherry-picked to 2.4.x right?

PR #1680, opened 2020-03-24
Some minor things regarding good style

AndreasFuchsSIT:
- Use modern flex-array style
- make getenv() returns const
- Remove unused macro
- Use strtok_r instead of strtok

### DIFF
--- a/src/tss2-fapi/api/Fapi_Provision.c
+++ b/src/tss2-fapi/api/Fapi_Provision.c
@@ -508,7 +508,7 @@ Fapi_Provision_Finish(FAPI_CONTEXT *context)
 
         statecase(context->state, PROVISION_READ_CERT);
             TPM2B_PUBLIC public_key;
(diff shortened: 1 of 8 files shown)


AndreasFuchsSIT:
P.S. Nothing for backporting here...

tstruk:
main FreeBSD:freebsd-13-0-snap build failed with:
```
../../test/integration/sapi-session-util.c:117:5: error: 'memcpy' will always overflow; destination buffer has size 0, but size argument is 1 [-Werror,-Wfortify-source]
    memcpy(&sessionAttributesByteBuffer.buffer, &pSessionDataIn->sessionAttributes,
    ^
1 error generated.
```

AndreasFuchsSIT:
Yes, I'm stupid... ;-)
But at least we don't have that variable of type `TPM2B` anymore now. :-)

Issue #1681, opened 2020-03-24
Doxygen warning


tstruk:
warning: tag INPUT: input source `../../test/integration/fapi-key-create-policy-password-sign.int.c' does not exist
tpm2-tss/doc/doxygen.dox:3145: warning: documented symbol `test_fapi_key_create_policy_password_sign' was not declared or defined.

The fapi-key-create-policy-password-sign.int.c file is missing in Makefile-test.am

tstruk:
Fixed with https://github.com/tpm2-software/tpm2-tss/pull/1689

PR #1683, opened 2020-03-24
FAPI: Add test function to doxygen

JuergenReppSIT:
The function test_fapi_key_create_policy_password_sign was added to dox file.
Fixes: #1681


tstruk:
Fixes: #1681

JuergenReppSIT:
> What about this file then:
> https://github.com/tpm2-software/tpm2-tss/blob/master/test/integration/fapi-key-create-policy-password-sign.int.c
> It is never used, except in here:
> https://github.com/tpm2-software/tpm2-tss/blob/master/Doxyfile.in#L865
> should it be deleted?

yes the file can be removed. The test test_fapi_key_create_policy_password_sign is covered in
fapi-key-create-policies-sign.int.c. I will reject the PR.

PR #1688, opened 2020-03-25
tpmclient: update swtpm return code to TSS2_TCTI_RC_BAD_REFERENCE

diabonas:
Commit bc70b835c2acbdff0c81ce97e21d9fb65354d309 changed the return code if a TCTI cannot be loaded from `TSS2_TCTI_RC_BAD_CONTEXT` to `TSS2_TCTI_RC_BAD_REFERENCE`, but did not update the [tpmclient](https://github.com/tpm2-software/tpm2-tss/blob/master/test/tpmclient/tpmclient.int.c) test. This leads to the following error when running the integration test suite with ibmswtpm2 instead of swtpm:
```
Execute the test script
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:241:Tss2_Tcti_Swtpm_Reset() Failed to reset TPM: 0xa0005
Script returned 1
FAIL test/tpmclient/tpmclient.int (exit status: 1)
```

### DIFF
--- a/test/tpmclient/tpmclient.int.c
+++ b/test/tpmclient/tpmclient.int.c
@@ -151,8 +151,8 @@ static TSS2_RC TpmReset()
 #ifdef TCTI_SWTPM
     rval = Tss2_Tcti_Swtpm_Reset( resMgrTctiContext );
 
-    /* If TCTI is not swtpm, bad context is returned. */
-    if (rval != TSS2_TCTI_RC_BAD_CONTEXT) {
+    /* If TCTI is not swtpm, bad reference is returned. */
+    if (rval != TSS2_TCTI_RC_BAD_REFERENCE) {
         return rval;
     }
 #endif /* TCTI_SWTPM */


AndreasFuchsSIT:
Could you tell me, why current master is not failing right now though ?

diabonas:
The CI runs the full test suite with swtpm because it is available in the Docker container and swtpm is the preferred simulator choice for the integration tests. There are [some additional tests](https://github.com/tpm2-software/tpm2-tss/blob/3fe4cc103afccd7daa55d0a861c5215f333c2efb/.ci/docker.run#L102-L124) that cover the different TCTIs, but they only run the respective unit test, not the integration tests (including tpmclient). In short, the error only occurs if tpmclient is run with the ibmswtpm2 simulator, but that currently does not happen in the CI build.

tstruk:
Is the ibmswtpm2 what we call mssim? The FreeBSD build integration tests run against mssim only.

diabonas:
Yes, `ibmswtpm2` == `mssim`. The FreeBSD integration tests indeed use mssim, but they [completely disable swtpm with `./configure --enable-tcti-swtpm=no`](https://github.com/tpm2-software/tpm2-tss/blob/3fe4cc103afccd7daa55d0a861c5215f333c2efb/.cirrus.yml#L26), so the problematic code is not built because it is wrapped in an [`#ifdef TCTI_SWTPM`](https://github.com/tpm2-software/tpm2-tss/blob/master/test/tpmclient/tpmclient.int.c#L151-L158).

diabonas:
Not sure why the status indicator got stuck, all checks have already passed: https://travis-ci.org/github/tpm2-software/tpm2-tss/builds/666748974

tstruk:
@diabonas I noticed this yesterday. There must be something wrong this travis.

tstruk:
I see why it is failing, and it is not in tpm_client. According to the spec TSS2_TCTI_RC_BAD_REFERENCE should only be returned if the context is NULL.
In this case it is not, it just not the right tcti (mssim instead of swtpm).
Instead of changing the tpm_client we need to fix the cast functions.
This applies to all tcti_device, tcti_swtpm, and tcti_mssim.

AndreasFuchsSIT:
So most tctis don't check or don't differentiate between context == NULL and a bad magic number ?

tstruk:
> So most tctis don't check or don't differentiate between context == NULL and a bad magic number ?

Yes, until now BAD_CONTEXT is returned in both cases. I changed it in https://github.com/tpm2-software/tpm2-tss/commit/bc70b835c2acbdff0c81ce97e21d9fb65354d309 to return BAD_REFERENCE in case of ctx == NULL, but the tcti_cast function like this:
https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-tcti/tcti-swtpm.c#L67
still didn't handle it correctly and if the magic number we wrong it returned NULL, which caused the common_checks function to return wrong RC. I will have a PR ready with a fix soon.

tstruk:
@diabonas please see a PR https://github.com/tpm2-software/tpm2-tss/pull/1690
What we are missing in our CI though, is a test with all TCTIs enabled and a setup forcing it to use mssim via TPM20_TCTI_* variables. I will add one.

diabonas:
@tstruk Unfortunately #1690 currently doesn't work with tpmclient in combination with mssim at the moment either, see my comment there. Anyway, I am closing my PR because a different approach to fixing the issue is necessary.

PR #1690, opened 2020-03-25
tcti: fix invalid error code in case of wrong magic

tstruk:
The commit bc70b83 changed TCTIs to return TSS2_TCTI_RC_BAD_REFERENCE
instead of TSS2_TCTI_RC_BAD_CONTEXT when the ctx was NULL, to make it
spec compliant. After that another problem was found, where the
tcti cast functions still didn't handle it correctly.
The problem was that if the magic number was wrong a NULL was returned,
which caused the common_checks function to return wrong return code.
This PR changes the casts function to overwrite the magic to a known
bad value and common_check function to check for the expected magic
explicitly passed by each tcti type and return TSS2_TCTI_RC_BAD_CONTEXT
with compliance with the specification.

### DIFF
--- a/src/tss2-tcti/tcti-swtpm.c
+++ b/src/tss2-tcti/tcti-swtpm.c
@@ -56,18 +56,17 @@
 
 /*
  * This function wraps the "up-cast" of the opaque TCTI context type to the
- * type for the swtpm TCTI context. The only safeguard we have to ensure this
- * operation is possible is the magic number in the swtpm TCTI context.
- * If passed a NULL context, or the magic number check fails, this function
- * will return NULL.
+ * type for the mssim TCTI context. If passed a NULL context the function
+ * returns a NULL ptr. The function doesn't check magic number anymore
+ * It should checked by the appropriate tcti_common_checks.
  */
 TSS2_TCTI_SWTPM_CONTEXT*
 tcti_swtpm_context_cast (TSS2_TCTI_CONTEXT *tcti_ctx)
 {
-    if (tcti_ctx != NULL && TSS2_TCTI_MAGIC (tcti_ctx) == TCTI_SWTPM_MAGIC) {
-        return (TSS2_TCTI_SWTPM_CONTEXT*)tcti_ctx;
-    }
-    return NULL;
+    if (tcti_ctx == NULL)
+        return NULL;
+
+    return (TSS2_TCTI_SWTPM_CONTEXT*)tcti_ctx;
 }
 /*
  * This function down-casts the swtpm TCTI context to the common context
@@ -120,6 +119,10 @@ TSS2_RC tcti_control_command (
 
     if (tcti_swtpm == NULL) {
         return TSS2_TCTI_RC_BAD_REFERENCE;
+    }
+
+    if (TSS2_TCTI_MAGIC (tcti_swtpm) != TCTI_SWTPM_MAGIC) {
+        return TSS2_TCTI_RC_BAD_CONTEXT;
     }
 
     if (cmd_sdu == NULL && cmd_sdu_len != 0) {
@@ -255,7 +258,7 @@ tcti_swtpm_transmit (
     tpm_header_t header;
     TSS2_RC rc;
 
-    rc = tcti_common_transmit_checks (tcti_common, cmd_buf);
+    rc = tcti_common_transmit_checks (tcti_common, cmd_buf, TCTI_SWTPM_MAGIC);
     if (rc != TSS2_RC_SUCCESS) {
         return rc;
     }
@@ -316,7 +319,7 @@ tcti_swtpm_set_locality (
     TSS2_TCTI_COMMON_CONTEXT *tcti_common = tcti_swtpm_down_cast (tcti_swtpm);
     TSS2_RC rc;
 
-    rc = tcti_common_set_locality_checks (tcti_common);
+    rc = tcti_common_set_locality_checks (tcti_common, TCTI_SWTPM_MAGIC);
     if (rc != TSS2_RC_SUCCESS) {
         return rc;
     }
@@ -376,7 +379,7 @@ tcti_swtpm_receive (
     TSS2_RC rc;
     int ret;
 
-    rc = tcti_common_receive_checks (tcti_common, response_size);
+    rc = tcti_common_receive_checks (tcti_common, response_size, TCTI_SWTPM_MAGIC);
     if (rc != TSS2_RC_SUCCESS) {
         return rc;
     }
--- a/src/tss2-tcti/tcti-common.c
+++ b/src/tss2-tcti/tcti-common.c
@@ -38,10 +38,15 @@ tcti_common_down_cast (TSS2_TCTI_COMMON_CONTEXT *ctx)
 
 TSS2_RC
 tcti_common_cancel_checks (
-    TSS2_TCTI_COMMON_CONTEXT *tcti_common)
+    TSS2_TCTI_COMMON_CONTEXT *tcti_common,
+    uint64_t magic)
 {
     if (tcti_common == NULL) {
         return TSS2_TCTI_RC_BAD_REFERENCE;
+    }
+
+    if (TSS2_TCTI_MAGIC(tcti_common) != magic) {
+        return TSS2_TCTI_RC_BAD_CONTEXT;
     }
 
     if (tcti_common->state != TCTI_STATE_RECEIVE) {
@@ -53,10 +58,15 @@ tcti_common_cancel_checks (
 TSS2_RC
 tcti_common_transmit_checks (
     TSS2_TCTI_COMMON_CONTEXT *tcti_common,
-    const uint8_t *command_buffer)
+    const uint8_t *command_buffer,
+    uint64_t magic)
 {
     if (command_buffer == NULL || tcti_common == NULL) {
         return TSS2_TCTI_RC_BAD_REFERENCE;
+    }
+
+    if (TSS2_TCTI_MAGIC(tcti_common) != magic) {
+        return TSS2_TCTI_RC_BAD_CONTEXT;
     }
 
     if (tcti_common->state != TCTI_STATE_TRANSMIT) {
@@ -69,10 +79,15 @@ tcti_common_transmit_checks (
 TSS2_RC
 tcti_common_receive_checks (
     TSS2_TCTI_COMMON_CONTEXT *tcti_common,
-    size_t *response_size)
+    size_t *response_size,
+    uint64_t magic)
 {
     if (response_size == NULL || tcti_common == NULL) {
         return TSS2_TCTI_RC_BAD_REFERENCE;
+    }
+
+    if (TSS2_TCTI_MAGIC(tcti_common) != magic) {
+        return TSS2_TCTI_RC_BAD_CONTEXT;
     }
 
     if (tcti_common->state != TCTI_STATE_RECEIVE) {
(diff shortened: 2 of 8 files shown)


src/tss2-tcti/tcti-swtpm.c
```diff
@@ -56,18 +56,23 @@
...
 {
-    if (tcti_ctx != NULL && TSS2_TCTI_MAGIC (tcti_ctx) == TCTI_SWTPM_MAGIC) {
-        return (TSS2_TCTI_SWTPM_CONTEXT*)tcti_ctx;
-    }
-    return NULL;
+    if (tcti_ctx == NULL)
+	return NULL;
+
+    if (TSS2_TCTI_MAGIC (tcti_ctx) != TCTI_SWTPM_MAGIC)
+        TSS2_TCTI_MAGIC (tcti_ctx) = TCTI_BAD_MAGIC;
```
diabonas: Unfortunately this doesn't work for tpmclient because it [first tries to call `Tss2_Tcti_Swtpm_Reset`](https://github.com/tpm2-software/tpm2-tss/blob/3879a0b41bbecd9b4e229b3a73b6fd3835d0370a/test/tpmclient/tpmclient.int.c#L152) on the given TCTI context and [falls back to `tcti_platform_command(...,MS_SIM_POWER_OFF/ON)`](https://github.com/tpm2-software/tpm2-tss/blob/3879a0b41bbecd9b4e229b3a73b6fd3835d0370a/test/tpmclient/tpmclient.int.c#L161) in case that fails. This cast irrevocably destroys non-swtpm TCTI contexts if they are passed into swtpm TCTI functions, so they are unusable afterwards even when used with their appropriate TCTI functions. This seems dangerous because other applications might also expect these functions not to alter the passed TCTI context.

diabonas:
Another issue with this is that `Tss2_Tcti_Swtpm_Reset` now returns [`TSS2_TCTI_RC_IO_ERROR`](https://github.com/tpm2-software/tpm2-tss/blob/2dd9be51866129584bf81e31fe253744cca18b0d/src/tss2-tcti/tcti-swtpm.c#L141) if a mssim TCTI context is used. This might make sense, but the error check in `tpmclient.int.c` must be updated from the current [`TSS2_TCTI_RC_BAD_CONTEXT`](https://github.com/tpm2-software/tpm2-tss/blob/3879a0b41bbecd9b4e229b3a73b6fd3835d0370a/test/tpmclient/tpmclient.int.c#L155).

Without this change, tpmclient currently fails with the following error:
```
Execute the test script
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:145:tcti_control_command() Failed to connect to control socket. 
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:246:Tss2_Tcti_Swtpm_Reset() Failed to reset TPM: 0xa000a 
Script returned 1
FAIL test/tpmclient/tpmclient.int (exit status: 1)
```

When changing `TSS2_TCTI_RC_BAD_CONTEXT` to `TSS2_TCTI_RC_IO_ERROR` in [l. 155](https://github.com/tpm2-software/tpm2-tss/blob/3879a0b41bbecd9b4e229b3a73b6fd3835d0370a/test/tpmclient/tpmclient.int.c#L155), the test still fails with the following output due to my comment above:
```
Execute the test script
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:145:tcti_control_command() Failed to connect to control socket. 
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:246:Tss2_Tcti_Swtpm_Reset() Failed to reset TPM: 0xa000a 
ERROR:tcti:src/util/io.c:107:write_all() failed to write to fd -1: Bad file descriptor 
ERROR:tcti:src/tss2-tcti/tcti-mssim.c:93:tcti_platform_command() Failed to send platform command 2 with error: 0 
Script returned 1
FAIL test/tpmclient/tpmclient.int (exit status: 1)
```

src/tss2-tcti/tcti-swtpm.c
tstruk: Yes, but that's yet another problem, because tpm_client assumes that if TCTI_SWTPM is enabled then it can directly call Tss2_Tcti_Swtpm_Reset() on the tcti it gets, which doesn't necessarily need to be swtpm_tcti at all if we set TPM20_TCTI_NAME to mssim for instance.
The problem with the reset call is that it is not part of the TCTI interface as defined by the spec. The spec only defines these: https://github.com/tpm2-software/tpm2-tss/blob/master/include/tss2/tss2_tcti.h#L88-L132
In this case the ctx becomes invalid.

diabonas:
I suggest getting rid of the destructive context magic overwrite since the value of `TCTI_BAD_MAGIC` is not actually used anywhere and adding a check for the correct magic value to `tcti_control_command`. The following diff makes the test suite pass with mssim:
```diff
diff --git a/src/tss2-tcti/tcti-common.h b/src/tss2-tcti/tcti-common.h
index e1e1235b..23dbffb5 100644
--- a/src/tss2-tcti/tcti-common.h
+++ b/src/tss2-tcti/tcti-common.h
@@ -12,7 +12,6 @@
 #include "tss2_tcti.h"
 
 #define TCTI_VERSION 0x2
-#define TCTI_BAD_MAGIC 0xBADBADBADBADBADBULL
 
 #define TPM_HEADER_SIZE (sizeof (TPM2_ST) + sizeof (UINT32) + sizeof (UINT32))
 
diff --git a/src/tss2-tcti/tcti-device.c b/src/tss2-tcti/tcti-device.c
index 71befae4..8bc811b5 100644
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -82,9 +82,6 @@ tcti_device_context_cast (TSS2_TCTI_CONTEXT *tcti_ctx)
     if (tcti_ctx == NULL)
 	return NULL;
 
-    if (TSS2_TCTI_MAGIC (tcti_ctx) != TCTI_DEVICE_MAGIC)
-        TSS2_TCTI_MAGIC (tcti_ctx) = TCTI_BAD_MAGIC;
-
     return (TSS2_TCTI_DEVICE_CONTEXT*)tcti_ctx;
 }
 /*
diff --git a/src/tss2-tcti/tcti-mssim.c b/src/tss2-tcti/tcti-mssim.c
index 3aa292b2..99e66cea 100644
--- a/src/tss2-tcti/tcti-mssim.c
+++ b/src/tss2-tcti/tcti-mssim.c
@@ -43,10 +43,6 @@ tcti_mssim_context_cast (TSS2_TCTI_CONTEXT *tcti_ctx)
     if (tcti_ctx == NULL)
 	return NULL;
 
-    if (TSS2_TCTI_MAGIC (tcti_ctx) != TCTI_MSSIM_MAGIC) {
-        TSS2_TCTI_MAGIC (tcti_ctx) = TCTI_BAD_MAGIC;
-    }
-
     return (TSS2_TCTI_MSSIM_CONTEXT*)tcti_ctx;
 }
 /*
diff --git a/src/tss2-tcti/tcti-swtpm.c b/src/tss2-tcti/tcti-swtpm.c
index 9e6a5d99..73eeb379 100644
--- a/src/tss2-tcti/tcti-swtpm.c
+++ b/src/tss2-tcti/tcti-swtpm.c
@@ -69,9 +69,6 @@ tcti_swtpm_context_cast (TSS2_TCTI_CONTEXT *tcti_ctx)
     if (tcti_ctx == NULL)
 	return NULL;
 
-    if (TSS2_TCTI_MAGIC (tcti_ctx) != TCTI_SWTPM_MAGIC)
-        TSS2_TCTI_MAGIC (tcti_ctx) = TCTI_BAD_MAGIC;
-
     return (TSS2_TCTI_SWTPM_CONTEXT*)tcti_ctx;
 }
 /*
@@ -127,6 +124,10 @@ TSS2_RC tcti_control_command (
         return TSS2_TCTI_RC_BAD_REFERENCE;
     }
 
+    if (TSS2_TCTI_MAGIC(tctiContext) != TCTI_SWTPM_MAGIC) {
+       return TSS2_TCTI_RC_BAD_CONTEXT;
+    }
+
     if (cmd_sdu == NULL && cmd_sdu_len != 0) {
         return TSS2_TCTI_RC_BAD_VALUE;
     }
diff --git a/src/tss2-tcti/tcti-tbs.c b/src/tss2-tcti/tcti-tbs.c
index 253aef64..18c629ff 100644
--- a/src/tss2-tcti/tcti-tbs.c
+++ b/src/tss2-tcti/tcti-tbs.c
@@ -40,10 +40,6 @@ tcti_tbs_context_cast (TSS2_TCTI_CONTEXT *tcti_ctx)
         return NULL;
     }
 
-    if (TSS2_TCTI_MAGIC (tcti_ctx) != TCTI_TBS_MAGIC) {
-        TSS2_TCTI_MAGIC (tcti_ctx) = TCTI_BAD_MAGIC;
-    }
-
     return (TSS2_TCTI_TBS_CONTEXT*)tcti_ctx;
 }
```

tstruk:
I added a check to tcti_control_command() in swtpm and platform_command() in mssim and updated tpm_client to skip the test if BAD_CONTEXT is returned in TpmReset(). That should make it passing with mssim.

diabonas:
The check in `tcti_control_command` works, though it would be safer to use `TSS2_TCTI_MAGIC(tctiContext)`: `tcti_swtpm` is of type `TSS2_TCTI_SWTPM_CONTEXT`, so `TSS2_TCTI_MAGIC(tcti_swtpm)` only works by accident because the `TSS2_TCTI_COMMON_CONTEXT` is the first member of the `TSS2_TCTI_SWTPM_CONTEXT` struct.

The check in tpmclient doesn't work because the return value is not used directly as the exit status of the test, but compared to `TPM2_RC_SUCCESS`  by `CheckPassed`. Even if it worked, it would be conceptually wrong: you really don't want to skip the whole test if swtpm is not used, you just want to use a different reset method appropriate for mssim instead of swtpm. The test works completely fine with mssim when applying my diff on top of the previous iteration bbab96026030b8fc142c37f277a8ed2982384ec7 of your PR.

src/tss2-tcti/tcti-swtpm.c
diabonas: Sure, you could avoid the problem by checking the magic number of the TCTI context before calling `Tss2_Tcti_Swtpm_Reset` in tpmclient. Still, I don't really see the purpose of invalidating the context: `TCTI_BAD_MAGIC` is never used for comparison, so what's the point?

tstruk:
> The check in tpmclient doesn't work because the return value is not used directly as the exit status of the test, but compared to `TPM2_RC_SUCCESS` by `CheckPassed`. Even if it worked, it would be conceptually wrong: you really don't want to skip the whole test if swtpm is not used, you just want to use a different reset method appropriate for mssim instead of swtpm. 

Well but how do you know that it is a mssim tcti and not a device tcti or something completely different?

diabonas:
> Well but how do you know that it is a mssim tcti and not a device tcti or something completely different?

I don't think tpmclient would work with a physical TPM anyway (if it is officially supported to run the test suite on an actual TPM at all, I don't know the current project policy on that) since it simulates a power cycle of the TPM several times. Hence I think concentrating on the two supported simulators is enough for this. I suppose it would be possible to switch according to the magic number of the TCTI context, but that seems a bit ugly and is not really necessary if you get a usable error code and TCTI context from `Tss2_Tcti_Swtpm_Reset`.

tstruk:
> I don't think tpmclient would work with a physical TPM anyway (if it is officially supported to run the test suite on an actual TPM at all, I don't know the current project policy on that) since it simulates a power cycle of the TPM several times. Hence I think concentrating on the two supported simulators is enough for this.

Ok, I think we can assume that it will be either mssim or swtpm. I added a skip if the two reset methods fail just in case. There is nothing that can stop me to set my TPM20_TEST_TCTI to tcit_device. I also removed the overwrite of magic. As you said it is not needed since we have checks for it in all the appropriate functions. I had to update the fuzz_tcti also. Now all should work fine.

PR #1708, opened 2020-04-27
sys: Minor static analysis cleanup in Tss2_Sys_CertifyCreation_Complete

snits:
No need to assign to rval since we are returning here.

Signed-off-by: Jerry Snitselaar <jsnitsel@redhat.com>


AndreasFuchsSIT:
@snits Since you say "covscan" do you need this backported to 2.4.x ?

snits:
Our build process runs a coverity scan against the source of a package when the package is updated, and this popped up when updating to 2.3.2. It probably can just go into master.

AndreasFuchsSIT:
Just wanted to make sure you don't need it backported.
Thanks for the patch...

PR #1722, opened 2020-05-08
Makefile.am: add --enable-debug to distcheck flags

williamcroberts:
make check should be run with a debug build so add --enable-debug
to distcheck flags. This will fix make distcheck with clang 6.0 not
being able to use ld --wrap to hijack various libc routines needed for
cmocka tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -190,6 +190,7 @@ endif #DOXYMAN
 include src_vars.mk
 
 # Add test definitions
+AM_DISTCHECK_CONFIGURE_FLAGS="--enable-debug"
 include Makefile-test.am
 
 # Add fuzz definitions


williamcroberts:
> I would prefer to do this for CI only, instead of forcing it on everyone.
> 
> ```
> diff --git a/.ci/docker.run b/.ci/docker.run
> index 9b4b9aaf..a0964ec4 100755
> --- a/.ci/docker.run
> +++ b/.ci/docker.run
> @@ -5,7 +5,7 @@ set -e
>  
>  source $TRAVIS_BUILD_DIR/.ci/docker-prelude.sh
>  
> -export CONFIGURE_OPTIONS=
> +export CONFIGURE_OPTIONS="--enable-debug=yes"
>  
>  if [ -d build ]; then
>    rm -rf build
> ```

What exactly does this force on everyone? That their compiler support -g? Or can we at least add debug when --configure-<test option>? Right now picking clang + make distcheck without debug is broken.

tstruk:
> What exactly does this force on everyone? That their compiler support -g? Or can we at least add debug when --configure-`<test option>`? Right now picking clang + make distcheck with tests enabled and without debug is broken.

With this in the Makefile.am it will not be possible to build the libraries without debug symbols, and not everybody wants that.
How is that broken? All clang builds in our CI run distcheck without debug enabled:
https://github.com/tpm2-software/tpm2-tss/blob/master/.ci/docker.run#L83
and it is working fine: 
https://travis-ci.org/github/tpm2-software/tpm2-tss/jobs/685547874#L4034

williamcroberts:
> > What exactly does this force on everyone? That their compiler support -g? Or can we at least add debug when --configure-`<test option>`? Right now picking clang + make distcheck with tests enabled and without debug is broken.
> 
> With this in the Makefile.am it will not be possible to build the libraries without debug symbols, and not everybody wants that.

Huh? It's only for the temporary directory used to check the dist tarball... make distcheck will create a tarball, unpack the tarball into a sub directory (tpm2-tss-3.0.0-dev), create a directory with a build variant (tpm2-tss-3.0.0-dev/_build) and then built it and run make check. So it's only for this temporary distcheck build where it will have the debug symbols. All other build artifacts will remain exactly the same.

> How is that broken? All clang builds in our CI run distcheck without debug enabled:

Its specific to clang 6.0.1, which ubuntu 16.04, 18.04 do not have, the others are on clang versions
way ahead of that. Granted, its bug, for a specific compiler, for a specific version, that's imposing -fsemantic-interposition details when it shouldn't be.

> https://github.com/tpm2-software/tpm2-tss/blob/master/.ci/docker.run#L83
> and it is working fine:
> https://travis-ci.org/github/tpm2-software/tpm2-tss/jobs/685547874#L4034

tstruk:
If it ain't broke, don't fix it.

williamcroberts:
> If it ain't broke, don't fix it.

It is broke, but I don't care. I just wanted to make sure that you understood that this doesn't affect the build artifacts.

Issue #1702, opened 2020-04-21
Test results with Ubuntu 20.04


KishoreRedlapalli:
Are the compilation and CI tests certified to run on ubuntu 20.04?

AndreasFuchsSIT:
No, not yet.

We'd need a `ubuntu-20.04.docker` file over at
https://github.com/tpm2-software/tpm2-software-container

And then an entry at
https://github.com/tpm2-software/tpm2-tss/blob/0783420b3db7c6c66ce597ab3dd0df5b053a2be2/.travis.yml#L10-L17

I guess #1701 gives us the necessary changes for OpenSSL version

PR #1723, opened 2020-05-08
travis: add ubuntu-20.04 build

tstruk:
Fixes: #1702

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
   - DOCKER_TAG=ubuntu-16.04
    # ubuntu 18.04
   - DOCKER_TAG=ubuntu-18.04
+   # ubuntu 20.04
+  - DOCKER_TAG=ubuntu-20.04
     # fedora-30
   - DOCKER_TAG=fedora-30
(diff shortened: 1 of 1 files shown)


tstruk:
We need one more for Fedora 31. I can build one this week. And I think we could drop ubuntu 18.04. and Fedora 30 then.

AndreasFuchsSIT:
18.04 is an LTS and we should keep it for another year or so.
I'd rather drop 16.04 first... ;-)

williamcroberts:
> We need one more for Fedora 31. I can build one this week. And I think we could
> drop ubuntu 18.04. and Fedora 30 then.

I would drop 16.04, we probably want to keep the last 2 stable things, LTS and what not.
Probably keep fedora 30 and 31, or maybe do 31 and rawhide...

>

williamcroberts:
I guess drop 30, it reached EOL on May 7
  - https://endoflife.date/fedora
  - https://fedoraproject.org/wiki/Fedora_Release_Life_Cycle

Maybe we could do 31 and rawhide, so we can track next as we go.

PR #1727, opened 2020-05-12
[RFC] Network TCTI

williamcroberts:
I still have a bunch of cleanup to do, but wanted to solicit feedback. I attached a general diagram below, The idea is that it takes raw tpm command buffers and just forwards them to the TCTI of choice.

Required by:
https://github.com/tpm2-software/tpm2-tools/pull/2018

[network-tcti.pdf](https://github.com/tpm2-software/tpm2-tss/files/4616033/network-tcti.pdf)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -334,6 +334,25 @@ src_tss2_tcti_libtss2_tcti_mssim_la_SOURCES  = \
     src/tss2-tcti/tcti-mssim.c
 endif # ENABLE_TCTI_MSSIM
 
+# tcti library for Raw TPM2 Network Interface
+if ENABLE_TCTI_NETWORK
+libtss2_tcti_network = src/tss2-tcti/libtss2-tcti-network.la
+tss2_HEADERS += $(srcdir)/include/tss2/tss2_tcti_network.h
+lib_LTLIBRARIES += $(libtss2_tcti_network)
+pkgconfig_DATA += lib/tss2-tcti-network.pc
+EXTRA_DIST += lib/tss2-tcti-network.map \
+              lib/tss2-tcti-network.def
+
+if HAVE_LD_VERSION_SCRIPT
+src_tss2_tcti_libtss2_tcti_network_la_LDFLAGS  = -Wl,--version-script=$(srcdir)/lib/tss2-tcti-network.map
+endif # HAVE_LD_VERSION_SCRIPT
+src_tss2_tcti_libtss2_tcti_network_la_LIBADD   = $(libutil)
+src_tss2_tcti_libtss2_tcti_network_la_SOURCES  = \
+    src/tss2-tcti/tcti-common.c \
+    src/tss2-tcti/tcti-network.c \
+    src/tss2-tcti/tcti-network.h
+endif # ENABLE_TCTI_NETWORK
+
 ### TCG TSS SAPI spec library ###
 libtss2_sys = src/tss2-sys/libtss2-sys.la
 tss2_HEADERS += $(srcdir)/include/tss2/tss2_sys.h
@@ -388,7 +407,7 @@ src_tss2_esys_libtss2_esys_la_SOURCES = $(TSS2_ESYS_SRC) $(TSS2_ESYS_SRC_CRYPTO)
     src/tss2-tcti/tctildr.c src/tss2-tcti/tctildr.h \
     src/tss2-tcti/tctildr-interface.h
 if NO_DL
-src_tss2_esys_libtss2_esys_la_LIBADD += $(libtss2_tcti_device) $(libtss2_tcti_mssim)
+src_tss2_esys_libtss2_esys_la_LIBADD += $(libtss2_tcti_device) $(libtss2_tcti_mssim) $(libtss2_tcti_network)
 src_tss2_esys_libtss2_esys_la_SOURCES += src/tss2-tcti/tctildr-nodl.c src/tss2-tcti/tctildr-nodl.h
 else
 src_tss2_esys_libtss2_esys_la_LIBADD += $(LIBADD_DL)
@@ -589,6 +608,7 @@ endef
 man3_MANS = \
     man/man3/Tss2_Tcti_Device_Init.3 \
     man/man3/Tss2_Tcti_Mssim_Init.3 \
+    man/man3/Tss2_Tcti_Network_Init.3 \
     man/man3/Tss2_TctiLdr_Finalize.3 \
     man/man3/Tss2_TctiLdr_FreeInfo.3 \
     man/man3/Tss2_TctiLdr_GetInfo.3 \
@@ -599,6 +619,7 @@ man7_MANS = \
     man/man7/tss2-tcti-device.7 \
     man/man7/tss2-tcti-swtpm.7 \
     man/man7/tss2-tcti-mssim.7 \
+    man/man7/tss2-tcti-network.7 \
     man/man7/tss2-tctildr.7
 
 man/man3/%.3 : man/%.3.in $(srcdir)/man/man-postlude.troff
@@ -632,6 +653,7 @@ EXTRA_DIST += \
     man/man-postlude.troff \
     man/Tss2_Tcti_Device_Init.3.in \
     man/Tss2_Tcti_Mssim_Init.3.in \
+    man/Tss2_Tcti_Network_Init.3.in \
     man/Tss2_TctiLdr_Finalize.3.in \
     man/Tss2_TctiLdr_FreeInfo.3.in \
     man/Tss2_TctiLdr_GetInfo.3.in \
@@ -639,6 +661,7 @@ EXTRA_DIST += \
     man/tss2-tcti-device.7.in \
     man/man7/tss2-tcti-swtpm.7 \
     man/tss2-tcti-mssim.7.in \
+    man/tss2-tcti-network.7.in \
     man/tss2-tctildr.7.in
 
 CLEANFILES += \
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])]) #Backward compatible setti
 
 AC_CONFIG_HEADERS([config.h])
 
-AC_CONFIG_FILES([Makefile Doxyfile lib/tss2-sys.pc lib/tss2-esys.pc lib/tss2-mu.pc lib/tss2-tcti-device.pc lib/tss2-tcti-mssim.pc lib/tss2-tcti-swtpm.pc lib/tss2-rc.pc lib/tss2-tctildr.pc lib/tss2-fapi.pc])
+AC_CONFIG_FILES([Makefile Doxyfile lib/tss2-sys.pc lib/tss2-esys.pc lib/tss2-mu.pc lib/tss2-tcti-device.pc lib/tss2-tcti-mssim.pc lib/tss2-tcti-swtpm.pc lib/tss2-rc.pc lib/tss2-tctildr.pc lib/tss2-fapi.pc lib/tss2-tcti-network.pc])
 
 # propagate configure arguments to distcheck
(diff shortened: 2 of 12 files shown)


williamcroberts:
And it looks like p11 broke as well as LD_PRELOAD for asan on executables, we need to fix up the LD_PRELOAD so we can run a command, like openssl with pkcs11 library loaded.
```
Detected clang version: 10.0.0
Looking for libasan to LD_PRELOAD at: /usr/lib/llvm-10.0/lib/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
Found libasan at: /usr/lib64/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so
export LD_PRELOAD="/usr/lib64/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so"
turning off asan detection for running commands...
ERROR: ld.so: object '/usr/lib64/clang/10.0.0/lib/linux/libclang_rt.asan-x86_64.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
engine "pkcs11" set.
/workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so: undefined symbol: __asan_option_detect_stack_use_after_return
Unable to load module /workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so
/workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so: undefined symbol: __asan_option_detect_stack_use_after_return
Unable to load module /workspace/tpm2-pkcs11/build/src/.libs/libtpm2_pkcs11.so
PKCS11_get_private_key returned NULL
cannot load Private Key from engine
140139537544512:error:81065401:libp11:pkcs11_CTX_load:Unable to load PKCS#11 module:p11_load.c:79:
140139537544512:error:26096080:engine routines:ENGINE_load_private_key:failed loading private key:../crypto/engine/eng_pkey.c:77:
unable to load Private Key
Failed to run the setup script
/P_RSA/HE:/P_RSA/HE/EK:/P_RSA/LOCKOUT:/P_RSA/HS:/P_RSA/HS/SRK
Execution of ../test/integration/nss-tests.sh failed: 1
FAIL test/integration/nss-tests.sh (exit status: 1)
```

tstruk:
> Depends on:
> [tpm2-software/tpm2-tools#2018](https://github.com/tpm2-software/tpm2-tools/pull/2018)

If something in TSS depends on anything in tools then we have a circular dependency, which is not good.

williamcroberts:
> > Depends on:
> > [tpm2-software/tpm2-tools#2018](https://github.com/tpm2-software/tpm2-tools/pull/2018)
> 
> If something in TSS depends on anything in tools then we have a circular dependency, which is not good.

Sorry I didn't write what I wanted (copy + paste + edit error). Required by tpm2-tools.

AndreasFuchsSIT:
So I was thinking about more of a separate project (similar to tcti-uefi) with the following things:
1. libtss2-tcti-cmd.so with `fork(); exec(conf);` and transmit and receive handed to stdin/stdout of the child
2. tcti-cmd (executable): with --tcti="device:/dev/tpm0" Hands stdin/out to a tcti.

Then we could use it as follows:
```
export TPM2TOOLS_TCTI="cmd:ssh remotehost -c \"tcti-cmd --tcti=device:/dev/tpm0\"" tpm2_getrandom
```
And we would have a remote connection.

Since it has a server and a client component I'd put it into a separate project instead of here.

Thoughts ?

AndreasFuchsSIT:
P.S. The unsecure networking example could then work like this:

Server:
```
socat tcp-l:12345,reuseaddr,fork system:'tcti-cmd --tcti=device:/dev/tpm0'
```

Client:
```
TPM2TOOLS_TCTI="cmd:socat tcp:target:12345" tpm2_getrandom
```

Since I'm no expert on socat, the parameters are probably wrong, but it grasps the idea.

williamcroberts:
> -----Original Message-----
> From: Andreas Fuchs [mailto:notifications@github.com]
> Sent: Wednesday, May 13, 2020 8:46 AM
> Cc: Roberts, William C <william.c.roberts@intel.com>; Author
> <author@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tss] [RFC] Network TCTI (#1727)
> 
> So I was thinking about more of a separate project (similar to tcti-uefi) with the
> following things:
> 
> 1.	libtss2-tcti-cmd.so with fork(); exec(conf); and transmit and receive
> handed to stdin/stdout of the child
> 2.	tcti-cmd (executable): with --tcti="device:/dev/tpm0" Hands stdin/out to
> a tcti.
> 
> Then we could use it as follows:
> 
> export TPM2TOOLS_TCTI="cmd:ssh remotehost -c \"tcti-cmd --
> tcti=device:/dev/tpm0\"" tpm2_getrandom

At that point what's the benefit of doing this, why not just 'ssh -c tpm2_getrandom -Tdevice'

AndreasFuchsSIT:
>> export TPM2TOOLS_TCTI="cmd:ssh remotehost -c \"tcti-cmd -- tcti=device:/dev/tpm0\"" tpm2_getrandom

> At that point what's the benefit of doing this, why not just 'ssh -c tpm2_getrandom -Tdevice'

Because of HMAC-sessions and parameter encryption that we get end-to-end with the TPM without relying on the remotehost OS security.
(Granted not for tpm2_getrandom, but for anything with a Session and FAPI)

williamcroberts:
> So I was thinking about more of a separate project (similar to tcti-uefi) with the following things:
> 
> 1. libtss2-tcti-cmd.so with `fork(); exec(conf);` and transmit and receive handed to stdin/stdout of the child
> 2. tcti-cmd (executable): with --tcti="device:/dev/tpm0" hands stdin/out to a tcti.
> 
> Then we could use it as follows:
> 
> ```
> TPM2TOOLS_TCTI="cmd:ssh remotehost -c \"tcti-cmd --tcti=device:/dev/tpm0\"" tpm2_getrandom
> ```
> 
> And we would have a remote connection.
> 
> For testing we could even
> 
> ```
> TPM2TOOLS_TCTI="cmd:tcti-cmd --tcti=device:/dev/tpm0" tpm2_getrandom
> ```
> 
> Since it has a server and a client component I'd put it into a separate project instead of here.
> 
> Thoughts ?

I think we already have item 2, tpm2_send should do it.

williamcroberts:
> > > export TPM2TOOLS_TCTI="cmd:ssh remotehost -c "tcti-cmd -- tcti=device:/dev/tpm0"" tpm2_getrandom
> 
> > At that point what's the benefit of doing this, why not just 'ssh -c tpm2_getrandom -Tdevice'
> 
> Because of HMAC-sessions and parameter encryption that we get end-to-end with the TPM without relying on the remotehost OS security.
> (Granted not for tpm2_getrandom, but for anything with a Session and FAPI)

I think an issue would be trying to use it with an RM. If it's a separate exec'd process, then at the end of each command invocation objects would be gone. You would need to always perform a load + cmd on the server side in the same process.

AndreasFuchsSIT:
> > > > export TPM2TOOLS_TCTI="cmd:ssh remotehost -c "tcti-cmd -- tcti=device:/dev/tpm0"" tpm2_getrandom
> > 
> > 
> > > At that point what's the benefit of doing this, why not just 'ssh -c tpm2_getrandom -Tdevice'
> > 
> > 
> > Because of HMAC-sessions and parameter encryption that we get end-to-end with the TPM without relying on the remotehost OS security.
> > (Granted not for tpm2_getrandom, but for anything with a Session and FAPI)
> 
> I think an issue would be trying to use it with an RM. If it's a separate exec'd process, then at the end of each command invocation objects would be gone. You would need to always perform a load + cmd on the server side in the same process.

The idea is that libtss2-tcti-cmd opens the sub-process on initialize, then does several transmits and receives via stdin/out according to the (to-be-defined) network protocol encoding and only kills the sub-process on tcti_finalize. Thus the behavior is exactly the same as with a direct tcti connection; including object contexts in the RM.

williamcroberts:
> > > > > export TPM2TOOLS_TCTI="cmd:ssh remotehost -c "tcti-cmd -- tcti=device:/dev/tpm0"" tpm2_getrandom
> > > 
> > > 
> > > > At that point what's the benefit of doing this, why not just 'ssh -c tpm2_getrandom -Tdevice'
> > > 
> > > 
> > > Because of HMAC-sessions and parameter encryption that we get end-to-end with the TPM without relying on the remotehost OS security.
> > > (Granted not for tpm2_getrandom, but for anything with a Session and FAPI)
> > 
> > 
> > I think an issue would be trying to use it with an RM. If it's a separate exec'd process, then at the end of each command invocation objects would be gone. You would need to always perform a load + cmd on the server side in the same process.
> 
> The idea is that libtss2-tcti-cmd opens the sub-process on initialize, then does several transmits and receives via stdin/out according to the (to-be-defined) network protocol encoding and only kills the sub-process on tcti_finalize. Thus the behavior is exactly the same as with a direct tcti connection; including object contexts in the RM.

Ahh it keeps the subprocess alive until finalize... Interesting approach.

As of right now (with a plain network tcti) I can kill tpm2_netlisten, because their is a program called tscpserver and we can just plumb that tpm2_send, and modify tpm2_send so it has a "stream mode".

With the cmd approach, we could just kill network and use command all together.

Interesting, I get what you're putting down now. Will re-spin, but I am not sure if we want every TCTI to be a separate project (especially considering tpm2-tss has many TCTIs). We could start like tpm2-tcti-aux project and plop whatever in there.

tstruk:
We could also leverage netcat (man 1 nc), to setup an SSL tunnel for command transmission.

tobyp:
Is there a more recent discussion about this topic @tstruk @williamcroberts? I have a similar use case (TPM module on a RaspberryPi that I'd like to use from my Desktop).

It occurs to me that adding a little conf string parsing, getaddrinfo, and a socket call in the tcti-device can already achieve this feature on the TSS side (e.g. -T"device:host=localhost,port=2321,protocol=ipv4", low chance of conflict because any sensible real device config string is an absolute path). On the remote side, a simple socat call can do as @AndreasFuchsSIT mentioned above (I use `socat TCP-LISTEN:${TPM_SERVER_PORT},bind=${TPM_SERVER_HOST},fork,reuseaddr OPEN:${TPM_DEVICE_REAL},nonblock`).

Note that for the TSS side, a normal device can be emulated too by giving the device TCTI a pty device node instead of a real TPM. They're both character devices, and the TCTI only uses basic operations like read and write currently, which behave more or less the same: `socat PTY,link=${TPM_DEVICE_FAKE},rawer TCP:${TPM_SERVER_HOST}:${TPM_SERVER_PORT}`.

In a further step, TLS with mutual authentication can be added on top of this. I have code for this too, though it's too much to post here. If there's interest, I can write it up as an example for documentation/the github.io blog.

AndreasFuchsSIT:
Yes, it was reimplemented under a different name: `tcti-cmd`

See this manpage: https://github.com/tpm2-software/tpm2-tss/blob/master/man/Tss2_Tcti_Cmd_Init.3.in

You can basically do `-T "cmd:ssh somehost.local tpm2_send -T device:/dev/tpm0"`

That being said, a writeup of this mechanism using ssh socat and whatnot would be highly appreciated for the tpm2-software.github.io tutorial blog thingy. :+1:

williamcroberts:
> Yes, it was reimplemented under a different name: `tcti-cmd`
> 
> See this manpage: https://github.com/tpm2-software/tpm2-tss/blob/master/man/Tss2_Tcti_Cmd_Init.3.in
> 
> You can basically do `-T "cmd:ssh somehost.local tpm2_send -T device:/dev/tpm0"`
> 
> That being said, a writeup of this mechanism using ssh socat and whatnot would be highly appreciated for the tpm2-software.github.io tutorial blog thingy.

Yeah socat would be nice. I tried various things, albeit briefly, but my bash foo wasn't good enough.

PR #1733, opened 2020-05-19
Fapi backport fixes

JuergenReppSIT:
* Several FAPI fixes which also should be back ported to 2.4.x
* Fix handle usage in Fapi_ChangeAuth.
* Fix setting of default flags for key creation.
* Fix path usage in Fapi_Import.
* Fix get version with Fapi_GetInfo.
*  Fix path checking for keys.
* Fix NV index and path handling in NV creation.
* Add content of the config file to FAPI Info.

### DIFF
--- a/src/tss2-fapi/api/Fapi_ChangeAuth.c
+++ b/src/tss2-fapi/api/Fapi_ChangeAuth.c
@@ -181,6 +181,7 @@ Fapi_ChangeAuth_Async(
         strdup_check(command->authValue, "", r, error_cleanup);
     }
     command->handle = ESYS_TR_NONE;
+    context->loadKey.parent_handle = ESYS_TR_NONE;
 
     /* Get a session for further authorizing and integrity checking the
        subsequent ChangeAuth calls. */
@@ -386,6 +387,9 @@ Fapi_ChangeAuth_Finish(
 
                 context->loadKey.parent_handle = ESYS_TR_NONE;
                 return TSS2_FAPI_RC_TRY_AGAIN;
+            } else {
+                /* No need to flush the parent key */
+                context->loadKey.parent_handle = ESYS_TR_NONE;
             }
 
             /* Store information about whether the new authorization is an
@@ -444,6 +448,13 @@ Fapi_ChangeAuth_Finish(
             fallthrough;
 
         statecase(context->state, ENTITY_CHANGE_AUTH_WAIT_FOR_NV_AUTH)
+            /* NV_ChangeAuth is only possible for objects with policy. */
+            if (!command->object.misc.nv.public.nvPublic.authPolicy.size) {
+                goto_error(r, TSS2_FAPI_RC_AUTHORIZATION_FAILED,
+                           "NV object has no policy. "
+                           "NV_ChangeAuth is not possible.", error_cleanup);
+            }
+
             /* Authorize the object with with the policies
                auth value and command code */
             r = ifapi_authorize_object(context, object, &auth_session);
@@ -452,7 +463,7 @@ Fapi_ChangeAuth_Finish(
 
             /* Change the NV index's AuthValue. */
             r = Esys_NV_ChangeAuth_Async(context->esys,
-                    context->nv_cmd.nv_object.handle,
+                    command->object.handle,
                     auth_session,
                     ESYS_TR_NONE,
                     ESYS_TR_NONE,
--- a/src/tss2-fapi/api/Fapi_CreateKey.c
+++ b/src/tss2-fapi/api/Fapi_CreateKey.c
@@ -158,10 +158,13 @@ Fapi_CreateKey_Async(
     LOG_TRACE("authValue: %s", authValue);
 
     TSS2_RC r;
+    TPMA_OBJECT *attributes;
 
     /* Check for NULL parameters */
     check_not_null(context);
     check_not_null(path);
+
+    attributes = &context->cmd.Key_Create.public_templ.public.publicArea.objectAttributes;
 
     /* Reset all context-internal session state information. */
     r = ifapi_session_init(context);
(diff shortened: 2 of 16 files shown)


src/tss2-fapi/api/Fapi_CreateKey.c
```diff
@@ -158,6 +158,8 @@ Fapi_CreateKey_Async(
     LOG_TRACE("authValue: %s", authValue);
 
     TSS2_RC r;
+    TPMA_OBJECT *attributes
+        = &context->cmd.Key_Create.public_templ.public.publicArea.objectAttributes;
```
tstruk, line 162: The assignment should be moved down below the check_not_null(context);

src/tss2-fapi/api/Fapi_CreateKey.c
```diff
@@ -179,6 +181,14 @@ Fapi_CreateKey_Async(
                             &context->cmd.Key_Create.public_templ);
     return_if_error(r, "Set key flags for key");
 
+    /* If neither sign nor decrypt is set both flags will
+       sign_encrypt and decrypt have to be set. */
+    if ( !(*attributes & TPMA_OBJECT_SIGN_ENCRYPT) &&
```
tstruk, line 186: No need for the space after '(' and before ')'. The style should be consistent with the rest.

test/integration/fapi-duplicate.int.c
```diff
@@ -112,6 +112,7 @@ test_fapi_duplicate(FAPI_CONTEXT *context)
     return EXIT_SUCCESS;
 
 error:
+    Fapi_Delete(context, "/");
```
tstruk, line 115: We should keep library changes and tests changes in separate commits.

test/integration/fapi-check-wrong-paths.int.c
```diff
@@ -0,0 +1,81 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
```
tstruk, line 1: Again - this should probably be in a separate commit, especially if it needs to be back-ported - just for the future.

src/tss2-fapi/api/Fapi_CreateNv.c
```diff
@@ -214,6 +214,12 @@ Fapi_CreateNv_Async(
                            policyPath);
     goto_if_error(r, "Set key flags for NV object", error_cleanup);
 
+    if (nvCmd->public_templ.public.nvIndex) {
+        /* NV index was defined by the user, has to be checked. */
+        r = ifapi_check_nv_index(path, nvCmd->public_templ.public.nvIndex);
+        goto_if_error(r, "Check NV path and NV index", error_cleanup);
```
tstruk, line 220: This is checking just the index, not the path, right?

src/tss2-fapi/ifapi_helpers.c
```diff
@@ -1870,6 +1870,135 @@ ifapi_get_nv_start_index(const char *path, TPM2_HANDLE *start_nv_index)
...
+{
+    TSS2_RC r = TSS2_RC_SUCCESS;
+    NODE_STR_T *dir_list = split_string(path, IFAPI_FILE_DELIM);
+
+    return_if_null(dir_list, "Out of memory.", TSS2_FAPI_RC_MEMORY);
+    if (dir_list->next && strcmp(dir_list->str, "nv") == 0) {
+        if (strcmp(dir_list->next->str, "TPM") == 0) {
+            if (nv_index >= 0x01000000 && nv_index <= 0x013fffff)
+                goto cleanup;
+            goto_error(r, TSS2_FAPI_RC_BAD_VALUE,
```
tstruk, line 1896: This goto cleanup; followed by the goto_error(), which goes to the same label looks a bit strange. We should probably reverse the logic and just do:
if (nv_index < 0x01000000 || nv_index > 0x013fffff)
    goto_error(r, TSS2_FAPI_RC_BAD_VALUE, ...)
...

and get remove the goto cleanup;

src/tss2-fapi/ifapi_json_serialize.c
```diff
@@ -805,3 +806,94 @@ ifapi_json_IFAPI_EVENT_serialize(const IFAPI_EVENT *in, json_object **jso)
...
+ * @param[out] jso pointer to the json object.
+ * @retval TSS2_RC_SUCCESS if the function call was a success.
+ * @retval TSS2_FAPI_RC_MEMORY: if the FAPI cannot allocate enough memory.
+ * @retval TSS2_FAPI_RC_BAD_VALUE if the value is not of type IFAPI_KEY.
+ * @retval TSS2_FAPI_RC_BAD_REFERENCE a invalid null pointer is passed.
+ */
+TSS2_RC
+ifapi_json_IFAPI_CONFIG_serialize(const IFAPI_CONFIG *in, json_object **jso)
+ {
+     return_if_null(in, "Bad reference.", TSS2_FAPI_RC_BAD_REFERENCE);
```
tstruk, line 822: This should go below the variable declaration block.

src/tss2-fapi/api/Fapi_ChangeAuth.c
```diff
@@ -386,6 +387,9 @@ Fapi_ChangeAuth_Finish(
 
                 context->loadKey.parent_handle = ESYS_TR_NONE;
                 return TSS2_FAPI_RC_TRY_AGAIN;
+            } else {
```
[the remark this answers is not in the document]
JuergenReppSIT: If the parent is persistent, the second part before  && 
!context->loadKey.parent_handle_persistent
will be false. So in this case context->loadKey.parent_handle might be not equal to ESYS_TR_NONE

src/tss2-fapi/api/Fapi_CreateNv.c
JuergenReppSIT: It will be checked whether the index selected by the user is appropriate for the provided path. If no index is selected by the user an index appropriate for the path will be determined. I will describe this in the comment.

PR #1734, opened 2020-05-19
Command TCTI

williamcroberts:
Things like this are now possible:
  - tpm2_getrandom -T "cmd:ssh localhost tpm2_send" --hex 4

This code is atrocious, so it needs a ton of clean up, but I wanted to start a discussion on 2 things:
1. Where should it live?
2. How should we handle the conf string

### Where Does It Live:
I'm in favor of keeping this in the tpm2-tss repo because:
1. We can use the shared logging and code among tctis, notably the tcti-common.h and util headers.
2. Other tctis are found here, like for swtpm and missim. So if we split this out, why not the others?
3. Folks already know where to find tctis, go to this project and look in folder X. I don't want to throw them more curve-balls. I've already had to field the where is tcti-tabrmd numerous times.

### Config String

The tcti is specified by `--tcti=cmd:<command>`. Where command, can be anything that reads to stdin and writes to stdout.

That `<command>` string eventually makes it to the exec family of functions, which  take an argv array or varargs (depending on exec variant). So a string that is the command, like "tpm2_send -s --tcti=foo" would need to be broken down into:
`[ "tpm2_send", "-s", "--tcti", "foo" ]`

That level of parsing, like what a shell does, can be a nightmare, considering that we have to deal with a lot of issues. For instance nested kvp options following commands.

Currently I just take the string and build an array like:
`[ "bash", "-c", "<command>" ]`

So the shell can do the sub processing. But then this leads to the, do want the shell to be configurable. Usually "sh" is the system shell, on Ubuntu it's dash (dash sucks). Folks could always play PATH tricks to get a different shell.

So then if you make the shell configurable, you also might need a different set of options, etc....
Down the rabbit hole you go.

Or do we just say, it's like system and make it: `execl("/bin/sh", "sh", "-c", "<command>") and call it good?

Their's also shutdown, we want to make sure that we reap children, so we need to call wait. I call kill sigterm, set an alarm for 5 seconds and then waitpid. If waitpid wakes up from EINTR, Ill invoke SIGKILL and then waitpid again... So then are the 5 seconds configurable? Do we just always want to SIGKILL?

Thoughts?


williamcroberts:
FYI this uses -s in tpm2_send, so if you want that you need:
  - https://github.com/williamcroberts/tpm2.0-tools/tree/tpm2_send-stream

AndreasFuchsSIT:
I'd like to see tpm2_send-stream executable be part of this project living under the same configure switch as the tcti-cmd. Thus you get both sides of the medallion together without having to look for a second project.
(Note: it would probably lose the tpm2_ prefix when compiled from there ?)

williamcroberts:
> Things like this are now possible:
> 
> * tpm2_getrandom -T "cmd:ssh localhost tpm2_send -s" --hex 4
> 
> This code is atrocious, so it needs a ton of clean up, but I wanted to start a discussion on 2 things:
> 
> 1. Where should it live?
> 2. How should we handle the conf string
> 
> ### Where Does It Live:
> I'm in favor of keeping this in the tpm2-tss repo because:
> 
> 1. We can use the shared logging and code among tctis, notably the tcti-common.h and util headers.
> 2. Other tctis are found here, like for swtpm and missim. So if we split this out, why not the others?
> 3. Folks already know where to find tctis, go to this project and look in folder X. I don't want to throw them more curve-balls. I've already had to field the where is tcti-tabrmd numerous times.

Oh and 4, it doesn't add any dependencies like the tcti tpm2-abrmd would have on dbus.

williamcroberts:
> I'd like to see tpm2_send-stream executable be part of this project living under the same configure switch as the tcti-cmd. Thus you get both sides of the medallion together without having to look for a second project.
> Also, this makes testing it via `TPM2TSS_TCTI="cmd:tpm2_send-stream --tcti=$TPM2TSS_TCTI"` very simple.

We don't actually need the `-s`. I initially added it over backwards compat concerns, but right now it reads the first command from a file stream and outputs a response, so if you feed it a file with N > 1 commands it only does the first one. I can drop -s and just do it for as many commands in the file.

> (Note: it would probably lose the tpm2_ prefix when compiled from there ?)

The command is arbitrary and no other tcti dependent things live here. For instance, tpm2-tss doesn't give you swtpm or ibmtpm, I don't see why this would be different.

I can add a stripped down copy of the tool for the test-suite, but I don't want to make it an install artifact and then have 2 send commands and have them diverge in any way.

williamcroberts:
Although we could create a tcti directory in tools and drop it in there:
- The tcti common code isn't much, and no one really uses most of it outside of state/sanity checking.
    - The header is populated and never used
- It breaks the circular dependency for a need for a tcti aware send executable

src/tss2-tcti/tcti-cmd.c
```diff
@@ -0,0 +1,453 @@
...
+    			tcti_cmd->child_pid,
+    			strerror(errno));
+    }
+
+    /*
+     * Probably should do a micro sleep and try waitpid with WNOHANG
+     * if it hasn't exited sent a SIGKILL and reap.
+     * XXX This should be an option...
+     */
+    alarm(5);
```
tstruk, line 177: Where is the SIGALRM signal handled?

src/tss2-tcti/tcti-cmd.c
```diff
@@ -0,0 +1,453 @@
...
+    TSS2_TCTI_CANCEL (tcti_common) = tcti_cmd_cancel;
+    TSS2_TCTI_GET_POLL_HANDLES (tcti_common) = tcti_cmd_get_poll_handles;
+    TSS2_TCTI_SET_LOCALITY (tcti_common) = tcti_cmd_set_locality;
+    TSS2_TCTI_MAKE_STICKY (tcti_common) = tcti_make_sticky_not_implemented;
+    tcti_common->state = TCTI_STATE_TRANSMIT;
+    tcti_common->locality = 0;
+    memset (&tcti_common->header, 0, sizeof (tcti_common->header));
+}
+
+static bool popen_w_pipes(const char *cmd, pid_t *pid,
```
tstruk, line 326: Why bool? This should return an int and if fork() fails it should return the errno.

src/tss2-tcti/tcti-cmd.c
```diff
@@ -0,0 +1,453 @@
...
+	    dup2(stderr_pipefd[1], STDERR_FILENO);
+
+	    pid_t p = getpid();
+	    (void)p;
+
+	    // ask kernel to deliver SIGTERM in case the parent dies
+	    prctl(PR_SET_PDEATHSIG, SIGTERM);
+
+	    //replace tee with your process
+	    execlp("bash", "bash", "-c", cmd, (char*) NULL);
```
tstruk, line 353: Is it a good idea to assume bash? Just "sh" might be safer.

src/tss2-tcti/tcti-cmd.c
```diff
@@ -0,0 +1,453 @@
...
+    rc = tcti_common_set_locality_checks (tcti_common, TCTI_COMMAND_MAGIC);
+    if (rc != TSS2_RC_SUCCESS) {
+        return rc;
+    }
+
+    tcti_common->locality = locality;
+    return TSS2_RC_SUCCESS;
+}
+// TODO do we need a define for this, this is coupled to device TCTI?
+#define TCTI_ASYNC
```
tstruk, line 123: Don't need that. The TCTI_ASYNC was only to handle kernel versions, where the driver didn't support poll()

tstruk:
> * It breaks the circular dependency for a need for a tcti aware send executable
There is no circular dependency really, as it doesn't directly refer to or depend on tpm2_send command.

williamcroberts:
> -----Original Message-----
> From: Tadeusz Struk [mailto:notifications@github.com]
> Sent: Wednesday, May 20, 2020 9:44 PM
> Cc: Roberts, William C <william.c.roberts@intel.com>; Author
> <author@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tss] [RFC] Command TCTI (#1734)
> 
> @tstruk commented on this pull request.
> 
> I like the approach. It makes it quite flexible. Maybe we should add some
> constrains for what the cmd is so that someone doesn't try to do "--tcti=cmd:rm -
> rf /" for example.

If they can do that, the tcti interface isn't stopping them from doing it.
Random path based checks don't work anyways and with hard links,
you can't trust the file name. This problem would be better solved. This
would be a problem better solved by SE Linux.

williamcroberts:
> -----Original Message-----
> From: Tadeusz Struk [mailto:notifications@github.com]
> Sent: Wednesday, May 20, 2020 9:47 PM
> Cc: Roberts, William C <william.c.roberts@intel.com>; Author
> <author@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tss] [RFC] Command TCTI (#1734)
> 
> 	*	It breaks the circular dependency for a need for a tcti aware send
> executable
> 		There is no circular dependency really, as it doesn't directly refer
> to or depend on tpm2_send command.

There is if you want to test it, you'll need something to fill in for the cmd.

AndreasFuchsSIT:
I agree to allow any command. I mean that's the purpose so be sure what you're doing.

Question remains: where do we put the `tpm2_send` executable ? I'd still put it into tpm2-tss and call it `tcti_send`.
That also allows us to test this in as a whole in the CI of tpm2-tss.

williamcroberts:
> I agree to allow any command. I mean that's the purpose so be sure what you're doing.
> 
> Question remains: where do we put the `tpm2_send` executable ? I'd still put it into tpm2-tss and call it `tcti_send`.
> That also allows us to test this in as a whole in the CI of tpm2-tss.

@tstruk see, I told you @AndreasFuchsSIT wanted a full integration test. @AndreasFuchsSIT I think we just create it, but don't distribute the executable. Just keep it for testing. Then we just keep send in the tools. This way they can never diverge or get confused.

AndreasFuchsSIT:
But `libtss2-tcti-cmd` and `tcti-send` come as a pair (2 sides of a stdin/out channel). So IMHO they should live together, either in their own project or here in tpm2-tss (which I prefer).

Rational: I don't want to make tpm2-tools a dependency for using tcti-send on a target machine.

williamcroberts:
> But `libtss2-tcti-cmd` and `tcti-send` come as a pair (2 sides of a stdin/out channel). So IMHO they should live together, either in their own project or here in tpm2-tss (which I prefer).
> 
> Rational: I don't want to make tpm2-tools a dependency for using tcti-send on a target machine.
I don't get that rational, distros have the tools anyways, and you could just keep that one tool on the target machine, so it's not like its adding a large dependency bloat.
I'll just take this tcti into the tools then.

AndreasFuchsSIT:
> > But `libtss2-tcti-cmd` and `tcti-send` come as a pair (2 sides of a stdin/out channel). So IMHO they should live together, either in their own project or here in tpm2-tss (which I prefer).
> > Rational: I don't want to make tpm2-tools a dependency for using tcti-send on a target machine.

> I don't get that rational, distros have the tools anyways, and you could just keep that one tool on the target machine, so it's not like its adding a large dependency bloat.
> I'll just take this tcti into the tools then.

I'm not talking about distros. My use cases currently evolve around embedded systems, IoT, etc.
There (during production) you will want to (net-)boot minimal systems for provisionings. In my case, just a sshd with tcti-send and libtss2-tcti-device, no tools, nothing.

And coming as a pair, thus testing them in CI together should IMHO be convincing anyways.

williamcroberts:
> > > But `libtss2-tcti-cmd` and `tcti-send` come as a pair (2 sides of a stdin/out channel). So IMHO they should live together, either in their own project or here in tpm2-tss (which I prefer).
> > > Rational: I don't want to make tpm2-tools a dependency for using tcti-send on a target machine.
> 
> > I don't get that rational, distros have the tools anyways, and you could just keep that one tool on the target machine, so it's not like its adding a large dependency bloat.
> > I'll just take this tcti into the tools then.
> 
> I'm not talking about distros. My use cases currently evolve around embedded systems, IoT, etc.
> There (during production) you will want to (net-)boot minimal systems for provisionings. In my case, just a sshd with tcti-send and libtss2-tcti-device, no tools, nothing.

You're still installing a tool, the tool is just in tpm2-tss. Then we have 2 tools that send stuff. So you could easily just install the single tpm2_send tool in your net image.

The tpm2-tss project has always been about providing spec compliant libraries, not non-spec compliant tools. That's been in the purview of the tools project. If that wasn't the case, we could just have one giant repo.

> 
> And coming as a pair, thus testing them in CI together should IMHO be convincing anyways.

The tpm2-tss project, since it contains spec compliant tcti's should just test that the command tcti works. The tpm2-tools project, would test that the tpm2_send and tcti-cmd items work together. So we would still get full coverage, and keep the projects at their intended scopings.

tstruk:
> > I agree to allow any command. I mean that's the purpose so be sure what you're doing.
> > Question remains: where do we put the `tpm2_send` executable ? I'd still put it into tpm2-tss and call it `tcti_send`.
> > That also allows us to test this in as a whole in the CI of tpm2-tss.
> 
> @tstruk see, I told you @AndreasFuchsSIT wanted a full integration test. @AndreasFuchsSIT I think we just create it, but don't distribute the executable. Just keep it for testing. Then we just keep send in the tools. This way they can never diverge or get confused.

Just for CI we could add an tpm2_send executable tpm2-tss/test/helper/ or the docker images or something that could be just used for the CI builds. I would not mix TSS code and tpm2_tools code together.

tstruk:
> The tpm2-tss project, since it contains spec compliant tcti's should just test that the command tcti  works. The tpm2-tools project, would test that the tpm2_send and tcti-cmd items work together. So we would still get full coverage, and keep the projects at their intended scopings.

+1

tstruk:
> I can work with that rational as well.
> LGTM
> I guess travis is only failing on indentation mixing.

No, the build is failing with:
'sys/prctl.h' file not found
https://cirrus-ci.com/task/5974609201725440?command=main#L582

williamcroberts:
> > I can work with that rational as well.
> > LGTM
> > I guess travis is only failing on indentation mixing.
> 
> No, the build is failing with:
> 'sys/prctl.h' file not found
> https://cirrus-ci.com/task/5974609201725440?command=main#L582

Thanks for taking a peek, ontop of that, their is a lot of cleanup to do as well. I think I don't close all the unused ends of the pipefd's. Probably also need to scan through and ensure all return codes are being checks, make sure their is no superfluous messages as well. Im not 100% sure when I'll swing back to this, could be a couple of weeks.

tstruk:
Might be worth running coverity scan against it. It should catch things like unchecked return codes etc.

src/tss2-tcti/tcti-cmd.c
williamcroberts: In any syscall below, we don't actually need to explicitly handle it. We send kill and waitpid, indefinitely. This is so waitpid will wake up with EINTR if the process doesn't respond to SIGTERM. If it doesn't respond in 5 seconds, we send SIGKILL and then reap.

I guess their is a race if the alarm comes in before the waitpid gets called, but the solutions get much messier. I don't want to register a handler, or adjust the sigmask, as the main process we're getting linked into, that's usually their thing.
- https://stackoverflow.com/questions/282176/waitpid-equivalent-with-timeout

src/tss2-tcti/tcti-cmd.c
williamcroberts: That's exactly what I ended up changing it to, thanks.

src/tss2-tcti/tcti-cmd.c
williamcroberts: Yep, I ended up modeling system(3): `execlp("/bin/sh", "sh", "-c", "cmd", NULL);`

williamcroberts:
> Might be worth running coverity scan against it. It should catch things like unchecked return codes etc.

Good idea, I still need to do this, maybe tomorrow.

williamcroberts:
Current TODO List:
- [ ] Static Analysis Scan
- [ ] Test harness

williamcroberts:
@tstruk what were you thinking for unit testing?

### Test to Verify stdin/stdout IO

I was thinking of using wrap to hook execlp() and have that waiting to read on a pipe from the unit test so it would look like:

1. parent process Unit test writes expected response to child response pipe.
2. child process blocks reading on stdin, waiting for tcti data to be sent.
3. child process read from response buffer pipe (what the unit test says to return)
4. child process write data from 3 to stdout.
5. child process loop to 1.
6. parent process verifies response data matches sent data.

Any syscall with EINTR will cause it to exit. I'll register a signal handler in the child process for SIGTERM and exit(0) as well.

### Test to verify conf string is passed to execlp

  - Ensure that execlp sees command string unadulterated

### Null conf string test

  - Should fail

### Test GetPollHandles

  - Test that it returns the read end of the stdout pipe with the child.

## Test Cancel

  - Unsupported

### Negative testing:

  - Test fork fail
  - test pipe fail
  - test dup2 fail
  - test execlp fail
  - test fdopen fail
  - test setvbuf fail
  - test fread fail
  - test fwrite fail
  - test short read fail scenarios (short header and short data)

williamcroberts:
The tcti-cmd tests are stalling on the Travis CI and causing timeout's. Not sure why, the tests work locally.

williamcroberts:
Well it is running tests now:
```
ERROR:tcti:../src/tss2-tcti/tcti-cmd.c:411:tcti_cmd_transmit() Transmitting to subprocess failed: Bad file descriptor 
[       OK ] tcti_cmd_test_transmit_fail
[==========] 12 test(s) run.
[  PASSED  ] 12 test(s).
FAIL test/unit/tcti-cmd (exit status: 255)
```

But for some reason 12 out of 12 tests passing results in a fail...

williamcroberts:
> test/unit/tcti-cmd

Additionally cmocka returns `return total_failed + total_errors;`, not sure how that's getting to 255, perhaps -1 return code. But tracing through the path I don't see how. Running locally on travis image 18.04 seems ok as well.

diabonas:
> Additionally cmocka returns `return total_failed + total_errors;`, not sure how that's getting to 255, perhaps -1 return code. But tracing through the path I don't see how. Running locally on travis image 18.04 seems ok as well.

This only happens when code coverage is enabled (`./configure --enable-code-coverage`), because saving the coverage data calls the [wrapped file I/O system calls](https://github.com/tpm2-software/tpm2-tss/blob/6bc3caa20f2c225cb97a6001438d18278fe5b683/test/unit/tcti-cmd.c#L50-L116) and there is no mockup data available for `mock_type (int)`, see e.g. https://github.com/tpm2-software/tpm2-tss/pull/1576#issuecomment-569672847. You need to somehow distinguish whether the wrapped file call does not come from a unit test and call the corresponding `__real_` function in this case. By trial and error I found this solution:
```diff
diff --git a/test/unit/tcti-cmd.c b/test/unit/tcti-cmd.c
index a4de99c9..4d83c36e 100644
--- a/test/unit/tcti-cmd.c
+++ b/test/unit/tcti-cmd.c
@@ -71,6 +71,11 @@ int __wrap_fork (void)

 FILE *__wrap_fdopen (int fd, const char *mode)
 {
+    /* allow fdopen() calls necessary for code coverage */
+    if (fd == 3) {
+        return __real_fdopen (fd, mode);
+    }
+
     int rc = mock_type (int);
     if (!rc) {
         return __real_fdopen (fd, mode);
@@ -105,6 +110,11 @@ int __wrap_ferror (FILE *stream)
 size_t __wrap_fwrite (const void *ptr, size_t size, size_t nmemb,
         FILE *stream)
 {
+    /* allow fwrite() calls necessary for code coverage */
+    if (size > 1)  {
+        return __real_fwrite (ptr, size, nmemb, stream);
+    }
+
     int rc = mock_type (int);
     if (!rc) {
         return __real_fwrite (ptr, size, nmemb, stream);
```
It works, see https://github.com/diabonas/tpm2-tss/commit/bc475984a4d34208e045126ffc9b477e9ddad26c (the clang failures are unrelated, code coverage is [only enabled for GCC](https://github.com/tpm2-software/tpm2-tss/blob/5cee99c90300eebeba58967abe486297bb70ffa2/.ci/docker.run#L68-L70)), but is obviously quite fragile, so I do hope there is a better approach than experimentally figuring out the parameters that the code coverage uses and hoping that they don't overlap with the test code or will change in the future...

diabonas:
> the clang failures are unrelated, code coverage is [only enabled for GCC](https://github.com/tpm2-software/tpm2-tss/blob/5cee99c90300eebeba58967abe486297bb70ffa2/.ci/docker.run#L68-L70)

The clang failures and the [whitespace error in the very last CI test](https://travis-ci.org/github/tpm2-software/tpm2-tss/jobs/701093022#L9119) are an easy fix, cf. the [successful CI build](https://travis-ci.org/github/diabonas/tpm2-tss/builds/701398136) of https://github.com/diabonas/tpm2-tss/commit/30c70e1c47932cbe4ba2cbd9ef750a1898eed2b5:
```diff
diff --git a/Makefile-test.am b/Makefile-test.am
index f6d376c7..cbc14018 100644
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -401,8 +401,8 @@ test_unit_tcti_cmd_LDFLAGS = -Wl,--wrap=pipe,--wrap=fork \
     -Wl,--wrap=fdopen,--wrap=sigprocmask,--wrap=fwrite,--wrap=ferror
 test_unit_tcti_cmd_SOURCES = test/unit/tcti-cmd.c \
     src/tss2-tcti/tcti-common.c \
-    src/tss2-tcti/tcti-cmd.c src/tss2-tcti/tcti-cmd.h
-endif 
+    src/tss2-tcti/tcti-cmd.c src/tss2-tcti/tcti-cmd.h test/unit/tcti-cmd-test.h
+endif

 test_unit_tctildr_CFLAGS = $(CMOCKA_CFLAGS) $(TESTS_CFLAGS)
 test_unit_tctildr_LDADD = $(CMOCKA_LIBS) $(libutil)
```

williamcroberts:
@diabonas I owe you so many beers, thank you.

williamcroberts:
> > Additionally cmocka returns `return total_failed + total_errors;`, not sure how that's getting to 255, perhaps -1 return code. But tracing through the path I don't see how. Running locally on travis image 18.04 seems ok as well.
> 
> This only happens when code coverage is enabled (`./configure --enable-code-coverage`), because saving the coverage data calls the [wrapped file I/O system calls](https://github.com/tpm2-software/tpm2-tss/blob/6bc3caa20f2c225cb97a6001438d18278fe5b683/test/unit/tcti-cmd.c#L50-L116) and there is no mockup data available for `mock_type (int)`, see e.g. [#1576 (comment)](https://github.com/tpm2-software/tpm2-tss/pull/1576#issuecomment-569672847). You need to somehow distinguish whether the wrapped file call does not come from a unit test and call the corresponding `__real_` function in this case. By trial and error I found this solution:
> 
> ```diff
> diff --git a/test/unit/tcti-cmd.c b/test/unit/tcti-cmd.c
> index a4de99c9..4d83c36e 100644
> --- a/test/unit/tcti-cmd.c
> +++ b/test/unit/tcti-cmd.c
> @@ -71,6 +71,11 @@ int __wrap_fork (void)
> 
>  FILE *__wrap_fdopen (int fd, const char *mode)
>  {
> +    /* allow fdopen() calls necessary for code coverage */
> +    if (fd == 3) {
> +        return __real_fdopen (fd, mode);
> +    }
> +
>      int rc = mock_type (int);
>      if (!rc) {
>          return __real_fdopen (fd, mode);
> @@ -105,6 +110,11 @@ int __wrap_ferror (FILE *stream)
>  size_t __wrap_fwrite (const void *ptr, size_t size, size_t nmemb,
>          FILE *stream)
>  {
> +    /* allow fwrite() calls necessary for code coverage */
> +    if (size > 1)  {
> +        return __real_fwrite (ptr, size, nmemb, stream);
> +    }
> +
>      int rc = mock_type (int);
>      if (!rc) {
>          return __real_fwrite (ptr, size, nmemb, stream);
> ```
> 
> It works, see [diabonas@bc47598](https://github.com/diabonas/tpm2-tss/commit/bc475984a4d34208e045126ffc9b477e9ddad26c) (the clang failures are unrelated, code coverage is [only enabled for GCC](https://github.com/tpm2-software/tpm2-tss/blob/5cee99c90300eebeba58967abe486297bb70ffa2/.ci/docker.run#L68-L70)), but is obviously quite fragile, so I do hope there is a better approach than experimentally figuring out the parameters that the code coverage uses and hoping that they don't overlap with the test code or will change in the future...

Interesting. I'm not sure how you would wrap fdopen. It might be better to implement my own internal fdopen I can mock over using this one, same with fwrite and fread. Then I can avoid this.

diabonas:
No worries :) Not wrapping the syscalls directly but through a custom proxy instead sounds like a good idea to me. The only other non-fragile approach I can think of would be something like
```diff
diff --git a/test/unit/tcti-cmd.c b/test/unit/tcti-cmd.c
index a4de99c9..456ce1d3 100644
--- a/test/unit/tcti-cmd.c
+++ b/test/unit/tcti-cmd.c
@@ -41,6 +41,13 @@ int __real_sigprocmask (int how, const sigset_t *set, sigset_t *oldset);
 int __real_fork (void);
 int __real_pipe (int pipefd[2]);
 
+static bool mock = true;
+static int stop_mocking_me (void **state)
+{
+    mock = false;
+    return 0;
+}
+
 static inline
 TSS2_TCTI_CONTEXT *state_cast (void **state)
 {
@@ -71,6 +78,10 @@ int __wrap_fork (void)
 
 FILE *__wrap_fdopen (int fd, const char *mode)
 {
+    if (!mock) {
+        return __real_fdopen (fd, mode);
+    }
+
     int rc = mock_type (int);
     if (!rc) {
         return __real_fdopen (fd, mode);
@@ -105,6 +116,10 @@ int __wrap_ferror (FILE *stream)
 size_t __wrap_fwrite (const void *ptr, size_t size, size_t nmemb,
         FILE *stream)
 {
+    if (!mock) {
+        return __real_fwrite (ptr, size, nmemb, stream);
+    }
+
     int rc = mock_type (int);
     if (!rc) {
         return __real_fwrite (ptr, size, nmemb, stream);
@@ -411,5 +426,5 @@ main (int   argc,
             test_teardown),
     };
 
-    return cmocka_run_group_tests (tests, NULL, NULL);
+    return cmocka_run_group_tests (tests, NULL, stop_mocking_me);
 }
```
to disable the mocking before the coverage data is written, but that feels somewhat weird as well.

williamcroberts:
> ```diff
> +    mock = false;
> ```

I ended up creating shims for all the libc functions I want to hijack, and if unit testing, make them weak, else make them static. Then the unit test main can just overload them.

It's a departure form using wrap, but just seems to be a simple way of dealing with this common issue.

williamcroberts:
@diabonas I cannot say thank you enough, you saved me countless hours tomorrow. I just cleaned out some dead code and added a few more tests. Hopefully this gets the code coverage up. Unfortunately, code executed in the child doesn't count. I guess I could mock fork to return 0 and act as the child and cause execlp to fail. I think if execlp fails, calling exit (1) is a tad bit harsh, so I want to test that path.

After that the only thing that doesn't have the best code coverage is reap_child(), I could add behavior to the child to be misbehaved and test that reap code a little better.

williamcroberts:
> Unfortunately, code executed in the child doesn't count. I guess I could mock fork to return 0 and act as the child and cause execlp to fail. I think if execlp fails, calling exit (1) is a tad bit harsh, so I want to test that path.
> 

Scratch that, the code is too hard to test in that fashion...

> After that the only thing that doesn't have the best code coverage is reap_child(), I could add behavior to the child to be misbehaved and test that reap code a little better.

The problem here is that we always spawn /bin/sh as the first child, which is always well behaved. I did some manual tests to check the code when not using /bin/sh, and it worked as expected. I tried doing things like creating a new process group for the children, and killing that, but wait doesn't wait for the whole group, as you cannot get the signal of a grandchild. For details this was helpful:
  - https://jmmv.dev/2019/11/wait-for-process-group.html

I'm a big fan of less code, since we "always" have a well behaved child, we can just send sigterm and move on. The process group will exit with the leader anyways (process calling TCTI functions).

williamcroberts:
@AndreasFuchsSIT one last look over before merge?

williamcroberts:
> @AndreasFuchsSIT one last look over before merge?

Oh wait you're not around IIRC.

williamcroberts:
FYI their is a PR over in tpm2-tools with full end-to-end testing:
  - https://github.com/tpm2-software/tpm2-tools/pull/2118

PR #1735, opened 2020-05-20
Fix memory leaks in unit tests

JuergenReppSIT:
* Two leaks in  in swtpm unit test were fixed.
* Several leaks if the fapi json unit test were fixed.

### DIFF
--- a/test/unit/fapi-json.c
+++ b/test/unit/fapi-json.c
@@ -101,6 +101,79 @@ static void ifapi_cleanup_policy(TPMS_POLICY *policy)
     }
 }
 
+/* 6 copies of cleanup functions from ifapi_keystore.c */
+
+void
+cleanup_ifapi_duplicate(IFAPI_DUPLICATE * duplicate) {
+    if (duplicate != NULL) {
+        SAFE_FREE(duplicate->certificate);
+    }
+}
+
+void
+cleanup_ifapi_hierarchy(IFAPI_HIERARCHY * hierarchy) {
+    if (hierarchy != NULL) {
+        SAFE_FREE(hierarchy->description);
+    }
+}
+
+void
+cleanup_ifapi_ext_pub_key(IFAPI_EXT_PUB_KEY * key) {
+    if (key != NULL) {
+        SAFE_FREE(key->pem_ext_public);
+        SAFE_FREE(key->certificate);
+    }
+}
+
+void
+cleanup_ifapi_key(IFAPI_KEY * key) {
+    if (key != NULL) {
+        SAFE_FREE(key->policyInstance);
+        SAFE_FREE(key->serialization.buffer);
+        SAFE_FREE(key->private.buffer);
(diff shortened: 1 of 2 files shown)


test/unit/tcti-swtpm.c
```diff
@@ -308,6 +308,9 @@ tcti_swtpm_init_fail_connect_test (void **state)
     will_return (__wrap_connect, -1);
     ret = Tss2_Tcti_Swtpm_Init (ctx, &tcti_size, "host=127.0.0.1,port=666");
     assert_int_equal (ret, TSS2_TCTI_RC_IO_ERROR);
+
+    free(((TSS2_TCTI_SWTPM_CONTEXT*)ctx)->conf_copy);
+    free(ctx);
```
williamcroberts, line 313: if it really matters and you want to guarantee your free gets called, you can setup the test with cmocka_unit_test_setup_teardown()
and the setup function set the void **s so *s points to the tcti that's calloc'd. Then the teardown will always get called and execute the free.

or you could just point the TCTI on the heap, a 4096 buffer should always be big enough, and just assert that size < sizeof(buf).

test/unit/fapi-json.c
```diff
@@ -101,6 +101,79 @@ static void ifapi_cleanup_policy(TPMS_POLICY *policy)
...
+        SAFE_FREE(nv->event_log);
+    }
+}
+
+void
+cleanup_ifapi_object(
+    IFAPI_OBJECT * object)
+{
+    if (object != NULL) {
+        if (object->objectType != IFAPI_OBJ_NONE) {
```
tstruk, line 156: Just a nit-pick - if you would combine the two if statements into one:
if (object != NULL && object->objectType != IFAPI_OBJ_NONE) {
...
it would save you one level of indentation.

williamcroberts:
> -----Original Message-----
> From: Tadeusz Struk [mailto:notifications@github.com]
> Sent: Wednesday, May 20, 2020 1:28 PM
> Cc: Roberts, William C <william.c.roberts@intel.com>; Comment
> <comment@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tss] Fix memory leaks in unit tests (#1735)
> 
> @tstruk approved this pull request.
> 
> LGTM - I will hold on with merging in case you want to address Bill's feedback.

TL;DR you don't have to do it.

Just to be clear, that feedback is a nit. All the other test code assumes leaks on fail, so this one
doesn't need to be special. Extra points if it is. It also serves as historical reference
for anyone looking back in time that cmocka has this functionality and that’s how
you use it.

test/unit/tcti-swtpm.c
JuergenReppSIT: In all other tests the ctx was passed via state and freed in the teardown method. Here ctx was allocated locally. So I thought it's OK to free it locally, to achieve running the test without failure if the address sanitizer is activated.

PR #1743, opened 2020-06-01
Bugfix: Update the public key object with the response from the TPM.

LordOfDorks:
This fix will populate the returned public in the key object with the proper unique object.


LordOfDorks:
I'll re-submit with some more bug fixes.

PR #1744, opened 2020-06-01
2 bug fixes

LordOfDorks:
BugFix: Update the public key object with the response from the TPM in CreateLoaded.
BugFix: Shared secret calculation is not spec compliant.

### DIFF
--- a/src/tss2-esys/api/Esys_CreateLoaded.c
+++ b/src/tss2-esys/api/Esys_CreateLoaded.c
@@ -392,6 +392,8 @@ Esys_CreateLoaded_Finish(
             "in Public name not equal name in response", error_cleanup);
 
     /* Update the meta data of the ESYS_TR object */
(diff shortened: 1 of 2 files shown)


tstruk:
@LordOfDorks thanks for the PR. Some general feedback for the future:
From the CONTRIBUTING.md doc: "Commit subject lines should be prefixed with a string identifying the effected subsystem". Therefore "Bugfix" is not a valid prefix. Please see: https://github.com/tpm2-software/tpm2-tss/blob/master/CONTRIBUTING.md#patch-requirements
If you work on a bugfix it is recommended to open an issue in the issue tracker [1] first, describing the problem, and then push PRs referencing the issue.

[1] - https://github.com/tpm2-software/tpm2-tss/issues

AndreasFuchsSIT:
cherry-pick in #1764

PR #1745, opened 2020-06-02
dev-tcti: remove build-time flags for async and partial read modes

tstruk:
Enable async mode and partial reads by default for 3.0
This has been implemented for some time already and is
supported by the kernel driver since 4.20 AFAICR.
Currently the kernel is at 5.7 so that should make it
safe to assume that most of distros will have that
driver changes. One could always disable this at config
time if required.


williamcroberts:
Is their a way to detect it at open time or via some ioctl? run time is *almost* always better than configure time.

tstruk:
> > > Is their a way to detect it at open time or via some ioctl? run time is _almost_ always better than configure time or does this just build it in, and enable the run-time smartness?
> > 
> > 
> > Not that I can think of. The only difference would be that if one would want to read the header first, and then the rest of the response, the second read will fail.
> 
> That sounds painful. Maybe add a uname call():
> 
> ```c
> #include <sys/utsname.h>
> 
> int uname(struct utsname *buf);
> ```

That won't help because many distros tend to stick with the older LTS kernels and add tons of back-ports.

williamcroberts:
> > > > Is their a way to detect it at open time or via some ioctl? run time is _almost_ always better than configure time or does this just build it in, and enable the run-time smartness?
> > > 
> > > 
> > > Not that I can think of. The only difference would be that if one would want to read the header first, and then the rest of the response, the second read will fail.
> > 
> > 
> > That sounds painful. Maybe add a uname call():
> > ```c
> > #include <sys/utsname.h>
> > 
> > int uname(struct utsname *buf);
> > ```
> 
> That won't help because many distros tend to stick with the older LTS kernels and add tons of back-ports.

Whats the behavior if you enable this but the kernel doesn't have support? You're statement makes it sound like it would fail, is that right?. I wouldn't enable anything that assumes, and can break. Because it's going to lead to breaks.

The uname check would at least guarantee that you can enable it. On older kernels that have backports, it wouldn't use the feature, but it also wouldn't break anything.

tstruk:
> > > > > Is their a way to detect it at open time or via some ioctl? run time is _almost_ always better than configure time or does this just build it in, and enable the run-time smartness?
> > > > 
> > > > 
> > > > Not that I can think of. The only difference would be that if one would want to read the header first, and then the rest of the response, the second read will fail.
> > > 
> > > 
> > > That sounds painful. Maybe add a uname call():
> > > ```c
> > > #include <sys/utsname.h>
> > > 
> > > int uname(struct utsname *buf);
> > > ```
> > 
> > 
> > That won't help because many distros tend to stick with the older LTS kernels and add tons of back-ports.
> 
> Whats the behavior if you enable this but the kernel doesn't have support? You're statement makes it sound like it would fail, is that right?. I wouldn't enable anything that assumes, and can break. Because it's going to lead to breaks.
> 
> The uname check would at least guarantee that you can enable it. On older kernels that have backports, it wouldn't use the feature, but it also wouldn't break anything.

The decision whether to enabled it or not is really up to the distro maintainers, who know best what's the version of the kernel. We are talking about enabling a change in TSS 3.0, which should be out in few months that goes with a driver change that was released more than two years ago.

williamcroberts:
> > > > > > Is their a way to detect it at open time or via some ioctl? run time is _almost_ always better than configure time or does this just build it in, and enable the run-time smartness?
> > > > > 
> > > > > 
> > > > > Not that I can think of. The only difference would be that if one would want to read the header first, and then the rest of the response, the second read will fail.
> > > > 
> > > > 
> > > > That sounds painful. Maybe add a uname call():
> > > > ```c
> > > > #include <sys/utsname.h>
> > > > 
> > > > int uname(struct utsname *buf);
> > > > ```
> > > 
> > > 
> > > That won't help because many distros tend to stick with the older LTS kernels and add tons of back-ports.
> > 
> > 
> > Whats the behavior if you enable this but the kernel doesn't have support? You're statement makes it sound like it would fail, is that right?. I wouldn't enable anything that assumes, and can break. Because it's going to lead to breaks.
> > The uname check would at least guarantee that you can enable it. On older kernels that have backports, it wouldn't use the feature, but it also wouldn't break anything.
> 
> The decision whether to enabled it or not is really up to the distro maintainers, who know best what's the version of the kernel. We are talking about enabling a change in TSS 3.0, which should be out in few months that goes with a driver change that was released more than two years ago.

You're adding a configure time dependency between the package and the kernel version which means when a bug report comes in, you need to know how the package was configured and the release version. Configure time stuff is terrible. Figure it out dynamically and just make it work. Its simplest for everyone involved. Less options are better.

tstruk:
> You're adding a configure time dependency between the package and the kernel version which means when a bug report comes in, you need to know how the package was configured and the release version. Configure time stuff is terrible. Figure it out dynamically and just make it work. Its simplest for everyone involved. Less options are better.

There is no guarantee that the kernel/driver that the package is configured & built on will be the same as the target kernel/driver. Uname dynamic check would only work on Linux, and things like FreeBSD will need a manual modification. On Linux the check would probably be:
 if [ "kernel_ver" >= "5.0" ] then enable fi;
Which will disable it on e.g. RHEL8, or SLES 15, which is not what we want.

williamcroberts:
Also configure time options means a release is the combination of configure flags. Removing those, means you have less releases per version. Dynamic checks can also be overridden, so folks dealing with bugs can poke or prod their way around it rather than having to rebuild. You could add an option to the tcti for async=on,off,auto. Default to auto and use uname to sort it out and drop a configure option.

williamcroberts:
> > You're adding a configure time dependency between the package and the kernel version which means when a bug report comes in, you need to know how the package was configured and the release version. Configure time stuff is terrible. Figure it out dynamically and just make it work. Its simplest for everyone involved. Less options are better.
> 
> There is no guarantee that the kernel/driver that the package is configured & built on will be the same as the target kernel/driver. 

That's why configure checks suck. Im already dealing with the ESYS Handle changes. That could have been done in a dynamic fashion to figure it out since the old handles and new handle values are different and distinguishable. Now we need to see what ESYS versions we're configuring against.

> Uname dynamic check would only work on Linux, and things like FreeBSD will need a manual 

No it has it: https://www.freebsd.org/cgi/man.cgi?query=uname&sektion=3&n=1

modification. On Linux the check would probably be:
> if [ "kernel_ver" >= "5.0" ] then enable fi;
> Which will disable it on e.g. RHEL8, or SLES 15, which is not what we want.

The backport kernel case that contains support, it'll guess it wrong (but does it matter, it will still work), but by making it dynamic and adding the option to the tcti, you can have complete control without needing to explicitly configure it and build time. Giving you full, run time control over the behavior and no need to recompile.

tstruk:
> Giving you full, run time control over the behavior and no need to recompile.

Are we talking about the same thing here? The async mode and partial read are both conditional compilation changes enabled/disabled at config time. Every time it is changed it needs to be recompiled.

williamcroberts:
> > Giving you full, run time control over the behavior and no need to recompile.
> 
> Are we talking about the same thing here? The async mode and partial read are both conditional compilation changes enabled/disabled at config time. Every time it is changed it needs to be recompiled.

Yep, make them dynamic and ditch the configure time conditional stuff. Figure out a sane default at run time (Tcti initialize) and make it "twiddlable" (copyright pending) via config. Right now, the sanest thing I could thing of would be to add options to the tcti config. This way, when the code guess's wrong, or something breaks somewhere you can just say, "user do this config change". When it's statically determined, theirs nothing you can do to get yourself out of a bad condition, and it explodes the actual number of different versions of code you have shipped. Configure options are bad. We should always be looking to minimize them.

tstruk:
I could think of many more ways this dynamic guessing + users overwriting config goes wrong, than the simple configure time enable/disable done by distro maintainers really.
This supposed to be simple and we would over-engineer it big time.
This is really about if we allow the response the by read it more chunks, or if it needs to be read in one go. The async we work the same way for the client point of view.

williamcroberts:
> I could think of many more ways this dynamic guessing + users overwriting config goes wrong, 
than the simple configure time enable/disable done by distro maintainers really.
> This supposed to be simple and we would over-engineer it big time.

Simple for us, not for the end user. Again cons:
1. Configure flags increase the variance between versions, you need to know the config.
2. Folks can't just drop this on any kernel, IoT tends to lag.
3. You're default isn't the always works case. Defaults have to be the always works case.

It's not really over engineering, its a pretty simple addition to the kvp parser that sets a variable, and a uname call to set it to a default. Then the configure time option just goes away and Version X doesn't have any variance around that configure option. This has been the major push in a lot of software projects, including the kernel. Move away from an array of KConfig options, just figure it out if you can, the other golden rule is you cant set a default that isn't always going to work. Its ok to have a sub-optimal default, but it *has* to work. If it doesn't need a KConfig interface, don't add it. For instance, tinkering with hash table sizes on SELinux, no KConfig, just figure it out in the code what the optimal size is at policy load.

> This is really about if we allow the response the by read it more chunks, or if it needs to be read in one go. The async we work the same way for the client point of view.

I'm focused on the chunked reading, and maybe that got lost in the noise. For async, just turn it on and even better remove the config time option as that will just work.

williamcroberts:
Also, won't this break ubuntu-16.04 and 18.04? Kernel version is 4.10 and  4.15 respectively. We don't want to break the default for those folks, just like how we're still carrying OSSL compat. Many of those users are grabbing the source from here so they can actually build something that works.

tstruk:
> Simple for us, not for the end user. Again cons:
> 
> 1. Configure flags increase the variance between versions, you need to know the config.
> 2. Folks can't just drop this on any kernel, IoT tends to lag.
> 3. You're default isn't the always works case. Defaults have to be the always works case.

The the end user would just do: "sudo apt install tpm2-tss". That's easy enough.
If one would want to install from sources, one would probably read INSTALL.md file fist.
Also if IoT tends to lag on the kernel then IoT would probably lag on the libraries as well.
I could understand the "always works case" though. It will require splitting this into two PRs, one for async, and one for partial read. Might not be an optimal, but I will try to make it "always work".

tstruk:
> Also, won't this break ubuntu-16.04 and 18.04? Kernel version is 4.10 and 4.15 respectively. We don't want to break the default for those folks, just like how we're still carrying OSSL compat. Many of those users are grabbing the source from here so they can actually build something that works.

The async works. The difference is that the poll() always returns POLLIN even though there is nothing to read, but then the read() will block:
```
open("/dev/tpm0", O_RDWR|O_NONBLOCK)    = 3
write(3, "\200\1\0\0\0\26\0\0\1z\0\0\0\6\0\0\1\0\0\0\0\177", 22) = 22
poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
read(3, "\200\1\0\0\1\203\0\0\0\0\0\0\0\0\6\0\0\0.\0\0\1\0002.0\0\0\0\1\1\0"..., 4096) = 387
write(3, "\200\1\0\0\0\f\0\0\1{\0 ", 12) = 12
poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
read(3, "\200\1\0\0\0,\0\0\0\0\0 \370*\352\353\1\330A\372\0,\220\222\323\221cY7\37|3"..., 4096) = 44
fstat(1, {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 1), ...}) = 0
close(3)     
```
The partial read will return EOF on the subsequent read:
```
open("/dev/tpm0", O_RDWR|O_NONBLOCK)    = 3
write(3, "\200\1\0\0\0\26\0\0\1z\0\0\0\6\0\0\1\0\0\0\0\177", 22) = 22
poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
read(3, "\200\1\0\0\1\203\0\0\0\0", 10) = 10
poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
read(3, "", 377)                        = 0
write(2, "WARNING:tcti:src/tss2-tcti/tcti-"..., 97WARNING:tcti:src/tss2-tcti/tcti-device.c:293:tcti_device_receive() Got EOF instead of response. 
) = 97
```
and that's expected on an old kernel/driver, but again, if someone would want to install TSS-3.0 on ubuntu-16.04 from sources they would hopefully read the INSTALL.md at least.

williamcroberts:
> > Also, won't this break ubuntu-16.04 and 18.04? Kernel version is 4.10 and 4.15 respectively. We don't want to break the default for those folks, just like how we're still carrying OSSL compat. Many of those users are grabbing the source from here so they can actually build something that works.
> 
> The async works. The difference is that the poll() always returns POLLIN even though there is nothing to read, but then the read() will block:
> 
> ```
> open("/dev/tpm0", O_RDWR|O_NONBLOCK)    = 3
> write(3, "\200\1\0\0\0\26\0\0\1z\0\0\0\6\0\0\1\0\0\0\0\177", 22) = 22
> poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
> read(3, "\200\1\0\0\1\203\0\0\0\0\0\0\0\0\6\0\0\0.\0\0\1\0002.0\0\0\0\1\1\0"..., 4096) = 387
> write(3, "\200\1\0\0\0\f\0\0\1{\0 ", 12) = 12
> poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
> read(3, "\200\1\0\0\0,\0\0\0\0\0 \370*\352\353\1\330A\372\0,\220\222\323\221cY7\37|3"..., 4096) = 44
> fstat(1, {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 1), ...}) = 0
> close(3)     
> ```
> 
> The partial read will return EOF on the subsequent read:
> 
> ```
> open("/dev/tpm0", O_RDWR|O_NONBLOCK)    = 3
> write(3, "\200\1\0\0\0\26\0\0\1z\0\0\0\6\0\0\1\0\0\0\0\177", 22) = 22
> poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
> read(3, "\200\1\0\0\1\203\0\0\0\0", 10) = 10
> poll([{fd=3, events=POLLIN}], 1, -1)    = 1 ([{fd=3, revents=POLLIN}])
> read(3, "", 377)                        = 0
> write(2, "WARNING:tcti:src/tss2-tcti/tcti-"..., 97WARNING:tcti:src/tss2-tcti/tcti-device.c:293:tcti_device_receive() Got EOF instead of response. 
> ) = 97
> ```
> 
> and that's expected on an old kernel/driver, but again, if someone would want to install TSS-3.0 on ubuntu-16.04 from sources they would hopefully read the INSTALL.md at least.

People don't especially when they have already been doing it. You can document it, and you can point to the documentation, but documentation is limited. You can't opt people into broken configs. Async is less worry some, because, it will at least still work. But the partial read breaks callers. Again, when you release a version with config flags, you're releasing a lot of different versions.

tstruk:
Will split this into two and enable async for now. Will do the partial read later.

tstruk:
@coolfengyu I have added O_NONBLOCK flag to VXWORKS specific open() call to prevent blocking on partial read discovery, but I have no way to test it. It should be fine according to [1]. Could you please give it a try. If that breaks anything then we can switch it of with fcntl() or ioctl(), or however it is done on VXWORKS, after init.

[1] https://docs.windriver.com/bundle/vxworks_7_application_core_os_sr0630-enus/page/CORE/ioLib.html#open

diabonas:
The changes in this PR make the initialisation of libtss2-tcti-device fails sporadically on my system, see #1809 for more details.

PR #1748, opened 2020-06-03
Fix fapi persistent keys

JuergenReppSIT:
FAPI: Fix usage of persistent handles.
    
* Persistent handles could not be defined for Fapi_CreateKey. The persistent handle
  now can be passed with the key flags and EvictControl is executed for the persistent key.
* The cleanup for keys had to be adapted.
* Tests for derived persistent keys were added,

    
### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -270,13 +270,16 @@ TESTS_CFLAGS +=  -DTOP_SOURCEDIR"=\"$(top_srcdir)\""
 FAPI_TESTS_INTEGRATION = \
     test/integration/fapi-check-wrong-paths.fint \
     test/integration/fapi-data-crypt.fint \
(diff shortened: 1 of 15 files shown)


tstruk:
That failed:
```
../../src/tss2-fapi/fapi_util.c:3246:31: error: implicit conversion from enumeration type 'enum IFAPI_KEY_CREATE_STATE' to different enumeration type 'enum _FAPI_STATE' [-Werror,-Wenum-conversion]
    statecase(context->state, KEY_CREATE_FLUSH2);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
../../src/tss2-fapi/ifapi_macros.h:126:13: note: expanded from macro 'statecase'
        VAR=STATE;
           ~^~~~~
1 error generated.
```

Issue #1676, opened 2020-03-23
Should we get rid of libgcrypt


tstruk:
Libgcrypt was the first crypto backend for ESAPI that was added in v2.0. After that we introduced openssl in v2.1, and both are supported currently. In v3.0 with the introduction with FAPI the only usable crypto backend is openssl unless FAPI is disabled at config time. 
Since by default FAPI is enabled we have an unusable gcrypt code that needs to maintained.
Maybe we should just get rid of libgcrypt in 3.1?

AndreasFuchsSIT:
Hmmmm... since CI works, why not keep it ?
In the end, we even added mbed-tls / -crypto as backend for esys.
So I'd vote to keep it for now.
Maybe we warn about it being deprecated and thus allow us to remove it in any minor revision ?

AndreasFuchsSIT:
I guess, 3.0.0 would be a good time to drop this if we ever want to.

@tstruk @williamcroberts Thoughts ?

williamcroberts:
> I guess, 3.0.0 would be a good time to drop this if we ever want to.
> 
> @tstruk @williamcroberts Thoughts ?

I'm 100% Ok will killing libgcrypt. I would float a mailer on the list and see if any users object.

emmanuel-deloget:
As a user, I see no issue with this. On the contrary, as it would simplify my dependency-related headaches :)

diabonas:
I went through the distributions listed on [Repology](https://repology.org/project/tpm2-tss/versions) to see which backend they are using. The only distribution I could find that exclusively uses gcrypt is [Ataraxia Linux](https://github.com/ataraxialinux/ataraxia/commit/de886f93759429f6fcb124bd5e5b719567e8e37c#diff-1bc63a20d6b798cd980983ed1ca1787a) (pinging @ataraxialinux, @protonesso), and that has been a very recent change less than a week ago, apparently in an effort to get rid of a custom patch to build with LibreSSL instead of OpenSSL. All other distributions either offer a compile-time choice between the two backends like [Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/app-crypt/tpm2-tss/tpm2-tss-2.4.1.ebuild?id=ec6e9c276a25be30d330d524f7238eade26d24f3#n63), or (by far the majority) use the default OpenSSL backend.

Personally, as well as speaking as a packager for Arch Linux, I see no issue with removing support for gcrypt.

[...]

PR #1761, opened 2020-06-14
Support mbedtls as crypto-backend

drandreas:
I've been working on adding TPM 2.0 support to [Zephyr](http://github.com/zephyrproject-rtos/zephyr). Since Zephyr's primary TLS stack is based on a fork of [mbedtls](https://github.com/zephyrproject-rtos/mbedtls), I've chosen to add mbedtlse's crypto-backend to tpm2-tss. The current code lacks `iesys_cryptmbed_get_ecdh_point` but pass all unit tests not using ECDH.

I've noticed that you are considering (#1676) to drop support for all crypto backends except OpenSSL. My current work does not require ECDH support. Nevertheless, if there is interest to include mbedtls support in tpm2-tss willing to fill that gap.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -141,12 +141,13 @@ AS_IF([test "x$enable_tcti_partial_reads" = "xyes"],
 	AC_DEFINE([TCTI_PARTIAL_READ],[1], [TCTI PARTIAL READ MODE]))
 
 AC_ARG_WITH([crypto],
-            [AS_HELP_STRING([--with-crypto={gcrypt,ossl}],
+            [AS_HELP_STRING([--with-crypto={gcrypt,ossl,mbed}],
                             [sets the ESYS crypto backend (default is OpenSSL)])],,
             [with_crypto=ossl])
 
 AM_CONDITIONAL(ESYS_OSSL, test "x$with_crypto" = "xossl")
 AM_CONDITIONAL(ESYS_GCRYPT, test "x$with_crypto" = "xgcrypt")
+AM_CONDITIONAL(ESYS_MBED, test "x$with_crypto" = "xmbed")
 
 AS_IF([test "x$enable_esys" = xyes],
       [AS_IF([test "x$with_crypto" = xgcrypt], [
@@ -158,6 +159,11 @@ AS_IF([test "x$enable_esys" = xyes],
            AC_DEFINE([OSSL], [1], [OpenSSL cryptographic backend])
            TSS2_ESYS_CFLAGS_CRYPTO="$LIBCRYPTO_CFLAGS"
            TSS2_ESYS_LDFLAGS_CRYPTO="$LIBCRYPTO_LIBS"
+       ], [test "x$with_crypto" = xmbed], [
+           AC_CHECK_HEADER(mbedtls/md.h, [], [AC_MSG_ERROR([Missing required mbedTLS library])])
+           AC_DEFINE([MBED], [1], [mbedTLS cryptographic backend])
+           TSS2_ESYS_CFLAGS_CRYPTO="$LIBMBED_CFLAGS"
+           TSS2_ESYS_LDFLAGS_CRYPTO="-lmbedcrypto"
        ], AC_MSG_ERROR([Bad value for --with-crypto $with_crypto]))])
 AC_SUBST([TSS2_ESYS_CFLAGS_CRYPTO])
 AC_SUBST([TSS2_ESYS_LDFLAGS_CRYPTO])
--- a/src/tss2-esys/esys_crypto_mbed.c
+++ b/src/tss2-esys/esys_crypto_mbed.c
@@ -1,0 +1,928 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*******************************************************************************
+ * Copyright: 2020, Andreas Dröscher
+ * All rights reserved.
+ ******************************************************************************/
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdlib.h>
+#include <string.h>
+
+#include <mbedtls/md.h>
+#include <mbedtls/rsa.h>
+#include <mbedtls/ecdh.h>
+#include <mbedtls/aes.h>
+#include <mbedtls/entropy.h>
+#include <mbedtls/ctr_drbg.h>
+
+#include "esys_iutil.h"
+#include "esys_mu.h"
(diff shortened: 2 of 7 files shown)


drandreas:
I've completed the [Zephyr PoC](https://github.com/drandreas/zephyr-tpm2-poc). I hope it gives a little more context.

AndreasFuchsSIT:
I guess you'll need to add mbedcrypto to the docker containers in https://github.com/tpm2-software/tpm2-software-container in order to CI to work.

drandreas:
Thank you for your suggestion. I find it little strange that the CI fails in the first place , since I haven't added any new test cases where mbedTLS is enabled. The new code should not be reached in the first place. I assume that there must be code in the CI environement that tries all configuration options automatically. But I haven't found it yet. However since adding a 3rd crypto backend is a significant change and Moreover there are discussions in dropping gcrypt. I would like to know if there is interest in this change in the first place before I spend a significant amount of time for cleaning up the code, tests and documentation.

drandreas:
I found it `bootstrap` has some magic that puts all files inside esys into a generic esys make target.

AndreasFuchsSIT:
> I found it `bootstrap` has some magic that puts all files inside esys into a generic esys make target.

Yes, that makes sense... So I guess you'll be altering that file too.

Nevertheless, I think it would be good to have a CI version running where the new crypto backend is actually tested.

drandreas:
I've opened a PR in [tpm2-software-container](https://github.com/tpm2-software/tpm2-software-container/pull/34) that should lead to successful CI results.

configure.ac
```diff
@@ -156,6 +157,10 @@ AS_IF([test "x$enable_esapi" = xyes],
            AC_DEFINE([OSSL], [1], [OpenSSL cryptographic backend])
            TSS2_ESYS_CFLAGS_CRYPTO="$LIBCRYPTO_CFLAGS"
            TSS2_ESYS_LDFLAGS_CRYPTO="$LIBCRYPTO_LIBS"
+       ], [test "x$with_crypto" = xmbed], [
+           AC_DEFINE([MBED], [1], [mbedTLS cryptographic backend])
+           TSS2_ESYS_CFLAGS_CRYPTO="$LIBMBED_CFLAGS"
+           TSS2_ESYS_LDFLAGS_CRYPTO="-lmbedcrypto"
        ], AC_MSG_ERROR([Bad value for --with-crypto $with_crypto]))])
```
tstruk, line 164: If someone chooses "--with-crypto=mbed" then the configure script should also check if the mbedcrypto library is installed and fail gracefully if it isn't. Please add something like:
```
AC_CHECK_HEADER(mbedtls/md.h, [], [AC_MSG_ERROR([Please install mbed])])
```

src/tss2-esys/esys_crypto_mbed.c
```diff
@@ -0,0 +1,785 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+
```
tstruk, line 2: No copyrights?

tstruk:
I restarted the build after a new ubuntu20 image was built, and it worked fine, except one integration test:
```
FAIL: test/integration/esys-create-session-auth-ecc.int
```
Didn't have time to investigate this farther, but if mbedtls is missing something, then this test needs to be disabled.

drandreas:
@tstruk thank you for enabling mbedTLS support on CI and reviewing my code. 

> I restarted the build after a new ubuntu20 image was built, and it worked fine, except one integration test:
> 
> ```
> FAIL: test/integration/esys-create-session-auth-ecc.int
> ```
> 
> Didn't have time to investigate this farther, but if mbedtls is missing something, then this test needs to be disabled.

That test failed because `iesys_cryptmbed_get_ecdh_point` is not yet finished. I had decided to ask if there is interest in the mbedTLS support before wasting my time in figuring out how ec-points have to be converted from mbedTLS to tpm-format (see initial PR message). I'm currently "on" the ELC NA, nevertheless I will try to address all issues by the end of the week.

tstruk:
Maybe you can follow what's done in iesys_cryptossl_get_ecdh_point(). It would be good to have full functionality implemented before merging that in.

drandreas:
I did rebase on the latest master and therefore had to do force push. I kept the change log intact to ease the review process.

src/tss2-esys/esys_crypto_mbed.c
drandreas: I addressed it in b9f8d8f

configure.ac
drandreas: thank you for the suggestion, I added it with 50e50e6

src/tss2-esys/esys_crypto_mbed.c
```diff
@@ -1,4 +1,8 @@
 /* SPDX-License-Identifier: BSD-2-Clause */
+/*******************************************************************************
+ * Copyright: 2020, Andreas Dröscher
+ * All rights reserved.
+ ******************************************************************************/
```
tstruk, line 5: What about esys_crypto_mbed.h?

tstruk:
> Just a few comments:
> The commits [4db4a5f](https://github.com/tpm2-software/tpm2-tss/commit/4db4a5f6130dafb2e19fbe0f2f183c76c543354f), [b9f8d8f](https://github.com/tpm2-software/tpm2-tss/commit/b9f8d8f529c5efecbb502a2bd5ce7e7a8a8993a1), [3141ba6](https://github.com/tpm2-software/tpm2-tss/commit/3141ba66fd88ec736e20fbaa929023d157f4da14) should have been squashed into [796dac9](https://github.com/tpm2-software/tpm2-tss/commit/796dac972fc62aad9effe075ae0a502d8ba5bd26). It doesn't look good when there is a commit that fixes a previous commit in the same PR. Also it might heart bisection.

@drandreas please let me know if you need help with rearranging the commits.

drandreas:
I've now squashed most of the commits, I consider 995f71d being worthy to be separate since it highlights an unusually API behaviour. I will squash it if you insist.

drandreas:
I changed my mind and sqashed it.

src/tss2-esys/esys_crypto_mbed.c
drandreas: Sorry I missed that one (I'm not used to having copyright headers in each file). It's fixed with 07c34f19a7e29f6be64e2be7a0bc2747b984911d

PR #1768, opened 2020-06-29
2.4.x fixes

JuergenReppSIT:
* The hierarchy usage in Fapi_Provision was fixed.
* A segfault Fapi_GetCertificate for objects which are not of type key was fixed.
* Some segfault and leak problems were fixed.

### DIFF
--- a/src/tss2-fapi/api/Fapi_NvRead.c
+++ b/src/tss2-fapi/api/Fapi_NvRead.c
@@ -260,8 +260,7 @@ Fapi_NvRead_Finish(
         command->esys_handle = object->handle;
         command->nv_obj = object->misc.nv;
 
-        if (size)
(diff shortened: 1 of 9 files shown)


src/tss2-fapi/api/Fapi_NvRead.c
```diff
@@ -261,7 +261,7 @@ Fapi_NvRead_Finish(
         command->nv_obj = object->misc.nv;
 
         if (size)
```
tstruk, line 263: Do we still need the check here then?

JuergenReppSIT: Thanks for the hint. I have removed the check.

tstruk:
Are these changes already in master? I assume yes.

PR #1773, opened 2020-06-30
esys: fapi: add base_rc macro and number_rc

tstruk:
Almost every file in esys and fapi check for the base RC
code by masking the layer bits. This adds a new base_rc
and number_rc macros that hide the masking away.


williamcroberts:
Theirs code for this in the rcdecose library we could pull out and add to the spec. This would be useful in a lot of places.

tstruk:
> Theirs code for this in the rcdecose library we could pull out and add to the spec. This would be useful in a lot of places.

You mean to put this into tss2_tpm2_types.h or some other standard header?

PR #1774, opened 2020-07-01
TPM2 spec 1.59 changes

tstruk:


tstruk:
Wireshark TPM2.0 dissector will also support the two new commands.
https://code.wireshark.org/review/37943

test/integration/esys-certifyX509.int.c
```diff
@@ -0,0 +1,243 @@
...
+                           &creationHash, &creationTicket);
+    goto_if_error(r, "Error esys create primary", error);
+
+    /* For the partial cert we need these fields:
+     * 1) Signature Algorithm Identifier (optional)
+       2) Issuer (mandatory)
+       3) Validity (mandatory)
+       4) Subject Name (mandatory)
+       5) Extensions (mandatory)
+       Data generated with IBM's tpmutils
```
williamcroberts, line 151: What was the exact command? Is their a way to generate this with openssl?

tstruk: I used this tool: https://sourceforge.net/p/ibmtpm20tss/tss/ci/master/tree/utils/certifyx509.c
Openssl will always create a "full cert", which will not work with TPM2_Certifyx509 command unless is it manually truncated and tweaked.

tstruk: Now we can use this tool: https://github.com/tpm2-software/tpm2-tools/pull/2151/files#diff-c0e10ea13892dafc56e08bdcc6b7d32e

PR #1777, opened 2020-07-02
Fapi several extensions

JuergenReppSIT:
* Creation of keys in the NULL hierarchy is supported.
* The check of access rights during provisioning is fixed.
* A cleanup for a failed provisioning was added.
* Some minor fixes.


src/tss2-fapi/ifapi_keystore.c
```diff
@@ -445,7 +445,7 @@ ifapi_keystore_initialize(
     }
 
     /* Create user directory if necessary */
-    r = ifapi_io_check_create_dir(keystore->userdir);
+    r = ifapi_io_check_create_dir(keystore->userdir, W_OK);
```
tstruk, line 448: Should we add FAPI #defines for these access specifiers? Something like:
```
#define FAPI_WRITE W_OK
#define FAPI_READ R_OK
```
these are Linux/Unix specific and won't work on Windows.

JuergenReppSIT: I have changed it and added two more more commits. One to delete invalid null hierarchy keys and an update for doxygen.

PR #1785, opened 2020-07-13
Build: Issue warning on FAPI directory creation

AndreasFuchsSIT:
Issue a warning during install-data-hook if the FAPI keystore and eventlog directories could not be created.


tstruk:
Cherry pick into 2.4.x

PR #1789, opened 2020-07-13
2.4.x fix pcr log and encrypt

JuergenReppSIT:
* Fix setting of app data.
* Enable all PCR registers for SHA256 bank in the distribution profiles.
* Fix computation of PCR logs and PCR digest of PCR logs.
* Fix size check for Fapi_Encrypt.
* Improve comment for authentication.


tstruk:
> Looks good, but needs rebasing to get rid of the first commit.

never mind. I rebased it for you.

Issue #1750, opened 2020-06-04
Make esys change the hierarchy type to spec backwards compat


williamcroberts:
Since:
9d48cc7922ad esys: change the hierarchy type according to the spec

Interface types changed from TPMI_RH_HIERARCHY and TPMI_RH_ENABLES to ESYS_TR, which from the API and ABI perspective nothing changed, they are still just UINT32, it sonly the accepted values.

The types have no overlap, can we just keep the backwards compat version and print a warning?

This way we won't actually break existing users, their code will work, but we can urge them along to fix it.

tstruk:
There is no requirement to maintain backwards compatibility when the major version number changes, but we have to maintain it between all minor versions. If we add this "compatibility layer" and release it in 3.0.0, then we will have to maintain it for all 3.x releases.

williamcroberts:
> There is no requirement to maintain backwards compatibility when the major version number changes, but we have to maintain it between all minor versions. If we add this "compatibility layer" and release it in 3.0.0, then we will have to maintain it for all 3.x releases.

Sure, so I am saying, we don't do 3.0. IIRC/IIUC, this is the only breaking change, no? 

It's not hard to maintain, we just make it backwards compat and put a test in place.

AndreasFuchsSIT:
What shall be do here ?
I'd like to close down on 3.0 and we should make a decision.
Has anyone checked if the TPMI_RH_HIERARCHY range is within a reserved range of ESYS_TR or is that far enough apart ?
Then I don't mind a compat version, since the rest of ESYS_TR can be assigned by ESYS to its liking.

williamcroberts:
IIUC, for RH values you have 0x40000000 to 0x4000010F and for ESYS_TR the code currently does:
`esys_handle_cnt = ESYS_TR_MIN_OBJECT + (rand() % 6000000);`. So you would be assigning at most 0x1000 + 6000000 which is 0x5B9D80. So you have a gap of 0x40000000 - 0x5B9D80 = 0x3FA46280 (over 1 billion). I don't know exactly why/how you start handles there with a rand() (doesn't seem like a robust way to namesapce), but perhaps some logic could be added to make sure ESYS_TR's never flow into values above 0x40000000 - 1? 

```C
#define TPM2_RH_FIRST       ((TPM2_RH) 0x40000000) /* R */
#define TPM2_RH_SRK         ((TPM2_RH) 0x40000000) /* R */
[...]

PR #1796, opened 2020-07-18
esys: make hiearchy API value change backwards compat

williamcroberts:
Since:
  - 9d48cc7 esys: change the hierarchy type according to the spec

The interface types changed from TPMI_RH_HIERARCHY and TPMI_RH_ENABLES
to ESYS_TR, which from the ABI perspective nothing changed, asthey are
still just UINT32. However, ESYS_TR's do not equal the older TPM2_RH
values they replaced. However, their is no overlap, so we can detect
when a non ESYS_TR hiearchy value is passed and accept it. However, we
want to be noisy about it and the resulting logs will result:

ERROR:esys:src/tss2-esys/esys_iutil.c:389:iesys_handle_to_tpm_handle() Error: Esys invalid ESAPI handle (40000001).
WARNING:esys:src/tss2-esys/esys_iutil.c:411:iesys_is_platform_handle() Convert handle from TPM2_RH to ESYS_TR, got: 0x40000001

This will allow users/distro's to easily upgrade packages without having
to immediatley make code changes. However, they will be encouraged to
use the proper handle values.

Fixes: #1750

### DIFF
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -61,8 +61,6 @@ typedef uint32_t ESYS_TR;
 #define ESYS_TR_RH_AUTH_FIRST  0x110U
 #define ESYS_TR_RH_AUTH(x) (ESYS_TR_RH_AUTH_FIRST + (ESYS_TR)(x))
 
-#define ESYS_TR_MIN_OBJECT (TPM2_RH_LAST + 1 + 0x1000)
-
 typedef struct ESYS_CONTEXT ESYS_CONTEXT;
 
 /*
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -61,7 +61,7 @@ typedef uint32_t ESYS_TR;
 #define ESYS_TR_RH_AUTH_FIRST  0x110U
 #define ESYS_TR_RH_AUTH(x) (ESYS_TR_RH_AUTH_FIRST + (ESYS_TR)(x))
 
-#define ESYS_TR_MIN_OBJECT 0x1000U
+#define ESYS_TR_MIN_OBJECT (TPM2_RH_LAST + 1 + 0x1000)
 
 typedef struct ESYS_CONTEXT ESYS_CONTEXT;
 
--- a/src/tss2-esys/api/Esys_Hash.c
+++ b/src/tss2-esys/api/Esys_Hash.c
@@ -155,8 +155,12 @@ Esys_Hash_Async(
     }
 
     r = iesys_handle_to_tpm_handle(hierarchy, &tpm_hierarchy);
-    if (r != TSS2_RC_SUCCESS)
-        return r;
+    if (r != TSS2_RC_SUCCESS) {
+        if (!iesys_is_platform_handle(hierarchy)) {
+            return r;
+        }
+        tpm_hierarchy = hierarchy;
+    }
 
     r = iesys_check_sequence_async(esysContext);
     if (r != TSS2_RC_SUCCESS)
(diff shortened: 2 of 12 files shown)


williamcroberts:
@AndreasFuchsSIT and @tstruk don't we want the backwards compat fix for this for 3.0? I think it will just make the life of end users and distro packagers much more pleasant.

AndreasFuchsSIT:
Are we sure that there is no collision between the TPM2_RH_* and the TSS2_TR numberspace ?
Or would this line need to learn a filter ?
https://github.com/tpm2-software/tpm2-tss/blob/ec48d5af2e46b1fcda49259fe441c50891fd5481/src/tss2-esys/esys_context.c#L78

All ESYS_TR assignments follow the scheme of `ESYS_TR handle = esys_context->esys_handle_cnt++`

williamcroberts:
> Are we sure that there is no collision between the TPM2_RH_* and the TSS2_TR numberspace ?
> Or would this line need to learn a filter ?
> https://github.com/tpm2-software/tpm2-tss/blob/ec48d5af2e46b1fcda49259fe441c50891fd5481/src/tss2-esys/esys_context.c#L78
> 
> All ESYS_TR assignments follow the scheme of `ESYS_TR handle = esys_context->esys_handle_cnt++`

So I was thinking of just bumping the start past the RH region. But then I saw the ESYS_TR_MIN_OBJECT in the header file. But I couldn't find that definition in the spec:
 - https://trustedcomputinggroup.org/wp-content/uploads/TSS_ESAPI_v1p0_r08_pub.pdf

So assuming we can change that #define since it's not in the spec (perhaps we should remove it from public header on this 3.0 break). I doubt anyone is using it, so small break at worst.

All the code write now just does a ++ on handle, so ESAPI has no guarantees of wrap safety. We could make all the handle modifiers call into a routine to get the next handle and gap from RH_FIRST to RH_LAST, but this seemed to be the simpler approach, but does reduce the UINT32 space.
From:  4194299902 minimum possible values before rollover (over 4 billion)
TO: 3215225199 minimum possible values before rollover (over 3 billion)

It's not an order of magnitude change so I would be ok with that. But overflow safety would be a good idea to handle.

williamcroberts:
Diff for clarity of my proposal:
```diff
diff --git a/include/tss2/tss2_esys.h b/include/tss2/tss2_esys.h
index e06230ffe713..892c0c708262 100644
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -61,7 +61,7 @@ typedef uint32_t ESYS_TR;
 #define ESYS_TR_RH_AUTH_FIRST  0x110U
 #define ESYS_TR_RH_AUTH(x) (ESYS_TR_RH_AUTH_FIRST + (ESYS_TR)(x))
 
-#define ESYS_TR_MIN_OBJECT 0x1000U
+#define ESYS_TR_MIN_OBJECT (TPM2_RH_LAST + 1)
 
 typedef struct ESYS_CONTEXT ESYS_CONTEXT;
```

AndreasFuchsSIT:
P.S. RH_LAST may change, so maybe we can say `RH_LAST + 0x1000` or similar...

williamcroberts:
> P.S. RH_LAST may change, so maybe we can say `RH_LAST + 0x1000` or similar...

Good point.

williamcroberts:
> I'm ok with both, moving it to a private header / code as well as the new value.
> 
> We never did the roll-over because 4bio seemed like more than enough. I wouldn't oppose save rollover, but I don't think it's needed.

We don't need the rollover in this series, but I will open a bug. That would be a nightmare to track in a long running service. that's using the ESYS.

williamcroberts:
Sorry gitfoo fail, little CI thrash has this marked as failing...

Issue #1784, opened 2020-07-13
FAPI: Default profile ECC


AndreasFuchsSIT:
Just wanted to discuss, whether we want to switch the default FAPI profile from RSA2k to ECCP256 upon the 3.0 release.
It can still be changed by the user. It's merely the default config file:
https://github.com/tpm2-software/tpm2-tss/blob/ebfe77b41e677de6cd808e3155fe43d606951143/dist/fapi-config.json.in#L2

williamcroberts:
+1 vote for EC

tstruk:
Security wise ECC-P256 provides the same security level as RSA-3072 so it would make more sense to make the ECC-P256 the default. The question is does all the HW out there support ECC since TCG recommends that at least one of them is implemented? Probably we could go the ECC by default, and document somewhere that if the TPM returns TPM2_RC_KEY the user should fallback to RSA.

AndreasFuchsSIT:
I just realized a potential issue here:
For this to work, we need TPMs to carry ECC-based EK certificates. Otherwise, provisioning will fail; or you have to set the ek-certless option in the config.
Can we assume either of this ?
Maybe we should stick with RSA for a year or two, until ECC-EK-certs are the norm.

williamcroberts:
> I just realized a potential issue here:
> For this to work, we need TPMs to carry ECC-based EK certificates. Otherwise, provisioning will fail; or you have to set the ek-certless option in the config.
> Can we assume either of this ?
> Maybe we should stick with RSA for a year or two, until ECC-EK-certs are the norm.

Why make it so static, could you have a list of things to try or an auto keyword? Probe around and look for the "strongest" cert.

AndreasFuchsSIT:
FAPI follows the idea of crypto cleanness. So if you choose ECC in the profile, there is no RSA anywhere in the stack and vice versa. It's the idea of being save against one crypto breaking. So there is no "strongest cert".

We'd have to change the logic and basic assumptions of FAPI crypto profiles if we wanted something like this and rethink the crypto profile idea principles.

AndreasFuchsSIT:
Ok, I extended the log output in the patch so users get some idea of what's happening.
I am ok with failing provisioning with this info, so PR sent.

PR #1798, opened 2020-07-20
FAPI: Switch default profile to ECC

AndreasFuchsSIT:
Switch the default profile of FAPI to be ECC instead of RSA and provide a good error message if no ECC EK cert was found.

Fixes: #1784

### DIFF
--- a/src/tss2-fapi/api/Fapi_Provision.c
+++ b/src/tss2-fapi/api/Fapi_Provision.c
@@ -1407,7 +1407,11 @@ Fapi_Provision_Finish(FAPI_CONTEXT *context)
             } else {
                 /* No certificate was stored in the TPM and ek_cert_less was not set.*/
                 goto_error(r, TSS2_FAPI_RC_NO_CERT,
-                           "No certifcate was stored in the TPM.", error_cleanup);
+                           "No EK certifcate found for current crypto profile found. "
+                           "You may want to switch the profile in fapi-config or "
+                           "set the ek_cert_less or ek_cert_file options in fapi-config. "
+                           "See also https://tpm2-software.github.io/2020/07/22/Fapi_Crypto_Profiles.html",
+                           error_cleanup);
             }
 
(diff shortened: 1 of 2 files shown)


AndreasFuchsSIT:
Last open issue for 3.0, let's go... :-)

williamcroberts:
> Last open issue for 3.0, let's go... :-)

+1 from me, but I don't know how much this will break things. Defaults in general should be safe bets of things that generally work. I would hate to see the majority of users having to modify their config.

I guess we can roll out a longer than normal RC process and announce thing change on the list? Just to garner what type of pain and suffering this might inflict on the masses.

AndreasFuchsSIT:
Fine for me. Good thing is though, you can change it back to by editing that one line in the config file.

williamcroberts:
> Fine for me. Good thing is though, you can change it back to by editing that one line in the config file.

Yeah super simple change thankfully. We may want to have a github io write up on it to be proactive.

src/tss2-fapi/api/Fapi_Provision.c
```diff
@@ -1407,7 +1407,10 @@ Fapi_Provision_Finish(FAPI_CONTEXT *context)
             } else {
                 /* No certificate was stored in the TPM and ek_cert_less was not set.*/
                 goto_error(r, TSS2_FAPI_RC_NO_CERT,
-                           "No certifcate was stored in the TPM.", error_cleanup);
+                           "No EK certifcate found for current crypto profile found. "
+                           "You may want to switch the profile in fapi-config or "
+                           "set the ek_cert_less or ek_cert_file options in fapi-config.",
+                           error_cleanup);
```
williamcroberts, line 1413: One thought, what if had a link to the github io writeup here? The links should be stable for a long time.

tstruk:
So now we just need to add a link https://tpm2-software.github.io/2020/07/22/Fapi_Crypto_Profiles.html to the err message in Fapi_Provision.c

PR #1799, opened 2020-07-20
CHANGELOG: change ABI to API for the TPM2_RH_ change

williamcroberts:
The change from TPM2_RH constants is an API break, not an ABI break. The
function paramters remain exactly the same, a uint32_t, passed as the
same argument to the function.


AndreasFuchsSIT:
I'd have argued the other way around.
API stays the same, because it's still a uint32.
ABI changes, because the uint32 is interpreted differently.

But I have no idea if I'm correct. If you're sure then let's merge it.
Are you sure ?

williamcroberts:
> I'd have argued the other way around.
> API stays the same, because it's still a uint32.
> ABI changes, because the uint32 is interpreted differently.
> 
> But I have no idea if I'm correct. If you're sure then let's merge it.
> Are you sure ?

100% sure. ABI is the actual type and call information.

PR #1800, opened 2020-07-21
Revert "build: bump soname of libtss2-esys"

diabonas:
PR #1796 makes libtss2-esys ABI compatible with tpm2-tss 2.4, so applications using this library don't need to be recompiled right away, hence no need for a soname version bump. Since no version with the bumped soname has been released, we can just revert it.

This reverts commit c94a14e65cd482cb9b3767a49535976cfd66a687.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -399,11 +399,6 @@ src_tss2_esys_libtss2_esys_la_LDFLAGS = $(AM_LDFLAGS) $(LIBSOCKET_LDFLAGS) \
 if HAVE_LD_VERSION_SCRIPT
 src_tss2_esys_libtss2_esys_la_LDFLAGS += -Wl,--version-script=$(srcdir)/lib/tss2-esys.map
 endif # HAVE_LD_VERSION_SCRIPT
-
-# Bump soname due to commit 9d48cc7922ad8dc3496444b04c94439c5bfbec8e breaking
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
Maybe we should make it 1:0:1 ?

diabonas:
Generally that would be correct, signalling a backwards-compatible change with a soname of `libtss2-esys.so.0.1.0` (instead of the current `libtss2-esys.so.0.0.0`). To be useful, we would have to increase the first and last part of the soname version number every time there is an ABI change (like newly added functions) for any of the libraries (so we would have to start using `-version-info` for the other libraries as well). On the other hand, only the first part of the soname version is used on Linux by default anyway, so this feels like something that is quite easily missed while cutting a new release and which most applications will not make use of anyway.

Personally I would just use the first component of the soname version to signal backwards-incompatible changes for *existing* applications that will have to be recompiled immediately. The other components of the soname only concern *new* functionality and could theoretically be used by applications to see whether a minimum required version of the library is available. However I feel it is easier to use the version information in the pkg-config file at compile time for this purpose, like we do e.g. for tpm2-tools.

williamcroberts:
LGTM but Ill leave the final call to @AndreasFuchsSIT

AndreasFuchsSIT:
I'm ok.

Issue #1763, opened 2020-06-25
make install data hook failing


williamcroberts:
Since: 8765814c64505163afe6c6614a99c45df9f994a2
```
commit 8765814c64505163afe6c6614a99c45df9f994a2
Author: Tadeusz Struk <tadeusz.struk@intel.com>
Date:   Mon May 18 15:20:10 2020 -0700

    build: no need to add user or change perms on FreeBSD
    
    This is done automagically by pkg
    
    Signed-off-by: Tadeusz Struk <tadeusz.struk@intel.com>
```
make install has been failing

```mv /usr/local/lib/udev/rules.d/tpm-udev.rules /usr/local/lib/udev/rules.d/tpm-udev.rules
mv: '/usr/local/lib/udev/rules.d/tpm-udev.rules' and '/usr/local/lib/udev/rules.d/tpm-udev.rules' are the same file
Makefile:24070: recipe for target 'install-data-hook' failed
```

On tip of master, to reproduce:
```bash
./boostrap
./configure --enable-debug
make -j4
sudo make install
```

williamcroberts:
If anyone else bumps into this, as a work-around I just did a
`git revert 8765814`

tstruk:
That's probably when you want to overwrite an existing install. I will add a force flag to mv

williamcroberts:
I am seeing this fail again:
```
git describe --always --dirty
3.0.0-rc0-8-g566afa722f8f
```

```
(    (id -g tss || groupadd -r tss) && (id -u tss || useradd -r -g tss tss) &&     (    (    if [ ! -d /usr/local/var/run/tpm2-tss/eventlog/ ]; then mkdir -p /usr/local/var/run/tpm2-tss/eventlog/; fi) || true) && (    (    if [ ! -d /usr/local/var/lib/tpm2-tss/system/keystore/ ]; then mkdir -p /usr/local/var/lib/tpm2-tss/system/keystore/; fi)) && (all set_fapi_permissions)) || true
if [ ! -d "/usr/local/var/run/tpm2-tss/eventlog/" ]; then echo "*** WARNING *** Directory /usr/local/var/run/tpm2-tss/eventlog/ could not be created"; fi
if [ ! -d "/usr/local/var/lib/tpm2-tss/system/keystore/" ]; then echo "*** WARNING *** Directory /usr/local/var/lib/tpm2-tss/system/keystore/ could not be created"; fi
[...]

PR #1805, opened 2020-07-23
build: change install-hook to mv udev file only if prefix not empty

tstruk:
Use install -C to install the udev files.
This will take care of a situation when the install
is trying to overwrite an existing installation and the files are the same.

Fixes: #1763

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -675,7 +675,9 @@ if HOSTOS_LINUX
 endif
 
 install-data-hook: install-dirs
-	-mv -f $(DESTDIR)$(udevrulesdir)/tpm-udev.rules $(DESTDIR)$(udevrulesdir)/$(udevrulesprefix)tpm-udev.rules
+	-if [ ! -z "$(udevrulesprefix)" ]; then \
+		mv $(DESTDIR)$(udevrulesdir)/tpm-udev.rules $(DESTDIR)$(udevrulesdir)/$(udevrulesprefix)tpm-udev.rules; \
+	fi
 
 uninstall-local:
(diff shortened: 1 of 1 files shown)


williamcroberts:
Still seems to fail
```
(systemd-sysusers && systemd-tmpfiles --create) || \
(    (id -g tss || groupadd -r tss) && (id -u tss || useradd -r -g tss tss) &&     (    (    if [ ! -d /usr/local/var/run/tpm2-tss/eventlog/ ]; then mkdir -p /usr/local/var/run/tpm2-tss/eventlog/; fi) || true) && (    (    if [ ! -d /usr/local/var/lib/tpm2-tss/system/keystore/ ]; then mkdir -p /usr/local/var/lib/tpm2-tss/system/keystore/; fi)) && (all set_fapi_permissions)) || true
if [ ! -d "/usr/local/var/run/tpm2-tss/eventlog/" ]; then echo "*** WARNING *** Directory /usr/local/var/run/tpm2-tss/eventlog/ could not be created"; fi
if [ ! -d "/usr/local/var/lib/tpm2-tss/system/keystore/" ]; then echo "*** WARNING *** Directory /usr/local/var/lib/tpm2-tss/system/keystore/ could not be created"; fi
install -C /usr/local/lib/udev/rules.d/tpm-udev.rules /usr/local/lib/udev/rules.d/tpm-udev.rules
install: '/usr/local/lib/udev/rules.d/tpm-udev.rules' and '/usr/local/lib/udev/rules.d/tpm-udev.rules' are the same file
Makefile:24404: recipe for target 'install-data-hook' failed
make[2]: [install-data-hook] Error 1 (ignored)
```

williamcroberts:
I think the problem is the src and dest are the same file

PR #1811, opened 2020-07-30
FAPI: Remove duplicate json_object_put in event log processing.

JuergenReppSIT:
If an log string with invalid json is passed to Fapi_Extend fapi_object_put is
called twice. Depending on the version of lib-json-c this causes an error.

### DIFF
--- a/src/tss2-fapi/ifapi_eventlog.c
+++ b/src/tss2-fapi/ifapi_eventlog.c
@@ -313,7 +313,6 @@ ifapi_eventlog_append_finish(
 
         r = ifapi_json_IFAPI_EVENT_serialize(&eventlog->event, &event);
         if (r) {
-            json_object_put(eventlog->log);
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/ifapi_eventlog.c
```diff
@@ -313,7 +313,6 @@ ifapi_eventlog_append_finish(
 
         r = ifapi_json_IFAPI_EVENT_serialize(&eventlog->event, &event);
         if (r) {
-            json_object_put(eventlog->log);
```
tstruk, line 316: Then we should we remove line 331 also.

JuergenReppSIT: I will add:  eventlog->log = NULL;
because in the success case the json_object_put would not be called if the line is removed.
Thanks for the hint.

PR #1812, opened 2020-07-30
FAPI: Remove duplicate json_object_put in event log processing.

JuergenReppSIT:
If an log string with invalid json is passed to Fapi_Extend fapi_object_put is
called twice. Depending on the version of lib-json-c this causes an error.


JuergenReppSIT:
It can be cherry picked from #1811 with the change.

Issue #1810, opened 2020-07-30
Files and directories under /var/lib/tpm2-tss/system/keystore are owned by the user that created them


diabonas:
The FAPI directory `/var/lib/tpm2-tss/system/keystore` is owned by the user and group `tss` thanks to the [tmpfiles.d configuration](https://github.com/tpm2-software/tpm2-tss/blob/master/dist/tmpfiles.d/tpm2-tss-fapi.conf.in) (or the [manual fallback](https://github.com/tpm2-software/tpm2-tss/blob/965ef058d8df57bc9b6277ca22aed16b5a856a39/Makefile.am#L599-L602)):
```
$ ls -l /var/lib/tpm2-tss /var/lib/tpm2-tss/system
/var/lib/tpm2-tss:
total 0
drwxr-xr-x 1 tss tss 16 2020-07-30 15:40 system

/var/lib/tpm2-tss/system:
total 0
drwxrwxr-x 1 tss tss 42 2020-07-30 15:40 keystore
```
However, files created under these directories, e.g. using `Fapi_Provision`, are owned by the user that created them:
```
$ ls -l /var/lib/tpm2-tss/system/keystore
total 0
drwxr-xr-x 1 diabonas diabonas 26 2020-07-30 14:00 P_ECCP256SHA256
drwxr-xr-x 1 diabonas diabonas  0 2020-07-30 14:00 policy
```
Since this is the system key store, all files should be modifiable by anybody in the `tss` group. The easiest way to achieve this would be setting the permissions of `/var/lib/tpm2-tss/system/keystore` to `2775`, the `setgid` bit makes sure that all files and directories inherit the group of the parent directory (recursively):
```
$ ls -l /var/lib/tpm2-tss/system/keystore
total 0
drwxr-xr-x 1 diabonas tss 26 2020-07-30 14:00 P_ECCP256SHA256
drwxr-xr-x 1 diabonas tss  0 2020-07-30 14:00 policy
```
However this means if a user created system keys, but is no longer part of the `tss` group, they are still able to modify or delete these keys, which is suboptimal. I see no easy way to avoid this issue (using the `setuid` bit to inherit the owner of the parent directory seems to work under BSD, but not under Linux) other than to remove write access to the group to enforce that all files are created by the `tss` user. This would mean that users will have to take care to run library functions using the `tss` user themselves.

AndreasFuchsSIT:
[...]

PR #1814, opened 2020-07-30
FAPI: use setgid for system keystore

diabonas:
Files in the system keystore should be modifiable by any user in the `tss` group, not only by the user that created them. Apply the setgid bit to the keystore directory to make sure that all directories and files under the keystore directory (recursively) are owned by the `tss` group.

This change will be applied to existing keystore directories as well, so newly created files will be owned by the `tss` group from now on. Modifying the owner of already existing files is out of scope for this PR.

Using setgid makes sure that all newly created directories and files in the system keystore are owned by the `tss` group. However, permissions for new files are controlled by the current umask, which might not grant write access to the group. To solve this, set a default ACL for the system keystore directory that allows read/write access for members of the `tss` group. The default ACL is inherited for newly created directories and files, granting access to the `tss` group regardless of the umask setting.

As for the setgid change, the default ACL is applied to an existing keystore directory as well, but not to existing directories and files below it.

Fixes: #1810

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -598,7 +598,8 @@ endef
 
 define set_tss_permissions
     (chown -R tss:tss "$1") && \
-    (chmod -R 2775 "$1")
+    (chmod -R 2775 "$1") && \
+    (setfacl -m default:group:tss:rwx "$1")
 endef
 
 define make_fapi_dirs
--- a/Makefile.am
+++ b/Makefile.am
@@ -598,7 +598,7 @@ endef
 
 define set_tss_permissions
     (chown -R tss:tss "$1") && \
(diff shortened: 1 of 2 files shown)


diabonas:
Hm, unfortunately this is not enough. While files and directories directly under `/var/lib/tpm2-tss/system/keystore` will be modifiable by members of the `tss` group, this does not necessarily apply to files in nested subdirectories: these are created using the default umask, which is `022` for many Linux distribution, meaning that files and directories are not writeable by the group.

It might be possible to achieve what we want using ACLs, I'll have to play around with that a bit.

diabonas:
ACLs appear to be the way to go, see the updated pull request.

Issue #1809, opened 2020-07-30
Sporadic initialisation failures of libtss2-tcti-device with tpm2-tss 3.0.0-rc1: "Failed to read response header fd 3, got errno 0: Success"


diabonas:
When running tpm2-tss 3.0.0-rc1 on my system (Arch Linux, kernel 5.7.11) with a discrete (Infineon) TPM, initialisation of `libtss2-tcti-device` sporadically fails with the error message "Failed to read response header fd 3, got errno 0: Success":
```
$ TSS2_LOG=all+trace tpm2_getrandom 1
[...]
debug:tcti:src/tss2-tcti/tctildr-dl.c:245:tctildr_get_default() Attempting to connect using standard TCTI: Access libtss2-tcti-device.so.0 with /dev/tpmrm0
trace:tcti:src/tss2-tcti/tctildr-dl.c:136:tcti_from_file() Attempting to load TCTI file: libtss2-tcti-device.so.0
trace:tcti:src/tss2-tcti/tctildr.c:96:tcti_from_info() Attempting to load TCTI info
trace:tcti:src/tss2-tcti/tctildr.c:103:tcti_from_info() Loaded TCTI info named: tcti-device
trace:tcti:src/tss2-tcti/tctildr.c:104:tcti_from_info() TCTI description: TCTI module for communication with Linux kernel interface.
trace:tcti:src/tss2-tcti/tctildr.c:105:tcti_from_info() TCTI config_help: Path to TPM character device. Default value is: TCTI_DEVICE_DEFAULT
trace:tcti:src/tss2-tcti/tctildr.c:61:tcti_from_init() Initializing TCTI for config: /dev/tpmrm0
debug:tcti:src/tss2-tcti/tcti-device.c:437:Tss2_Tcti_Device_Init() Trying to open specified TCTI device file /dev/tpmrm0
trace:tcti:src/tss2-tcti/tcti-device.c:444:Tss2_Tcti_Device_Init() Successfully opened specified TCTI device file /dev/tpmrm0
debug:tcti:src/tss2-tcti/tcti-device.c:449:Tss2_Tcti_Device_Init() Probe device for partial response read support
debug:tcti:src/util/io.c:87:write_all() writing 12 bytes starting at 0x7ffd92fc87b4 to fd 3
debug:tcti:src/util/io.c:101:write_all() wrote 12 bytes to fd 3
[...]

PR #1815, opened 2020-07-30
Fix timing issue in partial read probe

tstruk:
Fixes: #1809


tstruk:
@diabonas please give it a try.

tstruk:
> Thank you for the quick fix, after applying this PR, I was able to run `tpm2_getrandom 1` a hundred times in a row without any issues, so the issue seems to be fixed.

Thanks. I tested it on FreeBSD also.

PR #1820, opened 2020-08-05
test/esys-certifyX509: use TPMA_OBJECT_X509SIGN depending on integration_tcti

diabonas:
Currently the `esys-certifyX509 integration` test enables `TPMA_OBJECT_X509SIGN`, which only works with the mssim TCTI at the moment, [depending on the existence of the `TCTI_MSSIM` macro](https://github.com/tpm2-software/tpm2-tss/blob/1361aa370994eea08b446f1de975f918f9574e8c/test/integration/esys-certifyX509.int.c#L82-L87). This macro is defined [if the mssim TCTI is enabled](https://github.com/tpm2-software/tpm2-tss/blob/1361aa370994eea08b446f1de975f918f9574e8c/configure.ac#L185-L191), i.e. if `libtss2-tcti-mssim` is built. This excludes that valid use case that the user wants to have the mssim TCTI enabled (e.g. for compatibility with other tpm2-software test suites), but also has swtpm available for running the tpm2-tss integration test suite with `libtss2-tcti-swtpm`. Introduce a new macro `INTEGRATION_TCTI_MSSIM` that is defined if [`$integration_tcti`](https://github.com/tpm2-software/tpm2-tss/blob/1361aa370994eea08b446f1de975f918f9574e8c/configure.ac#L310-L334) is set to `mssim` in configure so that users can run the test suite with swtpm without having to use `./configure --disable-tcti-mssim`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -328,7 +328,8 @@ AS_IF([test "x$enable_integration" = "xyes"],
               AC_MSG_NOTICE([Falling back to testing with tcti-mssim.])
               AC_CHECK_PROG([result_tpm_server], [tpm_server], [yes], [no])
               AS_IF([test "x$result_tpm_server" = "xyes"],
-                    [integration_tcti=mssim],
+                    [integration_tcti=mssim
+                     AC_DEFINE([INTEGRATION_TCTI_MSSIM], [1], [use mssim TCTI for integration tests])],
                     [AC_MSG_WARN([Executable tpm_server not found in PATH (fallback)])])])
        AS_IF([test "x$integration_tcti" = "xnone"],
              [AC_MSG_ERROR([No simulator executable found in PATH for testing TCTI.])])
--- a/test/integration/esys-certifyX509.int.c
+++ b/test/integration/esys-certifyX509.int.c
@@ -79,7 +79,7 @@ test_esys_certifyx509(ESYS_CONTEXT * esys_context)
                     TPMA_OBJECT_USERWITHAUTH |
                     TPMA_OBJECT_RESTRICTED |
                     TPMA_OBJECT_SIGN_ENCRYPT |
-#if defined TCTI_MSSIM
+#if defined INTEGRATION_TCTI_MSSIM
                     TPMA_OBJECT_X509SIGN | /* This requires mssim >= v1628 Other simulators
                                               that do not implement 1.59 spec return 0x000002e1
                                               tpm:parameter(2):reserved bits not set to zero as required


tstruk:
> Introduce a new macro `INTEGRATION_TCTI_MSSIM` that is defined if [`$integration_tcti`](https://github.com/tpm2-software/tpm2-tss/blob/1361aa370994eea08b446f1de975f918f9574e8c/configure.ac#L310-L334) is set to `mssim` in configure so that users can run the test suite with swtpm without having to use `./configure --disable-tcti-mssim`.

The swtpm simulator is used by default, and users don't need to do ./configure --disable-tcti-mssim to use it. Currently it is the opposite - if one wants to use mssim one needs to ./configure --disable-tcti-swtpm.
There is something else going on here. We do have a CI build that does that (./configure --disable-tcti-swtpm ...)
https://github.com/tpm2-software/tpm2-tss/blob/master/.ci/docker.run#L78
but still swtpm is used. I will look into that.

diabonas:
I don't fully understand either why this is working on Travis CI without an explicit `--disable-tcti-mssim`, but this PR appears to do the "right thing" in my eyes: `TCTI_MSSIM` is [always defined if tcti-mssim is enabled](https://github.com/tpm2-software/tpm2-tss/blob/4b7441bbcf760400fd846e090f984586d319d4ae/configure.ac#L191), but at that time it is not known yet which TCTI will be used for testing. Therefore it makes sense to use the result of the [`integration_tcti` auto-detection magic](https://github.com/tpm2-software/tpm2-tss/blob/master/configure.ac#L310-L3349) instead, which is not exposed in the form of a preprocessor macro yet.

diabonas:
I found out why this works in Travis CI, but not on my local machine: I am using the latest released version 0.7.3 of libtpms, which does not support `TPMA_OBJECT_X509SIGN`. On the other hand, the docker images use the [libtpms master branch](https://github.com/tpm2-software/tpm2-software-container/blob/bca8926a34ce555371e42bc73c15ef05a9f6a27d/ubuntu-20.04.docker#L89), where support for `TPMA_OBJECT_X509SIGN` was introduced in https://github.com/stefanberger/libtpms/commit/f743d7faaed86b071473d7fc6cfec793677386ef, but [this has not made it to a released version yet](https://github.com/stefanberger/libtpms/blob/v0.7.3/src/tpm2/TpmTypes.h#L581-L582).

So I see two options:
1. Use this PR to remove `TPMA_OBJECT_X509SIGN` when testing with swtpm, even though the current libtpms master branch that we are using for the CI actually has support for it.
2. Remove  [`#if defined TCTI_MSSIM`](https://github.com/tpm2-software/tpm2-tss/blob/master/test/integration/esys-certifyX509.int.c#L82) and thus make `TPMA_OBJECT_X509SIGN` unconditional since it works on the libtpms master branch, and tell everybody to use that branch for running the test suite.

tstruk:
I would lean toward option 1. Should be sufficient to test this on mssim for now.
If we will make it unconditional it might trigger too traffic on the mailing list because the default configuration will start failing for people.

PR #1834, opened 2020-08-27
Fapi several fixes

JuergenReppSIT:
* Policy execution was fixed.
* Pathname check was added for policy files.
* New provisioning after deleting /HS was enabled.
* A missing return code check for ChangeAuth was added.


AndreasFuchsSIT:
I'll merge #1837 
The added error messages will help us in the future.
I'll blame the tpm simulator for now, since the FlashContext responses are as short as it gets...

PR #1838, opened 2020-08-27
tcti: fix collisions with multi-threading

joholl:
When multiple instances of tcti-pcap are writing to the same file at the same time collisions can appear. Fix the following three issues:

* The random host is seeded by the unix epoch in seconds. If multiple threads start roughly at the same time, their packages are associated with the same connection. Move to nanoseconds.
* A random port between 50k and 60k is prone to collisions. Instead use a fixed port and randomize both IP addresses.
* Wireshark sometimes seems to choke on the sequence and acknowledgement number not being implemented properly. Randomize their initial values and implement the acknowledge number properly.

### DIFF
--- a/src/tss2-tcti/tcti-pcap-builder.c
+++ b/src/tss2-tcti/tcti-pcap-builder.c
@@ -8,10 +8,10 @@
 #include <config.h>
 #endif
 
+#include <time.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
-#include <time.h>
 #include <errno.h>
 #include <unistd.h>
 #include <fcntl.h>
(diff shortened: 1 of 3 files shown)


tstruk:
> It fails cirrus build:
> error CC src/tss2-sys/api/libtss2_sys_la-Tss2_Sys_LoadExternal.lo
> : use of undeclared identifier 'CLOCK_MONOTONIC_RAW'

I think you just need to include <time.h>

tstruk:
Try this:
```
diff --git a/src/tss2-tcti/tcti-pcap-builder.c b/src/tss2-tcti/tcti-pcap-builder.c
index 640d14e1..79eb4871 100644
--- a/src/tss2-tcti/tcti-pcap-builder.c
+++ b/src/tss2-tcti/tcti-pcap-builder.c
@@ -8,8 +8,6 @@
 #include <config.h>
 #endif
 
-#undef _POSIX_C_SOURCE
-#define _POSIX_C_SOURCE 199309L
 #include <time.h>
 #include <stdlib.h>
 #include <stdint.h>
@@ -26,6 +24,10 @@
 #define LOGMODULE tcti
 #include "util/log.h"
 
+#ifdef __FreeBSD__
+#define CLOCK_MONOTONIC_RAW CLOCK_MONOTONIC
+#endif
+
 /* constants used */
 #define PCAP_MAJOR                          0x0001
 #define PCAP_MINOR                          0x0000
```

joholl:
Thank you for this hint! I'm not a BSD expert and tried some glibc defines, but to no avail. Builds now on Cirrus.

tstruk:
> Thank you for this hint! I'm not a BSD expert and tried some glibc defines, but to no avail. Builds now on Cirrus.

No problem. Looks good now. Thanks

PR #1840, opened 2020-08-29
FAPI: Fix cleanup of policy sessions in error case.

JuergenReppSIT:
For successful execution policy sessions are closed automatically. In error cases the policy
sessions have to be flushed in the cleanup function.


AndreasFuchsSIT:
Cirrus Image seems broken. Complains about missing tpm_server

Thus, I'll merge this

PR #1843, opened 2020-09-04
io.c: don't remap AGAIN to TSS2_TCTI_RC_TRY_AGAIN

### LINKED ISSUE #1842: Bad return value
This should be -1 since it's a ssize_t return type

https://github.com/tpm2-software/tpm2-tss/blob/e025f59723e0c0e2a6c9556643f1e71902af5df2/src/util/io.c#L61

Reported by @PeterHuewe

tstruk:
Don't need to remap the AGAIN to TPM2 AGAIN since
we implement all retry logic araound the poll() calls.
All the error cases from the socket_recv_buf()
translate to TSS2_TCTI_RC_IO_ERROR in both mssim_tcti
and swtpm_tcti. These two are the only users of the
socket_recv_buf() function.

Fixes #1842

### DIFF
--- a/src/util/io.c
+++ b/src/util/io.c
@@ -57,9 +57,6 @@ read_all (
 #else
         TEMP_RETRY (recvd, read (fd, &data [recvd_total], size));
         if (recvd < 0) {
-            if (errno == EAGAIN) {
-                return TSS2_TCTI_RC_TRY_AGAIN;
-            }
             LOG_WARNING ("read on fd %d failed with errno %d: %s",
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT (approved):

AndreasFuchsSIT:
@tstruk should this be backported ?

tstruk:
> @tstruk should this be backported or was that code only added uring the async support changes on master ?

That is only on master for the async change, as you said. No backport needed.

PR #1862, opened 2020-10-01
TCTI: Add generic platform independent SPI TCTI module

jpxd:
This PR adds a new SPI TCTI module for TPM communication over SPI e.g. in embedded systems like the ESP32.

The SPI TCTI uses user supplied methods for SPI and timing operations in order to be platform independent.
These methods are supplied to `Tss2_Tcti_Spi_Init` via the  `TSS2_TCTI_SPI_PLATFORM` struct.

Example code for using the SPI TCTI on an ESP32 can be found in `doc/tcti-spi.md`.


include/tss2/tss2_tcti_spi.h
```diff
@@ -0,0 +1,45 @@
...
+typedef struct TSS2_TCTI_SPI_PLATFORM TSS2_TCTI_SPI_PLATFORM;
+typedef void (*TSS2_TCTI_SPI_PLATFORM_SLEEP_MS_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform, int milliseconds);
+typedef void (*TSS2_TCTI_SPI_PLATFORM_START_TIMEOUT_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform, int milliseconds);
+typedef bool (*TSS2_TCTI_SPI_PLATFORM_TIMEOUT_EXPIRED_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform);
+typedef void (*TSS2_TCTI_SPI_PLATFORM_SPI_ACQUIRE_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform);
+typedef void (*TSS2_TCTI_SPI_PLATFORM_SPI_RELEASE_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform);
+typedef TSS2_RC (*TSS2_TCTI_SPI_PLATFORM_SPI_TRANSFER_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform, const void *data_out, void *data_in, size_t cnt);
+
+struct TSS2_TCTI_SPI_PLATFORM {
+    uint64_t magic;
```
AndreasFuchsSIT, line 26: This should not be necessary, since we have type safety whenever this struct is used.

include/tss2/tss2_tcti_spi.h
```diff
@@ -0,0 +1,45 @@
...
+#include <driver/spi_master.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define TSS2_TCTI_SPI_PLATFORM_MAGIC 0x61F7D74109DB0D8BULL
+
+typedef struct TSS2_TCTI_SPI_PLATFORM TSS2_TCTI_SPI_PLATFORM;
+typedef void (*TSS2_TCTI_SPI_PLATFORM_SLEEP_MS_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform, int milliseconds);
```
AndreasFuchsSIT, line 18: IMHO, this is a bit ugly. Instead of platform, we should only pass the user_data pointer to the callback functions.
Also applies to all subsequent function defs.

Question is, if all callbacks share one userdata or if we want to have a different userdata pointer for each callback ?

include/tss2/tss2_tcti_spi.h
```diff
@@ -0,0 +1,45 @@
...
+typedef void (*TSS2_TCTI_SPI_PLATFORM_START_TIMEOUT_FUNC) (TSS2_TCTI_SPI_PLATFORM* platform, int milliseconds);
```
AndreasFuchsSIT, line 19: All of these functions need some more documentation.
What's the service expected from this function ?

AndreasFuchsSIT (changes requested):
This is some very nice work !!!
Some technical comments inlined.

Besides this, I think this requires some more documentation in doc/
Also some more inline documentation (especially for the header file) would be needed to be useful.

As an example you could provide samples for the callbacks on an ESP and give some wireing information.

jpxd:
Thanks for the feedback @AndreasFuchsSIT!

I will add more documentation as requested.

Where should I put the example code for the ESP32 platform funcs & wiring information?
Should I force-push my combined changes as a single commit or should I add new commits to this PR?

AndreasFuchsSIT:
I'd just put some example code snippets into the documentation, nothing extra and no complete program needed.

Then you can just force-push.

include/tss2/tss2_tcti_spi.h
jpxd: Magic has been removed from TSS2_TCTI_SPI_PLATFORM.

include/tss2/tss2_tcti_spi.h
jpxd: We now only pass user_data to the callback functions.

In most cases the user_data will contain a struct with SPI device handle, SPI CS pin handle, timeout expiry time.

A shared user_data pointer is needed for e.g. the timeout functions which need to operate on the same timer / timestamp. I think different user data pointers per method would only be confusing and not helpful.

include/tss2/tss2_tcti_spi.h
jpxd: More documentation has been added in the header file and in doc/tcti-spi.md

AndreasFuchsSIT:
Awesome, that looks very good.
I just noticed however that we could even go ahead and compile and unit-test this on regular x86 linux platforms, right ?
I think we should do so.
I guess adding the library as noinst_LIBRARIES to the makefile, guarded by a if TEST and then a unit-test that links against the library, so we know that it works in the future.

AndreasFuchsSIT:
Maybe in the future, we could make it a inst LIBRARY because a ftdi-usb-spi tcti would also be kind of cool... ;-)

AndreasFuchsSIT:
P.S. I'd add the makefile integration for the library in the current commit and the unit-test into an extra commit.

jpxd:
I just pushed my attempt to add tcti-spi as a library to the build system (used tcti-pwd as reference).
But somehow `MAXLOGLEVEL` is not set in my build which leads to some unused-parameter and unused-but-set-variable errors, since the log functions are not inserted by the log macro:

```
In file included from src/tss2-tcti/tcti-spi.c:24:0:
./src/util/log.h:55:2: error: #error "MAXLOGLEVEL undefined"
 #error "MAXLOGLEVEL undefined"
```

Can you help me out? I used the Dockerfile to compile.

Because I'm using a completely different build system with ESP-IDF for my ESP32 builds, I'm not really familiar with the TSS build system.

PeterHuewe:
Very cool proposal - i hope I can find time to review it soon.

PeterHuewe:
@AndreasFuchsSIT : Is it possible to run the device tests against this tcti ? This would be a nice way for testing

doc/tcti-spi.md
```diff
@@ -0,0 +1,246 @@
...
+    spi_bus_config_t bus_cfg = {
+        .miso_io_num = PIN_NUM_MISO,
+        .mosi_io_num = PIN_NUM_MOSI,
+        .sclk_io_num = PIN_NUM_CLK,
+        .quadwp_io_num = -1,
+        .quadhd_io_num = -1,
+        .max_transfer_sz = 64
+    };
+    spi_device_interface_config_t dev_cfg = {
+        .clock_speed_hz = 10*1000, // 10 KHz
```
PeterHuewe, line 182: just an example, but 10khz is quite low.

include/tss2/tss2_tcti_spi.h
```diff
@@ -0,0 +1,96 @@
...
+typedef void (*TSS2_TCTI_SPI_PLATFORM_SLEEP_MS_FUNC) (void* user_data, int milliseconds);
+
+/* 
+ * Starts a timeout timer which expires in the specified amount of milliseconds.
+ * This can be done by storing the expire time (now + milliseconds) in the userdata.
+ */
+typedef void (*TSS2_TCTI_SPI_PLATFORM_START_TIMEOUT_FUNC) (void* user_data, int milliseconds);
+
+/* 
+ * Returns true if the timout started previously by START_TIMEOUT_FUNC already has expired, false otherwise.
```
PeterHuewe, line 34: timeout

lib/tss2-tcti-spi.map
```diff
@@ -0,0 +1,7 @@
+{
+    global:
+        Tss2_Tcti_Info;
+        Tss2_Tcti_Cmd_Init;
```
PeterHuewe, line 4: Tss2_Tcti_Spi_Init;

lib/tss2-tcti-spi.def
```diff
@@ -0,0 +1,4 @@
+LIBRARY tss2-tcti-spi
+EXPORTS
+    Tss2_Tcti_Info
+    Tss2_Tcti_Cmd_Init
```
PeterHuewe, line 4: Tss2_Tcti_Spi_Init

src/tss2-tcti/tcti-spi.c
```diff
@@ -0,0 +1,612 @@
...
+#include "tss2_tcti.h"
+#include "tss2_tcti_spi.h"
+#include "tss2_mu.h"
+#include "tcti-common.h"
+#include "tcti-spi.h"
+#include "util/io.h"
+#define LOGMODULE tcti
+#include "util/log.h"
+
+static void spi_tpm_delay_ms(TSS2_TCTI_SPI_CONTEXT* ctx, int milliseconds)
```
PeterHuewe, line 26: mark these helper functions as inline?
Code is probably identical, but still

src/tss2-tcti/tcti-spi.c
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    // Send header
+    uint8_t header_response[4];
+    rc = spi_tpm_spi_transfer(ctx, header, header_response, 4);
+    if (rc != TSS2_RC_SUCCESS) {
+        return TSS2_TCTI_RC_IO_ERROR;
+    }
+
+    // Wait until the TPM exits the wait state and sends a 1 bit
```
PeterHuewe, line 102: can be consolidated into one chunk, by movinf the byte&1 to the begining of the loop.

src/tss2-tcti/tcti-spi.c
```diff
@@ -0,0 +1,612 @@
...
+    
+    // The 1 bit is often already set in the last byte of the transaction header
+    byte = header_response[3];
+    if (byte & 1) {
+        return TSS2_RC_SUCCESS;
+    }
+    
+    // Otherwise try again a few times by reading ony byte at a time
+    uint8_t zero = 0;
+    for (int retries = 512; retries > 0; retries--) {
```
PeterHuewe, line 113: not sure what TCG specifies as max amount of waitstates.
is there any delay in between?

PeterHuewe (commented):
Looks good in general , a few nitpicks, but looks good

AndreasFuchsSIT:
There are some scan-build error in there (https://travis-ci.org/github/tpm2-software/tpm2-tss/jobs/733492007#L1697):
```
../../src/tss2-tcti/tcti-spi.c:215:5: warning: Undefined or garbage value returned to caller
    return status;
    ^~~~~~~~~~~~~
../../src/tss2-tcti/tcti-spi.c:233:5: warning: Undefined or garbage value returned to caller
    return access;
    ^~~~~~~~~~~~~
../../src/tss2-tcti/tcti-spi.c:565:21: warning: The left operand of '!=' is a garbage value
        if (did_vid != 0) break;
            ~~~~~~~ ^
../../src/tss2-tcti/tcti-spi.c:590:14: warning: Assigned value is garbage or undefined
    revision = rid;
             ^ ~~~
```

You can run `make clean; scan-build make` locally to get some more details.

doc/tcti-spi.md
```diff
@@ -0,0 +1,246 @@
+# SPI TCTI
+
+The SPI TCTI can be used for TPM communication over SPI e.g. in embedded systems like the ESP32.
+It uses user supplied methods for SPI and timing operations in order to be platform independent.
+This methods are supplied to `Tss2_Tcti_Spi_Init` via the `TSS2_TCTI_SPI_PLATFORM` struct.
```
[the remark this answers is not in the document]
jpxd: Fixed in next force-push.

doc/tcti-spi.md
jpxd: Yes, but a higher SPI clock speed occasionally produced bad SPI transfer results on the demo setup using ~15cm jumper cables between the ESP32 and the TPM (parasitic capacitance?) when 10Khz worked reliably. On a circuit board the speed could be a lot higher.
What is your suggestion for an example clock speed?

jpxd: Should I just add the supported clock speed range as a comment in the example config?

include/tss2/tss2_tcti_spi.h
jpxd: Fixed with the next force-push.

lib/tss2-tcti-spi.map
jpxd: Fixed with the next force-push.

lib/tss2-tcti-spi.def
jpxd: Fixed with the next force-push.

src/tss2-tcti/tcti-spi.c
jpxd: I figured gcc would be smart enough to automatically inline, but you're right and `spi_tpm_read_be32` was also explicitly marked as inline. So all platform helper methods will be inlined with the next force-push.

doc/tcti-spi.md
PeterHuewe: Hmm,
so basically we have seen a lot of spi masters with 'broken' frequencies - so it's a bit hard to tell - 10khz sounds really low though. did you try to scope higher frequencies to see whether it's an electrical issue or a controller/master issue?
I just also want to rule out that for whatever reason, some sleeps/timeouts are missing or incorrect which are only triggered at higher frequencies.

Most TPMs run up to 33mhz. On my rpi overlay I specified 32 mhz.
For a microcontroller I think 1-10 Mhz should be a suitable range for spi.

But of course cable length and overspeaking can be an issue.

I'll try to test your commit using the letstrust arduino adapter (https://buyzero.de/products/arduino-adapter-for-letstrust-tpm)  and my esp32 with arduino form factor and see what frequency works on a sensible connection.

PeterHuewe: you can keep it as 10khz, if this is your tested speed, but maybe add a comment about up to 33mhz max

jpxd: Update: I just tried to increase the clock rate up to 22MHz (the ESP32 does not support more using SPI on regular GPIO pins in full-duplex mode) and **it works**!
I just had to make a little fix in `Tss2_Tcti_Spi_Init` because at 22MHz and without logging (which slowed down the ESP just enough that it worked, but without logging it broke) we are too fast for the the TPM after claiming the locality. So now we wait for the STS_COMMAND_READY flag to be set after claiming the locality. And now it works reliably at 22MHz with and without logging.

src/tss2-tcti/tcti-spi.c
jpxd: Yes, but then we would not check the last read byte, right?!
For me this "fast path" before the retry loop also conveys a better intuition of the protocol when reading the code.

src/tss2-tcti/tcti-spi.c
jpxd: I could not find a general maximum amount of wait states in the specification. The spec only says:

> The TPM SHALL NOT insert more than 1 wait state on a Read cycle to the following registers: [some special registers]

No there is no delay in between. But there probably should be because 512 wait states mean something different at 1MHz than at 20MHz.

jpxd: What do you suggest? Waiting indefinitely for the wait states to end or some kind of timeout / retry count?
`tcti_spi_receive` does have a timeout parameter (although it is always set to `TSS2_TCTI_TIMEOUT_BLOCK`??) but `spi_tpm_start_transaction` is used everywhere, so we would need a sensible timeout even when not in `tcti_spi_receive`.
Also: What delay should we insert between retries? In my tests there never has been more than one wait state (I think) but I only tested with GetRandom and Quote... Would a 1ms delay be sufficient? We obviously want register transactions to be fast...

jpxd:
> There are some scan-build error in there (https://travis-ci.org/github/tpm2-software/tpm2-tss/jobs/733492007#L1697):

@AndreasFuchsSIT These are fixed by now.
Do you have any idea why the gcc-fedora builds are failing in the CI but not the clang-fedora builds?

AndreasFuchsSIT:
It's the addresssanitizer that's detecting something. No clue what it means though
```
ERROR: test/integration/fapi-quote-destructive
==============================================
Trying to start simulator swtpm
Starting simulator on port 62652
successfully started daemon: swtpm with PID: 31835
/workspace/tpm2-tss/build
simulator PID: 31835
LISTEN 0      1          127.0.0.1:62652      0.0.0.0:*    users:(("swtpm",pid=31835,fd=4))
LISTEN 0      1          127.0.0.1:62653      0.0.0.0:*    users:(("swtpm",pid=31835,fd=3))
Simulator with PID 31835 bound to port 62652 and  62653 successfully.
TPM20TEST_TCTI_NAME=swtpm
TPM20TEST_DEVICE_FILE=
TPM20TEST_SOCKET_ADDRESS=127.0.0.1
TPM20TEST_SOCKET_PORT=62652
TPM20TEST_TCTI=swtpm:host=127.0.0.1,port=62652
=================================================================
==2209==ERROR: AddressSanitizer: odr-violation (0x7fc910b8bd00):
  [1] size=40 'tss2_tcti_info' ../src/tss2-tcti/tcti-cmd.c:623:22
  [2] size=40 'tss2_tcti_info' ../src/tss2-tcti/tcti-spi.c:606:22
These globals were registered at these points:
  [1]:
    #0 0x7fc910f30cc8  (/lib64/libasan.so.6+0x37cc8)
    #1 0x7fc910b7bcae in _sub_I_00099_3 (/workspace/tpm2-tss/build/src/tss2-tcti/.libs/libtss2-tcti-cmd.so.0+0x1acae)
    #2 0x7fc9118e77e1 in call_init.part.0 (/lib64/ld-linux-x86-64.so.2+0x117e1)
  [2]:
    #0 0x7fc910f30cc8  (/lib64/libasan.so.6+0x37cc8)
    #1 0x7fc910b3cf90 in _sub_I_00099_3 (/workspace/tpm2-tss/build/src/tss2-tcti/.libs/libtss2-tcti-spi.so.0+0x18f90)
    #2 0x7fc9118e77e1 in call_init.part.0 (/lib64/ld-linux-x86-64.so.2+0x117e1)
==2209==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
SUMMARY: AddressSanitizer: odr-violation: global 'tss2_tcti_info' at ../src/tss2-tcti/tcti-cmd.c:623:22
==2209==ABORTING
TPM_StartUp failed
ERROR test/integration/fapi-quote-destructive.fint (exit status: 99)
```

src/tss2-tcti/tcti-spi.c
PeterHuewe: In your test there shouldn't be any waitstate at all ;) The 0x01 always comes at byte 4 for the tpm you were using :P
I'm fine with the current implementation - 512 seems a bit much, but in the end it's a don't care.

src/tss2-tcti/tcti-spi.c
PeterHuewe: I don't get the first part about.
I'm fine with the  current implementation.

doc/tcti-spi.md
PeterHuewe: Too fast for the TPM ? Probably not - the particular TPM is testes with 33Mhz+
Maybe your retry-loop is too fast and something expiring.
But yeah you fixed it.

As I said, sometimes the masters also have issues with certain clock rates.
22MHZ sound fine - 10 are probably less stress with the ESP :P

PeterHuewe:
odr-violation - one definition rule - it complains that tcti-cmd and tcti-spi both define the same global `tss2_tcti_info`, (as do all other tcti's.)
It might be the case (one would have to check) that for whatever reason tcti-spi pulls in tcti-cmd within the same compilation unit.

PeterHuewe:
the `tss2_tcti_info` should be static anyway.

doc/tcti-spi.md
jpxd: Okay, so in the example in `doc/tcti-spi.md` I will specify 10MHz but comment that it has been tested up to 22MHz.

src/tss2-tcti/tcti-spi.c
jpxd: Okay, I lowered the retry count to 256 and added a comment.

src/tss2-tcti/tcti-spi.c
jpxd: Okay, I lowered the retry count to 256 and added a comment.

jpxd:
> the `tss2_tcti_info` should be static anyway.

What do you mean with this @PeterHuewe ?

> It might be the case (one would have to check) that for whatever reason tcti-spi pulls in tcti-cmd within the same compilation unit.

I could not find why this happens (and only on these two fedora builds?!)... Any hints?

AndreasFuchsSIT:
> > the `tss2_tcti_info` should be static anyway.
> 
> What do you mean with this @PeterHuewe ?

The variable can be declared as static because it is only ever used inside that same C-file.

> > It might be the case (one would have to check) that for whatever reason tcti-spi pulls in tcti-cmd within the same compilation unit.
> 
> I could not find why this happens (and only on these two fedora builds?!)... Any hints?

Not all builds run the ASAN analyzer or not even in the same version. That's why it fails on those two boxes.

williamcroberts:
I wonder if this should handle being able to add in the TIS commands for communicating over SPI?

@iolivergithub any input?

AndreasFuchsSIT:
@jpxd Any progress on the unit test ?

jpxd:
@AndreasFuchsSIT 
I already started writing unit tests for the SPI-TCTI using a fake-SPI interface with cmocka, but had to pause that because I currently have to finish my bachelor's thesis. After that I can continue the work on the unit tests.

A few weeks ago I sent @PeterHuewe some ESP32 example projects, so he can also test the SPI-TCTI on his side.

About the linker / address sanitizer problem in the CI: I just could not replicate the error on my local machine running the unit tests with the sanitizer enabled... Any help here is greatly appreciated.
You said "the tss2_tcti_info should be static anyway", but I think changing the tss2_tcti_info to static in ALL TCTCIs is out of scope for this PR.

AndreasFuchsSIT:
If course, do your stuff.
If you want, you can switch this PR to "draft" status and post your intermediate versions as well.

Regarding the static: I guess making your info file static is enough for this PR, I agree.

Regarding asan: best is to use the scripts from the .ci folder directly to test. Of course if you have an older version of asan (Fedora is pretty much latest usually), then there might be errors that you cannot detect locally. A docker container (as done by the CI scripts) helps here.

PeterHuewe:
Hmm,
what if there is no way to hold the CS over multiple spi transfers ? (many spi masters have that limitation) and one does not want or cannot use a gpio. For certain tpms  (e.g. ifx) it would be possible to ignore the waitstate handling as no waitstates are generated in normal circumstances  (and thus avoid the limitiation of certain masters)

Would this be somehow possible (e.g. if the acquire function is NULL)?

PeterHuewe:
Hi,
since we have a 
`    TSS2_TCTI_SPI_PLATFORM_FINALIZE finalize;`
does it make sense to also include a `TSS2_TCTI_SPI_PLATFORM_INITIALIZE initialize` function ?
to include the posiblility for a one-time setup (a little bit like the initial part of your `create_tcti_ctx` function)

src/tss2-tcti/tcti-spi.c
```diff
@@ -0,0 +1,625 @@
...
+}
+
+static TSS2_RC spi_tpm_claim_locality(TSS2_TCTI_SPI_CONTEXT* ctx)
+{
+    uint8_t access;
+    access = spi_tpm_read_access_reg(ctx);
+    
+    // Check if locality 0 is busy
+    if (access & TCTI_SPI_TPM_ACCESS_ACTIVE_LOCALITY) {
+        LOG_ERROR("Locality 0 is busy, status: %#x", access);
```
PeterHuewe, line 253: If we already have the active locality this is FINE, not an error.
Localities are not really a concurrency mechanism, more a privilege level thing.
Some systems initially set once in user mode locality 0 as active and keep it this way.

PeterHuewe (commented):
@jpxd - ping :)

doc/tcti-spi.md
```diff
@@ -0,0 +1,250 @@
...
+
+    // Free resources inside user_data like SPI device handles here
+    // ...
+    
+    // Free user_data
+    free(platform_data);
+}
+
+// Creates a new platform struct with the given SPI device handle and chip select
+TSS2_TCTI_SPI_PLATFORM create_tcti_spi_platform(spi_device_handle_t spi, gpio_num_t cs_pin)
```
PeterHuewe, line 151: returning a full struct here makes it impossible to deal with any errors - e.g. if malloc fails.
For other tctis that use this as a reference, the create_tcti_spi_platform might be less trivial.
Probably makes sense to return a pointer instead.

src/tss2-tcti/tcti-spi.c
```diff
@@ -0,0 +1,625 @@
...
+    // Wait for the expected status with or without timeout
+    do {
+        status = spi_tpm_read_sts_reg(ctx);
+        // Return success on expected status
+        if ((status & status_mask) == status_expected) {
+            return TSS2_RC_SUCCESS;
+        }
+        // Delay next poll by 8ms to avoid spamming the TPM
+        spi_tpm_delay_ms(ctx, 8);
+    } while (blocking || spi_tpm_timeout_expired(ctx));
```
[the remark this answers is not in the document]
jpxd: Hi, you are right! This was the wrong way round.

jpxd:
@PeterHuewe 
> Hmm,
> what if there is no way to hold the CS over multiple spi transfers ? (many spi masters have that limitation) and one does not want or cannot use a gpio. For certain tpms (e.g. ifx) it would be possible to ignore the waitstate handling as no waitstates are generated in normal circumstances (and thus avoid the limitiation of certain masters)
> 
> Would this be somehow possible (e.g. if the acquire function is NULL)?

I see. The wolfTPM library by wolfSSL solves this by using only *one* SPI-callback function which is always responsible for *whole* transactions (see [the wolfTPM example](https://github.com/wolfSSL/wolfTPM/blob/master/examples/tpm_io.c#L215)). In my SPI-TCTI the transaction / wait state logic is hidden from the implemented "dumb" IO-methods. But this way I need multiple calls to the transfer method without loosing the CS, which leads to the extra acquire and release functions.

Doing it the wolfTPM-way, the application needs to decide for themselves, if they need to implement wait state handing. This can be good, but is also added complexity.

If we used an approach with setting the acquire and release functions to NULL in some cases, what would the expected transaction logic be? Assume no waitstates at all? This would need to be documented very well.

src/tss2-tcti/tcti-spi.c
jpxd: So the TCTI should just check if locality 0 is *not* already active, and only then request access via `TCTI_SPI_TPM_ACCESS_REQUEST_USE`? As I understand the PTP document requesting access to an already active locality does nothing, so we could just drop the check and error and just blindly request access every time?

PeterHuewe: I think this is a viable solution.

jpxd: The first or the second suggested one? ;)
1. check and request if not active
2. always request and do not check

doc/tcti-spi.md
```diff
@@ -0,0 +1,246 @@
...
+{
+    // Create required platform user data
+    PLATFORM_USERDATA* platform_data = malloc(sizeof(PLATFORM_USERDATA));
```
williamcroberts, line 150: Shouldn't our example NULL check this? Perhaps use calloc for belts and suspenders?

PeterHuewe:
@jpxd any updates on this one? I would really like to see this feature to be available.

jpxd:
Hi, as far as I can see, there were three points still open:

1. wait state handling
2. unit tests
3. ASAN error

**To point 1 (wait states):**
Here I was still waiting for feedback on how to proceed. As written in my last post there are two or even three variants:

**Variant 1**
The `TSS2_TCTI_SPI_PLATFORM` struct remains as it is, but we allow to disable the wait-state handling by passing NULL for the spi_acquire and spi_release functions.

**Variant 2a**
We follow wolfTPM's example: There would then be only one `spi_transfer_transaction` method in `TSS2_TCTI_SPI_PLATFORM` to be implemented by the user, which is called only once per register transaction. This method must take care of any wait states itself (e.g. with a GPIO-CS). In the examples I would include two sample codes with and without wait-state handling. The methods `spi_acquire` and `spi_release` are omitted from `TSS2_TCTI_SPI_PLATFORM`.

**Variant 2b**
In addition to variant 2a, we could streamline the interface even further by omitting the user-implemented timing methods (analogous to wolfTPM) and removing `TSS2_TCTI_SPI_PLATFORM` completely. Timing is then controlled via fixed retro count and via optionally-definable sleep method (as in wolfTPM, see [here](https://github.com/wolfSSL/wolfTPM/blob/cb70c00ebab3cedff6ac36d99aa8d5733fff4411/wolftpm/tpm2_types.h#L346) and [here](https://github.com/wolfSSL/wolfTPM/blob/cb70c00ebab3cedff6ac36d99aa8d5733fff4411/wolftpm/tpm2_types.h#L295)).
The Init method would then have the following signature without `TSS2_TCTI_SPI_PLATFORM`:
```
TSS2_RC Tss2_Tcti_Spi_Init (TSS2_TCTI_CONTEXT *tctiContext, size_t *size,
    TSS2_TCTI_SPI_PLATFORM_SPI_TRANSACTION_TRANSFER_FUNC spi_transaction_transfer, void *user_ctx);
```

I think variant 2a or 2b would be nicest. But in the end I don't care which of the variants it will be and therefore need feedback what is best / wanted here (@AndreasFuchsSIT @PeterHuewe). When this is decided, I can do the necessary changes.

To point 2 (unit tests):
I had already started to write unit tests that fake the SPI interface with cmocka, but this has to wait for the final decision on the user implemented spy methods, which have to bee mocked (see point one). I'll try to continue working with the limited time I have for this and can commit my intermediate state to the PR. Would it be theoretically possible to accept the PR without unit tests and handle the unit tests in a separate PR @AndreasFuchsSIT ?

Regarding point 3 (ASAN error):
Since I could not reproduce the ASAN problem from the Github-CI locally with Docker, I am dependent on assistance here. Could you maybe look into this @PeterHuewe ?

src/tss2-tcti/tcti-spi.c
```diff
@@ -0,0 +1,625 @@
...
+    // Use ctx as a shorthand for tcti_spi
+    TSS2_TCTI_SPI_CONTEXT* ctx = tcti_spi;
+
+    // Expected status bits for valid status and data availabe
+    uint32_t expected_status_bits = TCTI_SPI_TPM_STS_VALID | TCTI_SPI_TPM_STS_DATA_AVAIL;
+
+    // Check if we already have received the header
+    if (tcti_common->header.size == 0) {
+        // Wait for response to be ready
+        rc = spi_tpm_wait_for_status(ctx, expected_status_bits, expected_status_bits, timeout);
```
AndreasFuchsSIT, line 377: If timeout == NON_BLOCK we should still set a timeout of e.g. 1 second here.

AndreasFuchsSIT:
With my VERY VERY basic understanding of the issue, I have a tendency for Variante 1.
This serves the purpose of being a most flexible driver framework for the most use cases.

Maybe we could also allow the timeout-callbacks to be NULL and the implementation would then attempt a max-retries variant with e.g. 5 retries ? Or with timeout divided by 50ms or something.

jpxd:
@PeterHuewe Any comments on the proposed variants? If not, I can start working on variant 1.

@AndreasFuchsSIT Yes, we can also allow disabling the timeout functions and fallback to a retry count.

PeterHuewe:
Sorry I somehow missed that - variant 1 sounds fine.

AndreasFuchsTPM:
@jpxd Are you still planning to update this ?
Anyone else who is / wants to be working on this ?

wxleong:
Sorry for the reference spam, not a good idea to add issue references in commit message. I'll pick up what @jpxd had left off, a new PR is created https://github.com/tpm2-software/tpm2-tss/pull/2397

jpxd:
Hi all, unfortunately I currently don't have enough time to work on this project and get the PR over the finish line.

@wxleong Thank you a lot for continuing the work on this!!!

williamcroberts:
Where did we land with this @AndreasFuchsTPM ?

williamcroberts:
merged: https://github.com/tpm2-software/tpm2-tss/pull/2397/files

PR #1872, opened 2020-10-16
Fapi: Fix setting of the written flag of NV objects in FAPI NV commands

JuergenReppSIT:
* The written flag was not set for increment, set bits, and extend, which caused wrong computation
   of NV name after deserialization.
* For PolicyNV the WRITTEN bit is set during policy calculations, because during policy execution
    it will be set.
*  To check whether the nv written flag was set correctly after fapi increment this
    test was added.

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -311,6 +311,7 @@ FAPI_TESTS_INTEGRATION = \
     test/integration/fapi-key-create-policy-signed.fint \
     test/integration/fapi-key-create-policy-signed-ecc.fint \
     test/integration/fapi-key-create-policy-nv-sign.fint \
+    test/integration/fapi-key-create-policy-nv-counter-sign.fint \
     test/integration/fapi-key-create-policy-or-sign.fint \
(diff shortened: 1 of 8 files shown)


AndreasFuchsSIT (commented):
What happens if we create the key that references the counter before the first increment ? Then the policy will have a cleared NVWRITTEN bit, and fail after the first increment.
(it will also fail before the first increment).

Should we maybe check during PolicyNV_calculate if its a counter and in that case allways assume an NVWRITTEN bit ?

Same Question also applies to extend indexes, right ?

JuergenReppSIT:
> What happens if we create the key that references the counter before the first increment ? Then the policy will have a cleared NVWRITTEN bit, and fail after the first increment.
> (it will also fail before the first increment).
> 
> Should we maybe check during PolicyNV_calculate if its a counter and in that case allways assume an NVWRITTEN bit ?
> 
> Same Question also applies to extend indexes, right ?

Yes that's true. Wouldn't it make sense to set NVWRITTEN independent of the NV Type for computation of PolicyNV?

Issue #1888, opened 2020-10-30
FAPI not provisioned error (although everything works)


FadedCoder:
```
ERROR:fapi:src/tss2-fapi/api/Fapi_List.c:221:Fapi_List_Finish() FAPI not provisioned. 
ERROR:fapi:src/tss2-fapi/api/Fapi_List.c:81:Fapi_List() ErrorCode (0x00060034) Entities_List 
ERROR: Listing FAPI token objects failed.
```
I get this error whenever I'm trying to use the `tpm2_ptool`, like with SSH. Everything seems to work, but getting this error everytime is a bit annoying.

Any way to fix or disable this?

Using [the official tpm2-tss](https://www.archlinux.org/packages/community/x86_64/tpm2-tss/) and tpm2-pkcs11 tools for Arch Linux.

heobu:
I am getting the same error and was about to create an issue for it.
What could be the reason for this? For the most part everything seems to work fine, but this error message can be seen as a result of the execution all the time.

For example, I executed this on the terminal:
`tpm2pkcs11-tool --list-objects`

This error message was the result of that followed by the list of objects as expected.

AndreasFuchsSIT:
the tpm2-pkcs11 project allows for two different backends; ESYS and FAPI.
This means that FAPI is not provisioned. but it will still work with the ESYS backend.
I guess we could turn this into a WARNING, if you prefer this ?

joholl:
I agree that it should be a warning, since it is not a (fatal) error.

williamcroberts:
> the tpm2-pkcs11 project allows for two different backends; ESYS and FAPI.
> This means that FAPI is not provisioned. but it will still work with the ESYS backend.
> I guess we could turn this into a WARNING, if you prefer this ?

Yeah a warning would be nice, I get this question all the time in tpm2-pkcs11 land.

williamcroberts:
> > the tpm2-pkcs11 project allows for two different backends; ESYS and FAPI.
> > This means that FAPI is not provisioned. but it will still work with the ESYS backend.
> > I guess we could turn this into a WARNING, if you prefer this ?
> 
> Yeah a warning would be nice, I get this question all the time in tpm2-pkcs11 land.

[...]

PR #1898, opened 2020-11-16
Fapi backport fixes 3.0.x

JuergenReppSIT:
* Include policyRef into policy search for policy authorize.
* Add check for path of hierarchies in provisioning.
* Fix memory leak in policy execution.
* Adapt return codes of the test check wrong paths.
* Fix return codes of Fapi_List (Addresses #1888)
* Accept certificates with no CRL dist points


pdxjohnny:
> This passes coverity

PR #1900, opened 2020-11-16
Fapi backport fixes 2.4.x

JuergenReppSIT:
* Include policyRef into policy search for policy authorize.
* Fix memory leak in policy execution.
* Accept certificates with no CRL dist points


pdxjohnny:
This passes coverity

PR #1909, opened 2020-11-19
Fapi export policy compute digest

JuergenReppSIT:
* One bugfix for cleanup after policy evaluation with errors was added.
* The computation of policy digests was added to Fapi_ExportPolicy if
  the policy to be exported was read from the policy store.
* An integration test for all polices used in integration tests was added.

### DIFF
--- a/test/integration/fapi-export-policy.int.c
+++ b/test/integration/fapi-export-policy.int.c
@@ -1,0 +1,272 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*******************************************************************************
+ * Copyright 2017-2018, Fraunhofer SIT sponsored by Infineon Technologies AG
+ * All rights reserved.
+ *******************************************************************************/
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <inttypes.h>
+#include <string.h>
+#include <errno.h>
+#include <fcntl.h>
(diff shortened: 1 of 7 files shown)


AndreasFuchsSIT:
Scan-Build-Complaint:
```
../../src/tss2-fapi/ifapi_policy.c:157:31: warning: Access to field 'policy_elements' results in a dereference of a null pointer (loaded from variable 'eval_ctx')

    eval_ctx->policy_elements = NULL;

    ~~~~~~~~                  ^

1 warning generated.
```

test/integration/fapi-export-policy.int.c
```diff
@@ -0,0 +1,152 @@
...
+            goto error;
+
+        r = Fapi_Import(context, policies[i], json_policy);
+        goto_if_error(r, "Error Fapi_Import", error);
+
+        policy = NULL;
+        r = Fapi_ExportPolicy(context, policies[i], &policy);
+        goto_if_error(r, "Error Fapi_ExportPolicy", error);
+        assert(policy != NULL);
+        assert(strlen(policy) > ASSERT_SIZE);
```
AndreasFuchsSIT, line 129: IMHO this should contain a check for the policyDigest to be correct (i.e. strcmp() against a known good value), since that the reason of this test.

AndreasFuchsSIT (changes requested):

AndreasFuchsSIT (approved):

PR #1910, opened 2020-11-19
ESYS OSSL Crypto: Cleanups

williamcroberts:
* Reduce usages of OSSL _ex suffixed routines when not needed.
* Remove duplicate calls of EVP_EncryptInit and EVP_DecryptInit as key and iv are set in the initial call; no need for the piecemeal style setting of key and iv.


williamcroberts:
I <3 negative diffstats

AndreasFuchsSIT (approved):

PR #1913, opened 2020-11-20
FAPI: fix provisioning check in rel_path_to_abs_path()

diabonas:
Commit 05d9ebd1c3ffe7011fc5062211c3a7013f9b33df added a provisioning check to `rel_path_to_abs_path()` that always returns early with `TSS2_FAPI_RC_PATH_NOT_FOUND` if the FAPI is provisioned. This leads to dead code in the rest of the function and possibly wrong return codes, breaking software that relies on a return value of `TSS2_FAPI_RC_KEY_NOT_FOUND` instead of `TSS2_FAPI_RC_PATH_NOT_FOUND`, e.g. [tpm2-pkcs11](https://github.com/tpm2-software/tpm2-pkcs11/blob/8f2b96fbbbeb6c236d719857f86fdc672bf3b6fc/src/lib/backend_fapi.c#L385-L387).

### DIFF
--- a/src/tss2-fapi/ifapi_keystore.c
+++ b/src/tss2-fapi/ifapi_keystore.c
@@ -537,10 +537,7 @@ rel_path_to_abs_path(
         r = ifapi_check_provisioned(keystore, rel_path, &provision_check_ok);
         goto_if_error(r, "Provisioning check.", cleanup);
 
-        if (provision_check_ok) {
-            goto_error(r, TSS2_FAPI_RC_PATH_NOT_FOUND,
-            "Path not found: %s.", cleanup, rel_path);
-        } else {
+        if (!provision_check_ok) {
             goto_error(r, TSS2_FAPI_RC_NOT_PROVISIONED,
                        "FAPI not provisioned for path: %s.",
                        cleanup, rel_path);


JuergenReppSIT (changes requested):
yes TSS2_FAPI_RC_KEY_NOT_FOUND will not be returned. Additionally the error code for
the hierarchy check in this function should be changed to TSS2_FAPI_RC_PATH_NOT_FOUND. And the Part of the error messages ""FAPI not provisioned." for the object checks should be deleted.

diabonas:
> yes TSS2_FAPI_RC_KEY_NOT_FOUND will not be returned. Additionally the error code for
> the hierarchy check in this function should be changed to TSS2_FAPI_RC_PATH_NOT_FOUND. And the Part of the error messages ""FAPI not provisioned." for the object checks should be deleted.

I am not quite sure I understand this comment correctly: I believe these issues were already fixed by you in commit bd08b5a874f260f0aa4bcfade93e1bb9a6e57a8c on the master branch that this PR targets. Do you mean that commit should be backported to 3.0.x in #1914 as well?

JuergenReppSIT:
Sorry, I just was confused because I opened the 3.0.x version of this function in editor.
I will approve your PR. Perhaps we also should backport also this PR to 3.0.x because of
the wrong RC for the hierarchy check.

JuergenReppSIT (approved):

diabonas:
No worries :)

> Perhaps we also should backport also this PR to 3.0.x because of
the wrong RC for the hierarchy check.

I have opened #1914 for that, to which I also added bd08b5a874f260f0aa4bcfade93e1bb9a6e57a8c in order to have all changes related to `rel_path_to_abs_path` in one place.

PR #1924, opened 2020-11-25
FAPI: Improve error messages for reading of config and profile files.

JuergenReppSIT:
* If mandatory fields are not found an error message with the field name
  will be displayed instead of "BAD VALUE"
* If a field value  cannot be deserialized now here also the field name will
  be part of the error message.


AndreasFuchsSIT:
@JuergenReppSIT Could you rebase to master so we get CI-testing ?

AndreasFuchsSIT (approved):

Issue #1008, opened 2018-05-04
ESAPI: Have Esys_TR_SetAuth() handle large passwords


AndreasFuchsSIT:
Currently, Esys_TR_SetAuth() is restricted to authValues shorter or equal to the nameAlg digest size.
The TPM 2.0 library spec part 1 recommends hashing authVaules with the nameAlg if they exceed the nameAlg digest size, but to keep then in the plain if they are shorter.

The ESAPI should adopt this behavior. We shall implement this here and also propose it to TCG.

AndreasFuchsSIT:
This of course also has to touch Esys_Create, Esys_CreatePrimary, Esys_CreateLoaded, Esys_NV_DefineSpace

tstruk:
@AndreasFuchsSIT I don't get that. Since The TPM 2.0 spec recommends this, why do we need to propose it to TCG?

AndreasFuchsSIT:
@tstruk TPM2 spec recommends it. TSS spec does not contain it. We need to propose it as mandatory behaviour for the TSS spec.

williamcroberts:
So does PR #1929 address all of this? What about hierarchy authorizations? Do sequence objects come through that API?

PR #1929, opened 2020-11-27
ESYS: Enable large auth value for keys and NV objects (Addresses #1008).

JuergenReppSIT:
* The size of the auth value for keys and NV objects is restricted to the size
  of the name hash algorithm of these objects.  Esapi will use the
  hash of the user auth values as auth value for these object.
  The maximal size will be sizeof(TPMU_HA).
* Test cases for keys and NV objects with large auth values were added.

### DIFF
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -1554,3 +1554,55 @@ iesys_tpm_error(TSS2_RC r)
              (r & TSS2_RC_LAYER_MASK) == TSS2_RESMGR_TPM_RC_LAYER ||
              (r & TSS2_RC_LAYER_MASK) == TSS2_RESMGR_RC_LAYER));
 }
+
+
+/** Replace auth value with Hash for long auth values.
+ *
+ * if the size of auth value exceeds hash_size the auth value
+ * will be replaced with the hash of the auth value.
+ *
+ * @param[in,out] auth_value The auth value to be adapted.
+ * @param[in] hash_alg The hash alg used for adaption.
+ * @retval TSS2_RC_SUCCESS if the function call was a success.
+ * @retval TSS2_ESYS_RC_BAD_VALUE if an invalid hash is passed.
+ * @retval TSS2_ESYS_RC_MEMORY if the ESAPI cannot allocate enough memory.
+ * @retval TSS2_ESYS_RC_GENERAL_FAILURE for a failure during digest
+ *         computation.
+ */
+TSS2_RC
+iesys_hash_long_auth_values(
+    TPM2B_AUTH *auth_value,
+    TPMI_ALG_HASH hash_alg)
+{
+    TSS2_RC r;
+    IESYS_CRYPTO_CONTEXT_BLOB *cryptoContext;
+    TPM2B_AUTH hash2b;
+    size_t hash_size;
+
+    r = iesys_crypto_hash_get_digest_size(hash_alg, &hash_size);
+    return_if_error(r, "Get digest size.");
+
+    if (auth_value && auth_value->size > hash_size) {
+        /* The auth value has to be adapted. */
+        r = iesys_crypto_hash_start(&cryptoContext, hash_alg);
+        return_if_error(r, "crypto hash start");
+
+        r = iesys_crypto_hash_update(cryptoContext, &auth_value->buffer[0],
+                                     auth_value->size);
+        goto_if_error(r, "crypto hash update", error_cleanup);
+
+        r = iesys_crypto_hash_finish(&cryptoContext, &hash2b.buffer[0],
+                                     &hash_size);
+        goto_if_error(r, "crypto hash finish", error_cleanup);
+
+        memcpy(&auth_value->buffer[0], &hash2b.buffer[0], hash_size);
+        auth_value->size = hash_size;
+    }
+    return r;
+
+ error_cleanup:
+    if (cryptoContext) {
+        iesys_crypto_hash_abort(&cryptoContext);
+    }
+    return r;
+}
--- a/src/tss2-esys/esys_tr.c
+++ b/src/tss2-esys/esys_tr.c
@@ -381,6 +381,8 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
 {
     RSRC_NODE_T *esys_object;
     TSS2_RC r;
+    TPMI_ALG_HASH name_alg = TPM2_ALG_NULL;
+
     _ESYS_ASSERT_NON_NULL(esys_context);
     r = esys_GetResourceObject(esys_context, esys_handle, &esys_object);
     if (r != TPM2_RC_SUCCESS)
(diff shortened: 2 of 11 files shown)


AndreasFuchsSIT:
@JuergenReppSIT Could you rebase to master so we get CI-testing ?

src/tss2-esys/esys_tr.c
```diff
@@ -392,7 +394,20 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
         if (authValue->size > sizeof(TPMU_HA)) {
             return_error(TSS2_ESYS_RC_BAD_SIZE, "Bad size for auth value.");
         }
```
williamcroberts, line 396: Shouldn't this condition go away, because now we hash it?

test/integration/esys-create-session-auth.int.c
```diff
@@ -80,6 +80,12 @@ test_esys_create_session_auth(ESYS_CONTEXT * esys_context)
         .buffer = {1, 2, 3, 4, 5}
     };
 
+#ifdef TEST_LARGE_AUTH
+    for (int i = 0; i < 33; i++)
```
williamcroberts, line 84: shouldn't we really push these over size of TPMU_HA?

williamcroberts (changes requested):
See previous comments on auth value sizes

src/tss2-esys/esys_tr.c
JuergenReppSIT: No the maximal size still will be sizeof(TMU_HA).

test/integration/esys-create-session-auth.int.c
JuergenReppSIT: I would prefer to use 33 in this case, because it's the first size greater than the size of the name alg.

src/tss2-esys/esys_tr.c
williamcroberts: Only when you're not hashing because the resource doesn't have a name alg. All nameAlg digest sizes will be less then TPMU_HA by design.

I think the code needs to be:
```diff
diff --git a/src/tss2-esys/esys_tr.c b/src/tss2-esys/esys_tr.c
index 33d75fd10ad8..73a891c8bfd6 100644
--- a/src/tss2-esys/esys_tr.c
+++ b/src/tss2-esys/esys_tr.c
@@ -391,9 +391,6 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
     if (authValue == NULL) {
         esys_object->auth.size = 0;
     } else {
-        if (authValue->size > sizeof(TPMU_HA)) {
-            return_error(TSS2_ESYS_RC_BAD_SIZE, "Bad size for auth value.");
-        }
         /* Determine name alg of resource */
         if (esys_object->rsrc.rsrcType == IESYSC_KEY_RSRC) {
             name_alg = esys_object->rsrc.misc.rsrc_key_pub.publicArea.nameAlg;
@@ -402,11 +399,15 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
         } else {
             name_alg = TPM2_ALG_NULL;
         }
-        esys_object->auth = *authValue;
         /* Adapt auth value to hash for large auth values. */
         if (name_alg != TPM2_ALG_NULL) {
             r = iesys_hash_long_auth_values(&esys_object->auth, name_alg);
             return_if_error(r, "Adaption of out value failed.");
+        } else {
+            if (authValue->size > sizeof(TPMU_HA)) {
+                return_error(TSS2_ESYS_RC_BAD_SIZE, "Bad size for auth value.");
+            }
+            esys_object->auth = *authValue;
         }
     }
     return TSS2_RC_SUCCESS;
```

test/integration/esys-create-session-auth.int.c
williamcroberts: Yep you should try 1 over the HashAlg size and 1 over TPMU_HA size to trip that other condition for Esys_TR_SetAuth.

williamcroberts:
FYI the guidance I got from the TPM WG on:
- _Start() routines will truncate the authValue based on the value of property TPM2_PT_MAX_DIGEST.
- HiearchyChangeAuth will digest using the digest associated with property TPM2_PT_CONTEXT_HASH.

src/tss2-esys/esys_tr.c
```diff
@@ -392,7 +394,20 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
...
+        } else if (esys_object->rsrc.rsrcType == IESYSC_NV_RSRC) {
+            name_alg = esys_object->rsrc.misc.rsrc_nv_pub.nvPublic.nameAlg;
+        } else {
+            name_alg = TPM2_ALG_NULL;
+        }
         esys_object->auth = *authValue;
+        /* Adapt auth value to hash for large auth values. */
+        if (name_alg != TPM2_ALG_NULL) {
+            r = iesys_hash_long_auth_values(&esys_object->auth, name_alg);
+            return_if_error(r, "Adaption of out value failed.");
```
williamcroberts, line 409: is that really what you wanted for the message, or was it a copy paste error?

src/tss2-esys/esys_tr.c
```diff
@@ -392,7 +394,20 @@ Esys_TR_SetAuth(ESYS_CONTEXT * esys_context, ESYS_TR esys_handle,
         if (authValue->size > sizeof(TPMU_HA)) {
             return_error(TSS2_ESYS_RC_BAD_SIZE, "Bad size for auth value.");
         }
+        /* Determine name alg of resource */
+        if (esys_object->rsrc.rsrcType == IESYSC_KEY_RSRC) {
+            name_alg = esys_object->rsrc.misc.rsrc_key_pub.publicArea.nameAlg;
+        } else if (esys_object->rsrc.rsrcType == IESYSC_NV_RSRC) {
+            name_alg = esys_object->rsrc.misc.rsrc_nv_pub.nvPublic.nameAlg;
+        } else {
+            name_alg = TPM2_ALG_NULL;
```
williamcroberts, line 403: just initialize name_alg to TPM2_ALG_NULL and this else clause goes away.

src/tss2-esys/esys_iutil.c
```diff
@@ -1554,3 +1554,91 @@ iesys_tpm_error(TSS2_RC r)
...
+ * @retval TSS2_RC_SUCCESS if the function call was a success.
+ * @retval TSS2_ESYS_RC_MEMORY if the ESAPI cannot allocate enough memory.
+ * @retval TSS2_ESYS_RC_BAD_VALUE if an invalid hash is passed.
+ * @retval TSS2_ESYS_RC_GENERAL_FAILURE for a failure during digest
+ *         computation.
+ */
+TSS2_RC
+iesys_adapt_sensitive_to_long_auth_values(
+    TPM2B_SENSITIVE_CREATE *in_sensitive,
+    TPMI_ALG_HASH name_hash_alg)
```
williamcroberts, line 1629: You don't actually need this function. This actually duplicates checks and functionality of iesys_hash_long_auth_values.  iesys_hash_long_auth_values will check that the authValue is greater than the hash_size and replace it, or leave it alone.

williamcroberts (changes requested):

src/tss2-esys/esys_tr.c
williamcroberts: Disregard this comment, duh moment, we're bounded by the incoming TPM2B_AUTH, whic is TPMU_HA sized buffer anyways.

test/integration/esys-create-session-auth.int.c
williamcroberts: Disregard again.

williamcroberts:
For some reason I got it stuck in my head that incoming AuthValues can be larger than TPMU_HA, but that's not the case, only the size field can be as TPM2B_AUTH buffer is the sizeof TPMU_HA.

But I have some other comments, and I made the changes I suggested in this patch if you want it and squish it into your patch:
https://github.com/williamcroberts/TPM2.0-TSS/commit/63d105ec75e6daafb8dad6b2c92e5366e730f06f

JuergenReppSIT:
@williamcroberts I have squashed your code comment commit and added you to signed-off.
Thank you for the review.

williamcroberts:
LGTM

williamcroberts (approved):

PR #1931, opened 2020-12-03
TCTIs: Add missing events bits to pollhandles

AndreasFuchsSIT:
TCTIs are supposed to populate the pollevent structure such that it can be used with the `poll()` syscall.
However many did not set the events field, leading to infinite waits in poll.

I'm not entirely sure on the bits each of the tctis needs.

@williamcroberts what are the poll-events for tcti-cmd for example ?

### DIFF
--- a/src/tss2-tcti/tcti-cmd.c
+++ b/src/tss2-tcti/tcti-cmd.c
@@ -442,6 +442,7 @@ TSS2_RC tcti_cmd_get_poll_handles (TSS2_TCTI_CONTEXT *tctiContext,
             LOG_ERROR ("Could not get fileno: %s", strerror (errno));
             return TSS2_TCTI_RC_IO_ERROR;
         }
+        handles->events = POLLIN | POLLOUT;
     }
 
     return TSS2_RC_SUCCESS;
--- a/src/tss2-tcti/tcti-mssim.c
+++ b/src/tss2-tcti/tcti-mssim.c
@@ -288,6 +288,7 @@ tcti_mssim_get_poll_handles (
 #else
         handles->fd = tcti_mssim->tpm_sock;
 #endif
+        handles->events = POLLIN | POLLOUT;
     }
 
     return TSS2_RC_SUCCESS;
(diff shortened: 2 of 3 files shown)


src/tss2-tcti/tcti-cmd.c
```diff
@@ -442,6 +442,7 @@ TSS2_RC tcti_cmd_get_poll_handles (TSS2_TCTI_CONTEXT *tctiContext,
             LOG_ERROR ("Could not get fileno: %s", strerror (errno));
             return TSS2_TCTI_RC_IO_ERROR;
         }
+        handles->events = POLLIN | POLLOUT | POLLHUP;
```
williamcroberts, line 445: POLLHUP is only returned via revents in the struct pollfd. So you can drop that. POLLOUT might not give you a non-blocking result. If you write larger than available you'll block unless O_NONBLOCK is set in open(2). Definitely need POLLIN.

williamcroberts (changes requested):

src/tss2-tcti/tcti-mssim.c
```diff
@@ -288,6 +288,7 @@ tcti_mssim_get_poll_handles (
 #else
         handles->fd = tcti_mssim->tpm_sock;
 #endif
+        handles->events = POLLIN | POLLOUT | POLLHUP;
```
williamcroberts, line 291: No POLLHUP

williamcroberts (changes requested):

williamcroberts:
> TCTIs are supposed to populate the pollevent structure such that it can be used with the `poll()` syscall.
> However many did not set the events field, leading to infinite waits in poll.
> 
> I'm not entirely sure on the bits each of the tctis needs.
> 
> @williamcroberts what are the poll-events for tcti-cmd for example ?

You should just need POLLIN | POLLOUT since POLLHUB is a return value and ignored if set in events.

tcti-cmd only returns the fd mapped to the child process's stdout, so we really only need POLLIN.

We could return another fd for `cmd_tcti->sink` and have that event set for POLLOUT.

PR #1932, opened 2020-12-04
FAPI: Fix compilation error for integration tests in debugging mode.

JuergenReppSIT:
If the integration tests were compiled with --enable-debug in one integration test
undeclared variables were used.


AndreasFuchsSIT (approved):
Backport ?

JuergenReppSIT:
The test was not backported so no backport is needed.

PR #1934, opened 2020-12-09
FAPI: Fix determination of object type from path.

JuergenReppSIT:
The type of the path "/nv/" was not determined correctly as an NV path.
It did work e.g. for "/nv" and "/nv/..." but not for "/nv/".
So Fapi_List could not be called with path "/nv/".


williamcroberts (approved):

AndreasFuchsSIT:
@JuergenReppSIT Backport ?

PR #1941, opened 2020-12-11
FAPI: Improve integration tests.

JuergenReppSIT:
* The asserts which are activated by --enable-debug are replaced with ASSERT, which
  is always active, to ensure that for FAPI integration tests the asserts are
  always fulfilled.

* For JSON encoded results, if possible, the complete JSON value will be checked.
  If parts of the JSON object are variable only selected sub fields will be
  checked.

* The result of Fapi_List will be compared with an expected list of objects for
  typical use cases.

### DIFF
--- a/test/integration/main-fapi.c
+++ b/test/integration/main-fapi.c
@@ -13,6 +13,8 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <json-c/json_util.h>
+#include <json-c/json_tokener.h>
 
 #include "tss2_esys.h"
 #include "tss2_fapi.h"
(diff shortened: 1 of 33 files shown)


test/integration/main-fapi.c
```diff
@@ -43,6 +45,124 @@ bool file_exists (char *path) {
   return (stat (path, &buffer) == 0);
 }
 
+/* Determine integer number from json object. */
+static int64_t
+get_number(json_object *jso)
+{
```
cplappert, line 51: "{" should be one line above

test/integration/test-fapi.h
```diff
@@ -5,13 +5,127 @@
...
+             ASSERT(jso2); \
+             if (cmp_jso(jso1, jso2)) { \
+                break; \
+             } \
+             json_object_put(jso2); \
+        } \
+        if (i >=  n) { \
+            json_object_put(jso1);       \
+            json_object_put(jso2);          \
+            LOG_ERROR("Mismatch" ); \
```
cplappert, line 122: Unnecessary Whitespace

cplappert (changes requested):
I found a few minor things. Mostly formatting issues.
Otherwise, LGTM.

test/integration/test-fapi.h
JuergenReppSIT: @cplappert Thank you for the review.  I just pushed the changes.

cplappert (approved):

PR #1943, opened 2020-12-17
include/tss2_fapi.h: add stddef and stdint h files

williamcroberts:
Add missing stdint.h and stddef.h header files for things like uint32_t
and size_t respectively.

### DIFF
--- a/include/tss2/tss2_fapi.h
+++ b/include/tss2/tss2_fapi.h
@@ -6,6 +6,9 @@
 #ifndef TSS2_FAPI_H
 #define TSS2_FAPI_H
 
+#include <stddef.h>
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
Shouldn't these come from tss2_common.h / tss2_tcti.h ?
Or do you think we need to make them explicit here ?

williamcroberts:
> tss2_tcti.h

explicit since our uses are not tied to that API. Plus if you look at tcti header file, it includes both even though it includes the common header file which has stdint.  So if someone were to change anything in those included header files, like no longer use stdint.h, it would break the FAPI build. So for the sake of consistency and robustness yes.

PR #1944, opened 2020-12-18
configure.ac: fix compatibility with autoconf 2.70

phmccarty:
With autoconf 2.70, not quoting the second argument to one of the AS_IF
macro expansions leads to generation of invalid shell code affecting the
first nested ERROR_IF_NO_PROG expansion.

The invalid shell code leads to an error resembling:
```
  ./configure: line 18826: syntax error near unexpected token `newline'
  ./configure: line 18826: `    '''
```
Fix the issue by quoting the second argument to the affected AS_IF,
similar to the quoting found elsewhere in configure.ac.

Signed-off-by: Patrick McCarty <patrick.mccarty@intel.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -285,7 +285,7 @@ AC_ARG_ENABLE([integration],
         [build and execute integration tests])],,
     [enable_integration=no])
 AS_IF([test "x$enable_integration" = "xyes"],
-     AS_IF([test "$HOSTOS" = "Linux"],
+     [AS_IF([test "$HOSTOS" = "Linux"],
            [ERROR_IF_NO_PROG([ss])],
            [ERROR_IF_NO_PROG([sockstat])])
        ERROR_IF_NO_PROG([echo])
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
Sounds good.
Should we backport this to 3.0.x stable ?
Or even to 2.4.x major-stable ? Probably not needed ?
I'm asking you since you're the first to come here with 2.70

williamcroberts (approved):

phmccarty:
> Sounds good.
> Should we backport this to 3.0.x stable ?
> Or even to 2.4.x major-stable ? Probably not needed ?
> I'm asking you since you're the first to come here with 2.70

I am currently using 3.0.3 stable, and the same fix applies to that version. I also tried cherry-picking it to the 3.0.x branch, and it applies cleanly.

It's not needed for latest 2.4.x, because the second argument to the same AS_IF macro is properly quoted.

PR #1949, opened 2021-01-08
configure.ac: quote AS_IF arguments to avoid errors with Autoconf 2.70

diabonas:
Autoconf 2.70 [has been released](https://lists.gnu.org/archive/html/autoconf/2020-12/msg00002.html) and does not like it when `AS_IF` arguments are not quoted using square brackets, producing wrong macro expansions leading to errors like
```
checking for CURL... yes
./configure: line 18882: syntax error near unexpected token `newline'
./configure: line 18882: `    '''
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -120,9 +120,9 @@ AC_ARG_ENABLE([fapi-async-tests],
                [Force fapi to spin asynchronously. (NOT FOR PRODUCTION!)]),,
     [enable_fapi_async_tests=no])
 AS_IF([test "x$enable_fapi_async_tests" = "xyes"],
-    AC_DEFINE([TEST_FAPI_ASYNC], [1], [FAPI forced async spinning]))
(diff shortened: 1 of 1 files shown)


diabonas:
Whoops, I didn't see #1944 beforehand, sorry... I'll note that my PR is a bit more liberal regarding the quoting - not all of these are strictly needed at the moment, but perhaps they make the Autoconf script a bit more future-proof. Quoting from the [Autoconf 2.70 release announcement](https://lists.gnu.org/archive/html/autoconf/2020-12/msg00002.html):

> *** Many macros have become pickier about argument quotation.
>
>  If you get a shell syntax error from your generated configure script, or seemingly impossible misbehavior (e.g. entire blocks of the configure script not getting executed), check first that all macro arguments are properly quoted. The"Quotation" section of the manual explains how to quote macro arguments properly.
>
> It is unfortunately not possible for autoupdate to correct quotation errors.

PR #1953, opened 2021-01-13
tss2-fapi: add GlobalSign TPM Root CA certs

rvka:
Unlike other certificates in fapi_certificates.h, STM root certificates
are not self-signed.  Rather, they are signed by the respective (RSA or
ECC) GlobalSign TPM Root CA certificates.

As a result, the certificate chain built for STM devices does not
terminate in a self-signed certificate, and openssl's X509_verify_cert
refuses to verify it, so provisioning fails with ifapi_verify_ek_cert()
reporting "unable to get issuer certificate".

Arguably it's a bug in OpenSSL -- to try to search for issuer's
certificate in the chain prior to checking the current certificate's
trust status.  However, it appears easier to work it around here.

So add the proper root CAs to the trusted certificate store.  At the
time of this writing, the certificates are available (in DER format) at
https://secure.globalsign.com/cacert/gstpmroot.crt and
https://secure.globalsign.com/cacert/tpmeccroot.crt, respectively.

### DIFF
--- a/src/tss2-fapi/fapi_certificates.h
+++ b/src/tss2-fapi/fapi_certificates.h
@@ -77,7 +77,32 @@ static char * root_cert_list[] = {
     "VILmTjHwSL7uZBU=\n"
     "-----END CERTIFICATE-----",
 
-    /* STM RSA root certificate */
+    /* GlobalSign TPM RSA Root CA */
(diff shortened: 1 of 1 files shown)


williamcroberts (approved):

rvka:
The commit also applies to the stable branches.

Issue #1959, opened 2021-01-30
segmentation fault in Fapi_Finalize when log_dir is missing from FAPI config


fishilico:
When `"log_dir"` is missing from the FAPI configuration file, all `tss2_...` tools produce a segmentation fault in function `Fapi_Finalize`.

This is because with such a configuration, `out->log_dir = DEFAULT_LOG_DIR`: https://github.com/tpm2-software/tpm2-tss/blob/15b9b18c59cf01165032285f99d8216d6b925227/src/tss2-fapi/ifapi_config.c#L73-L78

and `DEFAULT_LOG_DIR` is a read-only string: https://github.com/tpm2-software/tpm2-tss/blob/15b9b18c59cf01165032285f99d8216d6b925227/src/tss2-fapi/fapi_int.h#L37

So `SAFE_FREE((*context)->config.log_dir);` tries to free a read-only string and fails: https://github.com/tpm2-software/tpm2-tss/blob/15b9b18c59cf01165032285f99d8216d6b925227/src/tss2-fapi/api/Fapi_Finalize.c#L81

I hit this bug while trying to use a software TPM in an Arch Linux container (`podman run --rm -ti docker.io/library/archlinux`, using tpm2-tss 3.0.3 and tpm2-tools 5.0) by running:
```sh
pacman -Syu --noconfirm swtpm tpm2-abrmd tpm2-tools
swtpm socket --tpm2 --server port=2321 --daemon --ctrl type=tcp,port=2322 \
    --flags not-need-init --tpmstate dir=/tmp --log file=/tmp/swtpm.log,level=5
mkdir -p /run/dbus && dbus-daemon --system --fork
tpm2-abrmd --allow-root --tcti swtpm:host=127.0.0.1,port=2321 &
export TPM2TOOLS_TCTI=tabrmd:bus_type=system

# Create a configuration without "log_dir"
echo > /etc/tpm2-tss/swtpm_fapi_config.json \
    '{"profile_name": "P_ECCP256SHA256",' \
    '"profile_dir": "/etc/tpm2-tss/fapi-profiles/",' \
    '"user_dir": "~/.local/share/tpm2-tss/user/keystore",' \
    '"system_dir": "/var/lib/tpm2-tss/system/keystore",' \
    '"tcti": "tabrmd:bus_type=system",' \
    '"system_pcrs": [],' \
    '"ek_cert_less": "yes"}'
[...]

PR #1960, opened 2021-02-01
FAPI: Fix initialization of default log_dir.

JuergenReppSIT:
When "log_dir" was missing from the FAPI configuration a segmentation fault was produced
 in Fapi_Finalize because free was called for a string constant.
 Fixes issue #1959.


williamcroberts (approved):
LGTM

PR #1962, opened 2021-02-05
code review: Why ABI version number was defined in 2 different places?

liuqun:
It should be fine to use the `TSS2_ABI_VERSION_CURRENT` defined in tss2_common.h


williamcroberts (approved):
LGTM

PR #1967, opened 2021-02-10
FAPI: initialise 'out' parameter in ifapi_json_IFAPI_CONFIG_deserialize

diabonas:
`out` needs to be fully initialised from the beginning because the `ifapi_json_IFAPI_CONFIG_deserialize` function [uses `return_if_error`](https://github.com/tpm2-software/tpm2-tss/blob/b211887403f646a7d09be1059a94ed943955c9b7/src/tss2-fapi/ifapi_config.c#L56), potentially leaving the struct only partly initialised when there is an error during JSON deserialisation. This causes a problem as `ifapi_config_initialize_finish` [tries to free all struct members in case of an error](https://github.com/tpm2-software/tpm2-tss/blob/b211887403f646a7d09be1059a94ed943955c9b7/src/tss2-fapi/ifapi_config.c#L260-L268). Attempting to free an uninitialised pointer can lead to a segfault crash.

This can be observed when running the [`fapi-config` unit test](https://github.com/tpm2-software/tpm2-tss/blob/master/test/unit/fapi-config.c) inside a systemd-nspawn container on Arch Linux, where the [`check_config_json_remove_field_not_allowed`](https://github.com/tpm2-software/tpm2-tss/blob/b211887403f646a7d09be1059a94ed943955c9b7/test/unit/fapi-config.c#L135) test crashes, leading to a test failure.

### DIFF
--- a/src/tss2-fapi/ifapi_config.c
+++ b/src/tss2-fapi/ifapi_config.c
@@ -44,28 +44,24 @@ ifapi_json_IFAPI_CONFIG_deserialize(json_object *jso, IFAPI_CONFIG *out)
     return_if_null(out, "out is NULL", TSS2_FAPI_RC_BAD_REFERENCE);
     return_if_null(jso, "jso is NULL", TSS2_FAPI_RC_BAD_REFERENCE);
 
+    memset(out, 0, sizeof(IFAPI_CONFIG));
+
     /* Deserialize the JSON object) */
     json_object *jso2;
     TSS2_RC r;
     LOG_TRACE("call");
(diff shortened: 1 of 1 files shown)


JuergenReppSIT (changes requested):
out->log_dir should still be initialized if the "log_dir"  field is not found in the json object.

diabonas:
> out->log_dir should still be initialized if the "log_dir" field is not found in the json object.

Oops, yeah, I didn't mean to delete that part, fixed now.

PR #1968, opened 2021-02-11
ci/cirrus: drop IGNORE_OSVERSION and set to 12-2

williamcroberts:
Rather than ignoring an error, set to the newest stable FreeBSD cirrus
version of 12-2. This seems to avoid the package/kernel mismatch errors.


JuergenReppSIT (approved):

JuergenReppSIT:
@williamcroberts Thank you for the improvement.

PR #1969, opened 2021-02-11
ci: run tpm2-tools test suite

williamcroberts:
To ensure changes in the tpm2-tss don't break dependent projects, run
the tpm2-tools test suite as part of the CI. This also builds tpm2-abrmd
since the tpm2-tools test suite is dependent on that project.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -83,8 +83,6 @@ else
   make -j check
 fi
 
-popd
-
 else # TEST_TCTI_CONFIG == true
 mkdir ./config_test
 pushd ./config_test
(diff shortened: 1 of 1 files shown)


williamcroberts:
It appears that python is not installed on Fedora-32 image, just python3. It should be compatible for our purposes, so Ill need to symlink it or do alternatives. Also, no option --enable-integration for tpm2-tools.

williamcroberts:
Actually I should probably fix this so it works on Fedora-32 without Python 2.X installed, since it appears that python might not always be available if a python is installed and theirs some inconsistency over that executable name. Ill fix the tools up and respin this when ready.

williamcroberts:
I started the tools fixes here: https://github.com/tpm2-software/tpm2-tools/pull/2332

tstruk:
Do we know how much this will increase the CI run time?

williamcroberts:
> Do we know how much this will increase the CI run time?

No, but I would like to do is add it to a specific included run rather than all the test runs. So this way it's more isolated.

PR #1972, opened 2021-02-12
Update ci 2.4.x

williamcroberts:


AndreasFuchsSIT:
So test 1 fail is #1975
Test 2 fail is `configure: error: Integration tests can not be enabled without the TCTI_MSSIM module`
Test 3 fail is `configure: error: Bad value for --with-crypto mbed`

@williamcroberts Seems like pretty easy fixes. Will you prepend them to the CI-Change commits ?

.ci/travis.run
```diff
@@ -2,71 +2,40 @@
```
tstruk, line 2: This file should be changed to "ci.run" or something

configure.ac
```diff
@@ -258,6 +258,7 @@ AS_IF([test "x$enable_integration" = "xyes" -a "$HOSTOS" = "Linux"],
        ERROR_IF_NO_PROG([grep])
        ERROR_IF_NO_PROG([env])
        ERROR_IF_NO_PROG([rm])
+       ERROR_IF_NO_PROG([sha256sum])
```
tstruk, line 261: For compatibility reasons we use "shasum -a 256". This is on master. Looks like on 2.4.x we have a mix.

tstruk (changes requested):
Some minor comments. @williamcroberts Let me know if you want to fix that and push or I can merge this and fix it myself.

.ci/travis.run
williamcroberts: It should, I totally agree, but right now on master its called travis.run. I would take it as it is now, and then we can fixup all the names, across all the projects for uniformity next.

configure.ac
williamcroberts: This is on 2.4.X, so it's a mix right, this isn't necessarily wrong? Its just we should probably kill all sha256sum and swap them with shasum -a sha256? I would imagine that would be a follow up patch.

tstruk: > This is on 2.4.X, so it's a mix right, this isn't necessarily wrong? Its just we should probably kill all sha256sum and swap them with shasum -a sha256? I would imagine that would be a follow up patch.

Ok, I will take it as is and work from there. Thanks

williamcroberts:
> So test 1 fail is #1975
> Test 2 fail is `configure: error: Integration tests can not be enabled without the TCTI_MSSIM module`
> Test 3 fail is `configure: error: Bad value for --with-crypto mbed`
> 
> @williamcroberts Seems like pretty easy fixes. Will you prepend them to the CI-Change commits ?

@tstruk I see Andreas's mentions of things currently blocking as priority to get the CI working.

For item 1, looks like it was already backported as df0ce3ea2608e
For items 2 I think it's just the tests aren't in line with what was there.
For  3,  just drop the test from the patch in the .github folder

williamcroberts:
> > So test 1 fail is #1975
> > Test 2 fail is `configure: error: Integration tests can not be enabled without the TCTI_MSSIM module`
> > Test 3 fail is `configure: error: Bad value for --with-crypto mbed`
> > @williamcroberts Seems like pretty easy fixes. Will you prepend them to the CI-Change commits ?
> 
> @tstruk I see Andreas's mentions of things currently blocking as priority to get the CI working.
> 
> For item 1, looks like it was already backported as [df0ce3e](https://github.com/tpm2-software/tpm2-tss/commit/df0ce3ea2608e21bb9b6af42830b2aaa1a1f6aa9)
> For items 2 I think it's just the tests aren't in line with what was there.
> For 3, just drop the test from the patch in the .github folder

Looks like for item 2, we can just drop that test as it was never performed on the 2.4.x branch.

williamcroberts:
@tstruk, some checks were failing, I just force pushed that branch after you merged it.

williamcroberts:
You can diff that for any hunks you might want to pull in

PR #1977, opened 2021-02-19
marshal: remove double define

joholl:
Remove double define from API header `tss2/tss2_mu.h`.

They are also defined here:
https://github.com/tpm2-software/tpm2-tss/blob/27d3bd8295a58d5788cf4ea220f614603d5c2fee/include/tss2/tss2_mu.h#L1998-L2010

### DIFF
--- a/include/tss2/tss2_mu.h
+++ b/include/tss2/tss2_mu.h
@@ -21,20 +21,6 @@ extern "C" {
 #endif
 
 TSS2_RC
-Tss2_MU_BYTE_Marshal(
-    BYTE           src,
-    uint8_t         buffer[],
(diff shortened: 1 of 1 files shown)


joholl:
I don't see , how the CI error is caused by my patch: `Fapi_Sign.c:130:Fapi_Sign() ErrorCode (0x00000990)`. Maybe this is a CI problem?

JuergenReppSIT (approved):

JuergenReppSIT:
@joholl Yes it seems to be a CI problem. I tried to reproduce the error with FreeBSD in a VM on my pc  to debug it. But the error did not occur.

PR #1990, opened 2021-02-25
headers: fix typedef/struct declarations

williamcroberts:
Currently, the struct/union typedef definitions are in the form of:
typedef struct {
   int thing1;
   int thing2;
} TYPE_NAME;

This has the side affect of causing an anonomous structure definition,
which causes code that uses the 'struct' keyword to fail with errors
like:
  - expected ‘TPM2B_DIGEST **’ {aka ‘struct <anonymous> **’} but argument is of type ‘struct TPM2B_DIGEST **

To correct this, make them non-anonymous by declaring the typedef fully
and then defining the structure.

Further reading:
  - https://stackoverflow.com/questions/32999739/gcc-gives-confusing-warning-message-when-using-pointer-to-struct-typedef
  - https://stackoverflow.com/questions/17720223/c-typedef-struct-name-vs-typedef-struct-name/23660072

This will take updates to the TCG TSS spec as well.

### DIFF
--- a/include/tss2/tss2_common.h
+++ b/include/tss2/tss2_common.h
@@ -29,12 +29,13 @@ typedef int64_t     INT64;
 /*
  * ABI runtime negotiation definitions
  */
-typedef struct {
+typedef struct TSS2_ABI_VERSION TSS2_ABI_VERSION;
+struct TSS2_ABI_VERSION {
     uint32_t tssCreator;
     uint32_t tssFamily;
     uint32_t tssLevel;
     uint32_t tssVersion;
-} TSS2_ABI_VERSION;
+};
 
 #define TSS2_ABI_VERSION_CURRENT {1, 2, 1, 108}
 
--- a/include/tss2/tss2_sys.h
+++ b/include/tss2/tss2_sys.h
@@ -28,15 +28,17 @@ typedef struct _TSS2_SYS_OPAQUE_CONTEXT_BLOB TSS2_SYS_CONTEXT;
 #define TSS2_SYS_MAX_SESSIONS 3
 
 /* Input structure for authorization area(s). */
-typedef struct {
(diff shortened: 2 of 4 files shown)


AndreasFuchsSIT:
Why not esys.h and fapi.h ?

williamcroberts:
> Why not esys.h and fapi.h ?

The two structures are by design Opaque:
```bash
67:typedef struct ESYS_CONTEXT ESYS_CONTEXT;
20:typedef struct FAPI_CONTEXT FAPI_CONTEXT;
```
Thus don't suffer this issue.

The FAPI_POLL_HANDLE is just a typedef to pollfd, so that doesn't need to change either.

AndreasFuchsSIT:
> > Why not esys.h and fapi.h ?
> 
> The two structures are by design Opaque:
> 
> ```shell
> 67:typedef struct ESYS_CONTEXT ESYS_CONTEXT;
> 20:typedef struct FAPI_CONTEXT FAPI_CONTEXT;
> ```
> 
> Thus don't suffer this issue.
> 
> The FAPI_POLL_HANDLE is just a typedef to pollfd, so that doesn't need to change either.

So true, I'm stupid...

williamcroberts:
> > > Why not esys.h and fapi.h ?
> > 
> > 
> > The two structures are by design Opaque:
> > ```shell
> > 67:typedef struct ESYS_CONTEXT ESYS_CONTEXT;
> > 20:typedef struct FAPI_CONTEXT FAPI_CONTEXT;
> > ```
> > 
> > 
> > Thus don't suffer this issue.
> > The FAPI_POLL_HANDLE is just a typedef to pollfd, so that doesn't need to change either.
> 
> So true, I'm stupid...

I was like good question, so I had to go look and justify it, so I am right there with you :-p

PR #1999, opened 2021-03-01
ESYS: Fix binding of ESYS_TR_RH_NULL 2.4.x

### LINKED ISSUE #1993: The Null seed fails as a parent for any session based operation
Trying to do any session based operation on something with the NULL seed as a parent results in authentication errors.  This is a trace from trying to do TPM2_CreatePrimary with the NULL seed as a primaryHandle:

WARNING:esys:src/tss2-esys/api/Esys_CreatePrimary.c:393:Esys_CreatePrimary_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:135:Esys_CreatePrimary() Esys Finish ErrorCode (0x000009a2) 
TSS_CreatePrimary failed with 2466
tpm:session(1):authorization failure without DA implications

The session that's failing was created as a bound session on the NULL permanent handle, so it's not like this has security implications, but it's something that should work.

If you do the same operation with TPM_RS_PW instead of a real authentication session, everything works, so I think something is wrong in the HMAC routines when the NULL seed is specified either for binding or as a parent.

JuergenReppSIT:
If ESYS_TR_RH_NULL was used as bind parameter for Esys_StartAuthSession the wrong HMAC
was computed, because the session was not treated as unbound session.
It will be checked whether no tpm key is passed and ESYS_TR_RH_NULL is used as bind
parameter to skip KDF computation of the session key.
Fixes #1993.

### DIFF
--- a/src/tss2-esys/api/Esys_StartAuthSession.c
+++ b/src/tss2-esys/api/Esys_StartAuthSession.c
@@ -445,45 +445,47 @@ Esys_StartAuthSession_Finish(
             LOG_ERROR("Error: initialize auth session (%x).", r);
             return r;
         }
-        /* compute session key */
-        size_t secret_size = 0;
-        if (tpmKey != ESYS_TR_NONE)
-            secret_size += keyHash_size;
-        if (bind != ESYS_TR_NONE && bindNode != NULL)
-            secret_size += bindNode->auth.size;
-        /*
-         * A non null pointer for secret is required by the subsequent functions,
-         * hence a malloc is called with size 1 if secret_size is zero.
-         */
-        uint8_t *secret = malloc(secret_size ? secret_size : 1);
-        if (secret == NULL) {
-            LOG_ERROR("Out of memory.");
-            return TSS2_ESYS_RC_MEMORY;
-         }
-        if  (bind != ESYS_TR_NONE && bindNode != NULL
-             && bindNode->auth.size > 0)
-             memcpy(&secret[0], &bindNode->auth.buffer[0], bindNode->auth.size);
-        if (tpmKey != ESYS_TR_NONE)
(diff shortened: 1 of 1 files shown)


lp9921 (commented):
src/tss2-esys/api/Esys_StartAuthSession.c

src/tss2-esys/api/Esys_StartAuthSession.c
```diff
@@ -445,45 +445,47 @@ Esys_StartAuthSession_Finish(
             LOG_ERROR("Error: initialize auth session (%x).", r);
```
lp9921, line 445: src/tss2-esys/api/Esys_StartAuthSession.c

lp9921 (commented):
https://raw.githubusercontent.com/tpm2-software/tpm2-tss/df0ce3ea2608e21bb9b6af42830b2aaa1a1f6aa9/doc/doxygen.dox

lp9921:
Note | Remote repositories can be on your local machine.It is entirely possible that you can be working with a “remote” repository that is, in fact, on the same host you are. The word “remote” does not necessarily imply that the repository is somewhere else on the network or Internet, only that it is elsewhere. Working with such a remote repository would still involve all the standard pushing, pulling and fetching operations as with any other remote.
-- | --

lp9921:
Note | Remote repositories can be on your local machine.It is entirely possible that you can be working with a “remote” repository that is, in fact, on the same host you are. The word “remote” does not necessarily imply that the repository is somewhere else on the network or Internet, only that it is elsewhere. Working with such a remote repository would still involve all the standard pushing, pulling and fetching operations as with any other remote.
-- | --

williamcroberts:
cirrus is fixed (https://github.com/tpm2-software/tpm2-tss/pull/2003) so you can just rebase and force push.

tstruk (approved):

PR #2000, opened 2021-03-01
FAPI: use FAPI_TEST_EK_CERT_LESS with --disable-self-generated-certificate

diabonas:
Since commit 199b4edc265b2f4758aa22ebf4ed6472a34b9a7a ("FAPI: Fix reading of the root certificate for provisioning.") it is required to specify `--enable-self-generated-certificate` in order to make the FAPI integration tests pass. This is an option that should usually not be enabled in production builds for security reasons, but still some form of integration testing might be desirable in this case to verify whether the compiled library works as expected. Use `FAPI_TEST_EK_CERT_LESS` in this case to run the tests without EK certificate validation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -462,6 +462,8 @@ AS_IF([test "x$enable_self_generated_certificate" = xyes],
 	[AC_DEFINE([SELF_GENERATED_CERTIFICATE], [1], [Allow usage of self generated root certificate])],
 	[AS_IF([test "x$integration_tcti" != "xdevice"], [AC_DEFINE([FAPI_TEST_EK_CERT_LESS], [1], [Perform integration tests without EK certificate verification])])])
 
+AS_IF([test "x$enable_integration" = "xyes" && test "x$enable_self_generated_certificate" != "xyes" && test "x$integration_tcti" != "xdevice"],
+      [AC_MSG_WARN([Running integration tests without EK certificate verification, use --enable-self-generated-certificate for full test coverage])])
 
 AC_SUBST([PATH])
 
--- a/configure.ac
+++ b/configure.ac
@@ -458,8 +458,9 @@ AC_ARG_ENABLE([self-generated-certificate],
             [AS_HELP_STRING([--enable-self-generated-certificate],
                             [Alllow usage of self generated root certifcate])],,
             [enable_self_generated_certificate=no])
-AS_IF([test "x$enable_self_generated_certificate" == xyes],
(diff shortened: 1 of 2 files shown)


JuergenReppSIT:
Wouldn't this change disable EK verification if you run the integration tests with a physical tpm with the standard configure parameters?

diabonas:
Possibly. I have added a test whether `$integration_tcti` is set to `device` before defining `FAPI_TEST_EK_CERT_LESS` and can confirm that `config.h` looks as expected (`./configure --enable-integration` has `FAPI_TEST_EK_CERT_LESS` set, `./configure --enable-integration --with-device=/dev/tpm0` has it unset). However I don't have a way to run the integration tests on a physical TPM, so I cannot verify whether they work as expected at the moment (but at least they should be unaffected by this PR now).

AndreasFuchsSIT:
How about we just issue a `if integration_test && !enable_self_cert then AC_WARNING()` in configure.ac ?

AndreasFuchsSIT:
And @diabonas: Congratz ! You opened Issue/PR Number 2000 ! ;-D

diabonas:
> How about we just issue a `if integration_test && !enable_self_cert then AC_WARNING()` in configure.ac ?

The use case for this change is that I want to be able to run the integration test suite while packaging tpm2-tss for Arch Linux. This is pretty useful to catch e.g. linking errors and has unearthed distribution-specific packaging errors in the past. With the change in commit 199b4edc265b2f4758aa22ebf4ed6472a34b9a7a, the only way to run the integration tests on a TPM simulator is to pass `--enable-self-generated-certificate`, which is undesirable for multiple reasons:

1. Passing this configure option for production builds is a possible security issue, so not an option.
2. Building with `--enable-self-generated-certificate` first to be able to run the integration tests, then building the production assets without this options means we need to build everything twice and don't run the tests on the actual libraries that we ship in the end.
3. The way Arch Linux packaging generally works is that we build everything once (in the `build()` function of the [`PKGBUILD` file](https://github.com/archlinux/svntogit-community/blob/packages/tpm2-tss/trunk/PKGBUILD)), then run all tests on the built artefacts (in the `check()` function). Compiling in `check()` should generally be avoided if possible.

diabonas:
Regardless I agree that it's a good idea to add a warning when `--enable-integration` is used without `--enable-self-generated-certificate` because it doesn't give full test coverage, but I'd like to get the tests working without this option as well for the reasons outlined above.

JuergenReppSIT (approved):
@diabonas Thank you for fixing the problem. I did check your PR with a physical TPM and the verification was carried out.

PR #2001, opened 2021-03-02
de-utf8 the source

williamcroberts:
I have three patches, in order of how invasive they are:
1. The external headers
2. The internal source and headers (.c and .h)
3. The rest of it

The primary motivations was item 1. If you have an ascii string and try to append a utf-8 string in python, it blows up if there are non-ascii characters in the utf-8 string. There were a bunch of characters in the headers that didn't do much, a NO BREAK SPACE character (0xC2A0). However, as we move through the rest it changes some minor formatting in the changlog (line character) and it does change the character pronunciation marks for those whom have them in their name.

I'm okay with dropping 2 and 3, but we probably want to keep our external headers UTF8 free to avoid this types of issues. Perhaps I can add a CI check to whatever parts we want to keep UTF8 "clean".


src/tss2-esys/tss2-esys.vcxproj.filters
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+?<?xml version="1.0" encoding="utf-8"?>
```
AndreasFuchsSIT, line 1: Are you sure about this one ?

src/tss2-mu/tss2-mu.vcxproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+?<?xml version="1.0" encoding="utf-8"?>
```
AndreasFuchsSIT, line 1: Same

src/tss2-rc/tss2-rc.vcxproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+?<?xml version="1.0" encoding="utf-8"?>
```
AndreasFuchsSIT, line 1: same

src/tss2-sys/tss2-sys.vcxproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+?<?xml version="1.0" encoding="utf-8"?>
```
AndreasFuchsSIT, line 1: same

src/tss2-tcti/tss2-tcti-mssim.vcxproj.filters
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+?<?xml version="1.0" encoding="utf-8"?>
```
AndreasFuchsSIT, line 1: same

src/tss2-tcti/tss2-tcti-tbs.vcxproj.filters
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+?<?xml version="1.0" encoding="utf-8"?>
```
AndreasFuchsSIT, line 1: same

PeterHuewe:
Except for Andreas' valid comments, I think it is fine and safe to change (1-3).
Also for the umlauts - (having this problem myself) - I think this is okay.

src/tss2-esys/tss2-esys.vcxproj.filters
williamcroberts: Nope, I thought I got all those. When it doesn't know what to do, the tool throws a ? in there.

williamcroberts:
gah, sed PEBUAK

williamcroberts:
Im going to actually drop the MS files, they claim utf8 encoding right at the top, so why mess with them.

tstruk:
@williamcroberts looks good now. Good idea no to change the MS files as they are VS auto-generated and changing them manually may only cause trouble. So you may also drop tpm2-tss.sln from the patch.

williamcroberts:
> @williamcroberts looks good now. Good idea no to change the MS files as they are VS auto-generated and changing them manually may only cause trouble. So you may also drop tpm2-tss.sln from the patch.

Plus the png file, I thought I got all those changes out.

tstruk (approved):

Issue #2002, opened 2021-03-02
Cirrus builds fail randomly


tstruk:
The random build failures seems to be related to IO errno 35: Resource temporarily unavailable error.
https://cirrus-ci.com/task/5455668787281920?command=main#L2013
Handling for retries is done in select on Linux, but on FreeBSD select() returns immediately. See related:
https://github.com/tpm2-software/tpm2-tss/commit/00f4ec151495e7b7c473ae14779f39b3fd1c1a9f 
Need to add read() retry handling for FreeBSD specifically.

PR #2003, opened 2021-03-02
tcti: add read eagain handling for freeBSD

tstruk:
The random build failures on Cirrus CI seems to be related
to IO errno 35: Resource temporarily unavailable error.
https://cirrus-ci.com/task/5455668787281920?command=main#L2013

Handling for retries is done in select() on Linux,
but on FreeBSD select() returns immediately.
Need to add read() retry handling for FreeBSD specifically.

Fixes: #2002

### DIFF
--- a/src/util/io.h
+++ b/src/util/io.h
@@ -33,6 +33,13 @@ typedef SSIZE_T ssize_t;
         __ret = exp; \
     } while (__ret == SOCKET_ERROR && WSAGetLastError() == WSAEINTR); \
     dest = __ret; }
+#elif defined (__FreeBSD__)
+#define TEMP_RETRY(dest, exp) \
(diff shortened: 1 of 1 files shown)


williamcroberts (approved):

tstruk:
> This needs to be back ported to 2.4 branch as well as any stable 3.X branches

I'm not sure about that. Looking at https://github.com/tpm2-software/tpm2-tss/blob/2.4.x/src/tss2-tcti/tcti-mssim.c and https://github.com/tpm2-software/tpm2-tss/blob/3.0.x/src/tss2-tcti/tcti-mssim.c I think it's only really needed in master.

williamcroberts:
Ahh I just assumed, I should have looked.

Issue #2005, opened 2021-03-03
OSS Fuzz build is failing


tstruk:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30179&sort=-opened&can=1&q=proj%3Atpm2-tss
The reason is that when fuzzing is enabled we check if $CC == "clang". That was fine before, but now the OSS fuzz compiler is called CC=/src/aflplusplus/afl-clang-fast.
See https://oss-fuzz-build-logs.storage.googleapis.com/log-97a9190d-8048-4383-97e4-2be2f1b4322c.txt

PR #2006, opened 2021-03-03
configure: fix oss fuzz build

tstruk:
When fuzzing is enabled we enforce clang and check if $CC == "clang".
That was fine before, but now the OSS fuzz compiler is called
/src/aflplusplus/afl-clang-fast.

Fixes: #2005


williamcroberts:
I am glad this was as simple as it was.

tstruk:
> I am glad this was as simple as it was.

Yes, it was easy and the build passed this morning:
https://oss-fuzz-build-logs.storage.googleapis.com/index.html#tpm2-tss

PR #2008, opened 2021-03-04
FAPI: Support ~ in paths.

### LINKED ISSUE #1980: Fapi: support ~ in all paths
Virtually all common shells interpret `~` as the home directory. The Fapi follows this convention, but only for the user directory:

https://github.com/tpm2-software/tpm2-tss/blob/9b42cab879819019b63aef46f8394e587c9b2d8e/src/tss2-fapi/ifapi_keystore.c#L436-L442

and:
https://github.com/tpm2-software/tpm2-tss/blob/9b42cab879819019b63aef46f8394e587c9b2d8e/src/tss2-fapi/ifapi_config.c#L215-L219

Enable the use of `~` for all paths, maybe via a common function for parsing paths.

JuergenReppSIT:
* $HOME and ~ in files defined in the FAPI config file will be interpreted
  as home directory.
  Fixes #1980
* The unit test fapi-config was adapted to check this expansion.
* The unnecessary expansion in ifapi_keystore_initialize was removed.

### DIFF
--- a/src/tss2-fapi/ifapi_config.c
+++ b/src/tss2-fapi/ifapi_config.c
@@ -154,6 +154,53 @@ ifapi_config_initialize_async(IFAPI_IO *io)
 }
 
 /**
+ * Expand user symbol in path.
+ *
+ * "~" and "$HOME" will be replaces by the value of the HOME environment
+ * variable.
+ *
+ * @param[in, out] The path.
+ * @retval TSS2_RC_SUCCESS on success
+ * @retval TSS2_FAPI_RC_BAD_VALUE if path is NULL.
+ * @retval TSS2_FAPI_RC_MEMORY if not enough memory can be allocated.
+ * @retval TSS2_FAPI_RC_BAD_PATH if the home directory can't be determined.
+ */
+
+static TSS2_RC
+expand_home(char **path) {
+    size_t startPos = 0;
+    TSS2_RC r;
+
+    return_if_null(path, "Null passed for path", TSS2_FAPI_RC_BAD_VALUE);
+
+    /* Check whether usage of home directory in pathname */
+    if (strncmp("~", *path, 1) == 0) {
+        startPos = 1;
+    } else if (strncmp("$HOME", *path, 5) == 0) {
+        startPos = 5;
+    }
+
+    /* Replace home abbreviation in path. */
+    char *newPath = NULL;
+    if (startPos != 0) {
+        LOG_DEBUG("Expanding path %s to user's home", *path);
+        char *homeDir = getenv("HOME");
+        return_if_null(homeDir, "Home directory can't be determined.",
+                       TSS2_FAPI_RC_BAD_PATH);
+        if (strncmp(&(*path)[startPos], IFAPI_FILE_DELIM, strlen(IFAPI_FILE_DELIM)) == 0) {
+            startPos += strlen(IFAPI_FILE_DELIM);
+        }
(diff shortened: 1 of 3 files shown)


src/tss2-fapi/ifapi_config.c
```diff
@@ -153,6 +153,49 @@ ifapi_config_initialize_async(IFAPI_IO *io)
...
+        }
+        r = ifapi_asprintf(&newPath, "%s%s%s", homeDir, IFAPI_FILE_DELIM,
+                           &(*path)[startPos]);
+        return_if_error(r, "Out of memory.");
+
+        SAFE_FREE(*path);
+        *path = newPath;
+    }
+    return TSS2_RC_SUCCESS;
+}
```
williamcroberts, line 197: This only expands the first portion, which is likely to cover the majority of use cases. Perhaps something in the future using wordexp.h, but thats a GNUC thing IIUC.

JuergenReppSIT: Tank you for the review. I have fixed your first two points and the CI error. Should I change anything related to your point three?

williamcroberts: More of an FYI, depends on how feature rich you want this to be, but this seems fine.

williamcroberts:
@tstruk looks like cirrus isn't finding tpm_server:
```
configure: WARNING: Executable tpm_server not found in PATH (fallback)
configure: error: No simulator executable found in PATH for testing TCTI.
```

it looks like we can end up in a state where gmake doesn't complete the build (the killed lines seem worrisome to me), but it doesn't specifically say their was a compilation error, which just seems weird.
```
2 warnings generated.
2 warnings generated.
gmake: *** [makefile:88: X509_ECC.o] Killed
gmake: *** Waiting for unfinished jobs....
2 warnings generated.
gmake: *** [makefile:88: AlgorithmCap.o] Killed
2 warnings generated.
cd -
rm -fr $ibmtpm_name $ibmtpm_name.tar.gz
```
Any insight on this? Is this just a re-trigger?

JuergenReppSIT:
@tstruk @williamcroberts  Also sometimes FAPI errors occur with cirrus CI, which do not occur after restart. I tried to reproduce the errors with FreeBSD in a VM on my pc to debug it. But the errors did not occur.

tstruk:
> @tstruk looks like cirrus isn't finding tpm_server:
> Any insight on this? Is this just a re-trigger?

I have seen these from time to time. It's cirrus related issue. Just re-trigger.

tstruk:
> @tstruk @williamcroberts Also sometimes FAPI errors occur with cirrus CI, which do not occur after restart. I tried to reproduce the errors with FreeBSD in a VM on my pc to debug it. But the errors did not occur.

No worries, cirrus looks good now.

tstruk (approved):

src/tss2-fapi/ifapi_config.c
```diff
@@ -153,6 +153,53 @@ ifapi_config_initialize_async(IFAPI_IO *io)
...
+        startPos = 1;
+    } else if (strncmp("$HOME", *path, 5) == 0) {
+        startPos = 5;
+    }
+
+    /* Replace home abbreviation in path. */
+    char *newPath = NULL;
+    if (startPos != 0) {
+        LOG_DEBUG("Expanding path %s to user's home", *path);
+        char *homeDir = getenv("HOME");
```
tstruk, line 187: That will be a problem on windows as it uses HOMEPATH, but tough luck...

PR #2011, opened 2021-03-05
FAPI: Fix leak in keystore load function.

JuergenReppSIT:
If access to the file to be loaded was not possible the computed path was not freed.


JuergenReppSIT:
Fix will be added to another PR.

PR #2012, opened 2021-03-05
FAPI: Fix invalid check of read access rights.

JuergenReppSIT:
* The check could be removed. Errors will be handled by fopen.
* The error messages related to fopen were improved.
* One leak related to file opening was fixed: If access to the file to
  be loaded was not possible the computed path string was not freed.

### DIFF
--- a/src/tss2-fapi/ifapi_io.c
+++ b/src/tss2-fapi/ifapi_io.c
@@ -57,11 +57,6 @@ ifapi_io_read_async(
         return TSS2_FAPI_RC_IO_ERROR;
     }
 
-    /* Check read access. */
-    if (!(statbuf.st_mode & R_OK)) {
-        LOG_ERROR("Can't read \"%s\".", filename);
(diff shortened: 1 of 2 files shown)


src/tss2-fapi/ifapi_io.c
```diff
@@ -57,11 +57,6 @@ ifapi_io_read_async(
         return TSS2_FAPI_RC_IO_ERROR;
     }
 
-    /* Check read access. */
-    if (!(statbuf.st_mode & R_OK)) {
-        LOG_ERROR("Can't read \"%s\".", filename);
-        return TSS2_FAPI_RC_IO_ERROR;
-    }
```
williamcroberts, line 64: much better this eliminates a TOCU issue.

williamcroberts (approved):

williamcroberts:
I'll let @tstruk merge this, as he is release planning at the moment, but I think we would want this on master.

tstruk (approved):

PR #2026, opened 2021-03-12
configure: Add pathname of config file to systemd-tmpfiles

### LINKED ISSUE #2025: RHEL 79. make install fails with Failed to create the FAPI directories
/bin/mkdir -p '/usr/local/lib/udev/rules.d'
 /bin/install -c -m 644 dist/tpm-udev.rules '/usr/local/lib/udev/rules.d'
make  install-data-hook
make[2]: Entering directory `/home/kgold/tpm2-tss'
(id -g tss 2>/dev/null || groupadd -r tss) && (id -u tss 2>/dev/null || useradd -r -g tss tss) || echo "*** WARNING *** Failed to create the tss user and group"
59
59
systemd-tmpfiles --create || echo "*** WARNING *** Failed to create the FAPI directories with the correct permissions"
(    if [ ! -d "/usr/local/var/run/tpm2-tss/eventlog/" ]; then echo "*** WARNING *** Directory /usr/local/var/run/tpm2-tss/eventlog/ could not be created"; fi)
*** WARNING *** Directory /usr/local/var/run/tpm2-tss/eventlog/ could not be created
(    if [ ! -d "/usr/local/var/lib/tpm2-tss/system/keystore/" ]; then echo "*** WARNING *** Directory /usr/local/var/lib/tpm2-tss/system/keystore/ could not be created"; fi)
*** WARNING *** Directory /usr/local/var/lib/tpm2-tss/system/keystore/ could not be created
if [ ! -z "" ]; then \
	mv /usr/local/lib/udev/rules.d/tpm-udev.rules /usr/local/lib/udev/rules.d/tpm-udev.rules; \
fi

JuergenReppSIT:
The name of the file with the configuration data had to be added as parameter
to the commands systemd-tmpfiles and systemd-sysusers.
Fixes #2025.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -638,7 +638,7 @@ define set_fapi_permissions
 endef
 
 define check_dir
-    if [ ! -d "$1" ]; then echo "*** WARNING *** Directory $1 could not be created"; fi
+    if [ ! -d "$1" ]; then echo "WARNING Directory $1 could not be created"; fi
 endef
 
 define check_fapi_dirs
@@ -691,16 +691,21 @@ EXTRA_DIST += dist/tpm-udev.rules
 install-dirs:
 if HOSTOS_LINUX
 if SYSD_SYSUSERS
-	systemd-sysusers || echo "*** WARNING *** Failed to create the tss user and group"
+	@echo "systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf"
+	@systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf || echo "WARNING Failed to create the tss user and group"
 else
-	$(call make_tss_user_and_group) || echo "*** WARNING *** Failed to create the tss user and group"
+	@echo "call make_tss_user_and_group"
+	@$(call make_tss_user_and_group) || echo "WARNING Failed to create the tss user and group"
 endif
 if SYSD_TMPFILES
-	systemd-tmpfiles --create || echo "*** WARNING *** Failed to create the FAPI directories with the correct permissions"
+	@echo "systemd-tmpfiles --create $(DESTDIR)$(sysconfdir)/tmpfiles.d/tpm2-tss-fapi.conf"
+	@systemd-tmpfiles --create $(DESTDIR)$(sysconfdir)/tmpfiles.d/tpm2-tss-fapi.conf|| echo "WARNING Failed to create the FAPI directories with the correct permissions"
 else
-	-$(call make_fapi_dirs) && $(call set_fapi_permissions) || echo "*** WARNING *** Failed to create the FAPI directories with the correct permissions"
+	@echo "(call make_fapi_dirs) && (call set_fapi_permissions)"
+	@-$(call make_fapi_dirs) && $(call set_fapi_permissions) || echo "WARNING Failed to create the FAPI directories with the correct permissions"
 endif
-	$(call check_fapi_dirs)
+	@echo "call check_fapi_dirs"
+	@$(call check_fapi_dirs)
 endif
 
 install-data-hook: install-dirs


JuergenReppSIT:
@tstruk Sorry I did approve your PR related to this problem because the installation on Ubuntu 16.04 worked for me. I did not recognize that the fapi files already existed.

tstruk:
@JuergenReppSIT Thanks for the fix. It was how the systemd helpers were originally invoked:
https://github.com/tpm2-software/tpm2-tss/pull/2019/files#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afL693
so it was wrong since the beginning. We will eventually get it right ;)
Since you are on it. Could you make it a little bit less noisy as requested in https://github.com/tpm2-software/tpm2-tss/issues/2025
Maybe something like:
```
if SYSD_SYSUSERS
	@echo "systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf"
	@systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf || echo "WARNING. Failed to create the tss user and group"
else
	@echo "call make_tss_user_and_group"
	@$(call make_tss_user_and_group) || echo "WARNING. Failed to create the tss user and group"
emdif
```

JuergenReppSIT:
@tstruk I hope it's enough less noisy now ? ;-)

Makefile.am
```diff
@@ -691,14 +691,14 @@ EXTRA_DIST += dist/tpm-udev.rules
 install-dirs:
 if HOSTOS_LINUX
 if SYSD_SYSUSERS
-	systemd-sysusers || echo "*** WARNING *** Failed to create the tss user and group"
+	systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf
```
diabonas, line 694: This is going to make the installation fail if `systemd-sysusers` cannot be successfully executed, e.g. because the installation is not done as root. Please use
```
-systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf
```
or
```
systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf || true
```
to make the installation continue in this case, see #2022.

As a side note, adding the configuration file to `systemd-sysusers` or `systemd-tmpfiles` is not really necessary, if left out, all configuration files in the default paths will be considered, but specifying our own config file shouldn't hurt either.

diabonas (changes requested):

diabonas (approved):

tstruk:
What I wanted to do is to silence it completely with '@' as in:
https://www.gnu.org/software/make/manual/html_node/Echoing.html
so that we don't see these || and so on. Then just add a @echo statement which would print what it's doing:
```
if SYSD_SYSUSERS
	@echo "systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf"
	@systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf || echo "WARNING. Failed to create the tss user and group"
else
	@echo "call make_tss_user_and_group"
	@$(call make_tss_user_and_group) || echo "WARNING. Failed to create the tss user and group"
endif
if SYSD_TMPFILES
....
endif
```
That way we won't see the offending word "WARNING" at all if all works good.
@diabonas will that work for you?

diabonas:
Sure, as long as creating the user or directories doesn't fail the installation, anything is fine by me :) I'm not particularly attached to the warning messages either, I just thought it might be nice to let the users know there is something they might need to do manually if this part of the installation didn't work.

williamcroberts:
> What I wanted to do is to silence it completely with '@' as in:
> https://www.gnu.org/software/make/manual/html_node/Echoing.html
> so that we don't see these || and so on. Then just add a @echo statement which would print what it's doing:
> 
> ```
> if SYSD_SYSUSERS
> 	@echo "systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf"
> 	@systemd-sysusers $(DESTDIR)$(sysconfdir)/sysusers.d/tpm2-tss.conf || echo "WARNING. Failed to create the tss user and group"
> else
> 	@echo "call make_tss_user_and_group"
> 	@$(call make_tss_user_and_group) || echo "WARNING. Failed to create the tss user and group"
> endif
> if SYSD_TMPFILES
> ....
> endif
> ```
> 
> That way we won't see the offending word "WARNING" at all if all works good.
> @diabonas will that work for you?

The only downside with that is that it wont follow the semantics of V=1.

I think it would be better to follow what automake has documented at the bottom of:
https://www.gnu.org/software/automake/manual/html_node/Automake-Silent-Rules.html

tstruk (approved):

Issue #2009, opened 2021-03-04
Quiet TCTI Ldr noise


williamcroberts:
When you use TCTI Ldr, its super noisy during the probe, we should device a way to silence this. Perhaps we could do something like a 'silent=true' option to all tcti's and add that to the user config if silent is not specified.

We only care that NO tcti connection is made, and then perhaps a list of attempted tctis used would be nice, extra points for somehow aggregating all the erros and dumping them only on failure. Perhaps override the logging handler and buffer all of this as an approach?

```
ERROR:tcti:src/tss2-tcti/tcti-device.c:441:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpmrm0: No such file or directory 
WARNING:tcti:src/tss2-tcti/tctildr.c:79:tcti_from_init() TCTI init for function 0x7fed0b01692a failed with a000a 
WARNING:tcti:src/tss2-tcti/tctildr.c:109:tcti_from_info() Could not initialize TCTI named: tcti-device 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-device.so.0 
ERROR:tcti:src/tss2-tcti/tcti-device.c:441:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpm0: Permission denied 
WARNING:tcti:src/tss2-tcti/tctildr.c:79:tcti_from_init() TCTI init for function 0x7fed0b01692a failed with a000a 
WARNING:tcti:src/tss2-tcti/tctildr.c:109:tcti_from_info() Could not initialize TCTI named: tcti-device 
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:154:tcti_from_file() Could not initialize TCTI file: libtss2-tcti-device.so.0 
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:223:tcti_control_command() Control command failed with error: 0 
ERROR:tcti:src/tss2-tcti/tcti-swtpm.c:331:tcti_swtpm_set_locality() Failed to set locality: 0xa000a 
WARNING:tcti:src/tss2-tcti/tcti-swtpm.c:600:Tss2_Tcti_Swtpm_Init() Could not set locality via control channel: 0xa000a 
WARNING:tcti:src/tss2-tcti/tctildr.c:79:tcti_from_init() TCTI init for function 0x7fed0b017ec8 failed with a000a 
[...]

PR #2041, opened 2021-03-23
tcti: quiet tcti ldr noise

tstruk:
Fixes: #2009

### DIFF
--- a/src/tss2-tcti/tctildr.c
+++ b/src/tss2-tcti/tctildr.c
@@ -74,9 +74,22 @@ tcti_from_init(TSS2_TCTI_INIT_FUNC init,
         return TSS2_ESYS_RC_MEMORY;
     }
 
+    /* Unless tcti loglevel is log_debug or higher
+     * (i.e. TSS2_LOG=tcti+debug) turn the logging
+     * from loaded tctis off completely, including warnings
+     * and error logs. It makes too much noise when tcti
+     * loader tries them all one by one and what we want
+     * use is the last one.
+     */
+    log_level old_loglevel = LOGMODULE_status;
+    if (LOGMODULE_status < LOGLEVEL_INFO)
+        LOGMODULE_status = LOGLEVEL_NONE;
+
     r = init(*tcti, &size, conf);
+    LOGMODULE_status = old_loglevel;
+
     if (r != TSS2_RC_SUCCESS) {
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT (approved):
Would we maybe want to add a LOG_INFO with the successfully loaded tcti ?

src/tss2-tcti/tctildr.c
```diff
@@ -74,15 +74,23 @@ tcti_from_init(TSS2_TCTI_INIT_FUNC init,
         return TSS2_ESYS_RC_MEMORY;
     }
 
+    /* Keep the level of logging in the loaded tcti low.
+     * It makes too much noise when tcti loader tries them
+     * all one by one and what we want use is the last one.
+     */
+    log_level old_loglevel = LOGMODULE_status;
+    LOGMODULE_status = LOGLEVEL_NONE;
```
AndreasFuchsSIT, line 82: By doing this, will we ever be able to enable LOGGING inside the init at all anyways ?

tstruk: No, this will quiet all output from any loaded tcti, only tctildr will log messages.

tstruk:
> Would we maybe want to add a LOG_INFO with the successfully loaded tcti ?

done

src/tss2-tcti/tctildr.c
AndreasFuchsSIT: And there's no way to turn those warning back on, right ?
So if you want to know why tcti-device failed (e.g.permissions or whatnot) there's no way to get to this info by increasing the debug level, right ?

But I don't have a solution, I just realized that this is the case...

AndreasFuchsSIT: How about this code:
```
    log_level old_loglevel = LOGMODULE_status;
    if (LOGMODULE_status <= LOGLEVEL_INFO)
        LOGMODULE_status = LOGLEVEL_NONE;
    r = init(*tcti, &size, conf);
    LOGMODULE_status = old_loglevel;
```

That way we don't silence `TSS2_LOG=tcti+debug` for the tcti's inside.

tstruk: Yeah, that will work better. Updated to comment accordingly.

AndreasFuchsSIT (approved):

PR #2042, opened 2021-03-24
Enable lower maxlog levels

AndreasFuchsSIT:
This patch fixed the setting of maxloglevel during configure wrt unused variables throughout the stack.

No clue how far this needs to be backported


tstruk:
Looks like we need a CI run with ./configure --with-maxloglevel=none

tstruk:
We should introduce something like:
#define MAYBE_UNUSED  __attribute__((unused))
as in https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes
We already have UNUSED(param) defined in pcap tcti so we can put these in a common header.

tstruk:
Have a PR here https://github.com/tpm2-software/tpm2-tss/pull/2044

PR #2044, opened 2021-03-24
Add unused and maybe_unused macros

tstruk:
* Add a CI build with --with-maxloglevel=none configuration
* Add unused and maybe_unused macros
* Change the tss lib code to use the new macros
* Change test code to use the new macros

### DIFF
--- a/src/util/aux_util.h
+++ b/src/util/aux_util.h
@@ -7,7 +7,9 @@
 #define AUX_UTIL_H
 
 #include <stdbool.h>
-
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
 #include "tss2_tpm2_types.h"
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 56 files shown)


AndreasFuchsSIT (approved):
LGTM.

The first will have to be backported though...

tstruk:
> LGTM.
> 
> The first will have to be backported though...

You mean this one? https://github.com/tpm2-software/tpm2-tss/pull/2044/commits/26db3573233939c815233b56b49ae841fc16ef1b
I pushed it to 3.0.x and 2.4.x already.
This commit is not supposed to be part of this PR anyway. I need to rebase the PR before merging.

src/util/aux_util.h
```diff
@@ -19,6 +19,9 @@ extern "C" {
 #define TPM2_ERROR_TEXT(r) "Error", "Code", r
 #define SIZE_OF_ARY(ary) (sizeof(ary) / sizeof(ary[0]))
 
+#define UNUSED(x) (void)(x)
+#define MAYBE_UNUSED __attribute__((unused))
```
williamcroberts, line 23: Do we really need MAYBE_UNUSED? I'm trying to figure out what this really does? Either the variable gets referenced or it doesn't. Sure their may be code paths that don't reference it, but that's normal.

williamcroberts: Also don't we want COMPILER_ATTR(unused)?

If loglevel=none is something that supposed to remove all logging code and disable it (ie no runtime level changes), shouldn't the macro be ifdef'd on the log level being none and the name be something like LOG_UNUSED() or something?

williamcroberts: I think what we want it is:
```C
#if MAXLOGLEVEL == LOGL_NONE
LOG_MAYBE_UNUSED COMPILER_ATTR(unused)
#else
LOG_MAYBE_UNUSED
#endif

This way we don't mark thinks always unused when they are MAYBE_UNUSED. We don't want to suppress legitimate unused warnings.

tstruk: MAYBE_UNUSED also applies to variables we use for assertions i.e. only in debug mode, mainly in test code.

williamcroberts: You can use `ifndef NDEBUG` from config.h for detecting debug builds. We just don't want a maybe macro that hides actual unused vars. I don't care what you call them.

williamcroberts: We just don't want a maybe macro that hides actual unused vars ... that we expect to have used paths.

tstruk: I can add some #ifdefs around MAYBE_UNUSED, no problem.
The main point of this PR is to catch early (i.e. in CI) PRs that break the TSS build with values for --with-loglevel flag different 
 that "warning" (the default) and/or --with-debug=no. Currently the master build fails when configured that way.
From the other hand having an unused variable marked and MAYBE_UNUSED doesn't cause any harm.

williamcroberts: It doesn't really cause any harm, except you can end up in conditions where their are extra variables floating around the code base; i'd like to avoid that as much as possible. Especially considering all the work we (mostly you) went through turning stuff on for code hygiene. Thanks.

tstruk: This: https://github.com/tpm2-software/tpm2-tss/pull/2044/commits/acc8b92890748a8b0e6d55e602f38017029ab630#diff-466b6f179d6c60bef3e5d528ebdc62bc54d25722b04f238d713fab2e8ff19515R31
should do the trick.

PR #2048, opened 2021-03-25
added supplementary examples for beginners

drvarun:
Added supplementary examples for beginners. These examples would help to understand how use the library for creating simple standalone applications harnessing TPM capabilities.


drvarun (commented):
Signed-off-by: drvarun <12458730+drvarun@users.noreply.github.com>

tstruk:
@drvarun thanks for the PR, but we would rather want to avoid adding references to people's private repositories as we don't have control over what goes in there, and if the content is peer reviewed and/or tested. I would encourage you to add the examples to our https://github.com/tpm2-software/tpm2-software.github.io repo, into the tutorial section, where it can be reviewed and built on by others.

PR #2049, opened 2021-03-25
FAPI: Perform URL encoding

### LINKED ISSUE #2046: Remote certificate retrival error during tss2_provision
During `tss2_provision` I encountered an error while retrieving the Root CA of my Nuvoton TPM, which is reported by the EK x509 as residing at:

```
Authority Information Access:
	CA Issuers - URI:http://www.nuvoton.com/security/NTC-TPM-EK-Cert/Nuvoton TPM Root CA 2110.cer
```

Two issues prevented the retrieval of the certificate:
- nutovon.com has deprecated `http://` and issue a 301 redirect to `https://`
- The whitespace in the URL parsed from the EK x509 is not URL encoded in the GET request, resulting in the remote resource being "not found".

I used a local proxy to work-around the 301 redirect-- while I am not sure if any redirects should be blindly followed-- and rendered this patch against 3.0.3 to handle the URL encoding of whitespace. I was then able to successfully use `tss2_provision`.

```diff
--- a/src/tss2-fapi/ifapi_helpers.c
+++ b/src/tss2-fapi/ifapi_helpers.c
@@ -2537,12 +2537,16 @@
     }
 
     CURL *curl = curl_easy_init();
+    CURLU *urlp = curl_url();
+    int res = 0;
     if (!curl) {
         LOG_ERROR("curl_easy_init failed");
         goto out_global_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_URL, url);
+    res = curl_url_set(urlp, CURLUPART_URL, url, CURLU_URLENCODE);
+
+    rc = curl_easy_setopt(curl, CURLOPT_CURLU, urlp);
     if (rc != CURLE_OK) {
         LOG_ERROR("curl_easy_setopt for CURLOPT_URL failed: %s",
                 curl_easy_strerror(rc));
@@ -2586,6 +2590,7 @@
     if (ret != 0)
         free(curl_buffer.buffer);
     curl_easy_cleanup(curl);
+    curl_url_cleanup(urlp);
 out_global_cleanup:
     curl_global_cleanup();
 out_memory:

```

mailhost:
Use the libcurl API to perform url encoding.

During operations such as tss2_provisioning, the parsed EK x509 cert may hold
a reference to a CA Issuer URI that does not adhere to rfc3986
"character encoding" requirements, a.k.a percent-encoding. We need to
perform url encoding before attempting to GET the resource.

Fixes #2046

### DIFF
--- a/src/tss2-fapi/ifapi_helpers.c
+++ b/src/tss2-fapi/ifapi_helpers.c
@@ -2415,12 +2415,16 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
     }
 
     CURL *curl = curl_easy_init();
+    CURLU *urlp = curl_url();
+    int res = 0;
     if (!curl) {
         LOG_ERROR("curl_easy_init failed");
         goto out_global_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_URL, url);
+    res = curl_url_set(urlp, CURLUPART_URL, url, CURLU_URLENCODE);
+
+    rc = curl_easy_setopt(curl, CURLOPT_CURLU, urlp);
     if (rc != CURLE_OK) {
         LOG_ERROR("curl_easy_setopt for CURLOPT_URL failed: %s",
                 curl_easy_strerror(rc));
@@ -2464,6 +2468,7 @@ out_easy_cleanup:
     if (ret != 0)
         free(curl_buffer.buffer);
     curl_easy_cleanup(curl);
+    curl_url_cleanup(urlp);
 out_global_cleanup:
     curl_global_cleanup();
 out_memory:


williamcroberts:
Please sign off your commit:
```bash
git commit --amend -s
git push -f mailhost
# or whatever your remote is named.
```

williamcroberts:
Somehow you ended up with 3 commits. You just want one commit. I'll clean up your history for you.

The CI is failing, becuase it appears that ubuntu18.04 version of curl is too old. 7.68.0 appears to missing the URL API:
  - https://github.com/curl/curl/commit/fb30ac5a2d63773c529c19259754e2b306ac2e2e
  
 Which I don't understand, because it shows the commit on 7.68.0 of the library:
```bash
$ git tag --contains fb30ac5a2
<snip>
curl-7_68_0
<snip>
```

But the header file:
```bash
pkg-config --cflags-only-I libcurl
-I/usr/include/x86_64-linux-gnu
```

```C
#define CURLVERSION_NOW CURLVERSION_SIXTH
```
Seems suspicious

src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2415,12 +2415,16 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
     }
 
     CURL *curl = curl_easy_init();
+    CURLU *urlp = curl_url();
```
williamcroberts, line 2418: null check curl_url(): https://curl.se/libcurl/c/curl_url.html

src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2415,12 +2415,16 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
...
+    int res = 0;
     if (!curl) {
         LOG_ERROR("curl_easy_init failed");
         goto out_global_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_URL, url);
+    res = curl_url_set(urlp, CURLUPART_URL, url, CURLU_URLENCODE);
```
williamcroberts, line 2425: check return value

src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2415,12 +2415,16 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
     }
 
     CURL *curl = curl_easy_init();
+    CURLU *urlp = curl_url();
+    int res = 0;
```
williamcroberts, line 2419: Why not just use rc?

williamcroberts (changes requested):

williamcroberts:
Seems curl on ubuntu-18.04 is only 7.58... So these aren't supported. https://packages.ubuntu.com/bionic/curl

AndreasFuchsSIT:
`AC_CHECK_LIB(curl, curl_url,[AC_DEFINE([HAVE_CURL_URL],[1],[Check for curl_url()])`
and
`#ifdef HAVE_CURL_URL`
?

williamcroberts:
> `AC_CHECK_LIB(curl, curl_url,[AC_DEFINE([HAVE_CURL_URL],[1],[Check for curl_url()])`
> and
> `#ifdef HAVE_CURL_URL`
> ?

curl.pc is provided by libcurl4-gnutls-dev (or openssl or nss variants). For some reason on my system it shows curl as being version 7.68 but the package itself provides 7.58. I don't ever recall building it by scratch and messing with it.

In docker of ubuntu 18.04, it shows the right one:
```
pkg-config --modversion libcurl
7.58.0
```

Wouldn't we want to find a fix that's going to scale across all of our active build flavors here?
Dang debian and their old packages!

src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2415,12 +2415,16 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
...
     if (!curl) {
         LOG_ERROR("curl_easy_init failed");
         goto out_global_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_URL, url);
```
williamcroberts, line 2423: Just use curl_easy_escape on URL and pass that to curl_easy_setopt.

AndreasFuchsSIT:
> Wouldn't we want to find a fix that's going to scale across all of our active build flavors here?
> Dang debian and their old packages!

I guess that would mean `url = str_replace(url, " ", "%20");` ?
There is no such function in standard-C though, right ?

williamcroberts:
Summary:
I think the right fix here is to take the URL and pass it through curl_easy_escape(). Then pass that to curl_easy_setopt(curl, CURLOPT_URL, url);

Just make sure to curl_free() the url returned from curl_easy_escape().

This is how tpm2-tools tpm2_getekcertificate.c does it for a reference:
  - https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_getekcertificate.c#L179
  - https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_getekcertificate.c#L239

williamcroberts:
> > Wouldn't we want to find a fix that's going to scale across all of our active build flavors here?
> > Dang debian and their old packages!
> 
> I guess that would mean `url = str_replace(url, " ", "%20");` ?
> There is no such function in standard-C though, right ?

curl_easy_escape, see my "summary" comment.

williamcroberts:
> > > Wouldn't we want to find a fix that's going to scale across all of our active build flavors here?
> > > Dang debian and their old packages!
> > 
> > 
> > I guess that would mean `url = str_replace(url, " ", "%20");` ?
> > There is no such function in standard-C though, right ?
> 
> curl_easy_escape, see my "summary" comment.

@tstruk can you make these changes using curl_easy_escape and get it staged for next and backport branches since you're working on the releases or is it too late?

williamcroberts:
Replaced by: https://github.com/tpm2-software/tpm2-tss/pull/2088

PR #2050, opened 2021-03-26
fix issues with MAYBE_UNUSED

williamcroberts:
Corrects:
- src/util/aux_util: drop NDEBUG from  MAYBE_UNUSED_MACRO
- test: conditionally include primaryHandle_AuthSession
- test: simplify conditional compilation expression
- test: fix preprocessor if statements
- fapi: remove UNUSED variables
- tpms-types.c: mark param as UNUSED not MAYBE_UNUSED

### DIFF
--- a/src/util/aux_util.h
+++ b/src/util/aux_util.h
@@ -33,7 +33,7 @@ extern "C" {
  * MAYBE_UNUSED macro should be used to mark variables used only
  * for assertions i.e. in debug mode, and/or for logging, which
  * might be compiled out. This shuldn't trigger 'unused variable'
- * or 'variable assigned, but not used' waraings when debug and
+ * or 'variable assigned, but not used' warnings when debug and
  * logging is disabled on configure time, but should trigger
  * warnings for variables that are not used for neither.
  */
--- a/src/util/aux_util.h
+++ b/src/util/aux_util.h
@@ -28,7 +28,7 @@ extern "C" {
 #endif
 
 #define UNUSED(x) (void)(x)
-#if (MAXLOGLEVEL == LOGL_NONE) || defined(NDEBUG)
+#if (MAXLOGLEVEL == LOGL_NONE)
 /* Note:
  * MAYBE_UNUSED macro should be used to mark variables used only
  * for assertions i.e. in debug mode, and/or for logging, which
--- a/src/tss2-esys/esys_crypto_mbed.c
+++ b/src/tss2-esys/esys_crypto_mbed.c
@@ -444,7 +444,7 @@ iesys_cryptmbed_hmac_abort(IESYS_CRYPTO_CONTEXT_BLOB ** context)
  * @param[om] buf_size Number of bytes to write to buffer.
  */
 static int get_random(void* context, unsigned char * buffer, size_t buf_size) {
-    (void)context;
+    UNUSED(context);
 
     int r = 0; //success in terms of mbedtls
 
--- a/src/tss2-fapi/api/Fapi_ChangeAuth.c
+++ b/src/tss2-fapi/api/Fapi_ChangeAuth.c
@@ -423,7 +423,7 @@ Fapi_ChangeAuth_Finish(
 
         statecase(context->state, ENTITY_CHANGE_AUTH_WRITE)
             /* Finish writing the object to the key store */
-            r = ifapi_keystore_store_finish(&context->keystore, &context->io);
+            r = ifapi_keystore_store_finish(&context->io);
             return_try_again(r);
             return_if_error_reset_state(r, "write_finish failed");
 
--- a/src/tss2-fapi/api/Fapi_CreateNv.c
+++ b/src/tss2-fapi/api/Fapi_CreateNv.c
@@ -190,8 +190,7 @@ Fapi_CreateNv_Async(
     return_if_error(r, "Initialize NV_CreateNv");
 
     /* First check whether an existing object would be overwritten */
-    r = ifapi_keystore_check_overwrite(&context->keystore, &context->io,
-                                       path);
+    r = ifapi_keystore_check_overwrite(&context->keystore, path);
     return_if_error2(r, "Check overwrite %s", path);
 
     /* Copy parameters to context for use during _Finish. */
@@ -467,7 +466,7 @@ Fapi_CreateNv_Finish(
 
         statecase(context->state, NV_CREATE_WRITE)
             /* Finish writing the NV object to the key store */
-            r = ifapi_keystore_store_finish(&context->keystore, &context->io);
+            r = ifapi_keystore_store_finish(&context->io);
             return_try_again(r);
             return_if_error_reset_state(r, "write_finish failed");
 
--- a/src/tss2-fapi/api/Fapi_Import.c
+++ b/src/tss2-fapi/api/Fapi_Import.c
@@ -234,7 +234,7 @@ Fapi_Import_Async(
             goto_if_error(r, "Serialize policy", cleanup_error);
 
             /* Check whether an existing object would be overwritten */
-            r = ifapi_policystore_check_overwrite(&context->pstore, &context->io,
+            r = ifapi_policystore_check_overwrite(&context->pstore,
                                                   command->out_path);
             goto_if_error_reset_state(r, "Check overwrite %s", cleanup_error,
                                       command->out_path);
@@ -257,8 +257,8 @@ Fapi_Import_Async(
             switch (object->objectType) {
             case IFAPI_EXT_PUB_KEY_OBJ:
                 /* Check whether an existing object would be overwritten */
-                r = ifapi_keystore_check_overwrite(&context->keystore, &context->io,
-                                                   command->out_path);
+                r = ifapi_keystore_check_overwrite(&context->keystore,
+                        command->out_path);
                 goto_if_error_reset_state(r, "Check overwrite %s", cleanup_error,
                                           command->out_path);
 
@@ -436,7 +436,7 @@ Fapi_Import_Finish(
             goto_if_error_reset_state(r, "Load", error_cleanup);
 
             /* Check whether object already exists in key store. */
-            r = ifapi_keystore_check_overwrite(&context->keystore, &context->io,
+            r = ifapi_keystore_check_overwrite(&context->keystore,
                                                command->out_path);
             goto_if_error_reset_state(r, "Check overwrite %s", error_cleanup,
                                       command->out_path);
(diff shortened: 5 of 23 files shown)


tstruk (commented):
NDEBUG is defined when debugging is disabled, in which case we also want to define the MAYBE_UNUSED.

williamcroberts:
> NDEBUG is defined when debugging is disabled, in which case we also want to define the MAYBE_UNUSED.

Are the use cases not:
1. This variable gets used  but only in debug builds?
2. This variable gets used but only in builds with logging?

tstruk:
> > NDEBUG is defined when debugging is disabled, in which case we also want to define the MAYBE_UNUSED.
> 
> Are the use cases not:
> 
> 1. This variable gets used  but only in debug builds?
> 2. This variable gets used but only in builds with logging?

exactly that

williamcroberts:
> > > NDEBUG is defined when debugging is disabled, in which case we also want to define the MAYBE_UNUSED.
> > 
> > 
> > Are the use cases not:
> > 
> > 1. This variable gets used  but only in debug builds?
> > 2. This variable gets used but only in builds with logging?
> 
> exactly that

OK, item 1:  This variable gets used  but only in debug builds. That's NOT NOT DEBUG (Release).
NDEBUG is defined on release, so to get debug it's, not NDEBUG or `!defined(NDEBUG)`.

tstruk:
> OK, item 1: This variable gets used but only in debug builds. That's NOT NOT DEBUG (Release).
> NDEBUG is defined on release, so to get debug it's, not NDEBUG or `!defined(NDEBUG)`.

It's simple - if NDEBUG is defined then we don't use the variable, and so we do want define MAYBE_UNUSED
if not (!defined(NDEBUG)) then we want to define MAYBE_UNUSED as empty.

You get a build errors because before some of the variables were marked as always unused, as (void) var;
In these cases we probably need to use UNUSED(var); instead of MAYBE_UNUSED.
You can leave that with me.

tstruk:
> > OK, item 1: This variable gets used but only in debug builds. That's NOT NOT DEBUG (Release).
> > NDEBUG is defined on release, so to get debug it's, not NDEBUG or `!defined(NDEBUG)`.
> 
> It's simple - if NDEBUG is defined then we don't use the variable, and so we do want define MAYBE_UNUSED
> if not (!defined(NDEBUG)) then we want to define MAYBE_UNUSED as empty.
> 
> You get a build errors because before some of the variables were marked as always unused, as (void) var;
> In these cases we probably need to use UNUSED(var); instead of MAYBE_UNUSED.
> You can leave that with me.

Les't fix that that first, and then we can remove the unused variables from FAPI.

williamcroberts:
> > OK, item 1: This variable gets used but only in debug builds. That's NOT NOT DEBUG (Release).
> > NDEBUG is defined on release, so to get debug it's, not NDEBUG or `!defined(NDEBUG)`.
> 
> It's simple - if NDEBUG is defined then we don't use the variable, and so we do want define MAYBE_UNUSED

Apparently not. If it's a release build, we do not want to mark anything MAYBE_UNUSED as unused except for the log level case. If you agree with that than:

NDEBUG | BUILD TYPE | ACTION |
| ----------- | -----------| ------------------- |
| defined  | release | mark as unused |
| undefined | debug | mark as used |

A release build is defined as !defined(NDEBUG).

> if not (!defined(NDEBUG)) then we want to define MAYBE_UNUSED as empty.
> 
> You get a build errors because before some of the variables were marked as always unused, as (void) var;
> In these cases we probably need to use UNUSED(var); instead of MAYBE_UNUSED.
> You can leave that with me.

I'm fixing those. Almost done.

src/util/aux_util.h
```diff
@@ -28,7 +28,7 @@ extern "C" {
 #endif
 
 #define UNUSED(x) (void)(x)
-#if (MAXLOGLEVEL == LOGL_NONE) || defined(NDEBUG)
+#if (MAXLOGLEVEL == LOGL_NONE) || !defined(NDEBUG)
```
tstruk, line 31: That is wrong. If NDEBUG is defined then we do want to have have MAYBE_UNUSED defined also.

tstruk (changes requested):

src/util/aux_util.h
williamcroberts: > That is wrong. If NDEBUG is defined then we do want to have have MAYBE_UNUSED defined also.

Because debug builds reference things for debug output that release builds wont, got it. But since I was able to make changes, combined we have done NDEBUG and non NDEBUG builds successfully, which *i think* means NDEBUG case ins't needed. Testing this hypothesis now.

tstruk:
Here is what we need for now: https://github.com/tpm2-software/tpm2-tss/pull/2051

williamcroberts:
> Here is what we need for now: #2051

Its more than just that:
1. The NDEBUG case in MAYBE_UNUSED is not needed.
2. The MAYBE_UNUSED in the test case papered over a bug in the conditionally compiled code.

Please use this PR.

tstruk:
> > Here is what we need for now: #2051
> 
> Its more than just that:
> 
> 1. The NDEBUG case in MAYBE_UNUSED is not needed.
> 2. The MAYBE_UNUSED in the test case papered over a bug in the conditionally compiled code.
> 
> Please use this PR.

Ok, that looks better. Since you are changing tpms-types.c could you also change the (void)() to UNUSED() in https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpms-types.c#L168
and maybe mark all params as UNUSED() in https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpms-types.c#L150

Also since we touch lots of ifapi code let's see if @JuergenReppSIT and @AndreasFuchsSIT have any opinion on that.
Thanks

williamcroberts:
> > > Here is what we need for now: #2051
> > 
> > 
> > Its more than just that:
> > 
> > 1. The NDEBUG case in MAYBE_UNUSED is not needed.
> > 2. The MAYBE_UNUSED in the test case papered over a bug in the conditionally compiled code.
> > 
> > Please use this PR.
> 
> Ok, that looks better. Since you are changing tpms-types.c could you also change the (void)() to UNUSED() in https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpms-types.c#L168
> and maybe mark all params as UNUSED() in https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpms-types.c#L150

Sure thing.

> 
> Also since we touch lots of ifapi code let's see if @JuergenReppSIT and @AndreasFuchsSIT have any opinion on that.
> Thanks

Definitely. This PR expanded rapidly.

williamcroberts:
> > > > Here is what we need for now: #2051
> > > 
> > > 
> > > Its more than just that:
> > > 
> > > 1. The NDEBUG case in MAYBE_UNUSED is not needed.
> > > 2. The MAYBE_UNUSED in the test case papered over a bug in the conditionally compiled code.
> > > 
> > > Please use this PR.
> > 
> > 
> > Ok, that looks better. Since you are changing tpms-types.c could you also change the (void)() to UNUSED() in https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpms-types.c#L168
> > and maybe mark all params as UNUSED() in https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-mu/tpms-types.c#L150
> 
> Sure thing.

Done

> 
> > Also since we touch lots of ifapi code let's see if @JuergenReppSIT and @AndreasFuchsSIT have any opinion on that.
> > Thanks
> 
> Definitely. This PR expanded rapidly.

Done

I also added in some other missed fapi (void) locations and a (void) in mbedtls backend.
FYI I used this to look for void casts...
`grep -rnE '\(void\)\w+;'`
Not sure if that's getting them all, but it's coming back empty now.

tstruk:
I would squash https://github.com/tpm2-software/tpm2-tss/pull/2050/commits/78d78868ea2d04035da5c88c77c144f100f23f44 and https://github.com/tpm2-software/tpm2-tss/pull/2050/commits/f00ce8f339ce9e613d618ac26ecc56b0aa2ad692 into https://github.com/tpm2-software/tpm2-tss/pull/2050/commits/a6b1b3e68d2dca738493664555ff76b40a0f225c
If you care. lgtm otherwise.

tstruk:
> I also added in some other missed fapi (void) locations and a (void) in mbedtls backend.
> FYI I used this to look for void casts...
> `grep -rnE '\(void\)\w+;'`
> Not sure if that's getting them all, but it's coming back empty now.

Be careful with that as it may change the fn(void) definitions to something that you don't want.
We have many of those in unit tests.

williamcroberts:
> > I also added in some other missed fapi (void) locations and a (void) in mbedtls backend.
> > FYI I used this to look for void casts...
> > `grep -rnE '\(void\)\w+;'`
> > Not sure if that's getting them all, but it's coming back empty now.
> 
> Be careful with that as it may change the fn(void) definitions to something that you don't want.
> We have many of those in unit tests.

That's why it's not a sed command, so I could audit the matches manually.

williamcroberts:
> I would squash [78d7886](https://github.com/tpm2-software/tpm2-tss/commit/78d78868ea2d04035da5c88c77c144f100f23f44) and [f00ce8f](https://github.com/tpm2-software/tpm2-tss/commit/f00ce8f339ce9e613d618ac26ecc56b0aa2ad692) into [a6b1b3e](https://github.com/tpm2-software/tpm2-tss/commit/a6b1b3e68d2dca738493664555ff76b40a0f225c)
> If you care. lgtm otherwise.

Generally I keep patches as small as possible within reason and keeping with "build atomicity" so it's git bisect-able.

tstruk:
> Generally I keep patches as small as possible within reason and keeping with "build atomicity" so it's git bisect-able.
I get that, and that's all important, but within reason is the key here.
So what's the reason of having the same change on the same file in 3 different commits?
That's a perfect candidate for cleanup with git rebase --interactive. There is even git commit --fixup command to help with exactly that.

williamcroberts:
> > Generally I keep patches as small as possible within reason and keeping with "build atomicity" so it's git bisect-able.
> > I get that, and that's all important, but within reason is the key here.
> > So what's the reason of having the same change on the same file in 3 different commits?

per function. File boundaries generally don't matter, you care more along lines of does this function work, not does this file work. These are incredibly low risk of breaking anything, so it's incredibly unlikely one would want to revert a single one.

> > That's a perfect candidate for cleanup with git rebase --interactive. There is even git commit --fixup command to help with exactly that.

Yep, well aware.

PR #2051, opened 2021-03-26
util: mark variables as unused

tstruk:
Mark some variables are always unused.
They are not used in nether debug nor logging statements.

### DIFF
--- a/src/tss2-fapi/ifapi_keystore.c
+++ b/src/tss2-fapi/ifapi_keystore.c
@@ -796,10 +796,12 @@ cleanup:
  */
 TSS2_RC
 ifapi_keystore_store_finish(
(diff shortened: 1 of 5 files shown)


williamcroberts:
This is just one part of the PR. Plus you didn't update the doxygen comments on paramters. Use PR #2050

williamcroberts:
Theirs also fixes on PR 2051 that fix some issues in the test code that was hidden by MAYBE_UNUSED, as well as removes the need for NDEBUG, among other changes

tstruk:
will go with #2050

williamcroberts:
> will go with #2050

No, this has the other fixes in it. Plus I found other cases of (void) in the fapi code that needs more removals.

PR #2054, opened 2021-03-30
tcti: fix partial read error when set selinux disable

whzhe51:
There is a partial read error, when i set selinux disable.And
i compile tss without "--enable-tcti-partial-reads" flags.

WARNING:tcti:src/tss2-tcti/tcti-device.c:321:tcti_device_receive()
TPM2 response size disagrees with number of bytes read from fd 3.
Header says 0 but we read 29 bytes.
ERROR:sys:src/tss2-sys/api/Tss2_Sys_Execute.c:118:
Tss2_Sys_ExecuteFinish() Malformed reponse: Invalid tag in response
 header: 0

I turn off the tcti-partial-read feature,but tcti_common is still
read the partiali.
p *tcti_common
$1 = {v2 = {v1 = {magic = 9881001429980276155, version = 2,
      transmit = 0x7ffff7fbb6b0 <tcti_device_transmit>,
      receive = 0x7ffff7fbb7d0 <tcti_device_receive>,
      finalize = 0x7ffff7fbbdc0 <tcti_device_finalize>,
      cancel = 0x7ffff7fbb640 <tcti_device_cancel>,
      getPollHandles = 0x7ffff7fbb650 <tcti_device_get_poll_handles>,
      setLocality = 0x7ffff7fbb660 <tcti_device_set_locality>},
    makeSticky = 0x7ffff7fbb390 <tcti_make_sticky_not_implemented>},
state = TCTI_STATE_RECEIVE,
  header = {tag = 0, size = 0, code = 0}, locality = 3 '\003', partial = 64}

So there is no need to get the header.

Signed-off-by: wanghongzhe <wanghongzhe@huawei.com>

### DIFF
--- a/src/tss2-tcti/tcti-device.c
+++ b/src/tss2-tcti/tcti-device.c
@@ -250,6 +250,7 @@ tcti_device_receive (
         LOG_INFO ("Poll timed out on fd %d.", tcti_dev->fd);
         return TSS2_TCTI_RC_TRY_AGAIN;
     } else if (fds.revents == POLLIN) {
+#ifdef TCTI_PARTIAL_READ
         if (tcti_common->partial == true) {
             memcpy(response_buffer, &tcti_common->header, TPM_HEADER_SIZE);
             TEMP_RETRY (size, read (tcti_dev->fd, response_buffer +
@@ -258,6 +259,10 @@ tcti_device_receive (
             TEMP_RETRY (size, read (tcti_dev->fd, response_buffer,
                                     *response_size));
         }
+#else
+        TEMP_RETRY (size, read (tcti_dev->fd, response_buffer,
+                                *response_size));
+#endif
         if (size < 0) {
             LOG_ERROR ("Failed to read response from fd %d, got errno %d: %s",
                tcti_dev->fd, errno, strerror (errno));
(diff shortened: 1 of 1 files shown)


whzhe51:
(gdb) bt
#0  tcti_device_receive (tctiContext=<optimized out>, response_size=0x7fffffffdfc0,
    response_buffer=0x555555567340 "\200\002", timeout=-1) at src/tss2-tcti/tcti-device.c:173
#1  0x00007ffff7fa43fe in Tss2_Sys_ExecuteFinish (sysContext=0x5555555672f0, timeout=<optimized out>)
    at src/tss2-sys/api/Tss2_Sys_Execute.c:86
#2  0x00007ffff7face02 in CommonOneCall (ctx=ctx@entry=0x5555555672f0,
    cmdAuthsArray=cmdAuthsArray@entry=0x7fffffffe120, rspAuthsArray=rspAuthsArray@entry=0x0)
    at src/tss2-sys/sysapi_util.c:157
#3  0x00007ffff7fa604e in Tss2_Sys_HierarchyChangeAuth (sysContext=0x5555555672f0,
    authHandle=<optimized out>, cmdAuthsArray=0x7fffffffe120, newAuth=<optimized out>,
    rspAuthsArray=0x0) at src/tss2-sys/api/Tss2_Sys_HierarchyChangeAuth.c:83
#4  0x0000555555555cf3 in ?? ()
#5  0x0000555555555559 in ?? ()
#6  0x00007ffff7aefc57 in __libc_start_main () from /usr/lib64/libc.so.6
#7  0x000055555555561a in ?? ()

tstruk:
We don't have any TCTI_PARTIAL_READ compile time condition anymore.
It has been removed in this commit: https://github.com/tpm2-software/tpm2-tss/commit/11d60b139bf304fbd6b070dcd2ed6ee09ff4c0f6
What version of TSS are you using?

jeffpferreira:
See #2055. If `partial` is used uninitialized and we find a bogus value of `true` then we will read the TPM response into `response_buffer + TPM_HEADER_SIZE` which explains the invalid header tag of 0.

whzhe51:
> See #2055. If `partial` is used uninitialized and we find a bogus value of `true` then we will read the TPM response into `response_buffer + TPM_HEADER_SIZE` which explains the invalid header tag of 0.

Thanks，it‘s better.’

PR #2058, opened 2021-04-03
cherry-pick: marshal: remove double define

whooo:
It would be good to get this commit into 3.0.x as the double BYTE marshal functions doesn't work well with python cffi.

### DIFF
--- a/include/tss2/tss2_mu.h
+++ b/include/tss2/tss2_mu.h
@@ -21,20 +21,6 @@ extern "C" {
 #endif
 
 TSS2_RC
-Tss2_MU_BYTE_Marshal(
-    BYTE           src,
-    uint8_t         buffer[],
-    size_t          buffer_size,
-    size_t         *offset);
-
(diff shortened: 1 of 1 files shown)


tstruk:
Tss2_MU_BYTE_Marshal is defined by the TCG specification https://trustedcomputinggroup.org/wp-content/uploads/TCG_TSS_Marshaling_Unmarshaling_API_v1p0_r07_pub.pdf

whooo:
The issue is the duplicate definitions of Tss2_MU_BYTE_Marshal and Tss2_MU_BYTE_Unmarshal, already fixed in master, I just would like to see it in 3.0.4 as well

williamcroberts:
> It would be good to get this commit into 3.0.x as the double BYTE marshal functions doesn't work well with python cffi.

We should take this, but in the CFFI bindings ee can drop them in the regex fixups of the headers.

tstruk (approved):

PR #2062, opened 2021-04-19
Update INSTALL.md

Mykrobin:
I met an error when I deploy the libs using the previous version “lib-dev” on ubuntu 18.04, I corrected the error when I changed it to "libcurl4-openssl-dev". Thus, I think it would be better after changing the dependencies.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -51,7 +51,7 @@ $ sudo apt -y install \
   doxygen \
   libjson-c-dev \
   libini-config-dev \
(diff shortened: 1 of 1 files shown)


tstruk:
libcurl-dev is a 'virtual' package that is provided by three different 'real' packages, each using different crypto libraries. libcurl4-openssl-dev is one of them. People may prefer different one so it's better to keep it generic.
See https://packages.ubuntu.com/bionic/libcurl-dev

Issue #1978, opened 2021-02-20
Insufficient error handling in file io


joholl:
The `ifapi_io.c` does not handle some error correctly.

## Reproduce

 Write a simple program, initializing the fapi and reading the fapi configuration file (by default `/etc/tpm2-tss/fapi-config.json`):

```
#include "tss2/tss2_fapi.h"

void main(void) {
    FAPI_CONTEXT *ctx;
    Fapi_Initialize(&ctx, NULL);
}
```
Compile:
```
gcc $(pkg-config --cflags --libs tss2-fapi) main.c -o main
```

Execute:
```
TSS2_LOG=all+trace TSS2_FAPICONF=. ./main
```

Or for tracking where the error is:

```
TSS2_LOG=all+trace TSS2_FAPICONF=. strace ./main
```

## The actual problem

Errors due to non-existent files are reported just fine, because `fopen` is error-checked correctly (in this instance). However, the return code of `fseek` is not checked. Therefore, if a directory is given, `fopen` succeeds, `fseek` fails with -1 (`7fffffffffffffff`), and malloc(`7fffffffffffffff + 1`) fails, resulting in an OOM error message.

Similarly, if the permissions prevent reading, the fapi fails much later when trying to parse JSON. This distracts from an actual permissions problem. I did not look further into why that is.

This applies to multiple io library calls:
 * fopen
   * e.g. src/tss2-fapi/ifapi_io.c:67  (as a race condition)
 * fclose
   * never checked (arguable)
 * fseek
   *  src/tss2-fapi/ifapi_io.c:63
 * ftell
   * src/tss2-fapi/ifapi_io.c:64
 * fcntl
    * e.g. src/tss2-fapi/ifapi_io.c:76

JuergenReppSIT:
@joholl Thank you for reporting this interesting case.
I could reproduce the malloc error with the "." file.  But fseek did return 0 in this case.
In my test ftell did return 7fffffffffffffff which is LONG_MAX.
So I will add the missing checks for -1 in ifapi_io.c and additionally check LONG_MAX for ftell.
I tried to reproduce an error related to the permissions for reading you did describe, but
the correct error messages were produced. Do you have a hint under what circumstances the
errors did occur?

joholl:
@JuergenReppSIT You're right. The permissions issue was a mistake on my end.

I just assumed fseek failed, because my strace looked like this:

```c
openat(AT_FDCWD, ".", O_RDONLY)         = 3
[...]

PR #2070, opened 2021-04-27
[RFC] fapi-io: simplify routines

williamcroberts:
- stat before fopen (TOUTOC issue)
  - extra fclose/fopen things
  - why fail if it's locked but continue if the lock fails (EAGAIN on
    lockf)?

### DIFF
--- a/src/tss2-fapi/ifapi_io.c
+++ b/src/tss2-fapi/ifapi_io.c
@@ -44,27 +44,8 @@ ifapi_io_read_async(
     struct IFAPI_IO *io,
     const char *filename)
 {
-    struct stat statbuf;
-
-    if (stat(filename, &statbuf) == -1) {
-        LOG_ERROR("File \"%s\" not found.", filename);
-        return TSS2_FAPI_RC_IO_ERROR;
-    }
-
-    /* Check whether file is a directory. */
-    if (S_ISDIR(statbuf.st_mode)) {
-        LOG_ERROR("\"%s\" is a directory.", filename);
-        return TSS2_FAPI_RC_IO_ERROR;
-    }
-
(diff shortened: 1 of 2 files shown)


JuergenReppSIT:
* The check whether path is a directory was added to solve issue #1978 because fopen works for directories.
* Yes the check of EAGAIN for lockf does not make sense. Perhaps we should switch from lockf to usage of fcntl with F_WRLCK and F_RDLCK. The man page of lockf states:   _On Linux, lockf() is just an interface on top of fcntl(2) locking.  Many other systems implement lockf() in this way,  but  note  that  POSIX.1 leaves the relationship between lockf() and fcntl(2) locks unspecified.  A portable application should probably avoid mixing calls to these interfaces._  Afterwards we could check whether the strange error in the unit tests with clang still occurs.

williamcroberts:
> * The check whether path is a directory was added to solve issue #1978 because fopen works for directories.

I love non documented features of glibc. It's still a Time of Check Time of Use issue. I would use fileno to produce the fd and pass it to fstat. So yoru check comes after open and you have a stable tether to the underlying filesystem resource.

> * Yes the check of EAGAIN for lockf does not make sense. Perhaps we should switch from lockf to usage of fcntl with F_WRLCK and F_RDLCK. The man page of lockf states:   _On Linux, lockf() is just an interface on top of fcntl(2) locking.  Many other systems implement lockf() in this way,  but  note  that  POSIX.1 leaves the relationship between lockf() and fcntl(2) locks unspecified.  A portable application should probably avoid mixing calls to these interfaces._  Afterwards we could check whether the strange error in the unit tests with clang still occurs.

williamcroberts:
#2072 antiquates this.

PR #2072, opened 2021-04-29
Fapi fix file locking

JuergenReppSIT:
* According to RFC PR #2070 unnecessary function calls were removed.
* According to the recommendation of the lockf man page fcntl was used instead
  of lockf to lock files to assure that the code is portable.
* The usage of lockf was replaced with usage of fcntl in the corresponding
  unit test.

### DIFF
--- a/src/tss2-fapi/ifapi_io.c
+++ b/src/tss2-fapi/ifapi_io.c
@@ -45,26 +45,10 @@ ifapi_io_read_async(
     const char *filename)
 {
     struct stat statbuf;
-
-    if (stat(filename, &statbuf) == -1) {
-        LOG_ERROR("File \"%s\" not found.", filename);
-        return TSS2_FAPI_RC_IO_ERROR;
-    }
-
-    /* Check whether file is a directory. */
-    if (S_ISDIR(statbuf.st_mode)) {
-        LOG_ERROR("\"%s\" is a directory.", filename);
-        return TSS2_FAPI_RC_IO_ERROR;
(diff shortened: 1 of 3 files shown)


src/tss2-fapi/ifapi_io.c
```diff
@@ -45,38 +45,44 @@ ifapi_io_read_async(
     const char *filename)
 {
     struct stat statbuf;
+    struct flock flock;
```
williamcroberts, line 48: Why not just initialize and zero over the memset?
```C
struct flock flock = { 0 };
```

src/tss2-fapi/ifapi_io.c
```diff
@@ -45,38 +45,44 @@ ifapi_io_read_async(
...
-        LOG_ERROR("Open file \"%s\": %s", filename, strerror(errno));
-        return TSS2_FAPI_RC_IO_ERROR;
-    }
+    /* Locking the file. Lock will be released upon close */
+    memset(&flock, 0, sizeof(flock));
+    flock.l_type = F_RDLCK;
+    flock.l_whence = SEEK_SET;
+    flock.l_start = 0;
+    flock.l_len = 0;
+    flock.l_pid = 0;
```
williamcroberts, line 80: It's already zeroed, no need to set it again, or dump the memset/initializer.

williamcroberts (commented):
Looks good, minor changes on struct lockf initialization and sets.

PR #2074, opened 2021-05-03
fix for possible dual freeing.

ihsinme:
in my opinion, the possibility of attempting a double freeing is visible. I propose to fix this with a simple solution.


williamcroberts (approved):

williamcroberts:
@ihsinme can you signoff your commit:
```bash
git commit --amend --s
```

williamcroberts:
> in my opinion, the possibility of attempting a double freeing is visible. I propose to fix this with a simple solution.

check_oom causes a return, so no double free. reverting.

PR #2076, opened 2021-05-04
fix of possible access outside the array.

ihsinme:
the specified condition will skip the value 16, which will cause a write outside the allocated memory.


AndreasFuchsSIT (approved):

williamcroberts:
backported as:
3.1.x: 5993c26aefe9f213d1fe5d07ac4b7d1b53a87121
3.0.x: ae823dae4260f77b211b1f41d58ca3c638259ade
2.4.x: 3caba69f98c4dc3c27b47f1a0ce9932b7ea146f7

And will be on subsequent RC's next week.

PR #2079, opened 2021-05-14
Avoid compiler warnings/errors in the unit test suite when building with GCC 11

diabonas:
GCC 11 tightens static checks, leading to additional compiler warnings that are treated as errors due to `-Werror`.

<s>Since `-Werror` is enabled in release builds as well, it would be good to get these changes into the pending 3.1.0 release as well.</s> `-Werror` is [disabled](https://github.com/tpm2-software/tpm2-tss/blob/2bae50a3551cd8b0123ef609ea973678339df5ae/configure.ac#L435-L436) for releases, so this shouldn't directly affect the upcoming 3.1.0 release.

### DIFF
--- a/test/unit/TPMS-marshal.c
+++ b/test/unit/TPMS-marshal.c
@@ -317,7 +317,7 @@ tpms_unmarshal_buffer_size_lt_data_nad_lt_offset(void **state)
 {
     TPMS_ALG_PROPERTY alg = {0};
     TPMS_CAPABILITY_DATA cap = {0};
-    uint8_t buffer[sizeof(alg) + sizeof(cap) + 1] = { 0 };
+    uint8_t buffer[sizeof(alg) + sizeof(cap) + 3] = { 0 };
     TPMS_ALG_PROPERTY *ptr;
     TPMS_CAPABILITY_DATA *ptr2;
     size_t offset = 3;
--- a/test/unit/TPM2B-marshal.c
+++ b/test/unit/TPM2B-marshal.c
@@ -326,7 +326,7 @@ tpm2b_unmarshal_buffer_null (void **state)
 void
 tpm2b_unmarshal_dest_null (void **state)
 {
-    uint8_t buffer [1];
+    uint8_t buffer [1] = { 0 };
     TSS2_RC rc;
 
     rc = Tss2_MU_TPM2B_DIGEST_Unmarshal (buffer, sizeof (buffer), NULL, NULL);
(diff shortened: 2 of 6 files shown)


test/unit/TPMS-marshal.c
```diff
@@ -317,7 +317,7 @@ tpms_unmarshal_buffer_size_lt_data_nad_lt_offset(void **state)
 {
     TPMS_ALG_PROPERTY alg = {0};
     TPMS_CAPABILITY_DATA cap = {0};
-    uint8_t buffer[sizeof(alg) + sizeof(cap) + 1] = { 0 };
+    uint8_t buffer[sizeof(alg) + sizeof(cap) + 3] = { 0 };
```
AndreasFuchsSIT, line 320: @diabonas Could you post the gcc-error that leads to this change ?
Looks kind of weird to me ?

diabonas: Sure, the error message is
```
test/unit/TPMS-marshal.c: In function ‘tpms_unmarshal_buffer_size_lt_data_nad_lt_offset’:
test/unit/TPMS-marshal.c:335:9: error: array subscript ‘TPMS_CAPABILITY_DATA[0]’ is partly outside array bounds of ‘uint8_t[1041]’ {aka ‘unsigned char[1041]’} [-Werror=array-bounds]
  335 |     ptr2->capability = HOST_TO_BE_32(TPM2_CAP_ECC_CURVES);
      |         ^~
test/unit/TPMS-marshal.c:320:13: note: while referencing ‘buffer’
  320 |     uint8_t buffer[sizeof(alg) + sizeof(cap) + 1] = { 0 };
      |             ^~~~~~
```
IIUC the problem is that [`ptr2` in l. 334](https://github.com/tpm2-software/tpm2-tss/blob/05beb087939024c2f1b640c58147a0c5e4e94fbf/test/unit/TPMS-marshal.c#L334) is outs of bounds of the `buffer`. I must admit I didn't really do the pointer arithmetic math myself, but `3` is the smallest value that removes the warning message.

diabonas: Seems to make sense though: the first `sizeof(alg)` bytes of the `buffer` are used to store a `TPMS_ALG_PROPERTY` struct (which `ptr` points to), while `ptr2` is a pointer to the offset position `sizeof(alg)+3` of the buffer. Since `ptr2` is supposed to store a `TPMS_CAPABILITY_DATA`, it needs `sizeof(cap)` bytes of space, but `buffer` only offers <code>sizeof(alg)+sizeof(cap)+*1*</code> bytes. In order to have enough room for `ptr2`, we therefore need exactly <code>sizeof(alg)+sizeof(cap)+*3*</code> bytes in `buffer` indeed to avoid a write overflow into adjacent memory.

AndreasFuchsSIT (approved):

PR #2082, opened 2021-05-31
Add tss2-dlopen-esys.c

AndreasFuchsSIT:
Add a tss2-dlopen-esys that can be copied into project that do not want to link against libtss2-esys at compile time.
Same for tctildr, mu, rc and fapi.
Also added tests for esys, mu, rc, fapi and failure.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -219,6 +219,14 @@ EXTRA_DIST += \
     src/tss2-tcti/tss2-tcti-tbs.vcxproj.filters \
     tpm2-tss.sln
 
+# tss2-dlopen wrappers
+EXTRA_DIST += \
+    tss2-dlopen/tss2-dlopen-rc.c \
+    tss2-dlopen/tss2-dlopen-tctildr.c \
+    tss2-dlopen/tss2-dlopen-mu.c \
+    tss2-dlopen/tss2-dlopen-esys.c \
+    tss2-dlopen/tss2-dlopen-fapi.c
+
 # Generate the AUTHORS file from git log
 AUTHORS :
 	$(AM_V_GEN)git log --format='%aN <%aE>' | grep -v 'users.noreply.github.com' | sort | \
--- a/tss2-dlopen/tss2-dlopen-esys.c
+++ b/tss2-dlopen/tss2-dlopen-esys.c
@@ -1,0 +1,2208 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*******************************************************************************
+ * Copyright 2021, Fraunhofer SIT
+ * All rights reserved.
+ *******************************************************************************/
+
+/**
+ * The purpose of this file is to copy it into your project and
+ * include it during compilation if you don't want to link against
+ * libtss2-esys at compile time.
+ * It will attempt to load libtss2-esys.so during runtime.
+ * It will either work similarly to directly linking to libtss2-esys.so
+ * at compile-time or return a NOT_IMPLEMENTED error.
+ *
+ * For new versions of this file, please check:
+ * http://github.com/tpm2-software/tpm2-tss/tss2-dlopen
+*/
+
+#include <dlfcn.h>
+#include <stdio.h>
+#include <tss2/tss2_esys.h>
+
+#define str(s) xstr(s)
+#define xstr(s) #s
+
+#ifdef ENABLE_WARN
+#define WARN(str, ...) do { fprintf(stderr, "WARNING: " str "\n", ## __VA_ARGS__); } while (0)
+#else /* ENABLE_WARN */
+#define WARN(...) do { } while (0)
+#endif /* ENABLE_WARN */
+
+#define LIB "libtss2-esys.so.0"
+static void *dlhandle = NULL;
+
+static TSS2_RC
+init_dlhandle(void)
+{
+    if (dlhandle)
+        return TSS2_RC_SUCCESS;
+    dlhandle = dlopen(LIB, RTLD_NOW | RTLD_LOCAL);
+    if (!dlhandle) {
+        WARN("Library " LIB " not found: %s.", dlerror());
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED;
+    }
+    return TSS2_RC_SUCCESS;
+}
+
+TSS2_RC
+Esys_Initialize(
+    ESYS_CONTEXT **esys_context,
+    TSS2_TCTI_CONTEXT *tcti,
+    TSS2_ABI_VERSION *abiVersion)
+{
+    if (init_dlhandle() != TSS2_RC_SUCCESS)
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED;
+
+    static TSS2_RC (*sym) (ESYS_CONTEXT **esys_context, TSS2_TCTI_CONTEXT *tcti, TSS2_ABI_VERSION *abiVersion) = NULL;
+    if (!sym)
+        sym = dlsym(dlhandle, "Esys_Initialize");
+    if (!sym) {
+        WARN("Function Esys_Initialize not found.");
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED;
+    }
+
+    return sym(esys_context, tcti, abiVersion);
+}
+
+void
+Esys_Finalize(ESYS_CONTEXT **ctx)
+{
+    if (!ctx || !*ctx)
+        return;
+    static TSS2_RC (*sym) (ESYS_CONTEXT **ctx) = NULL;
+    if (!sym)
+        sym = dlsym(dlhandle, "Esys_Finalize");
+    if (!sym) {
+        WARN("Function Esys_Finalize not found.");
+        return;
+    }
+    sym(ctx);
+}
+
+void
+Esys_Free(void *__ptr)
+{
+    if (!__ptr)
+        return;
+    static TSS2_RC (*sym) (void *__ptr) = NULL;
+    if (!sym)
+        sym = dlsym(dlhandle, "Esys_Free");
+    if (!sym) {
+        WARN("Function Esys_Free not found.");
+        return;
+    }
+    sym(__ptr);
+}
+
+#define MAKE_ESYS_0(fun) \
+TSS2_RC fun (ESYS_CONTEXT *ctx) { \
+    static TSS2_RC (*sym) (ESYS_CONTEXT *ctx) = NULL; \
+    if (!sym) \
+        sym = dlsym(dlhandle, str(fun)); \
+    if (!sym) { \
+        WARN("Function " str(fun) " not found."); \
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED; \
+    } \
+    return sym(ctx); \
+}
+
+#define MAKE_ESYS_1(fun, type1,parm1) \
+TSS2_RC fun (ESYS_CONTEXT *ctx, type1 parm1) { \
+    static TSS2_RC (*sym) (ESYS_CONTEXT *ctx, type1) = NULL; \
+    if (!sym) \
+        sym = dlsym(dlhandle, str(fun)); \
+    if (!sym) { \
+        WARN("Function " str(fun) " not found."); \
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED; \
+    } \
+    return sym(ctx, parm1); \
+}
+
+#define MAKE_ESYS_2(fun, type1,parm1, type2,parm2) \
+TSS2_RC fun (ESYS_CONTEXT *ctx, type1 parm1, type2 parm2) { \
+    static TSS2_RC (*sym) (ESYS_CONTEXT *ctx, type1, type2) = NULL; \
+    if (!sym) \
+        sym = dlsym(dlhandle, str(fun)); \
+    if (!sym) { \
+        WARN("Function " str(fun) " not found."); \
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED; \
+    } \
+    return sym(ctx, parm1, parm2); \
+}
+
+#define MAKE_ESYS_3(fun, type1,parm1, type2,parm2, type3,parm3) \
+TSS2_RC fun (ESYS_CONTEXT *ctx, type1 parm1, type2 parm2, type3 parm3) { \
+    static TSS2_RC (*sym) (ESYS_CONTEXT *ctx, type1, type2, type3) = NULL; \
+    if (!sym) \
+        sym = dlsym(dlhandle, str(fun)); \
+    if (!sym) { \
+        WARN("Function " str(fun) " not found."); \
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED; \
+    } \
+    return sym(ctx, parm1, parm2, parm3); \
+}
+
+#define MAKE_ESYS_4(fun, type1,parm1, type2,parm2, type3,parm3, type4,parm4) \
+TSS2_RC fun (ESYS_CONTEXT *ctx, type1 parm1, type2 parm2, type3 parm3, type4 parm4) { \
+    static TSS2_RC (*sym) (ESYS_CONTEXT *ctx, type1, type2, type3, type4) = NULL; \
+    if (!sym) \
+        sym = dlsym(dlhandle, str(fun)); \
+    if (!sym) { \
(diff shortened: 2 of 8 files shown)


AndreasFuchsSIT:
@poettering Would you like to review this as it might be an approach for simplifying https://github.com/systemd/systemd/blob/b69855e645a046d0860f2ae99464019e0f6adf81/src/shared/tpm2-util.c ?

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2171 @@
...
+static void *esys_dlhandle = NULL;
+
+TSS2_RC
+Esys_Initialize(
+    ESYS_CONTEXT **esys_context,
+    TSS2_TCTI_CONTEXT *tcti,
+    TSS2_ABI_VERSION *abiVersion)
+{
+    if (!esys_dlhandle) {
+        esys_dlhandle = dlopen("tss2_esys", RTLD_NOW & RTLD_LOCAL);
```
poettering, line 34: i figure you mean | rather than & here...

poettering (changes requested):

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2171 @@
...
+    if (!sym) {
+        fprintf(stderr, "WARNING: Function Esys_Finalize not found.");
+        return;
+    }
+    sym(ctx);
+}
+
+void
+Esys_Free(void *__ptr)
+{
```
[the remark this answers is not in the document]
AndreasFuchsSIT: Thanks for the review. You mean putting `if (!__ptr) return;` to the beginning of the function ?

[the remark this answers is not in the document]
poettering: yeah. precisely.

And Esys_Finalize() should probably do the same:

```c
if (!*ctx)  {
        *ctx = NULL;
        return;  
}
```

AndreasFuchsSIT:
Force-Push with Lennart's comments (hopefully) addressed.
If Peter and Bill are ok too, I'll go ahead and also make wrappers for tctildr, mu, rc and fapi.

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2171 @@
...
+TSS2_RC
+Esys_Initialize(
+    ESYS_CONTEXT **esys_context,
+    TSS2_TCTI_CONTEXT *tcti,
+    TSS2_ABI_VERSION *abiVersion)
+{
+    if (!esys_dlhandle) {
+        esys_dlhandle = dlopen("tss2_esys", RTLD_NOW & RTLD_LOCAL);
+        if (!esys_dlhandle) {
+            fprintf(stderr, "WARNING: Library libtss2_esys not found.");
```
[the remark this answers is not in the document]
AndreasFuchsSIT: Wrapped into a ifdef'd macro

tss2-dlopen/tss2-dlopen-esys.c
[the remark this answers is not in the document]
AndreasFuchsSIT: Without the `*ctx=NULL` NOP, but yes, done... ;-)
Thanks !

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2171 @@
...
+    }
+    sym(__ptr);
+}
+
+#define MAKE_ESYS_0(fun) \
+TSS2_RC fun (ESYS_CONTEXT *ctx) { \
+    TSS2_RC (*sym) (ESYS_CONTEXT *ctx) = \
+        dlsym(esys_dlhandle, str(fun)); \
+    if (!sym) { \
+        fprintf(stderr, "WARNING: Function " str(fun) " not found."); \
```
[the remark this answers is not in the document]
AndreasFuchsSIT: Does it really matter ?
I just took the same string was for symbol lookup.

Could be `__func__` there as well, I agree but maybe someone wants to rename the function names to eg `dl_ ## fun` and then it would not be correct here anymore.

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2201 @@
...
+ * For new versions of this file, please check:
+ * http://github.com/tpm2-software/tpm2-tss/tss2-dlopen
+*/
+
+#include <dlfcn.h>
+#include <stdio.h>
+#include <tss2/tss2_esys.h>
+
+#define str(s) xstr(s)
+#define xstr(s) #s
```
PeterHuewe, line 23: don't get it - why is xstr needed?

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2201 @@
...
+    if (!sym) \
+        dlsym(esys_dlhandle, str(fun)); \
+    if (!sym) { \
+        WARN("Function " str(fun) " not found."); \
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED; \
+    } \
+    return sym(ctx, parm1, parm2, parm3, parm4); \
+}
+
+#define MAKE_ESYS_5(fun, type1,parm1, type2,parm2, type3,parm3, type4,parm4, \
```
PeterHuewe, line 148: for the other macros e.g. MAKE_ESYS_4 you don't "group" type,param but skip the spaces alltogether.
please make this consisten.

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2201 @@
...
+MAKE_ESYS_1(Esys_GetTcti,
+    TSS2_TCTI_CONTEXT **,tcti);
+MAKE_ESYS_2(Esys_GetPollHandles,
+    TSS2_TCTI_POLL_HANDLE **,handles,
+    size_t *,count);
+MAKE_ESYS_1(Esys_SetTimeout,
+    int32_t, timeout);
+MAKE_ESYS_3(Esys_TR_Serialize,
+    ESYS_TR, object,
+    uint8_t **,buffer,
```
PeterHuewe, line 307: same here, sometimes you add spaces, sometimes you don't - please make consistent.

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2201 @@
...
+    if (!sym) { \
+        WARN("Function " str(fun) " not found."); \
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED; \
+    } \
+    return sym(ctx, parm1, parm2, parm3, parm4, parm5, parm6, parm7, parm8, \
+                    parm9, parm10, parm11, parm12, parm13); \
+}
+
+
+
```
PeterHuewe, line 297: what is the order of these functions below?

PeterHuewe:
Few nitpicks, but looks okay to me.

tss2-dlopen/tss2-dlopen-esys.c
AndreasFuchsSIT: The order is the same as in the header file which is the same as in the TPM spec part 3.

tss2-dlopen/tss2-dlopen-esys.c
AndreasFuchsSIT: It's a best practice that has to do with order of macro expansion. You want to expand the contents of s before you stringify it. Don't ask me exactly why, but I learned it somewhere... ;-)

PeterHuewe: ok

tss2-dlopen/tss2-dlopen-esys.c
PeterHuewe: resolved but not changed?

tss2-dlopen/tss2-dlopen-esys.c
PeterHuewe: ok

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2199 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*******************************************************************************
+ * Copyright 2021, Fraunhofer SIT
+ * All rights reserved. *******************************************************************************/
```
poettering, line 4: hmm, wasn#t there supposed to be one more line break in this license header?

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2199 @@
...
+
+#include <dlfcn.h>
+#include <stdio.h>
+#include <tss2/tss2_esys.h>
+
+#define str(s) xstr(s)
+#define xstr(s) #s
+
+#ifdef ENABLE_WARN
+#define WARN(str, ...) fprintf(stderr, "WARNING: " str "\n", ## __VA_ARGS__);
```
poettering, line 26: if you define it this way, then you will insert two ";" in the code. Which is a bit weird. just drop the ";"?

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2199 @@
...
+#else /* ENABLE_WARN */
+#define WARN(...) ;
```
poettering, line 28: change do: `do {} while(0)` or so?

That way people can put a semicolon at the end and will not get the right number of expressions and semicolons, so that you can then do this safely:

```c
if (x)
        WARN(…);
```

And you'll get exactly one ";" generated and no uncertainty what is done as part of the if branch and what not

tss2-dlopen/tss2-dlopen-mu.c
```diff
@@ -0,0 +1,296 @@
...
+#ifdef ENABLE_WARN
+#define WARN(str, ...) fprintf(stderr, "WARNING: " str "\n", ## __VA_ARGS__);
+#else /* ENABLE_WARN */
+#define WARN(...) ;
+#endif /* ENABLE_WARN */
+
+static void *mu_dlhandle = NULL;
+
+static TSS2_RC
+init_dlhandle()
```
poettering, line 34: this isn't valid C. `init_dlhandle()` → `init_dlhandle(void)`

tss2-dlopen/tss2-dlopen-tctildr.c
```diff
@@ -0,0 +1,132 @@
...
+#ifdef ENABLE_WARN
+#define WARN(str, ...) fprintf(stderr, "WARNING: " str "\n", ## __VA_ARGS__);
+#else /* ENABLE_WARN */
+#define WARN(...) ;
+#endif /* ENABLE_WARN */
+
+static void *tctildr_dlhandle = NULL;
+
+static TSS2_RC
+init_dlhandle()
```
poettering, line 34: ditto

Makefile.am
```diff
@@ -219,6 +219,14 @@ EXTRA_DIST += \
     src/tss2-tcti/tss2-tcti-tbs.vcxproj.filters \
     tpm2-tss.sln
 
+# tss2-dlopen wrappers
+EXTRA_DIST += \
+    tss2-dlopen/tss2-dlopen-rc.c \
+    tss2-dlopen/tss2-dlopen-tctildr.c \
+    tss2-dlopen/tss2-dlopen-mu.c \
+    tss2-dlopen/tss2-dlopen-esys.c \
+    tss2-dlopen/tss2-dlopen-fapi.c
```
williamcroberts, line 228: i'm fine with this, just need tests and negative test (cmocka).

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2199 @@
...
+        }
+    }
+
+    static TSS2_RC (*sym) (ESYS_CONTEXT **esys_context, TSS2_TCTI_CONTEXT *tcti, TSS2_ABI_VERSION *abiVersion) = NULL;
+    if (!sym)
+        sym = dlsym(esys_dlhandle, "Esys_Initialize");
+    if (!sym) {
+        WARN("Function Esys_Initialize not found.");
+        return TSS2_ESYS_RC_NOT_IMPLEMENTED;
+    }
```
williamcroberts, line 54: personally, I would do all the heavy lifting in `Esys_Initialize()` and do something like:
```C
malloc(esys_hdr + sizeof(function table);
```
So your structure for esys ctx pointer is:
```
Normal esys ctx
function table
```

Then in your subsequent calls no dlsym needed, you just:
```
if (esys_ctx) {
   esys_ctx->fn1()
}
```
However that would mean esys ctx would be incompatible between dlopen and standard versions, but I don't know how they would end up with both symbol names in their program without doing something weird to bring it on themselves. You could pass a dl-open ctx to the lib version but not vice versa.

williamcroberts: I guess you could just make the whole fn table static global and initialize it in Esys_Initialize, but you would want to make it thread safe, which it currently isn't with your use of static, but the dlopen/dlsym/dlclose are MT safe. However, the scenario of N>1 threads updating the sym variable which are not atomic.

AndreasFuchsSIT: Is it really worth moving all the dlsym() calls into the intiialize ?
It will end up being slower because we make a lot more dlsym() calls than the application really needs. So I see no benefit here.

Wrt being thread safe: The pointer writes are atomic and calling dlsym twice will also not be a problem. If we end up with two handles to libesys / etc it's also ok, because the libtss2_* do not have global state so we can even mix and mingle between those.
Thus, I'd keep the current lightweight version.
If people want to do it differently when they copy these files, they can of course rewrite them to their liking.

tss2-dlopen/tss2-dlopen-esys.c
```diff
@@ -0,0 +1,2208 @@
...
+#include <stdio.h>
+#include <tss2/tss2_esys.h>
+
+#define str(s) xstr(s)
+#define xstr(s) #s
+
+#ifdef ENABLE_WARN
+#define WARN(str, ...) do { fprintf(stderr, "WARNING: " str "\n", ## __VA_ARGS__); } while (0)
+#else /* ENABLE_WARN */
+#define WARN(...) do { ; } while (0)
```
poettering, line 29: you can drop the ";"...

AndreasFuchsSIT:
@williamcroberts I cannot reproduce the asan failure locally. Do you have any clue what that even means ?
```
FAIL: test/unit/dlopen-fail
===========================

==100267==AddressSanitizer CHECK failed: ../../../../src/libsanitizer/asan/asan_interceptors.cc:734 "((__interception::real_memcpy)) != (0)" (0x0, 0x0)
    <empty stack>

FAIL test/unit/dlopen-fail (exit status: 1)
```

williamcroberts:
> "((__interception::real_memcpy)) != (0)" (0x0, 0x0)

no idea, can't you fetch the Docker image and run in there, does it work with newer versions of clang?

I have these aliases in my bashrc:
```bash
_docker_run() {
  local b=$(basename $(pwd)) 
  #docker run --cap-add=SYS_PTRACE --ulimit core=0 -it --env-file .ci/docker.env -v "`pwd`:/workspace/$b" "$1" /bin/bash
  docker run --cap-add=SYS_PTRACE --memory=14GB --cpus=2 -it --env-file .ci/docker.env -v "`pwd`:/workspace/$b" "$1" /bin/bash
}

docker_run() {
  local r="ghcr.io/tpm2-software/${1:-ubuntu-16.04}"
  _docker_run "$r"
}
```

then I cd to the project:
```bash
cd ~/workspace/tpm2-tss
```

Then I run one of the containers:
```bash
docker_run ubuntu-18.04
```

Then you can do something like setting the env. you could optionally do these things before the `docker_run`.

```bash
export CC=clang
```

Then run the ci script or individual commands.

AndreasFuchsSIT:
For the record: The reason was that I did implement dlopen() and dlsym() instead of -ldl.
Apparently libasan has problems with that. So you always have to --wrap=dlopen/sym even if you don't link against -ldl in your test.
Just keeping this info here for the record.

tss2-dlopen/tss2-dlopen-esys.c
williamcroberts: Its MT-Safe so we're fine here.

williamcroberts (approved):

PR #2088, opened 2021-06-09
FAPI: Perform URL encoding

### LINKED ISSUE #2046: Remote certificate retrival error during tss2_provision

williamcroberts:
Use the libcurl API to perform url encoding.

During operations such as tss2_provisioning, the parsed EK x509 cert may hold
a reference to a CA Issuer URI that does not adhere to rfc3986
"character encoding" requirements, a.k.a percent-encoding. We need to
perform url encoding before attempting to GET the resource.

Modifications from code review done by:
  - William Roberts <william.c.roberts@intel.com>

Fixes #2046

Signed-off-by: Alex Waher <alex@waher.org>

### DIFF
--- a/src/tss2-fapi/ifapi_helpers.c
+++ b/src/tss2-fapi/ifapi_helpers.c
@@ -2397,6 +2397,11 @@ write_curl_buffer_cb(void *contents, size_t size, size_t nmemb, void *userp)
     return realsize;
 }
 
+static inline bool do_url_encode(const char *url) {
+
+    return !!strncmp(url, "file:", 5);
+}
+
 /** Get byte buffer from file system or web via curl.
  *
  * @param[in]  url The url of the resource.
@@ -2424,7 +2429,19 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
         goto out_global_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_URL, url);
+    char *urlp = NULL;
+    if (do_url_encode((char *)url)) {
+        urlp = curl_easy_escape(curl, (char *)url, 0);
+        if (!urlp) {
+            LOG_ERROR("curl_easy_escape failed: %s",
+                    curl_easy_strerror(rc));
+            goto out_easy_cleanup;
+        }
+    }
+
+    LOG_DEBUG("CURL get: %s", urlp ? urlp : (char *)url);
+
+    rc = curl_easy_setopt(curl, CURLOPT_URL, urlp ? urlp : (char *)url);
     if (rc != CURLE_OK) {
         LOG_ERROR("curl_easy_setopt for CURLOPT_URL failed: %s",
                 curl_easy_strerror(rc));
@@ -2468,6 +2485,7 @@ out_easy_cleanup:
     if (ret != 0)
         free(curl_buffer.buffer);
     curl_easy_cleanup(curl);
+    curl_free(urlp);
 out_global_cleanup:
     curl_global_cleanup();
 out_memory:


williamcroberts:
Ahh we can't URL encode file: `file%3A%2Fworkspace%2Ftpm2-tss%2Ftest%2Fintegration%2Ffapi-export-policy.fint_intermedecc-ca.crt`

JuergenReppSIT (changes requested):
It would be nice if you could add
 LOG_DEBUG("CURL get: %s", urlp ? urlp : (char *)url));

williamcroberts:
> It would be nice if you could add
> LOG_DEBUG("CURL get: %s", urlp ? urlp : (char *)url));

Done

JuergenReppSIT:
@williamcroberts curl_easy_escape(curl, (char *)url, 0) escapes the complete url:
Example:  `http%3A%2F%2Fpki.infineon.com%2FOptigaEccMfrCA042%2FOptigaEccMfrCA042.crt`
Which causes an error:
`ERROR:fapi:src/tss2-fapi/ifapi_helpers.c:2294:ifapi_get_curl_buffer() curl_easy_perform() failed: Couldn't resolve host name `
In the original PR ` rc = curl_url_set(urlp, CURLUPART_URL, (const char *)url, CURLU_URLENCODE);`
was used for escaping. 
From man page of curl_url_set:
`CURLU_URLENCODE
    When set, curl_url_set(3) URL encodes the part on entry, except for scheme, port and URL.`
So perhaps we should use this function again instead of curl_easy_escape?

williamcroberts:
> @williamcroberts curl_easy_escape(curl, (char *)url, 0) escapes the complete url:
> Example: `http%3A%2F%2Fpki.infineon.com%2FOptigaEccMfrCA042%2FOptigaEccMfrCA042.crt`
> Which causes an error:

Easy enough to fix...

> `ERROR:fapi:src/tss2-fapi/ifapi_helpers.c:2294:ifapi_get_curl_buffer() curl_easy_perform() failed: Couldn't resolve host name `
> In the original PR ` rc = curl_url_set(urlp, CURLUPART_URL, (const char *)url, CURLU_URLENCODE);`
> was used for escaping.
> From man page of curl_url_set:
> `CURLU_URLENCODE When set, curl_url_set(3) URL encodes the part on entry, except for scheme, port and URL.`
> So perhaps we should use this function again instead of curl_easy_escape?

I'd love to, I really would. The problem is that curl versions lag (notably on the debian side), for instance ubuntu-18.04. Which means we would essentially nuke the bugfix upgrade path for those distro maintainers. The rest of the house, Fedora-33+, and OpenSuse, FreeBSD seem fine.

williamcroberts:
So either we kill Ubuntu-18.04 (which I don't want to do), or we write the functionality to do it.

JuergenReppSIT:
> So either we kill Ubuntu-18.04 (which I don't want to do), or we write the functionality to do it.
yes that's true we should not kill Ubuntu-18.04. Perhaps a simple solution where we implement the escaping for a set of characters would be sufficient. Currently as far as I know only blank characters in the URL cause problems.

JuergenReppSIT:
@williamcroberts In the following example I have replaced
` urlp = curl_easy_escape(curl, (char *)url, 0);` with the call of url_encode, where currently only blanks are encoded.
Now the correct intermediate certificate was downloaded.
```
static inline void url_encode(unsigned char* url, char* urle) {
    for (; *url; url++) {
        if (*url == (int)' ')
            sprintf(urle, "%%%02X", *url);
        else
            sprintf(urle, "%c", (char)*url);
        /* Until the string terminator has been set. */
        while (*++urle);
    }
}

/** Get byte buffer from file system or web via curl.
 *
 * @param[in]  url The url of the resource.
 * @param[out] buffer The buffer retrieved via the url.
 * @param[out] buffer_size The size of the retrieved object.
 *
 * @retval 0 if buffer could be retrieved.
 * @retval -1 if an error did occur
 */
int
ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
                          size_t *buffer_size) {
    int ret = -1;
    struct CurlBufferStruct curl_buffer = { .size = 0, .buffer = NULL };
    char urle[strlen((char*)url) * 3 + 1];

    CURLcode rc = curl_global_init(CURL_GLOBAL_DEFAULT);
    if (rc != CURLE_OK) {
        LOG_ERROR("curl_global_init failed: %s", curl_easy_strerror(rc));
        goto out_memory;
    }

    CURL *curl = curl_easy_init();
    if (!curl) {
        LOG_ERROR("curl_easy_init failed");
        goto out_global_cleanup;
    }

    char *urlp = NULL;
    if (do_url_encode((char *)url)) {
        url_encode(url, urle);
        urlp = urle;
    }

    LOG_DEBUG("CURL get: %s", urlp ? urlp : (char *)url);

    rc = curl_easy_setopt(curl, CURLOPT_URL, urlp ? urlp : (char *)url);
    if (rc != CURLE_OK) {
        LOG_ERROR("curl_easy_setopt for CURLOPT_URL failed: %s",
                curl_easy_strerror(rc));
        goto out_easy_cleanup;
    }

    rc = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION,
                          write_curl_buffer_cb);
    if (rc != CURLE_OK) {
        LOG_ERROR("curl_easy_setopt for CURLOPT_URL failed: %s",
                curl_easy_strerror(rc));
        goto out_easy_cleanup;
    }

    rc = curl_easy_setopt(curl, CURLOPT_WRITEDATA,
                          (void *)&curl_buffer);
    if (rc != CURLE_OK) {
        LOG_ERROR("curl_easy_setopt for CURLOPT_URL failed: %s",
                curl_easy_strerror(rc));
        goto out_easy_cleanup;
    }

    if (LOGMODULE_status == LOGLEVEL_TRACE) {
        if (CURLE_OK != curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L)) {
            LOG_WARNING("Curl easy setopt verbose failed");
        }
    }

    rc = curl_easy_perform(curl);
    if (rc != CURLE_OK) {
        LOG_ERROR("curl_easy_perform() failed: %s", curl_easy_strerror(rc));
        goto out_easy_cleanup;
    }

    *buffer = curl_buffer.buffer;
    *buffer_size = curl_buffer.size;

    ret = 0;

out_easy_cleanup:
    if (ret != 0)
        free(curl_buffer.buffer);
    curl_easy_cleanup(curl);
out_global_cleanup:
    curl_global_cleanup();
out_memory:
    return ret;
}
```

williamcroberts:
FYI the most current patch isn't well tested. We need better tests around this, ideally some unit tests where we can call it with some differing values. I have so many questions about what the behavior should be here and what curl expects for things like `file://`. I also wonder what the next character that needs escaping will be. I don't want to get into the game of add, add, add over using something more generic to escape the bad chars.

JuergenReppSIT:
So we would need an URL parser. imho it would be sufficient to extend escaping of the proposed function `url_escape` with all characters except:
alphanumeric characters, "/", ":", ".", hyphen "-"

williamcroberts:
> So we would need an URL parser. imho it would be sufficient to extend escaping of the proposed function `url_escape` with all characters except:
> alphanumeric characters, "/", ":", ".", hyphen "-"

@JuergenReppSIT do you want to take this PR over? I'm running out of time to dedicate to it.

JuergenReppSIT:
@williamcroberts  @AndreasFuchsSIT  yes I can take over the PR.
* The encoding seems not to be a problem. I just downloaded the Nuvoton certificate in a test program without URL encoding but with following the redirection and disabling certificate verification.
* Can we implement the download in this way?
* The download also worked with certificate verification enabled and adding the Nuvoton root certificate temporally to /etc/ssl/certs/ca-certificates.crt

williamcroberts:
> @williamcroberts @AndreasFuchsSIT yes I can take over the PR.

Ill close this one.

> 
> * The encoding seems not to be a problem. I just downloaded the Nuvoton certificate in a test program without URL encoding but with following the redirection and disabling certificate verification.
> * Can we implement the download in this way?

I don't see why not. Ultimately we want to fix the root cause.

> * The download also worked with certificate verification enabled and adding the Nuvoton root certificate temporally to /etc/ssl/certs/ca-certificates.crt

Issue #2089, opened 2021-06-09
Fapi: policy action + sign results in double free


joholl:
When I use a signing key with a policy action, I get a double free. 

Callstack of first free:
```
src/tss2-fapi/api/Fapi_Sign.c:285:Fapi_Sign_Finish():             r = ifapi_load_key(context, command->keyPath,
src/tss2-fapi/fapi_until:2670:ifapi_load_key():                   r = ifapi_load_keys_finish(context, IFAPI_FLUSH_PARENT,
src/tss2-fapi/fapi_until:1828:ifapi_load_key_finish():            ifapi_cleanup_ifapi_object(context->loadKey.key_object);
src/tss2-fapi/ifapi_keystore.c:1734:ifapi_cleanup_ifapi_object(): ifapi_cleanup_policy(object->policy);
src/tss2-fapi/ifapi_helpers.c:1125:ifapi_cleanup_policy():        cleanup_policy_elements(policy->policy);
src/tss2-fapi/ifapi_helpers.c:1096:cleanup_policy_elements():     cleanup_policy_element(&policy->elements[i]);
src/tss2-fapi/ifapi_helpers.c:1069:cleanup_policy_element():      SAFE_FREE(policy->element.PolicyAction.action);
```

Callstack of second free:
```
src/tss2-fapi/api/Fapi_Sign.c:294:Fapi_Sign_Finish():             r = ifapi_key_sign(context, command->key_object,
src/tss2-fapi/fapi_until:2824:ifapi_key_sign():                   ifapi_cleanup_ifapi_object(context->Key_Sign.key_object);
src/tss2-fapi/ifapi_keystore.c:1734:ifapi_cleanup_ifapi_object(): ifapi_cleanup_policy(object->policy);
src/tss2-fapi/ifapi_helpers.c:1125:ifapi_cleanup_policy():        cleanup_policy_elements(policy->policy);
src/tss2-fapi/ifapi_helpers.c:1096:cleanup_policy_elements():     cleanup_policy_element(&policy->elements[i]);
src/tss2-fapi/ifapi_helpers.c:1069:cleanup_policy_element():      SAFE_FREE(policy->element.PolicyAction.action);
```

Commenting out `SAFE_FREE(policy->element.PolicyAction.action);` removes the double free issue (but is not the solution, obviously).

Also, maybe running multiple times into the `SIGN_WAIT_FOR_KEY` state in `ifapi_load_key` causes multiple frees?

PR #2090, opened 2021-06-10
FAPI: Fix double free for policy action

JuergenReppSIT:
The action string was not duplicated when the policy was copied.
Fixes: #2089


williamcroberts:
Looking at the CI, everything is failing but this path, seems FAPI is broken without this. Im hitting merge now.

Issue #1587, opened 2019-12-31
ideas for utility library


williamcroberts:
- Routines to convert from PEM files to TPM structures and back
  - Including key blob format for the engine
- routine for populating a TPM2B_PUBLIC with the provisioning spec details:
  - https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf
  - https://trustedcomputinggroup.org/wp-content/uploads/Credential_Profile_EK_V2.0_R14_published.pdf

AndreasFuchsSIT:
From #1567 
```
const char * _alg_id_to_string (TPM2_ALG_ID hash_kind)
uint32_t _alg_id_get_size (TPM2_ALG_ID hash_kind)
```

AndreasFuchsSIT:
TPM2B_Private from PEM and pfx
Mergeing Public keys into existing TPM2B_Publics

PR #2101, opened 2021-06-22
RFC: Add user-level mu-api call for key derivation

cplappert:
This commit introduces a new user-level mu-api call
(Tss2_MU_TPMT_PUBLIC_DERIVE_Marshal) that allows to
create a Derived Object with the command TPM2_CreateLoaded.

The rationale for a dedicated user-level call is that from a library
perspective it cannot be determined if a regular keyed-hash object
should be created or a derived object since the Marshalling API 
(Tss2_MU_TPMU_PUBLIC_ID_Marshal) selects the dedicated marshalling
method based on the algorithm type which is in both cases TPM2_ALG_KEYEDHASH 
(s. code snippet below).

```
TPMU_MARSHAL2(TPMU_PUBLIC_ID,
    TPM2_ALG_KEYEDHASH, ADDR, keyedHash, Tss2_MU_TPM2B_DIGEST_Marshal,
    TPM2_ALG_SYMCIPHER, ADDR, sym, Tss2_MU_TPM2B_DIGEST_Marshal,
    TPM2_ALG_RSA, ADDR, rsa, Tss2_MU_TPM2B_PUBLIC_KEY_RSA_Marshal,
    TPM2_ALG_ECC, ADDR, ecc, Tss2_MU_TPMS_ECC_POINT_Marshal)
```
Then we would need to add something like

```
 TPM2_ALG_KEYEDHASH, ADDR, derive, Tss2_MU_TPMS_DERIVE_Marshal,
```
which results in a duplicate case value.

Since the user knows which type of object he wants to create, I implemented the new
user-level call.

Any feedback on this approach?

Signed-off-by: Christian Plappert <christian.plappert@sit.fraunhofer.de>


cplappert:
Alternative: Could also be a candidate for the utility library (#1587)? But I do not know the status there.

cplappert:
Another possibility: TPM2 Specification Part 2 (01.59, 12.2.3.2 TPMU_PUBLIC_ID, p.140) does not specify a selector for derive. A new selector (e.g., TPM_ALG_DERIVE or TPM_ALG_KEYEDHASH_DERIVE) would make the user level call obsolete.

AndreasFuchsSIT:
But then we'd have a mismatch in the type selector of the surrounding TPMT_PUBLIC structure.
So TPMT_PUBLIC_DERIVE_mashall would still be needed, correct ?

williamcroberts:
> Alternative: Could also be a candidate for the utility library (#1587)? But I do not know the status there.

We're looking at it in the TSS WG now, I think this would be a candidate for the utility library. I'm confused at what the problem this is solving. Isn't a derived object one in where you specify the seed in the template and have sensitiveDataOrigin clear?

Issue #2098, opened 2021-06-18
Mark FAPI-create objects in keystore


AndreasFuchsSIT:
Given https://github.com/tpm2-software/tpm2-tss/pull/2081 we should differentiate in the FAPI keystore between objects created by FAPI and those that pre-existed.
In the latter case, we should not delete such an object upon FAPI_Delete("/");

AndreasFuchsSIT:
Done

PR #2102, opened 2021-06-23
FAPI: Prohibit deletion of already existing persistent SRK.

JuergenReppSIT:
If the SRK did already exist during provisioning the persistent SRK will not be removed
by Fapi_Delete. Only the corresponding files in the keystore will be deleted.
Addresses #2098.

### DIFF
--- a/src/tss2-fapi/api/Fapi_Delete.c
+++ b/src/tss2-fapi/api/Fapi_Delete.c
@@ -642,18 +642,27 @@ Fapi_Delete_Finish(
             fallthrough;
 
         statecase(context->state, ENTITY_DELETE_KEY_WAIT_FOR_AUTHORIZATION);
+            /* Delete persistent object if not prohibited. */
             if (object->misc.key.persistent_handle) {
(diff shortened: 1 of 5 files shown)


src/tss2-fapi/api/Fapi_Delete.c
```diff
@@ -642,7 +642,8 @@ Fapi_Delete_Finish(
             fallthrough;
 
         statecase(context->state, ENTITY_DELETE_KEY_WAIT_FOR_AUTHORIZATION);
-            if (object->misc.key.persistent_handle) {
+            /* Delete persistent object if not prohibited. */
+            if (object->misc.key.persistent_handle && !object->misc.key.delete_prohibited) {
```
AndreasFuchsSIT, line 646: IMHO; we should make this an explicit error message, like
"Failed to delete key because it was not created by the tss Feature API" or similar

AndreasFuchsSIT (changes requested):

AndreasFuchsSIT (approved):

PR #2111, opened 2021-06-25
FAPI: Allow following of redirects for curl downloads.

JuergenReppSIT:
For the download of some certificates 3xx redirects are used by the provider.
Thus redirects have to be allowed.
If https URLs are used the certificate needed for verification must exists in the ssl
certificate store.


williamcroberts (approved):

williamcroberts:
Thanks for putting in the time to root cause this

PR #2116, opened 2021-07-12
ESYS: Fix buffer overflow in xor parameter obfuscation.

### LINKED ISSUE #2115: Buffer overflow in LOGBLOB_TRACE
The second `LOGBLOB_TRACE` (line 521) causes ASan error (when logging is enabled). The `LOGBLOB_TRACE` is called after calling `data++` several times (line 520). Printing `data_size` bytes thus causes buffer overflow.

https://github.com/tpm2-software/tpm2-tss/blob/37c5bb566cf2016c49b6ab1305161ae108dbc912/src/tss2-esys/esys_crypto.c#L517-L521

JuergenReppSIT:
If trace is activated LOGBLOB_TRACE is called with a wrong pointer to display
the obfuscated data. Fixes #2115.


AndreasFuchsSIT (approved):
Can you provide the backports for 2.4, 3.0 and 3.1 as well ?

PR #2117, opened 2021-07-12
FAPI: Improve error messages for invalid JSON.

JuergenReppSIT:
For invalid JSON currently the position an type of the error is not displayed
in the error message. The user has to check the JSON data with another tool.
Now the line number and the column of the error will be displayed.


AndreasFuchsSIT (approved):
Very cool stuff !

PR #2123, opened 2021-07-16
FAPI: Fix cipherText logging in Fapi_Decrypt_Async

gotthardp:
The `cipherText` is a binary structure, so logging with `%s` often overflows the buffer as there is no NULL at the end, besides printing non-ASCII characters.


gotthardp:
The Cirrus CI check failure seems not relevant.

AndreasFuchsSIT (approved):

PR #2130, opened 2021-07-17
Require OpenSSL >= 1.1.0

gotthardp:
To reduce the amount of `#ifdefs` and version-specific code we drop support for OpenSSL prior 1.1.0, including all forks such as LibreSSL, which are not API compatible with OpenSSL >= 1.1.0. This simplifies code review and maintenance. Moreover, the old OpenSSL versions are not maintained anymore, so using these libraries is not recommended from security reasons anyway.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,7 +14,7 @@ following sections describe them for the supported platforms.
 * C library development libraries and header files
 * pkg-config
 * doxygen
-* OpenSSL development libraries and header files
+* OpenSSL development libraries and header files, version >= 1.1.0
 * libcurl development libraries
 * Access Control List utility (acl)
 * JSON C Development library
@@ -73,7 +73,6 @@ C Runtime (UCRT) version 10.0.16299.0. Building the type marshaling library
 (tss2-mu.dll) and the system API (tss2-sapi.dll) should be as simple as
 loading the tpm2-tss solution (tpm2-tss.sln) with a compatible and properly
 configured version of Visual Studio 2017 and pressing the 'build' button.
-Windows build setup requires OpenSSL >= v1.0.2 crypto library.
 
 ### References
 Visual Studio 2017 with "Clang for Windows": https://blogs.msdn.microsoft.com/vcblog/2017/03/07/use-any-c-compiler-with-visual-studio/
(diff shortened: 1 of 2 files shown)


gotthardp:
This was mentioned on the mailing list and nobody objected https://lists.01.org/hyperkitty/list/tpm2@lists.01.org/thread/AFY35L5AV2DBTKV456KVMAZFHZXR2JUV/

Python 3.10 will even require OpenSSL >= 1.1.1. The corresponding [PEP 644](https://www.python.org/dev/peps/pep-0644/) contains a detailed impact analysis which is also relevant here.

gotthardp:
This PR is to discuss whether and in which release shall the support for old OpenSSL be dropped.

Once agreed, additional PR with individual changes will follow, which will
* add support for OpenSSL 3.0 and above
* drop support for OpenSSL < 1.1.0

AndreasFuchsSIT:
First of all, thanks a lot for this effort! I very much appreciate it !

Two thinks I want to consider first though:
- Which versions of openssl are used in ubuntu LTS and debian stable. If they are >=1.1.0 for 16.04 and debian/raspbian stable, them I'm ok.
- Second: How can we add Openssl 3.0 to our test matrix ?

gotthardp:
Concerning the openssl versions: for Debian there are 2 active LTS versions (third is imminent) and for Ubuntu even 3 active LTS versions that have openssl 1.1.0 or newer:
- Debian: testing (stable in few months) and stable have 1.1.1, oldstable has 1.1.0
- Ubuntu: 18.04LTS has 1.1.0 (ports) or 1.1.1 (security), 20.04LTS as well as 20.10 and 21.04 have 1.1.1 (ports, security)

The only trouble is Ubuntu 16.04LTS and 14.04 LTS, which are both out of standard support (Ubuntu 16.04 had EOL April 2021), but not end of life. See [here](https://wiki.ubuntu.com/Releases).
I would say that commercial companies pay for extended Ubuntu support to get extra stability, so they will likely use some stable branch of tpm2-tss too. It is unlikely they will install a bleeding edge branch on these machines.

gotthardp:
Concerning the tests: the OpenSSL 3.0 is not released yet, so there are no stable images. We could either build it ourselves or wait for some more stable image, which will likely appear after the release of openssl.

Issue #2150, opened 2021-08-15
The library needs to create own OSSL_LIB_CTX with OpenSSL 3.0


gotthardp:
When an OpenSSL provider (like tpm2-openssl) invokes tpm2-tss to create a key, the `iesys_compare_name` tries to calculate a hash before `esysContext->state` is set back to `_ESYS_STATE_INIT`. For this digest operation a global context is used, where the provider is registered. Hence the provider gets called also to calculate this hash-- it tries to start a second esys operation and everything fails.

The tpm2-tss needs to create an own OSSL_LIB_CTX to ensure the default provider is used for such operations to avoid the loop back.

gotthardp:
I will take care of this issue. This is just a notification that the OpenSSL 3.0 support isn't quite ready yet.

PR #2152, opened 2021-08-25
Use default OpenSSL context for internal crypto operations

gotthardp:
The TPM2 provider may be loaded in the global library context. As we don't want the TPM to be called for some operations, we have to initialize own library context with the default provider. This is similar to the RAND_set_rand_method dance with older OpenSSL.

Also, I suggest to remove the second `struct` from the context union as the second is exactly the same as the first. I see no point in having the same struct twice.

Fixes: #2150


gotthardp:
UPDATE: I made the same modifications in the FAPI as well. I modified the ESYS changes a bit, so both FAPI and ESYS uses similar style.

JuergenReppSIT (approved):
Thank you for the quick update!

PR #2154, opened 2021-08-30
Add libltdl-dev to Ubuntu dependencies

morbitzer:
Otherwise ./bootstrap will throw the following error: 

configure.ac:33: error: possibly undefined macro: LT_LIB_DLLOAD
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,7 +53,8 @@ $ sudo apt -y install \
   doxygen \
   libjson-c-dev \
   libini-config-dev \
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT (changes requested):
Could you sQuash your PR into a single commit and add the Signed-off tag please ?

PR #2155, opened 2021-09-02
[RFC] add policy lib

williamcroberts:
This has a ton of todo's but we should probably work through what the
raw internal interface is compared to what we would want to export to
publicly to callers.

TODO: Need to remove duplicate structs between the public header and
internal headers.


src/tss2-policy/tss2_policy.c
```diff
@@ -0,0 +1,124 @@
...
+    return ifapi_policyeval_execute(esys_ctx, &context);
+}
+
+TSS2_RC
+Tss2_PolicyCalculate(
+        TPML_POLICYELEMENTS *policy,
+        TPML_DIGEST_VALUES *policyDigests,
+        TPMI_ALG_HASH hash_alg,
+        size_t hash_size,
+        size_t digest_idx)
```
williamcroberts, line 47: @AndreasFuchsSIT can you enlighten me as to what hash_size and digest_idx do, this is calling ifapi_calculate_policy, and I can't figure out what these inputs do. Also, why is hash_size needed if hash_alg is passed?

williamcroberts: Oh I get digest_idx, I see that it's a list of policies that get passed.

williamcroberts: So can a json file have N>1 policies contained in it?

williamcroberts: Kinda works, the FAPI async model is messing with me on execute. I guess I would have to do a little more work in the internals so make a blocking version of FAPI policy execution.

AndreasFuchsSIT: If I remember correctly, it's as follows:
A policy can have multiple digests, one for each hash-alg.
Thus the list of digests and the array-index we give here.

JuergenReppSIT: Yes e.g. Fapi_ExportPolicy computes the policy digest for every name alg used in the current fapi profiles if possible.

williamcroberts: Ok cool, so I think in this regard we just use index 0, is that always the hashalg specified by the caller of ifapi_calculate_policy?

AndreasFuchsSIT: yes, the caller decides which hashAlg goes into which index.

JuergenReppSIT:
@williamcroberts Your function `Tss2_PolicyInstantiate` has a functionality similar to `ifapi_calculate_tree`
 which has `FAPI_CONTEXT` as parameter and is implemented as a state machine.
Perhaps it would make sense to divide the function `ifapi_calculate_tree` into the two functions:
`ifapi_calculate_policy_prepare` and `ifapi_calculate_policy`.
The parameters of `ifapi_calculate_policy` could be:
```
TSS2_RC
ifapi_calculate_policy(
    IFAPI_POLICY_EVAL_INST_CTX *eval_ctx,
    IFAPI_IO io,
    IFAPI_POLICY_CTX *policy,
    struct IFAPI_POLICY_STORE *pstore;
    const char *policyPath,
    TPMS_POLICY *policy,
    TPMI_ALG_HASH hash_alg,
    size_t *digest_idx,
    size_t *hash_size);
```
The state so far stored in `FAPI_CONTEXT` could now be stored in `eval_ctx` and if
`pstore` is equal NULL `policyPath` could be used as absolute path.
For FAPI the initialization of `eval_ctx`  and the callbacks would be done by `ifapi_calculate_policy_prepare`.
In `tss2_policy.c` a specific version of the initialization and the call of `ifapi_io_poll`
has to be implemented e.g. in `Tss2_PolicyInstantiate_Prepare`.
`ifapi_calculate_policy` could be called in `Tss2_PolicyInstantiate_Finish` which only had to
implement the call of `ifapi_calculate_policy` and the retry handling.
If the current callbacks used by FAPI for instantiation are called without the FAPI context
as user data the policy instantiation would produce segfaults in several cases.
The FAPI default callback methods could be modified to return TSS2_FAPI_RC_BAD_VALUE if access to
the FAPI context is needed. So the already defined callback methods could be used even if FAPI context
is not available.
E.g. for policy authorize instantiation there are cases where the FAPI context is not
needed for instantiation, thus the policy could be computed without FAPI context. 
What do think about this suggestion?

williamcroberts:
> @williamcroberts Your function `Tss2_PolicyInstantiate` has a functionality similar to `ifapi_calculate_tree` which has 

I noticed this as well, but it wasn't clear to me at the moment I had that thought how to divide it.

> `FAPI_CONTEXT` as parameter and is implemented as a state machine. Perhaps it would make sense to divide the function `ifapi_calculate_tree` into the two functions: `ifapi_calculate_policy_prepare` and `ifapi_calculate_policy`. The parameters of `ifapi_calculate_policy` could be:
> 
> ```
> TSS2_RC
> ifapi_calculate_policy(
>     IFAPI_POLICY_EVAL_INST_CTX *eval_ctx,
>     IFAPI_IO io,
>     IFAPI_POLICY_CTX *policy,
>     struct IFAPI_POLICY_STORE *pstore;
>     const char *policyPath,
>     TPMS_POLICY *policy,
>     TPMI_ALG_HASH hash_alg,
>     size_t *digest_idx,
>     size_t *hash_size);
> ```

What does prepare look like?

> 
> The state so far stored in `FAPI_CONTEXT` could now be stored in `eval_ctx` and if `pstore` is equal NULL `policyPath` could be used as absolute path. For FAPI the initialization of `eval_ctx` and the callbacks would be done by `ifapi_calculate_policy_prepare`. In `tss2_policy.c` a specific version of the initialization and the call of `ifapi_io_poll` has to be implemented e.g. in `Tss2_PolicyInstantiate_Prepare`. `ifapi_calculate_policy` could be called in `Tss2_PolicyInstantiate_Finish` which only had to implement the call of `ifapi_calculate_policy` and the retry handling. If no callbacks are defined or if the current callbacks used by FAPI for instantiation are called without the FAPI context as user data the policy instantiation would produce segfaults in several cases. If the callback method is not defined or if the FAPI context is not available for the default method TSS2_FAPI_RC_BAD_VALUE should be returned. So Tss2_PolicyInstatiate_Finish would also work when no callbacks are defined or async callbacks are defined in Tss2_PolicyInstantiate_Prepare. What do think about this suggestion?

I don't fully grok this, and likely won't until I am deeper into splitting off `ifapi_calculate_tree`, but I can definitely say their won't be any of the async API exported in the policy interface at the moment. So with that said, I could just refactor ifapi_calculate_tree to take the IFAPI_POLICY_CTX and support a NULL pstore.

JuergenReppSIT:
@williamcroberts  The prepare only should initialize the callbacks and the state of the automaton. But if no async handling is required it can be done as you did implement it in Tss2_PolicyInstantiate, and the splitting to prevent code duplication is not needed.
But I think it should be checked in ifapi_policyeval_instantiate_finish whether the callbacks are defined.

williamcroberts:
OK, CI is passing, oss-fuzz will pass when this gets merged and they take:
  - https://github.com/williamcroberts/oss-fuzz/tree/tpm2-software-disable-policy

So any feedback on this before I start writing tons of tests?

williamcroberts:
> @williamcroberts The prepare only should initialize the callbacks and the state of the automaton. But if no async handling is required it can be done as you did implement it in Tss2_PolicyInstantiate, and the splitting to prevent code duplication is not needed. But I think it should be checked in ifapi_policyeval_instantiate_finish whether the callbacks are defined.

Actually, I was able to refactor the interface to ifapi_policy_calculate_tree and add the checks. So while at the moment no async is needed, it will be trivial to add in the future. Plus it reduced a ton of duplicate code and made everything simpler.

JuergenReppSIT:
Very good. It's really great that FAPI  policies now can be used in this way.
One remark: you could use digest_idx instead of 0 for the memcpy in Tss2_PolicyCalculate and remove the check
`digest_idx != 0`.
So `Tss2_PolicyCalculate` will also work for policies where the digest is already computed. E.g. produced by `Fapi_ExportPolicy`.

williamcroberts:
> Very good. It's really great that FAPI policies now can be used in this way. One remark: you could use digest_idx instead of 0 for the memcpy in Tss2_PolicyCalculate and remove the check `digest_idx != 0`. So `Tss2_PolicyCalculate` will also work for policies where the digest is already computed. E.g. produced by `Fapi_ExportPolicy`.

I still don't understand any of that. IIUC export generates a list of policy hashes based on the configuration of all valid hash algorithms FAPI is configured for. However, in this case, their is no FAPI configuration. I'm still not sure how you get a list of digests to populate given a policy JSON file.

JuergenReppSIT:
> I still don't understand any of that. IIUC export generates a list of policy hashes based on the configuration of all valid hash algorithms FAPI is configured for. However, in this case, their is no FAPI configuration. I'm still not sure how you get a list of digests to populate given a policy JSON file.

E.g. in https://github.com/tpm2-software/tpm2-tss/blob/59ccdb797c681aef45d1dfdf598330dfe5b31266/test/integration/fapi-export-policy.int.c the results of Fapi_ExportPolicy with more than one digest are checked. A policy created with Fapi_ExportPolicy could also be used as input. In this case the digest appropriate to the parameter hash_alg would be returned without calculation of the digest by `ifapi_calculate_tree_ex`. But normally digest_idx will be set to 0 by `ifapi_calculate_tree_ex`.

williamcroberts:
> > I still don't understand any of that. IIUC export generates a list of policy hashes based on the configuration of all valid hash algorithms FAPI is configured for. However, in this case, their is no FAPI configuration. I'm still not sure how you get a list of digests to populate given a policy JSON file.
> 
> E.g. in https://github.com/tpm2-software/tpm2-tss/blob/59ccdb797c681aef45d1dfdf598330dfe5b31266/test/integration/fapi-export-policy.int.c the results of Fapi_ExportPolicy with more than one digest are checked. A policy created with Fapi_ExportPolicy could also be used as input. In this case the digest appropriate to the parameter hash_alg would be returned without calculation of the digest by `ifapi_calculate_tree_ex`. But normally digest_idx will be set to 0 by `ifapi_calculate_tree_ex`.

I see, a policy can contain it's own policyDigests field for N hash algorithms. Is it guarenteed, that if I ask for SHA256 hash, whether it's precomputed or not, that that value is digest_idx 0?

williamcroberts:
> > I see, a policy can contain it's own policyDigests field for N hash algorithms. Is it guarenteed, that if I ask for SHA256 hash, whether it's precomputed or not, that that value is digest_idx 0?
> 
> The order of the digest list depends on the profiles used for Fapi_ExportPolicy. So it's not guaranteed that digest_idx 0 is used for SHA256. But if no precomputed digest exists always digest_idx will be 0 independent of the hash_alg.

So how do I guarantee that I get the digest for SHA256? Their is no hash alg in the list, just sizes and with SMx coming into play, I can't gaurantee that 32 bytes is SHA256.

williamcroberts:
@JuergenReppSIT a few more questions.

The callback cbpcr, for reading a PCR, I don't get it. What do all the parameters do?
```C
TSS2_RC policy_cb_pcr (
    TPMS_PCR_SELECT *pcrSelect,
    TPML_PCR_SELECTION *pcrBankSelect,
    TPML_PCRVALUES **pcrs,
    void *userdata)
```
Should we consider a simpler interface for PCR read that just does what the TPM does?

I also think I am missing the exec callbacks of ifapi_policyeval_EXEC_CB, is that right? If so, perhaps calculate and execute should take their own callback structure pointers as inputs.

JuergenReppSIT:
> 



> The callback cbpcr, for reading a PCR, I don't get it. What do all the parameters do?
> 
> ```c
> TSS2_RC policy_cb_pcr (
>     TPMS_PCR_SELECT *pcrSelect,
>     TPML_PCR_SELECTION *pcrBankSelect,
>     TPML_PCRVALUES **pcrs,
>     void *userdata)
> ```
> 
> Should we consider a simpler interface for PCR read that just does what the TPM does?

yes it a simple callback function could be implemented which just calls Esys_PCR_Read with the parameter prcBankSelect and
the esys context could be used as userdata.
The function should return BAD_VALUE if pcrSelect is initialized.
> 
> I also think I am missing the exec callbacks of ifapi_policyeval_EXEC_CB, is that right? If so, perhaps calculate and execute should take their own callback structure pointers as inputs.

Sorry, I forgot that for the policy execution the also the "policy util" functions are need.
Currently the exec callbacks are initialized in the function new_policy called by ifapi_policyutil_execute_prepare. I think for the functions  ifapi_policyutil_execute_prepare and ifapi_policy_util_execute also some refactoring is needed.

williamcroberts:
> > 
> 
> > The callback cbpcr, for reading a PCR, I don't get it. What do all the parameters do?
> > ```c
> > TSS2_RC policy_cb_pcr (
> >     TPMS_PCR_SELECT *pcrSelect,
> >     TPML_PCR_SELECTION *pcrBankSelect,
> >     TPML_PCRVALUES **pcrs,
> >     void *userdata)
> > ```
> > 
> > Should we consider a simpler interface for PCR read that just does what the TPM does?
> 
> yes it a simple callback function could be implemented which just calls Esys_PCR_Read with the parameter prcBankSelect and the esys context could be used as userdata. The function should return BAD_VALUE if pcrSelect is initialized.

I am starting to grok this a little better, essentially a policy can have 3 keywords to define PCR values:
"pcrs"
"currentPCRs"
"currentPCRandBanks"

Both JSON ifapi_json_TPMS_POLICYPCR serialize and deserialze routines for the type check that only one condition
may be satisfied.

To me, this implies that the callback could:
1. Take those as a union with a selector. This is checked at serialize and deserialize.
2. That BAD_VALUE return is dead code, as this condition is impossible as its checked at serialize and deserialize.
3. The caller could populate the data into a TPML_PCRVALUES structure.

williamcroberts:
<snip>
> To me, this implies that the callback could:
<snip>
> 3. The caller could populate the data into a TPML_PCRVALUES structure. The caller only needs to return a TPML_PCR_SELECTION.

It should have said the callee would just return... but it should probably be a TPML_DIGEST and let the caller construct the higher lever data structures.

JuergenReppSIT:
> 2\. That BAD_VALUE return is dead code, as this condition is impossible as its checked at serialize and deserialize.

williamcroberts:
> > 2. That BAD_VALUE return is dead code, as this condition is impossible as its checked at serialize and deserialize.
> 
> IMHO this refactoring is unnecessary. I would prefer the solution with returning BAD_VALUE if the bank is not specified. 

These are separate issues here:
1. The existing BAD_VALUE return as the code stands now is dead code. Only one selection is ever present due to the checks on serialize and deserialize.
2. To make this into a more consumable public API, rather than passing both selections and having the caller figure out which one to use by checking count or selection size, put in a union with a selector so they know which one to use.
3. The conversion from TPML_PCR_SELECTION and TPML_DIGESTS, ie from a PCR read call, can be handled by the caller.

> The intention of using currentPCRs was to use a selection appropriate to the FAPI profile. If FAPI is not used I would say it could be expected to specify the bank in the policy e.g.:

The whole point is to take this out of FAPI. **Any policy loaded into FAPI, should be loadable, unmodified, into this
policy library**. I think that **is** the most important feature.

> 
> `"currentPCRandBanks":{ "hash":"sha256", "pcrSelect":[ 16 ] }`
> 
> instead of using:
> 
> ` "currentPCRs":[ 16 ]`
> 
> with an implicit bank default sha256.

Than their is incompat with the policy language for FAPI vs tss2-policy, I don't want that.

JuergenReppSIT:
> The whole point is to take this out of FAPI. **Any policy loaded into FAPI, should be loadable, unmodified, into this policy library**. I think that **is** the most important feature.

So it should at least be possible to adjust also the bank instead of using the default sha256. To really achieve  compatibility and produce the same results as with FAPI for the same json policy it should to be possible to define the pcr selection as used from the FAPI profile. 
If **any** policy has to be loadable, will you also enable access to a FAPI keystore by the tss_policy functions?

williamcroberts:
> > The whole point is to take this out of FAPI. **Any policy loaded into FAPI, should be loadable, unmodified, into this policy library**. I think that **is** the most important feature.
> 
> So it should at least be possible to adjust also the bank instead of using the default sha256. To really achieve compatibility 

That callback is in my test, an application using this library can make their decision based on whatever they want.

and produce the same results as with FAPI for the same json policy it should to be possible to define the pcr selection as used from the FAPI profile. If **any** policy has to be loadable, will you also enable access to a FAPI keystore by the tss_policy functions?

Two things:
1. In the application callback, if they wanted to, they could go and grab the fapi config file if they desired to get that value.
2. Just because the policy loads, doesn't mean it will satisfy between FAPI and application X using tss2-policy. However, it will satisfy between invocations of application X. In the case a raw pcr selection is produced AND one wants FAPI + application to behave together, they need to ensure the same bank is selected.

JuergenReppSIT:
> e results as with FAPI for the same json policy it should to be possible to define the pcr selection as used from the FAPI profile. If **any** policy has to be loadable, will you also enable access to a FAPI keystore by the tss_policy functions?
> 
> Two things:...

I didn't think of policy pcr when I asked this question. I was just curious about your thoughts for handling policies containing FAPI pathnames ....

williamcroberts:
> > e results as with FAPI for the same json policy it should to be possible to define the pcr selection as used from the FAPI profile. If **any** policy has to be loadable, will you also enable access to a FAPI keystore by the tss_policy functions?
> > Two things:...
> 
> I didn't think of policy pcr when I asked this question. I was just curious about your thoughts for handling policies containing FAPI pathnames ....

I really am not sure yet. I started porting the exec callbacks. I noticed through the pol select callback that an IFAPI_OBJECT is needed. I only pass the ESYS_TR in the public callback (internal policy callback to fapi callback resolves the original object and the fapi path). But since this is linked into the FAPI internals, I'm assuming that FAPI ,under the hood, resolves all of this and I have a valid ESYS_TR. The question I need to figure out is how to use a custom fapi config. I think in policy "initialize", if you want to use a FAPI keystore to get resolution, I just need to pass the uri. Somewhere, somehow a FAPI_CONTEXT is getting made :-p

williamcroberts:
> I really am not sure yet. I started porting the exec callbacks. I noticed through the pol select callback that an IFAPI_OBJECT is needed. I only pass the ESYS_TR in the public callback (internal policy callback to fapi callback resolves the original object and the fapi path). But since this is linked into the FAPI internals, I'm assuming that FAPI ,under the hood, resolves all of this and I have a valid ESYS_TR. The question I need to figure out is how to use a custom fapi config. I think in policy "initialize", if you want to use a FAPI keystore to get resolution, I just need to pass the uri. Somewhere, somehow a FAPI_CONTEXT is getting made :-p

Then the question is, we also need this to work when fapi is not provisioned, so I have to walk down those error paths as well and likely add "is it policy lib or fapi" state to those calls to determine what to do.

williamcroberts:
> > e results as with FAPI for the same json policy it should to be possible to define the pcr selection as used from the FAPI profile. If **any** policy has to be loadable, will you also enable access to a FAPI keystore by the tss_policy functions?
> > Two things:...
> 
> I didn't think of policy pcr when I asked this question. I was just curious about your thoughts for handling policies containing FAPI pathnames ....

If you use FAPI pathnames then you need to have the logic to resolve FAPI paths... IE know where the store is.

williamcroberts:
FYI Fuzz target will fail until https://github.com/google/oss-fuzz/pull/7512

src/tss2-policy/tss2_policy.c
```diff
@@ -0,0 +1,464 @@
...
+            &state,
+            &context,
+            &eval_ctx,
+            NULL, /* don't use the fapi policy store *aka pstore */
+            &io,
+            NULL, /* cause it skip loading policy path */
+            &policy_ctx->policy,
+            esys_ctx,
+            policy_ctx->hash_alg,
+            false));
```
JuergenReppSIT, line 245: Similar to the other NULL parameters a comment  /* Session will not be flushed in error cases */ could be added.

src/tss2-fapi/ifapi_policy.h
```diff
@@ -34,14 +34,14 @@ ifapi_compute_policy_digest(
 static inline void ifapi_policy_ctx_init(FAPI_CONTEXT *context) {
 
     /* why are callbacks not set up in the eval_ctx before this?? */
```
JuergenReppSIT, line 36: The eval ctx is part of the fapi context. So I would say it's ok and the comment can be deleted?

test/unit/tss2_policy.c
```diff
@@ -0,0 +1,287 @@
...
+
+#include "tss2_policy.h"
+
+#define LOGMODULE test
+#include "util/log.h"
+
+#include "test/data/test-fapi-policies.h"
+#include "util/aux_util.h"
+
+#define TEST_LAYER        TSS2_RC_LAYER(0)
```
JuergenReppSIT, line 33: Because of `#define TSS2_TPM_RC_LAYER    TSS2_RC_LAYER(0)` there would be equal error codes.
We should use a value not already defined in tss2_common.h.

JuergenReppSIT (changes requested):

src/tss2-fapi/ifapi_policy_callbacks.c
```diff
@@ -578,42 +663,32 @@ ifapi_policyeval_cbauth(
...
     TSS2_RC r;
     FAPI_CONTEXT *fapi_ctx = userdata;
-    size_t i;
-    const char *names[8];
-    IFAPI_OBJECT *auth_object;
     const char* object_path;
 
     return_if_null(fapi_ctx, "Bad user data.", TSS2_FAPI_RC_BAD_REFERENCE);
 
-    if (!fapi_ctx->callbacks.branch) {
```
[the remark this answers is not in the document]
williamcroberts: I didn't I made it more streamlined on Line 681 and used:
```C
return_if_null(fapi_ctx->callbacks.branch, "No branch selection callback", TSS2_FAPI_RC_AUTHORIZATION_UNKNOWN);
```

test/unit/tss2_policy.c
williamcroberts: Whoops I meant to do that.

src/tss2-fapi/ifapi_policy.h
williamcroberts: Yep agreed, done.

src/tss2-fapi/ifapi_policyutil_execute.c
```diff
@@ -288,7 +288,8 @@ ifapi_policyutil_execute(FAPI_CONTEXT *context, ESYS_TR *session)
 
         statecase(pol_util_ctx->state, POLICY_UTIL_EXEC_POLICY);
             r = ifapi_policyeval_execute(context->esys,
-                                         pol_util_ctx->pol_exec_ctx);
+                                         pol_util_ctx->pol_exec_ctx,
+                                         true);
```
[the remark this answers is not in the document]
williamcroberts: I commented the functions with doxygen docs to help clarify this.

src/tss2-fapi/ifapi_policy_execute.c
```diff
@@ -1606,6 +1608,7 @@ execute_policy_element(
  *         is not set.
  * @retval TSS2_FAPI_RC_AUTHORIZATION_FAILED if the authorization attempt fails.
  */
+#include "ifapi_policyutil_execute.h"
```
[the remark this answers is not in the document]
williamcroberts: Yeah just some cruft from working.

src/tss2-fapi/ifapi_policy_callbacks.c
```diff
@@ -277,6 +277,91 @@ ifapi_get_nv_public(
     return r;
 }
 
+TSS2_RC ifapi_read_pcr2(
```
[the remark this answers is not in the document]
williamcroberts: Yep, will kill that dead code and refactor the naming.

williamcroberts:
@JuergenReppSIT sorry for the delay, I got COVID, but I got all your comments resolved hopefully.

JuergenReppSIT:
> @JuergenReppSIT sorry for the delay, I got COVID, but I got all your comments resolved hopefully.

@williamcroberts No problem. I hope you are fine again. Can I merge the PR now?

williamcroberts:
> > @JuergenReppSIT sorry for the delay, I got COVID, but I got all your comments resolved hopefully.
> 
> @williamcroberts No problem. I hope you are fine again. Can I merge the PR now?

I would hold, check your email. Some of the callbacks are not tested and I think they may need a tweak to the interface.

williamcroberts:
> > > @JuergenReppSIT sorry for the delay, I got COVID, but I got all your comments resolved hopefully.
> > 
> > 
> > @williamcroberts No problem. I hope you are fine again. Can I merge the PR now?
> 
> I would hold, check your email. Some of the callbacks are not tested and I think they may need a tweak to the interface.

@JuergenReppSIT We can proceed to merge, I understand the confusion on my part, the API looks good IMO. Ill add tests as we go.

JuergenReppSIT (approved):

PR #2160, opened 2021-09-13
Support SM2 and SM3

JerryDevis:
Add SM2 and SM3 support.

Signed-off-by: JerryDevis <seclab@huawei.com>


JuergenReppSIT:
* To support SM2 in esys also `TPM2_ECC_SM2_P256` (with `NID_sm2`) should be added in case statement in `iesys_cryptossl_get_ecdh_point` where the curve id is mapped to the OSSL NID.
* For determining the hash algorithm also the else case of `#if OPENSSL_VERSION_NUMBER < 0x30000000L` with the second version of `get_ossl_hash_md` should be changed.
* Related to SM2 it also would be nice to take into account migration guide to OSSL3: https://www.openssl.org/docs/manmaster/man7/migration_guide.html

Issue #1808, opened 2020-07-30
FAIL: test/unit/tctildr-dl with musl libc 1.2 on 32bit architectures (x86, armhf)


ncopa:
```
======================================
   tpm2-tss 2.4.1: ./test-suite.log
======================================

# TOTAL: 33
# PASS:  32
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: test/unit/tctildr-dl
==========================

[==========] Running 14 test(s).
[ RUN      ] test_info_from_handle_null
[       OK ] test_info_from_handle_null
[ RUN      ] test_info_from_handle_dlsym_fail
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:67:info_from_handle() Failed to get reference to TSS2_TCTI_INFO_SYMBOL: __wrap_dlerror 
[  ERROR   ] --- %s() has remaining non-returned values.
test/unit/tctildr-dl.c:104: note: remaining item was declared here
__wrap_dlsym: '%s' parameter still has values that haven't been checked.
test/unit/tctildr-dl.c:102: note: remaining item was declared here
'%s' parameter still has values that haven't been checked.
test/unit/tctildr-dl.c:103: note: remaining item was declared here

[  FAILED  ] test_info_from_handle_dlsym_fail
[ RUN      ] test_info_from_handle_success
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:67:info_from_handle() Failed to get reference to TSS2_TCTI_INFO_SYMBOL: __wrap_dlerror 
[  ERROR   ] --- 0 != 0xff8a6368
[   LINE   ] --- test/unit/tctildr-dl.c:122: error: Failure!
[  FAILED  ] test_info_from_handle_success
[ RUN      ] test_handle_from_name_null_handle
[       OK ] test_handle_from_name_null_handle
[ RUN      ] test_handle_from_name_first_dlopen_success
[       OK ] test_handle_from_name_first_dlopen_success
[ RUN      ] test_handle_from_name_second_dlopen_success
[       OK ] test_handle_from_name_second_dlopen_success
[ RUN      ] test_handle_from_name_third_dlopen_success
[       OK ] test_handle_from_name_third_dlopen_success
[ RUN      ] test_fail_null
ERROR:tcti:src/tss2-tcti/tctildr-dl.c:218:tctildr_get_default() tcticontext must not be NULL 
[...]

PR #2167, opened 2021-09-29
test: add wrapper fix for musl on 32 bit systems

whooo:
For some reason wrapping clock_gettime doesn't work with musl on 32 bit systems, but wrapping __clock_gettime64 does.
Tested (via alpine linux aports CI) on x86, x86_64, armv7, armhf, aarch64 and ppc64le.
Also tested on my laptop running debian unstable x86_64, but I haven't been able to test it on 32 bit systems using glibc.

Edit: added the same type of fix for dlsym, with the same level of testing, which fixes https://github.com/tpm2-software/tpm2-tss/issues/1808

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -447,7 +447,8 @@ if ENABLE_TCTI_PCAP
 test_unit_tcti_pcap_CFLAGS  = $(CMOCKA_CFLAGS) $(TESTS_CFLAGS)
 test_unit_tcti_pcap_LDADD   = $(CMOCKA_LIBS) $(libtss2_mu) $(libutil)
 test_unit_tcti_pcap_LDFLAGS = -Wl,--wrap=getenv -Wl,--wrap=rand -Wl,--wrap=clock_gettime \
-        -Wl,--wrap=open -Wl,--wrap=read -Wl,--wrap=write -Wl,--wrap=close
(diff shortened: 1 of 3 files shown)


AndreasFuchsSIT:
Maybe we can get https://github.com/tpm2-software/tpm2-tss-engine/pull/224 ported over to this project as well.
@williamcroberts Could you assist on that topic ?

whooo:
I was able to test it with 32 bit userspace but on x86_64 using glibc and it passes

williamcroberts:
> Maybe we can get [tpm2-software/tpm2-tss-engine#224](https://github.com/tpm2-software/tpm2-tss-engine/pull/224) ported over to this project as well. @williamcroberts Could you assist on that topic ?

Sure. We can just take this now and I'll add a 32 bit container into our build.

williamcroberts (approved):

williamcroberts:
> > Maybe we can get [tpm2-software/tpm2-tss-engine#224](https://github.com/tpm2-software/tpm2-tss-engine/pull/224) ported over to this project as well. @williamcroberts Could you assist on that topic ?
> 
> Sure. We can just take this now and I'll add a 32 bit container into our build.

See: https://github.com/tpm2-software/tpm2-tss/pull/2192

PR #2170, opened 2021-10-07
RFC: Fapi preview fw and ima logging cel

JuergenReppSIT:
FAPI: Add event logging for firmware and IMA events.

 Add serialization and deserialization for firmware events:

 * Serialization and deserialization of firmware events specified in the
   TCG PC Client Platform Firmware Profile was added.
 * Also for the firmare legacy format (sha1) the serialization and deserialization
   was added.

 Add serialization and deserialization for IMA events:

 * Serialization of IMA events as described on: https://sourceforge.net/p/linux-ima/wiki/Home/
   was added. The templates ima, img-sig, ima-ng, and ima-sig are supported.
 * Tests in fapi-jons.c were adapted.

 Integrate firmware and IMA eventlog:

 * The IMA event logging was integrated.
 * The PC-Client firmware logging was integrated.

CEL Events:

* The cel_version event was added as first event for the event log of PCR 0.
* The firmware_end event was added behind the last firmware event.
* Serialization and deserialization of CEL events was added.

Event logging:

* For invalid digests an error will only produced for quote verification.
  In other cases only a warning will be displayed.
* rename "type" to "content_type".
* rename "sub_event" to "content".

Tests for event logging for firmware and IMA events:

* An Unit test for several binary examples was added.
* Tests in fapi-jons.c were adapted.
* Copied functions were removed from unit test fapi-json.c
* An Integration test for The PC-Client firmware logging was added.
* An Integration test for The PC-Client firmware logging in the
  legacy format was added.


configure.ac
```diff
@@ -463,6 +463,7 @@ AS_IF([test "x$enable_defaultflags" = "xyes"],
       AS_IF([test "x$ax_is_release" = "xno"],
           [ADD_COMPILER_FLAG([-Werror])])
       ADD_COMPILER_FLAG([-fstack-protector-all])
+      ADD_COMPILER_FLAG([-Wno-error=address-of-packed-member])
```
AndreasFuchsTPM, line 495: IMHO this should not be turned of globally but locally with a #pragma

williamcroberts: I would ask why? Also pragma's are less portable. Ie if someone has to take the source and add a build system, that's one thing they can carry out of tree easily, but modifying the source to support pragma means carrying an out of tree inter-source file patch.

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+ * param[in] token the token with a potential prefix.
+ * @retval the position of the sub string after the prefix.
+ * @retval 0 if no prefix is found.
+ */
+static int
+get_token_start_idx(const char *token)
+{
+    int itoken = 0;
+    char *entry;
+    int i;
```
williamcroberts, line 67: i and itoken appear to never need to hold negative values, so wouldn't unsigned be a better choice? this way if this code gets attacked an attacker can not do negative offsets.

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+    if (digest == NULL) {
+        LOG_ERROR("digest cannot be NULL");
+        return false;
+    }
+
+    bool ret = true;
+    TSS2_RC r;
+
+    size_t i, j;
+    for (i = 0; i < count; ++i) {
```
williamcroberts, line 115: generally convention is `i++` unless you need that behavior explicitly.

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+        const TPMI_ALG_HASH alg = digest->AlgorithmId;
+        const size_t alg_size = ifapi_hash_get_digest_size(alg);
+        if (size < sizeof(*digest) + alg_size) {
+            LOG_ERROR("insufficient size for digest buffer");
+            return false;
+        }
+
+        uint8_t *pcr = NULL;
+        if (pcr_index > TPM2_MAX_PCRS) {
+            LOG_ERROR("PCR%d > max %d", pcr_index, TPM2_MAX_PCRS);
```
williamcroberts, line 130: is this supposed to keep going and not return false immediately?

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+ * @param[out] out the deserialzed binary object.
+ * @retval TSS2_RC_SUCCESS if the function call was a success.
+ * @retval TSS2_FAPI_RC_BAD_VALUE if the json object can't be deserialized.
+ */
+TSS2_RC
+ifapi_json_TCG_EVENT_TYPE_deserialize_txt(json_object *jso,
+        IFAPI_EVENT_TYPE *out)
+{
+    LOG_TRACE("call");
+    const char *token = json_object_get_string(jso);
```
williamcroberts, line 559: This can return null, granted it requires jso to be NULL if I grok the libjson code properly:
  - https://github.com/json-c/json-c/blob/9749b0cb66b32ddac4bbbca28dd9009968722a9f/json_object.c#L1281

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+    int64_t i64;
+    if (get_number(token, &i64)) {
+        *out = (IFAPI_EVENT_TYPE) i64;
```
williamcroberts, line 562: Why have get_number return an i64 when out is bounded on UINT32? Why not just use strtol and get a uint32_t?

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+        *verify = false;
+    }
+
+    LOG_TRACE("true");
+    return TSS2_RC_SUCCESS;
+
+ error:
+    SAFE_FREE(out->data.buffer);
+    return r;
+
```
williamcroberts, line 738: extra new line to delete

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+
+    char *hex_string = malloc((size) * 2 + 1);
+    return_if_null(hex_string, "Out of memory.", TSS2_FAPI_RC_MEMORY);
+
+    if (size > 0) {
+        for (size_t i = 0, off = 0; i < size; i++, off += 2)
+            sprintf(&hex_string[off], "%02x", buffer[i]);
+    }
+    hex_string[(size) * 2] = '\0';
+    jso_byte_string = json_object_new_string(hex_string);
```
williamcroberts, line 71: insert `SAFE_FREE(hex_string);` here and you can avoid the goto and just return if error.

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+    jso_number = json_object_new_int64(number);
+    return_if_null(jso_number, "Out of memory", TSS2_FAPI_RC_MEMORY);
+
+    json_object_object_add(jso, jso_tag, jso_number);
+    return TSS2_RC_SUCCESS;
+}
+
+static bool
+zero_digest(UINT8 *buffer, size_t size) {
+    size_t i;
```
williamcroberts, line 115: memcmp so we can take advantage of quicker instructions than a byte for byte compare.

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+    return r;
+}
+
+/** Callback for digest of old IMA format.
+ */
+static TSS2_RC
+sha_digest_json_cb(UINT8 *digest, UINT8 * buffer, size_t *offset, json_object *jso,
+                   IFAPI_IMA_TEMPLATE *template) {
+    TSS2_RC r;
+    (void)template;
```
williamcroberts, line 181: use: `src/util/aux_util.h:#define UNUSED(x) (void)(x)`

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+
+/** Callback to get digest with size field (UINT32).
+ */
+static TSS2_RC
+signature_cb(UINT8 *digest, UINT8 *buffer, size_t *offset, json_object *jso,
+             IFAPI_IMA_TEMPLATE *template) {
+    (void) digest;
+    (void) jso;
+    UINT32 digest_size;
+    (void)template;
```
williamcroberts, line 252: UNUSED macro here for all 3 (void) marks.

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+}
+
+/** Callback to get null terminated name with size field (n-ng).
+ */
+static TSS2_RC eventname_ng_json_cb(UINT8 *digest, UINT8 * buffer, size_t *offset, json_object *jso,
+                                    IFAPI_IMA_TEMPLATE *template) {
+    size_t size;
+    UINT32 size_from_buffer;
+    (void)digest;
+    (void)jso;
```
williamcroberts, line 267: UNUSED macro here as well

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+    return TSS2_RC_SUCCESS;
+}
+
+/** Callback to get null terminated name (n).
+ */
+static TSS2_RC eventname_cb(UINT8 *digest, UINT8 * buffer, size_t *offset, json_object *jso,
+                            IFAPI_IMA_TEMPLATE *template) {
+    size_t size;
+    (void)digest;
+    (void)jso;
```
williamcroberts, line 289: Unused macro here as well.

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+    }
+    LOG_TRACE("IMA name: %s", (const char *)&buffer[*offset]);
+    template->name = (char *)&buffer[*offset];
+    *offset += size + 1; /**< with 0 terminator */
+    return TSS2_RC_SUCCESS;
+}
+
+/** Callback to get null terminated name (n).
+ */
+static TSS2_RC eventname_cb(UINT8 *digest, UINT8 * buffer, size_t *offset, json_object *jso,
```
williamcroberts, line 285: I just noticed this now, inconsistent spacing on `*`. Use it tight to the variable everywhere. `UINT8 *buffer`. Here and throughout any others I missed. This seems not explicitly covered in the coding style doc, but its by and large the convention in this code base.

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+{
+    TSS2_RC r;
+    json_object *jso_digest, *jso_digest_type, *jso_ary, *jso_content;
+    char *hash_name;
+
+    *jso = json_object_new_object();
+    return_if_null(*jso, "Out of memory.", TSS2_FAPI_RC_MEMORY);
+
+    r = add_number_to_json(recnum, *jso, "recnum");
+    return_if_error(r, "Add number to json object.");
```
williamcroberts, line 333: is the caller responsible for freeing jso?

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+        case TPM2_SHA256_DIGEST_SIZE:
+            hash_name = "sha256";
+            break;
+        case TPM2_SHA384_DIGEST_SIZE:
+            hash_name = "sha384";
+            break;
+        case TPM2_SHA512_DIGEST_SIZE:
+            hash_name = "sha512";
+            break;
+        default:
```
williamcroberts, line 355: Won't all these assumptions break on SM3? As SM3 IIRC is 256 bits?

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+                            template->event_buffer, &offset, jso, template);
+        *name= template->name;
+        goto_if_error(r, "Get field", error);
+    }
+    if (jso) {
+        r = get_json_content(jso, &jso_content);
+        return_if_error(r, "Get sub event");
+
+        r = add_uint8_ary_to_json(template->event_buffer, template->event_size, jso_content, "template_value");
+        return_if_error(r, "Create data to be hashed");
```
williamcroberts, line 511: shouldn't these 2 return_if_error be goto_if_error to free copy_of_template_format_orig?

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+
+typedef struct {
+    TPMI_ALG_HASH hash_alg;
+    UINT32 read_n;
+    size_t digest_size;
+} digest_info;
+
+digest_info digest_info_tab[4] = {
+    { .hash_alg = TPM2_ALG_SHA1,
+      .read_n = sizeof(UINT32) + TPM2_SHA1_DIGEST_SIZE + sizeof(UINT32) + 3,
```
williamcroberts, line 587: Why? Comment these magic numbers.

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+
+        if (add_event) {
+            char *name;
+            r = convert_ima_event_buffer(&template, jso_current_event, &name);
+            goto_if_error(r, "Create json event.", error);
+        }
+
+        SAFE_FREE(template.event_buffer);
+
+    }
```
williamcroberts, line 717: how come we don't `fclose(fp)` here?

src/tss2-fapi/ifapi_ima_eventlog.c
```diff
@@ -0,0 +1,853 @@
...
+        LOG_ERROR("Field \"template_value\" not found.");
+        return TSS2_FAPI_RC_BAD_VALUE;
+    }
+    r = ifapi_json_UINT8_ARY_deserialize(jso2, &out->template_value);
+    return_if_error(r, "Bad value for field \"template_valuse\".");
+
+    LOG_TRACE("true");
+    return TSS2_RC_SUCCESS;
+}
+
```
williamcroberts, line 825: extra blank newline

src/tss2-fapi/ifapi_json_deserialize.c
```diff
@@ -969,19 +1075,180 @@ ifapi_json_IFAPI_EVENT_deserialize(json_object *jso,  IFAPI_EVENT *out)
...
+            return TSS2_FAPI_RC_BAD_VALUE;
+        }
+        return ifapi_json_UINT64_deserialize(jso2, &out->cel_timestamp);
+
+    default:
+        LOG_TRACE("false");
+        return TSS2_FAPI_RC_BAD_VALUE;
+    };
+}
+
```
williamcroberts, line 1182: extra newline or was that intentional as a visual break between functions for the static array data?

src/tss2-fapi/ifapi_json_serialize.c
```diff
@@ -951,5 +950,146 @@ ifapi_json_IFAPI_CONFIG_serialize(const IFAPI_CONFIG *in, json_object **jso)
...
+    json_object **jso)
+{
+    if (*jso == NULL) {
+        *jso = json_object_new_object();
+        return_if_null(*jso, "Out of memory.", TSS2_FAPI_RC_MEMORY);
+    }
+    TSS2_RC r;
+    json_object *jso2 = NULL;
+    r = ifapi_json_UINT16_serialize(in->major, &jso2);
+    return_if_error(r, "Serialize major version");
```
williamcroberts, line 992: so caller frees jso on error?

src/tss2-fapi/ifapi_json_serialize.c
```diff
@@ -951,5 +950,146 @@ ifapi_json_IFAPI_CONFIG_serialize(const IFAPI_CONFIG *in, json_object **jso)
...
+    };
+    CHECK_IN_LIST(TPMI_CELMGTTYPE, in, CEL_VERSION, FIRMWARE_END);
+    for (size_t i = 0; i < sizeof(jso_tab) / sizeof(jso_tab[0]); i++) {
+        if (jso_tab[i].in == in) {
+            *jso = json_object_new_string(jso_tab[i].name);
+            check_oom(*jso);
+            return TSS2_RC_SUCCESS;
+        }
+    }
+    return_error(TSS2_FAPI_RC_BAD_VALUE, "Undefined constant.");
```
williamcroberts, line 1060: caller frees jso on error?

src/tss2-fapi/ifapi_json_serialize.c
```diff
@@ -951,5 +950,146 @@ ifapi_json_IFAPI_CONFIG_serialize(const IFAPI_CONFIG *in, json_object **jso)
...
+        *jso = json_object_new_object ();
+    jso2 = NULL;
+    jso2 = NULL;
+    r = ifapi_json_TPMI_CELMGTTYPE_serialize(in->type, &jso2);
+    return_if_error(r, "Serialize TPMI_CELMGTTYPE");
+
+    json_object_object_add(*jso, "type", jso2);
+    jso2 = NULL;
+    r = ifapi_json_TPMU_CELMGT_serialize(&in->data, in->type, &jso2);
+    return_if_error(r,"Serialize TPMU_CELMGT");
```
williamcroberts, line 1089: caller frees jso on error?

williamcroberts (changes requested):
i'd like to see fuzz tests to catch issues with malicious log data as inputs.

src/tss2-fapi/ifapi_eventlog_system.c
```diff
@@ -0,0 +1,766 @@
...
+static bool
+get_number(const char *token, int64_t *num)
+{
+    int itoken = 0;
+    int pos = 0;
+    if (strncmp(token, "0x", 2) == 0) {
+        itoken = 2;
+        sscanf(&token[itoken], "%"PRIx64"%n", num, &pos);
+    } else {
+        sscanf(&token[itoken], "%"PRId64"%n", num, &pos);
```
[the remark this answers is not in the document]
JuergenReppSIT: No there is no special reason. if then would still be needed because of the second parameter of sscanf. And I would say the error handling of strtoll is more complicated.

src/tss2-fapi/ifapi_eventlog_system.c
JuergenReppSIT: I will add return false.

src/tss2-fapi/ifapi_eventlog_system.c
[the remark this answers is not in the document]
williamcroberts: > No there is no special reason. if then would still be needed because of the second parameter of sscanf. And I would say the error handling of strtoll is more complicated.

Yeah thats fine, I agree it would just be the if starts with 0x then increment itoken to hold 2 and then call scanf.

src/tss2-fapi/ifapi_eventlog_system.c
JuergenReppSIT: I did always use int64_t because we did serialize unit32 with  json_object_new_int64.

src/tss2-fapi/ifapi_ima_eventlog.c
JuergenReppSIT: sm3 is not possible for the template-hash it should be sha1. The function is prepared for cases where e.g. sha256 will be allowed.

JuergenReppSIT:
> i'd like to see fuzz tests to catch issues with malicious log data as inputs.
@williamcroberts The afl fuzzer revealed some bugs in the IMA part. I will test also the system event part and try to add the afl fuzzing to the ci.

AndreasFuchsTPM:
@williamcroberts @JuergenReppSIT Let's have the discussion here:
Should we roll a new major release (4.0) because we change the CEL format or do we consider the old format was still beta anyways and not much used and just roll 3.3 after this one ?

JuergenReppSIT:
> @williamcroberts @JuergenReppSIT Let's have the discussion here: Should we roll a new major release (4.0) because we change the CEL format or do we consider the old format was still beta anyways and not much used and just roll 3.3 after this one ?

@williamcroberts @AndreasFuchsTPM I think the new CEL format can be part of 3.3 because the old format was beta without IMA and without system events. An old log file could be converted to the new format with a simple sed script if really needed.

williamcroberts:
> > @williamcroberts @JuergenReppSIT Let's have the discussion here: Should we roll a new major release (4.0) because we change the CEL format or do we consider the old format was still beta anyways and not much used and just roll 3.3 after this one ?
> 
> @williamcroberts @AndreasFuchsTPM I think the new CEL format can be part of 3.3 because the old format was beta without IMA and without system events. An old log file could be converted to the new format with a simple sed script if really needed.

Yeah this was what @JuergenReppSIT and I discussed before and we landed on this approach. If we can prevent a major release I am for it.

JuergenReppSIT:
@williamcroberts @AndreasFuchsTPM AFL fuzzing has now been running for more than 5 days on 20 CPUs without finding more errors. So could the PR now be merged for 3.3?
@williamcroberts I could not get into the maintainer sync today. Was it canceled?

williamcroberts:
@JuergenReppSIT 
- [ ] Make sure all added, non-generated, files have the BSD-2-Clause SPDX identifier
- [ ] for the bin files, could we make a build rule that generates them from b64 encoded to avoid sticking a bunch of binary blobs in the repo?
- [ ] Fix the conflict in Makefile.am

Then I think we can merge this, right?

JuergenReppSIT:
> @JuergenReppSIT
> 
>     * [x]  Make sure all added, non-generated, files have the BSD-2-Clause SPDX identifier
> 
I have added a new commit to the PR which adds the SPDX identifier to all files without one or a wrong identifier.
>     * [ ]  for the bin files, could we make a build rule that generates them from b64 encoded to avoid sticking a bunch of binary blobs in the repo?
Is this really necessary? The tpm-tools also store binaries under test/integration/fixtures
> 
>     * [x]  Fix the conflict in Makefile.am
> 
> 
> Then I think we can merge this, right?
I would say yes.

JuergenReppSIT:
> We should never be stuffing binary files into git. If we have to frequently update the files it can/may consume a lot of space in the .git db. I guess in this scenario it's unlikely, but I think its best to avoid any issue. We can then just have the Makefile-test.am include them as a dependency and build them using b64decode.

OK I will add it.

williamcroberts:
> > We should never be stuffing binary files into git. If we have to frequently update the files it can/may consume a lot of space in the .git db. I guess in this scenario it's unlikely, but I think its best to avoid any issue. We can then just have the Makefile-test.am include them as a dependency and build them using b64decode.
> 
> OK I will add it.

Thanks, I know it's a pain.

JuergenReppSIT:
> Thanks, I know it's a pain.

I think next Friday will be the day of pain for me ;-)

JuergenReppSIT:
@williamcroberts I have added a commit where the fapi binary test files are created from base64 encoded files.

williamcroberts (approved):

PR #2174, opened 2021-10-09
Support SM2, SM3 and SM4

JerryDevis:
Support SM2, SM3 and SM4. Because TPM2_SM3_256_DIGEST_SIZE is also 32 bytes, hash algorithm cannot be determined by digest size in rsa_verify_signature and ifapi_get_hash_alg_for_size interface. Thus, we add a parameter, i.e. hash algorithm, in the following interfaces:
Fapi_Sign
Fapi_Sign_Async
Fapi_VerifySignature
Fapi_VerifySignature_Async
ifapi_verify_signature
rsa_verify_signature
ifapi_get_sig_scheme
Besides, we have added SM3 and SM4 in libtpms, and we will push our code later.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,6 +213,7 @@ AC_ARG_ENABLE([tcti-libtpms],
             [AS_IF([test "x$enable_tcti_libtpms" = "xyes"],
                    [AC_CHECK_HEADER(libtpms/tpm_library.h, [], [AC_MSG_ERROR([library libtpms missing])])])],
             [AC_CHECK_HEADER(libtpms/tpm_library.h, [enable_tcti_libtpms=yes],
+                                                    [enable_tcti_libtpms=no]
                                                     [AC_MSG_WARN([library libtpms missing])])])
 AM_CONDITIONAL([ENABLE_TCTI_LIBTPMS], [test "x$enable_tcti_libtpms" != xno])
 


williamcroberts:
> @AndreasFuchsSIT hello, the Cirrus CI failed. Maybe we need a maintainer to approve running workflows.

Cirrus has been failing with:
```
configure: Falling back to testing with tcti-mssim.
checking for tpm_server... no
configure: WARNING: Executable tpm_server not found in PATH (fallback)
configure: error: No simulator executable found in PATH for testing TCTI.
cat: /tmp/cirrus-ci-build/tpm2-tss-*/_build/sub/test-suite.log: No such file or directory
```
For some time now, not really sure why it can't find the simulator, ill have to look again.

williamcroberts:
> > I thought FAPI was supposed to use the config file for setting the algorithms for sign and everything so a crypto admin can set things up, not be directly user specified int he application?
> 
> Thanks for the suggestion, but SM3_256 and SHA_256 have the same hash length. ifapi_hash_get_digest_size, rsa_verify_signature, etc. cannot support SM3, do you have any good suggestion to solve this problem?

I think, and I'll let @AndreasFuchsSIT be authoritative here, is that in the config file you would configure a cyrptographic profile:
- https://github.com/tpm2-software/tpm2-tss/blob/master/doc/fapi-config.md

So in the profile_nameL it would be something like `P_ECCP256SM3_256` or `P_ECCP256SM3256`. Then your code just looks at the profile being used. That would be used for signing stuff. Now for verify, I think it just uses the hashalg size to guess, but you could check both things and then fail.

JerryDevis:
> @AndreasFuchsSIT To avoid API changes (new parameter hashAlg) a solution might be:
> 
> * In function `ifapi_get_sig_scheme` use the hashAlg from the keys sig scheme if the keys sig scheme is sm3
> * In function `rsa_verify_signature` currently we make several trials with different padding schemes. So also a trail for sm3 could be added.

However，hashAlg in ‘tpm2 sign’ command of tpm2-tools is also specified，it does’t used hashAlg from the keys sig scheme.

https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_sign.1.md

So do we need to be consistent with tpm2-tools？

JerryDevis:
> @AndreasFuchsSIT To avoid API changes (new parameter hashAlg) a solution might be:
> 
> * In function `ifapi_get_sig_scheme` use the hashAlg from the keys sig scheme if the keys sig scheme is sm3
> * In function `rsa_verify_signature` currently we make several trials with different padding schemes. So also a trail for sm3 could be added.

However, in tpm2-tools, the hashAlg in 'tpm2 sign' command is specified, not the hashAlg from the keys sig scheme.

https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_sign.1.md

so do we need to be consistent with tpm2-tools

JuergenReppSIT:
> In Fapi, we need to specify the hashAlg in sign scheme when create the sign key，right？
yes the scheme from the profile will be stored with the key.

williamcroberts:
> Hi，@williamcroberts , I submitted the code related to the SM algorithm, I found that Cirrue CI does not support SM. I use the following command to search:
> 
> ```
> openssl dgst --list
> openssl enc --list
> ```

So the openssl version on cirrus FreeBSD12-2 that gets fetched is `1.1.1l,1` and if I install it in my FreeBSD 12-2 VM I get 1.1.1h-freebsd with SM3 support, not sure what they did i their env to change all this. The header file has defines for things like NO_SM3_SUPPORT, I wonder if something with CRYPTO_CFLAGS ins't getting propagated to include this...

I would try cat'ing the evp.h file and see what's on the machine and look at the .pc file and see if they define any flags, i'd be curious to see what those look like before figuring out the best solution.

Offhand, my initial thoughts would be to have the configure script check for the functions with AC_CHECK_FUNC and make defines and you can ifdef the code based on that.

JerryDevis:
> > Hi，@williamcroberts , I submitted the code related to the SM algorithm, I found that Cirrue CI does not support SM. I use the following command to search:
> > ```
> > openssl dgst --list
> > openssl enc --list
> > ```
> 
> So the openssl version on cirrus FreeBSD12-2 that gets fetched is `1.1.1l,1` and if I install it in my FreeBSD 12-2 VM I get 1.1.1h-freebsd with SM3 support, not sure what they did i their env to change all this. The header file has defines for things like NO_SM3_SUPPORT, I wonder if something with CRYPTO_CFLAGS ins't getting propagated to include this...
> 
> I would try cat'ing the evp.h file and see what's on the machine and look at the .pc file and see if they define any flags, i'd be curious to see what those look like before figuring out the best solution.
> 
> Offhand, my initial thoughts would be to have the configure script check for the functions with AC_CHECK_FUNC and make defines and you can ifdef the code based on that.

Hi, @williamcroberts. I have added detection in the configure.ac in the newest PR. If  the cryptographic library is openssl , it will check whether there is an EVP_sm3 interface in libcrypto.so. If there is, define HAVE_EVP_SM3. Maybe it is a solution.

JuergenReppSIT:
@JerryDevis To achieve backward compatibility perhaps we should use the hash alg from the signing scheme only for digest size 32.
In all other cases the hash alg should be determined by `ifapi_get_hash_alg_for_size` as before.
So no changes for the integration tests would be needed.

PR #2181, opened 2021-10-20
fix: autoconf undefined LT_LIB_DLLOAD err

aaronUniontech:


williamcroberts:
Interesting, I thought this macro was autoconf-archive, but it's not:
https://packages.debian.org/sid/amd64/libltdl-dev/filelist

Its in:
/usr/share/aclocal/ltdl.m4

AndreasFuchsSIT (approved):

PR #2185, opened 2021-10-27
Build: remove all useless vc project filters files

uudiin:


williamcroberts (approved):

williamcroberts (commented):
Why are these useless?

uudiin:
> Why are these useless?

These are the temporary files of the vs project, which are automatically generated, and do not need to be edited manually. Maybe the contents of the files in different vs versions are different. These files are not suitable for being placed in the git repository.

PR #2192, opened 2021-11-11
ci: add runner for 32 bit builds

williamcroberts:


williamcroberts:
I think building tpm2-tss in the emulated env takes too long and it times out:

dead-ci

tpm2-abrmd uses these, but it disables ESAPI and FAPI... so it has a much simpler and quicker build.

PR #2203, opened 2021-11-18
tcti: support swtpm and mssim running over unix domain sockets

geoffthorpe:
On non-windows systems, passing port=0 to socket_connect() now causes it
to treat the 'hostname' parameter as a filesystem path to a unix domain
socket. For systems that want to support domain sockets using host/port
parameters in TCTI strings can in principle do this, if they don't prevent
the 'host' string from being a path nor prevent 'port' from being zero.
More naturally, swtpm/mssim TCTI strings can specify a 'path' attribute
directly, avoiding 'host' and 'port' altogether.

The convention for data/control pairs of sockets (with TCP) is to use the
same hostname and two consecutive port numbers. This change introduces a
'control' parameter to the socket_connect() API, updates existing callers
to use it, such that the control cases now pass a non-zero 'control'
instead of incrementing the port number caller-side. The increment of the
port number is now done within the API, so that the unix domain socket
case can implement it differently; by appending ".ctrl" to the socket path.

Signed-off-by: Geoff Thorpe <geoffrey@twosigma.com>

### DIFF
--- a/src/tss2-tcti/tcti-mssim.c
+++ b/src/tss2-tcti/tcti-mssim.c
@@ -513,12 +513,17 @@ mssim_kv_callback (const key_value_t *key_value,
     LOG_DEBUG ("key: %s / value: %s\n", key_value->key, key_value->value);
     if (strcmp (key_value->key, "host") == 0) {
         mssim_conf->host = key_value->value;
+        mssim_conf->path = NULL;
         return TSS2_RC_SUCCESS;
     } else if (strcmp (key_value->key, "port") == 0) {
         mssim_conf->port = string_to_port (key_value->value);
         if (mssim_conf->port == 0) {
             return TSS2_TCTI_RC_BAD_VALUE;
         }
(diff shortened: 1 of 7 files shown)


geoffthorpe:
OK, I can see from the CI pipeline failure and a subsequent examination of the unit-tests that my overloading of host/port behavior won't fly with the current assumptions. I'll rework the code and the approach a bit and update the MR once I'm done.

geoffthorpe:
BTW, I got some good feedback from Nico Williams, reproduced the unit tests (and initial errors my patch caused), and have updated the patch a couple of times. It's now passing the unit tests (which CI will hopefully confirm shortly) and it is also passing end-to-end testing in my attestation environment too. In the interests of clarity and accuracy (the description is now slightly out of date), I'll close this PR once CI has done its thing and reopen a new one. Apologies for the noise, but hopefully this will produce a more palatable/reviewable result.

PR #2212, opened 2021-11-24
tcti: support swtpm and mssim running over unix domain sockets

geoffthorpe:
socket_connect() provides TCP, so we add socket_connect_unix() for unix
domain sockets, which takes 'path' rather than 'host'+'port'.

The convention for data/control pairs of sockets with TCP is that they use
a common hostname and consecutive port numbers. This patch adds a
'control' parameter to both API functions, and updates existing callers of
socket_connect() to provide it. Calls to create control sockets no longer
increment the port number caller-side, it is incremented within the
implementation instead. For the socket_connect_unix() case, the control
socket is implemented by appending ".ctrl" to the path.

Signed-off-by: Geoff Thorpe <geoffrey@twosigma.com>

### DIFF
--- a/src/tss2-tcti/tcti-mssim.c
+++ b/src/tss2-tcti/tcti-mssim.c
@@ -513,12 +513,17 @@ mssim_kv_callback (const key_value_t *key_value,
     LOG_DEBUG ("key: %s / value: %s\n", key_value->key, key_value->value);
     if (strcmp (key_value->key, "host") == 0) {
         mssim_conf->host = key_value->value;
+        mssim_conf->path = NULL;
         return TSS2_RC_SUCCESS;
     } else if (strcmp (key_value->key, "port") == 0) {
         mssim_conf->port = string_to_port (key_value->value);
         if (mssim_conf->port == 0) {
             return TSS2_TCTI_RC_BAD_VALUE;
         }
+        return TSS2_RC_SUCCESS;
+    } else if (strcmp (key_value->key, "path") == 0) {
+        mssim_conf->path = key_value->value;
+        mssim_conf->host = NULL;
         return TSS2_RC_SUCCESS;
     } else {
         return TSS2_TCTI_RC_BAD_VALUE;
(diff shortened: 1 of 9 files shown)


geoffthorpe:
This is a repeat of a previously-submitted (and recently-closed) PR, in case you were getting a sense of déjà vu. The original patch got reworked, rendering the patch description out-of-date and the history a little messy. Thanks

geoffthorpe:
BTW, I just updated the patch to fix a white-space issue (two stray tab characters), otherwise no change.

AndreasFuchsSIT:
I like the feature, however we need some kind (unit) testing for the new function.

geoffthorpe:
@AndreasFuchsSIT , fair point - I will see what I can come up with. Thanks

geoffthorpe:
@AndreasFuchsSIT , thanks again for the suggestion. I managed to get the unit tests building and running, and in doing so I noticed a minor oversight in my original patch. This wasn't from running the tests, just from looking at them carefully, but chalk about another win for unit tests anyway. :-) So I've force-pushed an update to my branch, and separated the patch so the functionality and the tests are now separate commits.

BTW I am currently unable to get coverage numbers, but this should certainly be better than before. Though I will have to let you (and/or the CI) confirm whether or not I broke even. Also, I don't have a _WIN32 environment so can't comment on that, but I have done my best to do right by that platform, including putting corresponding ifdef's in the unit tests. (But I seem to have created the first use of "_WIN32" in test/unit/, so if the tests should be *nix-only, please let me know and I'll strip that out.) Thanks!

AndreasFuchsSIT:
The problem now is that you moved all changes to tests into the second commit, even those that just update the old tests with the new parameter, not only the new tests.
So right now, the compilation after the first commit will be failing, which we generally don't want...

I'm afraid, the best way would be to split the second commit into fixes and additions, and to sQuash the fixes back into the first commit...

geoffthorpe:
@AndreasFuchsSIT , ahh. I see. <groan> Thanks for clarifying - and apologies for the hassle. It somehow escaped my mind that I might be breaking bisectability when I separated the patch in two, which is unforgivably dumb of me. I'll fix this and force-push an update. Thanks again.

geoffthorpe:
@AndreasFuchsSIT, I've been away for a week so am just checking in to see if there was any news and it doesn't seem so. The code-coverage stats in this PR are out-of-date, and though I expect that my updates have fixed the code-coverage drop I don't currently have any way to confirm this - are you able to click a button to refresh the stats? And if those concerns are now addressed, is there anything else I can/should do to get these changes acceptable for merge? Thanks in advance.

AndreasFuchsSIT (approved):

PR #2213, opened 2021-11-26
ci: add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which tpm2-tss already runs.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'tpm2-tss'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'tpm2-tss'
+        fuzz-seconds: 600
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure() && steps.build.outcome == 'success'
+      with:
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
I am not too sure if we want this on every PR or just on merges into master... Seems too much to me...
@williamcroberts thoughts ?

williamcroberts:
> I am not too sure if we want this on every PR or just on merges into master... Seems too much to me... @williamcroberts thoughts ?

I'd run this with every PR to help catch regressions that get missed in typical testing. AFAICT it doesn't cost us anything or needlessly slow the CI, so I am in favor of adding it. I actually integrated fuzzing tests into PKCS11 with `--enable-fuzzing` and run it as part of my CI.

The fact that this is integrated with OSS fuzz seems nice, as it only flunks people for new failures, granted we don't have old failures. I like this a ton, although it's not clear to me whos server it's running on (is it triggering an OSS Fuzz server and reporting back, or it it running on GH and just querying OSS Fuzz).

DavidKorczynski:
> The fact that this is integrated with OSS fuzz seems nice, as it only flunks people for new failures, granted we don't have old failures. I like this a ton, although it's not clear to me whos server it's running on (is it triggering an OSS Fuzz server and reporting back, or it it running on GH and just querying OSS Fuzz).

I believe it's running on the OSS-Fuzz server. @jonathanmetzman can you confirm this?

jonathanmetzman:
> > The fact that this is integrated with OSS fuzz seems nice, as it only flunks people for new failures, granted we don't have old failures. I like this a ton, although it's not clear to me whos server it's running on (is it triggering an OSS Fuzz server and reporting back, or it it running on GH and just querying OSS Fuzz).
> 
> I believe it's running on the OSS-Fuzz server. @jonathanmetzman can you confirm this?

CIFuzz fuzzes in your CI, not on OSS-Fuzz.
The only interaction it has with CIFuzz is downloading the (public) corpus and coverage data.

PR #2219, opened 2021-12-01
configure: Check system with busybox.

JuergenReppSIT:
Busybox does not support usseradd and groupadd. So for systems with busybox
(e.g. Alpine Linux) adduser and addgroup have to be used for FAPI configuration.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -636,8 +636,21 @@ define make_parent_dir
 endef
 
 define make_tss_user_and_group
-    (id -g tss 2>/dev/null || groupadd -r tss) && \
-    (id -u tss 2>/dev/null || useradd -r -g tss tss)
+    if type -p groupadd > /dev/null; then \
+        id -g tss 2>/dev/null || groupadd --system tss; \
+    else \
+        id -g tss 2>/dev/null || \
+        addgroup --system tss; \
+    fi && \
+    if type -p useradd > /dev/null; then \
+        id -u tss 2>/dev/null || \
+        useradd --system --home-dir / --shell `type -p nologin` \
+                         --no-create-home -g tss tss; \
+    else \
+        id -u tss 2>/dev/null || \
+        adduser --system --home / --shell `type -p nologin` \
+                --no-create-home --ingroup tss tss; \
+    fi
 endef
 
 define make_tss_dir
--- a/configure.ac
+++ b/configure.ac
@@ -487,15 +487,17 @@ AC_CHECK_PROG(systemd_sysusers, systemd-sysusers, yes)
 AM_CONDITIONAL(SYSD_SYSUSERS, test "x$systemd_sysusers" = "xyes")
 AC_CHECK_PROG(systemd_tmpfiles, systemd-tmpfiles, yes)
 AM_CONDITIONAL(SYSD_TMPFILES, test "x$systemd_tmpfiles" = "xyes")
+
 # Check all tools used by make install
 AS_IF([test "$HOSTOS" = "Linux"],
-      [ERROR_IF_NO_PROG([groupadd])
-       ERROR_IF_NO_PROG([useradd])
-       ERROR_IF_NO_PROG([id])
-       ERROR_IF_NO_PROG([chown])
-       ERROR_IF_NO_PROG([chmod])
-       ERROR_IF_NO_PROG([mkdir])
-       ERROR_IF_NO_PROG([setfacl])])
+    [ AC_CHECK_PROG(useradd, useradd, yes)
+      AC_CHECK_PROG(groupadd, groupadd, yes)
+      AC_CHECK_PROG(adduser, adduser, yes)
+      AC_CHECK_PROG(addgroup, addgroup, yes)
+      AS_IF([test "x$addgroup" != "xyes" && test "x$groupadd" != "xyes" ],
+         [AC_MSG_ERROR([addgroup or groupadd are needed.])])
+      AS_IF([test "x$adduser" != "xyes" && test "x$useradd" != "xyes" ],
+         [AC_MSG_ERROR([adduser or useradd are needed.])])])
 
 AC_SUBST([PATH])
 


AndreasFuchsSIT:
Hmmm.... We should differentiate between the existence of busybox on a system and whether we are currently running within a busybox shell.

Also on my Ubuntu, I have both, adduser and useradd. Maybe we just AC_CHECK_PROG for both and base our decision on their existance.
On busybox, adduser should exist and be a symlink, right ?

configure.ac
```diff
@@ -488,10 +488,17 @@ AM_CONDITIONAL(SYSD_SYSUSERS, test "x$systemd_sysusers" = "xyes")
 AC_CHECK_PROG(systemd_tmpfiles, systemd-tmpfiles, yes)
 AM_CONDITIONAL(SYSD_TMPFILES, test "x$systemd_tmpfiles" = "xyes")
 # Check all tools used by make install
+
+AC_CHECK_PROG(busybox,busybox,yes)
+AM_CONDITIONAL(BUSYBOX, [test x"$busybox" = xyes])
```
williamcroberts, line 493: So on a lot of devices, busybox is sometimes renamed. I think this would be more generic if we look for adduser or addgroup during configure time over checking for busybox. What about in a side-by-size install where they have busybox and groupadd? We could still use groupadd.

JuergenReppSIT: I tried to install groupadd in Alpine but it was not available.  If the name of busybox is often changed we should determine the name of the running shell and compare it with readlink from the path of adduser.

JuergenReppSIT: It's possible to determine the shell on Linux via `/proc/$$/exe`. But perhaps it would be better to use the current method from this PR for systems which do not change the name (works for Alpine) and add a configure option --with-busybox for cases when the name was changed?

williamcroberts: We don't care what shell we're in, so long as its compatible, and we don't care about busybox. We care that we have adduser or useradd and the same for group. We just need one available in each set and you can run. So lets not worry about looking at our shell or looking for busybox, rather lets look for something executable that has what we need. I checked locally to make sure that the autoconf macros pick up symlinked executables, since that's how busybox is installed, and it does seem to work as expected.
```
AC_CHECK_PROG(addgroup, ...)
AC_CHECK_PROG(groupadd, ...)
AC_CHECK_PROG(adduser, ...)
AC_CHECK_PROG(useradd, ...)

if not useradd and not adduser:
   fail("Need adduser or useradd)

if not groupadd and not addgroup:
   fail("Need addgroup or groupadd)
```

JuergenReppSIT: The busybox adduser and e.g. ubuntu adduser have different options. So would you expect?:
```
if not groupadd and adduser:
   use busybox version of adduser for make installl
elsif groupadd
   use groupadd for make install
else 
   fail("Need addgroup or groupadd")
```

Makefile.am
```diff
@@ -636,8 +636,10 @@ define make_parent_dir
 endef
 
 define make_tss_user_and_group
-    (id -g tss 2>/dev/null || groupadd -r tss) && \
-    (id -u tss 2>/dev/null || useradd -r -g tss tss)
+    (if ! type -p groupadd > /dev/null; then (id -g tss 2>/dev/null || addgroup -S tss); fi) && \
+    (if ! type -p useradd > /dev/null; then (id -u tss 2>/dev/null || adduser -S -G tss tss); fi) && \
+    (if type -p groupadd > /dev/null; then (id -g tss 2>/dev/null || groupadd -r tss); fi) && \
+    (if type -p useradd > /dev/null; then (id -u tss 2>/dev/null || useradd -r -g tss tss); fi)
```
williamcroberts, line 642: Read my mind.

williamcroberts (approved):

Makefile.am
williamcroberts: Actually, I misread this. I would do this:
```
# Use the higher level friendly tools over non-friendly useradd/groupadd
if addgroup_defined:
  addgroup
else:
  addgroup linux-way || addgroup busybox-way

if adduser_defined:
  adduser 
else:
  adduser linux-way || adduser busybox-way
```

Extra points to keep track of stderr and only display if both commands fail.

williamcroberts (changes requested):

JuergenReppSIT:
I would prefer to use useradd and groupadd if available.
* Opensuse has no adduser and no addgroup command in our CI image.
* Fedora has adduser but no addgroup command in our CI image.

Or did you mean:
```
if groupadd_defined:
  groupadd
else:
  addgroup linux-way || addgroup busybox-way

if useradd_defined:
  useradd 
else:
  adduser linux-way || adduser busybox-way
```

williamcroberts:
> @williamcroberts I would prefer to use useradd and groupadd if available.
> 
> * Opensuse has no adduser and no addgroup command in our CI image.
> * Fedora has adduser but no addgroup command in our CI image.
> 
> Or did you mean:
> 
> ```
> if groupadd_defined:
>   groupadd
> else:
>   addgroup linux-way || addgroup busybox-way
> 
> if useradd_defined:
>   useradd 
> else:
>   adduser linux-way || adduser busybox-way
> ```

Yeah this way.

williamcroberts:
> > @williamcroberts I would prefer to use useradd and groupadd if available.
> > 
> > * Opensuse has no adduser and no addgroup command in our CI image.
> > * Fedora has adduser but no addgroup command in our CI image.
> > 
> > Or did you mean:
> > ```
> > if groupadd_defined:
> >   groupadd
> > else:
> >   addgroup linux-way || addgroup busybox-way
> > 
> > if useradd_defined:
> >   useradd 
> > else:
> >   adduser linux-way || adduser busybox-way
> > ```
> 
> Yeah this way.

My overall point is:
1. If present prefer the high level tools over the low level tools[1]
2. If using the conflicting tools, try one way and fallback to the other way

[1] Per my Ubuntu manpages adduser and addgroup are friendlier front ends over the low level tools useradd and groupadd.

JuergenReppSIT:
@williamcroberts I have compared the result of `make_tss_user_and_group` with the result of `systemd-sysusers` and changed the options to achieve the same result. This version now works on Fedora, Opensuse, Ubuntu, and Alpine. I hope it's OK now?

williamcroberts (approved):
LGTM

PR #2221, opened 2021-12-02
FAPI: Fix leak in fapi crypto with ossl3

JuergenReppSIT:
A leak in the case "out of memory" detected by scan-build was fixed.


gotthardp (approved):
Yeah, my bad. This is a copy-paste error.

PR #2240, opened 2021-12-16
Fix errors if log is disabled and check availability of sanitizer

JuergenReppSIT:
* esys: Mark variable only used for logging maybe-unused.
* tcti: Mark variables only used for logging maybe-unused.
* ci: Set fetch-depth to 0 fot github actions.
* ci: Check whether sanitizer libs are available.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -63,15 +63,19 @@ if [ "$CC" == "gcc" ]; then
   export CONFIGURE_OPTIONS="$CONFIGURE_OPTIONS --enable-code-coverage";
 fi
 
+if [ ldconfig -p 2>/dev/null| grep libasan > /dev/null && ldconfig -p 2>/dev/null| grep libubsan > /dev/null ]; then
+  SANITIZER_OPTION="--with-sanitizer=undefined,address"
+fi
+
 if [ "$SCANBUILD" == "yes" ]; then
   scan-build --status-bugs ../configure --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
 elif [ "$CC" == "clang" ]; then
   ../configure --enable-unit --enable-self-generated-certificate  --enable-integration --with-maxloglevel=none --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
 else
   if [ "$WITH_TCTI" == "mssim" ]; then
-    ../configure --with-sanitizer=undefined,address --disable-tcti-swtpm --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
+    ../configure $SANITIZER_OPTION --disable-tcti-swtpm --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
   else
-    ../configure --with-sanitizer=undefined,address --with-maxloglevel=none --enable-debug=yes --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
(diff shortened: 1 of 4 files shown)


williamcroberts:
Why do we still care about Alpine Linux, I thought we were ditching that in the CI?

.ci/docker.run
```diff
@@ -69,9 +69,15 @@ elif [ "$CC" == "clang" ]; then
   ../configure --enable-unit --enable-self-generated-certificate  --enable-integration --with-maxloglevel=none --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
 else
   if [ "$WITH_TCTI" == "mssim" ]; then
-    ../configure --with-sanitizer=undefined,address --disable-tcti-swtpm --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
+    if [ ldconfig -p 2>/dev/null| grep libasan > /dev/null && ldconfig -p 2>/dev/null| grep libubsan > /dev/null ]; then
+        SANITIZER_OPTION="--with-sanitizer=undefined,address"
+    fi
```
williamcroberts, line 74: Why not move this block up to something like line 65 and just set the variable there rather than repeating the code block to set it and then just keep the references to the option in the configure line?

JuergenReppSIT: @williamcroberts If we will include an alpine Image we cannot run these two tests with the sanitizer enabled. Currently the option sanitizer can be used even when the two libs are not available.

williamcroberts: Yeah I get that, but why duplicate the code to set the variable. This patch should be functionally equivelent:

```diff
diff --git a/.ci/docker.run b/.ci/docker.run
index 9d6914d3bc5f..e835b584298d 100755
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -63,20 +63,18 @@ if [ "$CC" == "gcc" ]; then
   export CONFIGURE_OPTIONS="$CONFIGURE_OPTIONS --enable-code-coverage";
 fi
 
+if [ ldconfig -p 2>/dev/null| grep libasan > /dev/null && ldconfig -p 2>/dev/null| grep libubsan > /dev/null ]; then
+    SANITIZER_OPTION="--with-sanitizer=undefined,address"
+fi
+
 if [ "$SCANBUILD" == "yes" ]; then
   scan-build --status-bugs ../configure --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
 elif [ "$CC" == "clang" ]; then
   ../configure --enable-unit --enable-self-generated-certificate  --enable-integration --with-maxloglevel=none --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
 else
   if [ "$WITH_TCTI" == "mssim" ]; then
-    if [ ldconfig -p 2>/dev/null| grep libasan > /dev/null && ldconfig -p 2>/dev/null| grep libubsan > /dev/null ]; then
-        SANITIZER_OPTION="--with-sanitizer=undefined,address"
-    fi
     ../configure $SANITIZER_OPTION --disable-tcti-swtpm --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
   else
-    if [ ldconfig -p 2>/dev/null| grep libasan > /dev/null && ldconfig -p 2>/dev/null| grep libubsan > /dev/null ]; then
-      SANITIZER_OPTION="--with-sanitizer=undefined,address"
-    fi
     ../configure $SANITIZER_OPTION --with-maxloglevel=none --enable-debug=yes --enable-unit --enable-self-generated-certificate --enable-integration --with-crypto=$WITH_CRYPTO $CONFIGURE_OPTIONS
   fi
 fi
```

JuergenReppSIT:
@williamcroberts That's true. I will change it.

AndreasFuchsSIT:
> Why do we still care about Alpine Linux, I thought we were ditching that in the CI?

AFAIR we ditch Alpine for OpenSSL3 (using Ubuntu with a custom OpenSSL compile there). But supporting Alpine in general would be nice, since they package us nicely and it's used often in embedded and container applications.

AndreasFuchsSIT (approved):

PR #2247, opened 2021-12-21
FAPI: Free certificate when failed 3.0.x

JuergenReppSIT:
In fapi-key-create-sign.int.c, when test_fapi_key_create_sign and test_fapi_key_create_sign fails, it will go to “error” to exit. the certifacate needs to free.

Signed-off-by: JerryDevis <JerryDevis@users.noreply.github.com>

### DIFF
--- a/test/integration/fapi-key-create-sign.int.c
+++ b/test/integration/fapi-key-create-sign.int.c
@@ -250,6 +250,7 @@ error:
     SAFE_FREE(policy);
     SAFE_FREE(publicKey);
     SAFE_FREE(signature);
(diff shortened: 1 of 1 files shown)


AndreasFuchsSIT:
This commit includes a complete new test. Was that intended ?

JuergenReppSIT:
> This commit includes a complete new test. Was that intended ?

oh sorry no. 3.0.x did not include this test I will remove the test from the PR.

AndreasFuchsSIT (approved):

PR #2254, opened 2021-12-27
FAPI: Fix the authorization of hierarchy objects.

### LINKED ISSUE #2253: FAPI crashes when creating key under the Edorsement Key.
I want to create an AIK under the EK. I tried using tss2_createKey and the FAPI function`Fapi_CreateKey`. both work for creating a key under the SRK or in the Endorsement hirarchy but when i supply the path to the EK I get a segmentation fault:
```bash
$ tss2_provision
$ tss2_createkey -p /SRK/AK -t sign -a ""
$ tss2_createkey -p /EK/AK -t sign -a ""
Segmentation fault (core dumped)
````
My C code using FAPI has the same results therefore i post this here and not in tools.
I am using a docker image based on ubuntu 18.04  which builds tss from source. It uses the ibmswtpm version 1332 which does not come with an EK certificate therefore i disabled the EK certificate check in the fapi config. Is this a bug or am I missing something?

### PR DESCRIPTION
**Pull request by JuergenReppSIT**
The default policy "policy secret" for the endorsement key is defined for the
endorsement hierarchy. The corresponding object had to be searched in the
keystore during policy execution. This search did not work, due to various
bugs.
* A erroneous cleanup of the EK was executed where the policy was deleted.
* The name needed for the search in the keystore was not stored in the
  hierarchy object.
* The esys handle stored in the hierarchy object was not correct.

To ensure compatibility with existing system keystores the initialization
of the hierarchy objects was fixed as follows:

* The initialization of the hierarchy object was moved to to object
  deserialization.
* Depending on the pathname the esys handle in the hierarchy initialized.
* The name of the hierarchy object was computed.
* The object needed for authorization during policy execution was initialized
  with the hierarchy object.
* The searching of objects in the keystore was fixed.
* Objects to be freed in Fapi_Import were initialzed with NULL.

The tests were adapted:
* The linking of the unit test for FAPI deserialization had to be adapted.
* An integration test where a key in the endorsement hierarchy was used
  for signing was added.
* PCR 16 was not reset int all tests expecting pcr 16 equal 0.

Fixes #2253

### DIFF
--- a/src/tss2-fapi/ifapi_helpers.c
+++ b/src/tss2-fapi/ifapi_helpers.c
@@ -922,12 +922,51 @@ append_object_to_list(void *object, NODE_OBJECT_T **object_list)
     return TSS2_RC_SUCCESS;
 }
 
+/**  Compute the name of a hierarchy object.
+ *
+ * The TPM handle will be computed from the esys handle and the name
+ * will be computed from the TPM handle.
+ *
+ * @param[in,out] hierarchy The hierarchy object.
+ */
+static void
+set_name_hierarchy_object(IFAPI_OBJECT *object)
+{
+    TPM2_HANDLE handle = 0;
+    size_t offset = 0;
+    switch (object->handle) {
+    case ESYS_TR_RH_NULL:
+        handle = TPM2_RH_NULL;
+        break;
+    case ESYS_TR_RH_OWNER:
+        handle = TPM2_RH_OWNER;
+        break;
+    case ESYS_TR_RH_ENDORSEMENT:
+        handle = TPM2_RH_ENDORSEMENT;
+        break;
+    case ESYS_TR_RH_LOCKOUT:
+        handle = TPM2_RH_LOCKOUT;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/tss2-fapi/ifapi_helpers.c [hunk 1]
```diff
@@ -922,12 +922,51 @@ append_object_to_list(void *object, NODE_OBJECT_T **object_list)
     return TSS2_RC_SUCCESS;
 }
 
+/**  Compute the name of a hierarchy object.
+ *
+ * The TPM will be computed from the esys handle and the name
```
AndreasFuchsSIT, line 927: This sentence appears werid

AndreasFuchsSIT (approved):
Looks good, just a comment about the function description...

src/tss2-fapi/ifapi_helpers.c
JuergenReppSIT: @AndreasFuchsSIT Indeed, I have changed it.

PR #2288, opened 2022-01-13
ossl: Fix check whether SM3 is available

### LINKED ISSUE #2287: EVP_sm3 not found even if HAVE_EVP_SM3
There is a little inconsistency in the SM3 related checks in `configure.ac`: first the `PKG_CHECK_MODULES(CRYPTO)` finds an openssl library desired by the user and then `AC_CHECK_LIB(crypto,[EVP_sm3])` searches for an openssl library with the `EVP_sm3` symbol, which can possibly be a different library than the first one. (Note the AC_CHECK_LIB does not take the discovered `pkg-config` settings.)
As a result, one can `HAVE_EVP_SM3` (because there is some openssl in your system with SM3 enabled), but get a build error (because SM3 is not enabled in the library used during build).

The openssl has `OPENSSL_NO_SM3`, which is used also in the header files and thus available to applications. One can use it in combinarion with `HAVE_EVP_SM3`, e.g. like this:
```
#if HAVE_EVP_SM3 && !defined(OPENSSL_NO_SM3)
...
#endif
```

What do you think?

JuergenReppSIT:
Use also OPENSSL_NO_SM3 to check whether SM3 is available and the correct library is used.
Fixes #2287.

### DIFF
--- a/src/tss2-esys/esys_crypto_ossl.c
+++ b/src/tss2-esys/esys_crypto_ossl.c
@@ -123,7 +123,7 @@ get_ossl_hash_md(TPM2_ALG_ID hashAlg)
         return EVP_sha384();
     case TPM2_ALG_SHA512:
         return EVP_sha512();
-#if HAVE_EVP_SM3
+#if HAVE_EVP_SM3 && !defined(OPENSSL_NO_SM3)
     case TPM2_ALG_SM3_256:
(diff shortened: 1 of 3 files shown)


gotthardp (approved):
It works. Thank you!

PR #2297, opened 2022-02-01
FAPI: Add an integration test for a high range EK policy.

JuergenReppSIT:
* TCG EK Credential Profile TPM 2.0, Version 2.4 Rev. 3, 2021 defines policies
  for the low and high address range of EK templates.
* A policy for a high range template with sha256 was added to the test policies
  and The policy digests for policy A, B, and C defined in the profile were computed
  in the integration test for Fapi_PolicyExport.
* This policy could be used in a FAPI profile as "ek_policy".

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -600,7 +600,8 @@ EXTRA_DIST +=  \
     test/data/fapi/policy/pol_pcr16_read.json \
     test/data/fapi/policy/pol_action.json \
     test/data/fapi/policy/pol_cphash.json \
-    test/data/fapi/policy/pol_or_read_write_secret.json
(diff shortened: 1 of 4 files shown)


williamcroberts (approved):

williamcroberts:
I'm rebasing this and handling the conflicts, will update soon, let the CI churn on it, and merge it.

test/integration/fapi-export-policy.int.c
```diff
@@ -29,6 +29,7 @@
 #include "util/log.h"
 #include "util/aux_util.h"
 
+
```
AndreasFuchsTPM, line 32: why the newline ?

williamcroberts (approved):

PR #2306, opened 2022-02-16
FAPI: Fix usage of curl url.

JuergenReppSIT:
Errors did occur for urls which where not escaped since curl version 7.80. Now the escaping
is executed before executing the download.

### DIFF
--- a/src/tss2-fapi/ifapi_helpers.c
+++ b/src/tss2-fapi/ifapi_helpers.c
@@ -2504,6 +2504,9 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
                           size_t *buffer_size) {
     int ret = -1;
     struct CurlBufferStruct curl_buffer = { .size = 0, .buffer = NULL };
+#ifdef CURLU_ALLOW_SPACE
+    CURLU *urlp = NULL;
+#endif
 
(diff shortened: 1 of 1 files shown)


williamcroberts (approved):

src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2517,7 +2517,19 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
         goto out_global_cleanup;
     }
 
+#ifdef CURLU_ALLOW_SPACE
+    CURLU *urlp = curl_url();
+    rc = curl_url_set(urlp, CURLUPART_URL, url, CURLU_ALLOW_SPACE | CURLU_URLENCODE);
+    if (rc != CURLE_OK) {
+        LOG_ERROR("curl_url_set for CURUPART_URL failed: %s",
+                  curl_easy_strerror(rc));
```
williamcroberts, line 2525: If this fails, we need a `curl_url_cleanup(urlp)` or AFAICT we will leak as we never transferred ownership to the curl pointer. Their is a chance we hit this case when curl_url() returns a NULL pointer, which curl_url_cleanup is safe to call with a NULL pointer.

williamcroberts (changes requested):

AndreasFuchsSIT (approved):

PR #2321, opened 2022-03-07
tcti: Use heap instead of stack when tcti initialize

JerryDevis:
Signed-off-by: JerryDevis <89641128+JerryDevis@users.noreply.github.com>

### DIFF
--- a/src/tss2-tcti/tctildr-dl.c
+++ b/src/tss2-tcti/tctildr-dl.c
@@ -79,7 +79,7 @@ TSS2_RC
 handle_from_name(const char *file,
                  void **handle)
 {
-    char file_xfrm [PATH_MAX] = { 0, };
+    char *file_xfrm = NULL;
     size_t size;
 
     if (handle == NULL) {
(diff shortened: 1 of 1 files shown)


src/tss2-tcti/tctildr-dl.c
```diff
@@ -91,36 +91,45 @@ handle_from_name(const char *file,
...
+    file_xfrm = calloc(PATH_MAX, sizeof(char));
+    if (file_xfrm == NULL) {
+        return TSS2_TCTI_RC_MEMORY;
+    }
     /* 'name' alone didn't work, try libtss2-tcti-<name>.so.0 */
     size = snprintf(file_xfrm,
-                    sizeof (file_xfrm),
+                    PATH_MAX,
                     TCTI_NAME_TEMPLATE_0,
                     file);
```
williamcroberts, line 103: To be less greedy on memory why not run snprintf twice:
1. With a NULL buffer and 0 buffer size to get needed size
2. bound check it to path max
3. calloc the buffer
4. call snprintf with buffer.

williamcroberts (changes requested):

src/tss2-tcti/tctildr-dl.c
JerryDevis: > To be less greedy on memory why not run snprintf twice:
> 
> 1. With a NULL buffer and 0 buffer size to get needed size
> 2. bound check it to path max
> 3. calloc the buffer
> 4. call snprintf with buffer.

Thanks for your review. I've revised the code according to your comments.

williamcroberts (approved):

PR #2351, opened 2022-04-19
Fixing an issue when Null deref occures in Esys_TR_SetAuth

Abdulmajid-A:
When the provided value of  esys_handle is ESYS_TR_NONE the call to esys_GetResourceObject will set esys_object to NULL, which get dereferenced below.
We mitigate the issue by returning TSS2_ESYS_RC_BAD_TR when the handle value is ESYS_TR_NONE


JuergenReppSIT (approved):
@Abdulmajid-A  Thank you for the fix

PR #2354, opened 2022-04-19
Makefile: Fix usage of DESTDIR for make install

### LINKED ISSUE #2344: makefile: install-dirs not behaving correctly
I'm packaging tpm2-tss and ran into a couple of things where the buildsystem isn't behaving as I would expect:

1. When building with systemd-sysusers (i.e. `SYSD_SYSUSERS` is set in the Makefile), the check for `adduser`, `addgroup`, etc ([here](https://github.com/tpm2-software/tpm2-tss/blob/master/configure.ac#L491-L500)) is unneeded because `make install` will never use these commands. 
2. When `DESTDIR` is set, `systemd-sysusers` and `systemd-tmpfiles` **should not** be executed! This probably also applies to the `make_tss_user_and_group` helper also. When `DESTDIR` is set, it doesn't make much sense to be creating users in the build environment. And `systemd-tmpfiles` will not handle `DESTDIR` for you either, meaning it will install the configuration specified into the build environment. In many build environments, this prevents the build from functioning because everything outside of `DESTDIR` and the checked-out sources is read-only. Please, leave executing these tools to the distributor
3. The call to `systemd-sysusers` and `systemd-tmpfiles` does not pass the right path in. If configured with `--with-sysusersdir` and `--with-tmpfilesdir`, the `install-dirs` target ignores these options and instead tries to execute sysusers/tmpfiles with the default paths in `/etc`, leading to "file not found" errors. In fact, this is why I can still build this package despite point 2

JuergenReppSIT:
* If DESTDIR is defined to create a staging installation  systemd-sysusers and systemd-tmpfiles
  should not be called.
* The directories defined by --with-sysusersdir and --with-tmpfilesdir have to be used for the
  corresponding systemd commands.
* If DESTDIR is provided and systemd-tmpfiles is available the eventlog and keystore directory
   will be created under DESTDIR but the owner will not be changed.

Fixes #2344

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -671,20 +671,22 @@ define make_parent_dir
 endef
 
 define make_tss_user_and_group
-    if type -p groupadd > /dev/null; then \
-        id -g tss 2>/dev/null || groupadd --system tss; \
-    else \
-        id -g tss 2>/dev/null || \
-        addgroup --system tss; \
-    fi && \
-    if type -p useradd > /dev/null; then \
-        id -u tss 2>/dev/null || \
-        useradd --system --home-dir / --shell `type -p nologin` \
-                         --no-create-home -g tss tss; \
-    else \
-        id -u tss 2>/dev/null || \
-        adduser --system --home / --shell `type -p nologin` \
-                --no-create-home --ingroup tss tss; \
+    if test -z "${DESTDIR}"; then \
+        if type -p groupadd > /dev/null; then \
+            id -g tss 2>/dev/null || groupadd --system tss; \
+        else \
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -631,25 +631,28 @@ endif #DOXYMAN
 endif #FAPI
 
 ### Helper Functions ###
+
```
williamcroberts, line 634: superfluous change, drop?

williamcroberts (changes requested):

williamcroberts (approved):

Issue #2360, opened 2022-05-20
CertifyCreation for Fapi_CreateKey


isergieienkov:
Allow CertifyCreation for keys created with Fapi. Currently There is no way to get creation hash and creation ticket for keys created with Fapi.

AndreasFuchsTPM:
Should be fixed by #2362

PR #2362, opened 2022-05-21
FAPI: Add CreationHash to keystore

JuergenReppSIT:
Currently only CreationData and CreationHash were stored in the FAPI keystore.
Creation hash is now added to provide the possibility for the tpm2 tool command
tss2_gettpm2object to extract also CreationData, CreationTicket, and CreationHash.
Also the initialization of cretionTicket during deserialization was fixed.
Addresses #2360


AndreasFuchsTPM (changes requested):
The commit message should read "Currently only CreationData and CreationTicket"... and not "Currently only CreationData and CreationHash"

williamcroberts (approved):

PR #2365, opened 2022-06-03
esapi: support runtime switchable crypto backends

williamcroberts:
ESAPI is currently configured at compile time for eithe the mbedtls or
openssl crypto backends. Support the ability to switch this at run time.
Note that the next path will be to support a configure option of none
for the crypto backend so no crypto is linked.

This will decouple ESAPI from needed direct integration with backends
that may not be publicly available and or easy to test due to
environmental constraints. This will also prevent folks from having to
carry out of tree patches for crypto backends.

The following scenarios are possible:

| configure time | run time set | outcome |
| -------------- | ------------ | ------- |
| mbedtls        | no           | mbedtls |
| mbedtls        | yes          | runtime |
| openssl        | no           | openssl |
| openssl        | yes          | runtime |
| none           | no           | none    |
| none           | yes          | runtime |

Thus runtime set is always used. To restore back to original call the
getter and save the callback function pointers before calling the set.

### DIFF
--- a/include/tss2/tss2_esys.h
+++ b/include/tss2/tss2_esys.h
@@ -66,6 +66,290 @@ typedef uint32_t ESYS_TR;
 
 typedef struct ESYS_CONTEXT ESYS_CONTEXT;
 
+typedef struct ESYS_CRYPTO_CONTEXT_BLOB ESYS_CRYPTO_CONTEXT_BLOB;
+
+/*
+ * Crypto Backend Support
+ */
+
+/** Provide the context for the computation of a hash digest.
+ *
+ * The context will be created and initialized according to the hash function.
+ * @param[out] context The created context (callee-allocated).
+ * @param[in] hashAlg The hash algorithm for the creation of the context.
+ * @param[in/out] userdata information.
+ * @retval TSS2_RC_SUCCESS on success.
+ * @retval USER_DEFINED user defined errors on failure.
+ */
+typedef TSS2_RC
+    (*ESYS_CRYPTO_HASH_START_FNP)(
+        ESYS_CRYPTO_CONTEXT_BLOB ** context,
+        TPM2_ALG_ID hashAlg,
+        void *userdata);
+
+/** Update the digest value of a digest object from a byte buffer.
+ *
+ * The context of a digest object will be updated according to the hash
+ * algorithm of the context. <
+ * @param[in,out] context The context of the digest object which will be updated.
+ * @param[in] buffer The data for the update.
+ * @param[in] size The size of the data buffer.
+ * @param[in/out] userdata information.
+ * @retval TSS2_RC_SUCCESS on success.
+ * @retval USER_DEFINED user defined errors on failure.
+ */
+typedef TSS2_RC
+    (*ESYS_CRYPTO_HASH_UPDATE_FNP)(
+        ESYS_CRYPTO_CONTEXT_BLOB * context,
+        const uint8_t *buffer,
+        size_t size,
+        void *userdata);
+
+/** Get the digest value of a digest object and close the context.
+ *
+ * The digest value will written to a passed buffer and the resources of the
+ * digest object are released.
+ * @param[in,out] context The context of the digest object to be released
+ * @param[out] buffer The buffer for the digest value (caller-allocated).
+ * @param[out] size The size of the digest.
+ * @param[in/out] userdata information.
+ * @retval TSS2_RC_SUCCESS on success.
+ * @retval USER_DEFINED user defined errors on failure.
+ */
+typedef TSS2_RC
+    (*ESYS_CRYPTO_HASH_FINISH_FNP)(
+        ESYS_CRYPTO_CONTEXT_BLOB **context,
+        uint8_t *buffer,
+        size_t *size,
+        void *userdata);
+
+/** Release the resources of a digest object.
+ *
+ * The assigned resources will be released and the context will be set to NULL.
+ * @param[in,out] context The context of the digest object.
+ * @param[in/out] userdata information.
+ */
+typedef void
+    (*ESYS_CRYPTO_HASH_ABORT_FNP)(
+        ESYS_CRYPTO_CONTEXT_BLOB **context,
+        void *userdata);
+
+/** Provide the context an HMAC digest object from a byte buffer key.
+ *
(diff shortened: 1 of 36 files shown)


williamcroberts:
@AndreasFuchsTPM could you look at this one too?

src/tss2-esys/esys_context.c
```diff
@@ -10,8 +10,9 @@
 #include <stdlib.h>
 
 #include "tss2_esys.h"
-#include "tss2_tctildr.h"
+#include "tss2_esys.h"
```
AndreasFuchsTPM, line 13: ???

include/tss2/tss2_tpm2_types.h
```diff
@@ -1991,4 +1991,5 @@ struct TPML_AC_CAPABILITIES {
     UINT32 count; /* Number of values in the acCapabilities list. May be 0 */
     TPMS_AC_OUTPUT acCapabilities[TPM2_MAX_AC_CAPABILITIES]; /* List of AC values */
 };
+
```
AndreasFuchsTPM, line 1994: ???

include/tss2/tss2_esys.h
```diff
@@ -3302,6 +3404,17 @@ Esys_GetSysContext(
...
     TSS2_SYS_CONTEXT **sys_context);
 
+TSS2_RC
+Esys_SetCryptoCallbacks(
+    ESYS_CONTEXT *esysContext,
+    ESYS_CRYPTO_CALLBACKS *callbacks,
+    void *userdata);
+
+ESYS_TR
+Esys_GetCryptoCallbacks(
```
AndreasFuchsTPM, line 3414: What is the purpose of GetCallbacks ?
IMHO a setter is enough, no need for the getter.

AndreasFuchsTPM (changes requested):
I only looked at the interface and have some minor comments there.

I guess I can live with this approach, even though I don't really like it.

As for the implementation details, maybe @JuergenReppSIT could have a look at this ?

include/tss2/tss2_esys.h
williamcroberts: I was initially leaning towards that, but then I was thinking about signal handlers that give you the current state so you can back up and restore. So in this case you could figure out at runtime if t he backend has no crypto linked and set your own, or back up and restore something for some reason.

src/tss2-esys/esys_context.c
williamcroberts: `¯\_(ツ)_/¯`

src/tss2-esys/esys_context.c
```diff
@@ -238,3 +239,46 @@ Esys_GetSysContext(ESYS_CONTEXT *esys_context, TSS2_SYS_CONTEXT **sys_context)
 
     return TSS2_RC_SUCCESS;
 }
+
+TSS2_RC
+Esys_SetCryptoCallbacks(
```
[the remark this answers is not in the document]
williamcroberts: Once we tweak any API's, ill submit a final change with the doxygen, no point now while we figure it out if something drastic changes.

include/tss2/tss2_esys.h
```diff
@@ -3302,6 +3404,17 @@ Esys_GetSysContext(
     ESYS_CONTEXT *esys_context,
     TSS2_SYS_CONTEXT **sys_context);
 
+TSS2_RC
+Esys_SetCryptoCallbacks(
+    ESYS_CONTEXT *esysContext,
+    ESYS_CRYPTO_CALLBACKS *callbacks,
+    void *userdata);
```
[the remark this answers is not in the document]
AndreasFuchsTPM: a userdata pointer per callback makes the pointer struct very large.
IMHO a single one is nicer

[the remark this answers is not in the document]
williamcroberts: Thats what I was leaning towards as well. But refreshing myself we have ESYS_CRYPTO_CONTEXT_BLOB which is opaque, so that's really our user data here. I think we can drop that and make everything a `void *`. The callback `ESYS_CRYPTO_PK_RSA_ENCRYPT_FNP` however, doesn't take a crypto blob, so we should update that to include the `void *`.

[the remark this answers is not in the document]
AndreasFuchsTPM: The purpose of the ESYS__CRYPTO_CONTEXT_BLOB was to keep track of the start/update/finish sequence info, right ?
Userdata would be more stable I guess.
So IMHO it is more convenient for the user to have these two things seperated.

include/tss2/tss2_esys.h
AndreasFuchsTPM: I guess this somewhat makes sense.
BTW, was happens, if the user calls this without setting before if the build-in thing is set ?
Are the esys-pointers returned or is null returned ?

include/tss2/tss2_esys.h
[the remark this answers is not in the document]
williamcroberts: So do we want:
1. opaque struct type for the CONTEXT BLOB 
2. `void *` for userdata
    - Update all functions to take userdata and pass that userdata set in the SetCryptoCallbacks to all routines?

include/tss2/tss2_esys.h
williamcroberts: > I guess this somewhat makes sense. BTW, was happens, if the user calls this without setting before if the build-in thing is set ? Are the esys-pointers returned or is null returned ?

Depends:
| backend | return |
| ----------- | -------- |
| ossl        | ossl esys pointers |
| mbed     | mbed esys pointers |
| none      | null |

So they can always restore what was originally configured.

AndreasFuchsTPM: I do see the use case of when you want to override only one of the crypto functions but use the default ones for all the rest, I guess.
But this could also be achieved by leaving the pointers inside the struct as NULL. (btw, is this supported)

In order to restore the original pointers, I though you could just call the setter with a NULL pointer instead of a struct ?

AndreasFuchsTPM: ok with me

williamcroberts:
@AndreasFuchsTPM anything else you see that we should resolve?

AndreasFuchsTPM (approved):

PR #2368, opened 2022-06-08
sapi: reduce code size

williamcroberts:
Some commits that reduce code size of the system API by about 5k total.


AndreasFuchsTPM (approved):
Looks good to me

PR #2370, opened 2022-06-10
policy: add null checks on exec callbacks

williamcroberts:


AndreasFuchsTPM (changes requested):
That a bad bug here, so good that it's fixed.

I just don't like the return code. NOT_IMPLEMENTED would be something where the user-called function is not implemented.

Since we don't have a better return code, I'd vote for adding a new one; i.e. CALLBACK_MISSING or similar.

williamcroberts:
> NOT_IMPLEMENTED would be something where the user-called function is not implemented.

This seems contradictory tot he fact you want a change, considering we are returning NOT_IMPLEMENTED where the user-called function is not implemented.

AndreasFuchsTPM:
My rational is that when you call Fapi_Sign() and a "NOT_IMPLEMENTED" is returned, you might easily assume that Sign is not implemented or that a certain algorithm is not implemented, but the case of "the callback was not set" does not intuitively come to mind.

williamcroberts:
@AndreasFuchsTPM or @JuergenReppSIT please review, thanks

PR #2381, opened 2022-07-07
SAPI: fix number of handles for FlushContext

### LINKED ISSUE #2382: SAPI: Tss2_Sys_GetCpBuffer returns zero parameter size for flushcontext
The flush handle is placed in the handle area instead of the parameter area.

williamcroberts:
Fixes #2382 

The lookup table for the number of command handles for
Tss2_Sys_FlushContext has the count set to 1, when in reality the
command takes no handles in the handle area but a handle as the input
parameter. This works currently because handles and parameters are just
concatenated and the parsing logic on the TPM just unpacks them, so in
this case, thet're in the same spot with the same value. This goes
unnoticed until you call Tss2_Sys_GetCpBuffer and the buffer is empty as
the logic things its a handle in the handle area and not a handle in the
parameter area.

### DIFF
--- a/src/tss2-sys/sysapi_util.c
+++ b/src/tss2-sys/sysapi_util.c
@@ -261,7 +261,7 @@ static int GetNumHandles(TPM2_CC commandCode, bool req)
         { TPM2_CC_FirmwareRead, 0, 0 },
         { TPM2_CC_ContextSave, 1, 0 },
         { TPM2_CC_ContextLoad, 0, 1 },
-        { TPM2_CC_FlushContext, 1, 0 },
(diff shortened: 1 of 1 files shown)


idesai:
Also update https://github.com/tpm2-software/tpm2-tss/blob/2846211e74a722d6ac2f05a38475bef7d20b9fa6/include/tss2/tss2_common.h#L12

williamcroberts:
@JuergenReppSIT @AndreasFuchsTPM

JuergenReppSIT (approved):

Issue #2383, opened 2022-07-07
Questions policy JSON encoding and callbacks with tss2-policy


whooo:
Hi,

After working on the python bindings for tss2-policy I some questions, TPMT_PUBLIC, TPM2B_PUBLIC and TPM2B_NV_PUBLIC is encoded with the size and nvPublic/publicArea fields, but my understanding of the specification is it should be encoded as TPMT_PUBLIC and TPMS_NV_PUBLIC.
Would it be possible to change the encoding, or would it break backward compatibility for FAPI?

Also, the callback TSS2_POLICY_CB_NVPUBLIC doesn't work with TPMS_POLICYNV if nvIndex is used instead of nvPath

williamcroberts:
Oh and TPMT_PUBLIC doesn't have a size, that's the data member in the TPM2B_PUBLIC. But I still get your point.

whooo:
> 
> AFAICT that is not exported publicly through FAPI Callbacks, so we can change that but we have to do it now.
> 
I was thinking of the on disk storage for FAPI and Fapi_ExportPolicy

> > Also, the callback TSS2_POLICY_CB_NVPUBLIC doesn't work with TPMS_POLICYNV if nvIndex is used instead of nvPath
> 
> Can you elaborate, what happens?
When executing a policy looking something like:
```json
{
  "description": "my policy",
  "policy": {
    "type": "nv",
    "nvIndex": "0x1000000",
    "operandB": "00"
  }       
} 
```
It fails with `
ERROR:fapi:src/tss2-fapi/ifapi_policy_instantiate.c:314:ifapi_policyeval_instantiate_finish() ErrorCode (0x00060028) No path for policy PolicyNv`.
I guess part of the problem is that there is no way for tss2-policy to get the matching TPMS_NV_PUBLIC with nvIndex as it can't be passed via TSS2_POLICY_CB_NVPUBLIC 
I can always create a PR which replaces nv_public with a struct containing a selector and an union with TPM2B_NV_PUBLIC and a TPMI_RH_NV_INDEX

JuergenReppSIT:
The current implementation did assume that the NV public info could be set in the json policy:
`  if (pol_element->element.PolicyNV.nvPublic.nvPublic.nvIndex) {`
`                /* nvIndex is already set in policy. Path will not be needed */`
But according to the Spec "JSON policy data types" only the NV index can be set.
[...]

PR #2384, opened 2022-07-08
fapi: fix usage of policy_nv.

JuergenReppSIT:
* Currently it was not possible to define a policy nv with a TPM nv index.
    The callback to get the public nv data related to the policy was extended
    to get public nv data from the TPM in this case.
    Addresses #2383
* Also an integration test was added.
 
### DIFF
--- a/src/tss2-fapi/ifapi_policy_callbacks.c
+++ b/src/tss2-fapi/ifapi_policy_callbacks.c
@@ -210,7 +210,11 @@ cleanup:
 
 /** Get public data of a NV object from keystore.
  *
- * @param[in] path The relative path of the NV object.
+ * @param[in] path The relative path of the NV object. The path will
+              be used to get the public from the keystore if no nv_index is
+              passed.
+ * @param[in] nv_index The index of the NV object. The index will be used to
+ *            get the public data if nv_index > 0.
  * @param[out] nv_public The caller allocated public structure.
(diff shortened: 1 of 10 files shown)


src/tss2-fapi/ifapi_policy_callbacks.c
```diff
@@ -210,7 +210,11 @@ ifapi_get_object_name(
 
 /** Get public data of a NV object from keystore.
  *
- * @param[in] path The relative path of the NV object.
+ * @param[in] path The relative path of the NV object. The path will
+              be used to get the public from the keystore if no nv_index is
+              passed.
+ * @param[in] nv_index The index of the NV object. The index will be used to
+ *            get the public data if nv_index > 0.
```
williamcroberts, line 217: ahh ok if nv_index is 0 look at path..., is the inverse thru if not path look at index?

williamcroberts (approved):

src/tss2-fapi/ifapi_policy_callbacks.c
JuergenReppSIT: > ahh ok if nv_index is 0 look at path..., is the inverse thru if not path look at index?

yes. It's ensured by `ifapi_json_TPMS_POLICYNV_deserialize` that either nv_index or path is set.

PR #2389, opened 2022-07-13
Backport 3.2.x tcti libtpms link

williamcroberts:


williamcroberts:
CL Entry: linking tcti for libtpms against tss2-tctildr. It should be linked against tss2-mu.

PR #2390, opened 2022-07-14
policy: drop complex TPM2B from callback interface

williamcroberts:
Rather then using types like TPM2B_NV_PUBLIC just use the TPMS_NV_PUBLIC
since the size field is never needed. Keep the ondisk representation
in JSON going through the same routines and convert between the two
types this way we don't break on disk representation of existin objects.

Fixes: #2383

### DIFF
--- a/src/tss2-fapi/ifapi_policy_callbacks.c
+++ b/src/tss2-fapi/ifapi_policy_callbacks.c
@@ -190,7 +190,7 @@ ifapi_get_object_name(
                                (TPM2B_NAME *)name);
             break;
         case IFAPI_NV_OBJ:
-            r = ifapi_nv_get_name(&object.misc.nv.public, name);
+            r = ifapi_nv_get_name(&object.misc.nv.public.nvPublic, name);
             break;
         default:
             goto_error(r, TSS2_FAPI_RC_BAD_VALUE, "Invalid object %s.",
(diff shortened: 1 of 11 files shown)


williamcroberts:
@whooo this should preserve on-disk formats and only change the in-memory representation through the callbacks.

src/tss2-fapi/ifapi_policy_callbacks.c
```diff
@@ -1366,9 +1366,14 @@ ifapi_exec_auth_nv_policy(
 
     switch (cb_ctx->cb_state) {
         statecase(cb_ctx->cb_state, POL_CB_EXECUTE_INIT)
-            /* Search a NV object with a certain NV indext stored in nv_public. */
+            /*
+             * Search a NV object with a certain NV indext stored in nv_public.
+             * TODO Consider modifying internal APIs to use TPMS_NV_INDEX over TPM2B variant.
```
williamcroberts, line 1371: Modifying this keystore search really starts to bleed into other areas of the codebase, might be better just to leave it alone, but I could modify if it needed.

JuergenReppSIT: > Modifying this keystore search really starts to bleed into other areas of the codebase, might be better just to leave it alone, but I could modify if it needed.

@williamcroberts I would agree to leave it alone. Also in the JSON policy spec TPM2B_PUBLIC is used for nvPublic in TPMS_POLICYAUTHORIZENV.

JuergenReppSIT (approved):
really nice to remove all this nvPublic.nvPublic

Issue #2171, opened 2021-10-07
Inconsistency caused by Esys_TR_FromTPMPublic


JuergenReppSIT:
Several ESYS objects with the same TPM handle with the function Esys_TR_FromTPMPublic. Some ESYS functions e.g. Esys_NV_Write update the internal meta for the object referenced by the passed ESYS handle. The consequence is that there might be objects with NW_WRITTEN set and NV_WRITTEN not set.
This caused an error in the TPM tools described in https://github.com/tpm2-software/tpm2-tools/issues/2846.
So all objects with the same TPM handle should be updated to avoid inconsistencies. Also for changing the auth values such an update might be needed.

williamcroberts:
@JuergenReppSIT is this fixed and can be closed? I see a lot of commit descriptions as "addresses" is that just a partial fix or does it fix the whole thing?

gfrankliu:
I have to questions:
1) Is the fix in tss 3.0, 3.1 and 3.2? or just 3.2? I am using Debian 11 that includes tss 3.0.3, and looking for the fix in that branch.
2) Does the fix make the [tools fix](https://github.com/tpm2-software/tpm2-tools/issues/2846#issuecomment-934563179) obsolete? Or do we need a combination of tools and tss fix?

williamcroberts:
> I have to questions:
> 
> 1. Is the fix in tss 3.0, 3.1 and 3.2? or just 3.2? I am using Debian 11 that includes tss 3.0.3, and looking for the fix in that branch.

- merged:
master: 6c88eea
3.2.x: 9239184

They are currently not released. Master is slated to be the upcoming 4.0 release and 3.2.x should be 3.2.1 release offhand.

> 2. Does the fix make the [tools fix](https://github.com/tpm2-software/tpm2-tools/issues/2846#issuecomment-934563179) obsolete? Or do we need a combination of tools and tss fix?

We could probably check the ESAPI version when building and compile out the patch, github.com/tpm2-software/tpm2-tools/pull/2847, but considering that that the tools CI runs against master and it works, I see no issues other than performance caused by always doing this.

williamcroberts:
I can mark this as closed after investigating this, but will open a bug on tpm2-tools to compile that out. See tpm2-tools bug https://github.com/tpm2-software/tpm2-tools/issues/3156 so we don't forget about it.

JuergenReppSIT:
@williamcroberts Thank you for the investigation and closing the issue.

PR #2393, opened 2022-07-16
esys: fix usage of HMAC sessions for Esys_TR_FromTPMPublic

JuergenReppSIT:
The ESAPI spec states that it must be possible to use HMAC sessions
for Esys_TR_FromTPMPublic.
Currently it was not implemented to call Esys_TR_FromPublic twice.
The first call has to be used to determine the object name which
will be used to compute the HMAC of the session.
If the esys object already exists the existing object will be used.
If a session is used in this case the name retrieved from the
TPM will be compared with name stored in the object.

The integration test for the test with Esys_TR_FromTPMPublic was
extended to add a call with a HMAC session.
Addresses: #2171

### DIFF
--- a/src/tss2-esys/esys_tr.c
+++ b/src/tss2-esys/esys_tr.c
@@ -133,11 +133,41 @@ Esys_TR_FromTPMPublic_Async(ESYS_CONTEXT * esys_context,
     _ESYS_ASSERT_NON_NULL(esys_context);
     ESYS_TR esys_handle = esys_context->esys_handle_cnt++;
     RSRC_NODE_T *esysHandleNode = NULL;
-    r = esys_CreateResourceObject(esys_context, esys_handle, &esysHandleNode);
-    goto_if_error(r, "Error create resource", error_cleanup);
+    RSRC_NODE_T *node_rsrc = NULL;
+    RSRC_NODE_T *next_node_rsrc;
 
-    esysHandleNode->rsrc.handle = tpm_handle;
-    esys_context->esys_handle = esys_handle;
+    for (node_rsrc = esys_context->rsrc_list; node_rsrc != NULL;
+         node_rsrc = next_node_rsrc) {
+         if (node_rsrc->rsrc.handle == tpm_handle) {
+             esysHandleNode = node_rsrc;
+             esys_context->esys_handle = node_rsrc->esys_handle;
+             break;
+         }
+         next_node_rsrc = node_rsrc->next;
+    }
+
+    if (!esysHandleNode) {
+        /* Object was already created */
+        esys_handle = esys_context->esys_handle_cnt++;
+        r = esys_CreateResourceObject(esys_context, esys_handle, &esysHandleNode);
+        goto_if_error(r, "Error create resource", error_cleanup);
+
+        /* In the first trial no session will be used to determine the object name. */
+        esys_context->sav_session1 = shandle1;
+        esys_context->sav_session2 = shandle2;
+        esys_context->sav_session3 = shandle3;
+        esys_context->session_tab[0] = NULL;
+        esys_context->session_tab[1] = NULL;
+        esys_context->session_tab[2] = NULL;
+        esysHandleNode->rsrc.handle = tpm_handle;
+        esys_context->esys_handle = esys_handle;
+        shandle1 = ESYS_TR_NONE;
+        shandle2 = ESYS_TR_NONE;
+        shandle3 = ESYS_TR_NONE;
+    } else {
+        esys_handle = esysHandleNode->esys_handle;
+        esys_context->esys_handle = esys_handle;
+    }
 
     if (tpm_handle >= TPM2_NV_INDEX_FIRST && tpm_handle <= TPM2_NV_INDEX_LAST) {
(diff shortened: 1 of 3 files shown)


src/tss2-esys/esys_tr.c
```diff
@@ -133,11 +133,35 @@ Esys_TR_FromTPMPublic_Async(ESYS_CONTEXT * esys_context,
...
+        esys_handle = esys_context->esys_handle_cnt++;
+        r = esys_CreateResourceObject(esys_context, esys_handle, &esysHandleNode);
+        goto_if_error(r, "Error create resource", error_cleanup);
+
+        /* In the first trial no session will be used to determine the object name. */
+        esys_context->sav_session = shandle1;
+        esys_context->session_tab[0] = NULL;
+        esysHandleNode->rsrc.handle = tpm_handle;
+        esys_context->esys_handle = esys_handle;
+        shandle1 = ESYS_TR_NONE;
```
williamcroberts, line 160: So this means that the user specified a session but the object is not in the ESYS_TR list and the call will now use no session handle to go to the TPM to get the name. Shouldn't this error instead as to not surprise the user that the tpm communications were unprotected? What about shandles 2 and 3?

williamcroberts: Ok lets add a check that if it's not an NV index error. But how is nv index ok? We still don't know the name until we get the public for the attribute and other properties right?

JuergenReppSIT: **Addition:** The spec says:
"The ‘optionalSession’ parameter can be used in order to pass an HMAC session, encryption
session or audit session. This parameter can also be ESYS_TR_NONE. It is recommended to
pass in one salted HMAC session for untrusted connections to a TPM."
with:
`ESYS_TR optionalSession1, /* optional (ESYS_TR_NONE) */`
which is a contradiction to the "shall" in:
For all other handles (besides NV-Index) this function shall
only be used if the path to the TPM can be trusted.

williamcroberts: > Yes in all cases it's necessary to determine the name for the HMAC session. So in the case of an NV index it's the users responsibility to provide a session. If the session is provided the public data of the two calls will be compared.

Wait, two calls? Do you mean this function calls readpublic twice or if the user issues 2 Esys_TR_FromTPMPublic calls?

The thing I want to prevent is the case where a user is thinking they're using an HMAC session and that HMAC session getting silently ignored. Despite what the documentation says, we can't silently be dropping sessions.

> For handles of persistent keys BAD_VALUE will be returned if no session is used. So calls of `Esys_TR_FromTPMPublic` for persistent keys which did work in the past now would fail.

Apparently I don't get whats going on here. If we're communicating with the TPM it's because we don't know the name and have to go get it. So how could anything requiring the name, like an HMAC session, work?
The only case I care about is when a user supplies a session and that session is changed to ESYS_TR_NONE.

I really see an API where a user can say make me an esys TR from this TPM handle *and* this name.

> Currently we have one integration test which has to be modified (esys-tr-fromTpmPublic-key.int.c). I also checked the tpm-tools. Here this function is only used with NV indexes. Should I add the test?

JuergenReppSIT: The ESAPI spec says:
> NOTE: If a session is passed into this function, then the ReadPublic command with the TPM will be called
> twice. The first call is executed without a session in order to retrieve the public name that is then used in
> the HMAC calculation for the second call. For the synchronous invocation the two TPM calls are made
> transparently for the application.
So that's what's implemented in the current PR. If the user is using a HMAC session the session is not ignored.
The second call with the HMAC session will be used to verify the result of the first call.

williamcroberts: > > transparently for the application.

Ahh ok, but the attacker can still give a bad object, but the path must be trusted in this case. That's fine.

williamcroberts: Actually it does support serialized EYS_TR's (I thought it did):
```bash
tpm2 createprimary -c primary.ctx
tpm2 evictcontrol -c primary.ctx -o primary.tr
tpm2 create -C primary.tr -u key.pub -r key.priv
```

JuergenReppSIT: yes the serialization is supported. I just checked the tools to see whether the discussed test of a session for persistent key handles for `Esys_TR_FromTPMPublic`would cause problems. 
But now I would say the check is not necessary because in the parameter description the usage of the session is only recommended. If the check is not needed the would the PR be ok?

williamcroberts (approved):
This seems fine, since it's a backport please make sure the other changes are back on master branch as well

JuergenReppSIT:
@williamcroberts I got the message "dismissed review" after I did rebase the PR and git push force. I also backported the PR to 3.2.x. So both PRs could be approved.

williamcroberts:
> @williamcroberts I got the message "dismissed review" after I did rebase the PR and git push force. I also backported the PR to 3.2.x. So both PRs could be approved.

That's because of one of the options checked in the settings. If a PR is updated all reviews are reset.

williamcroberts (approved):

PR #2396, opened 2022-07-23
fapi: Fix json encoding of policies with complex TPM2Bs.

JuergenReppSIT:
The json spec states that complex TPM2Bs do not have a JSON representation.
Currently it's possible to use the the representation of a complex TPM2B
in policy authorize nv, in policy duplication select, policy template.
The usage of the embedded TPMT or TPMS JSON representation, as required
by the spec, can't be used.
To achieve backward compatibility with existing key stores both cases now
can be used.
Policy duplication select is specified with TPMT_PUBLIC for newParentPublic
while the implementation uses TPM2B_PUBLIC. Now TPMT_PUBLIC is used but
also here the JSON representation of the TPM2B_PUBLIC can be deserialized.
Addresses #2383.


src/tss2-fapi/ifapi_json_deserialize.h
```diff
@@ -25,6 +25,29 @@
         return_if_error2(r, "Bad value for field \"%s\".", json_name);  \
     }
 
+
+/*
+ * Deserialize an embedded structure of an complex TPM2B.
+ * To achieve backward compatibility both JSON representations, of the complex
+ * TPM2B, and of the embedde structure will be allowed.
```
williamcroberts, line 32: spelling: "embedde" to "embedded".

williamcroberts (changes requested):

williamcroberts:
> The usage of the embedded TPMT or TPMS JSON representation, as required
by the spec, can't be used.

Why?

JuergenReppSIT:
> > The usage of the embedded TPMT or TPMS JSON representation, as required
> > by the spec, can't be used.
> 
> Why?
In the current version you will get an error if the JSON representation of the TPMS or TPMT structure is used instead of the complex TPM2B.

AndreasFuchsTPM:
I guess an example JSON of current behavior and expected would be nice for understanding.
Also I guess that this case is currently not covered by an integration test, so we might need one...

JuergenReppSIT:
> I guess an example JSON of current behavior and expected would be nice for understanding. Also I guess that this case is currently not covered by an integration test, so we might need one...

I have added integration tests and a unit test to cover these cases.

williamcroberts (approved):

PR #2406, opened 2022-08-25
Tss2 mu precision fixes

williamcroberts:
You can enabled this warnings on gcc/clang via -Wconversion. I guess do we want to enable this for libmu?

### DIFF
--- a/src/tss2-mu/base-types.c
+++ b/src/tss2-mu/base-types.c
@@ -64,13 +64,13 @@ Tss2_MU_##type##_Marshal ( \
         case 1: \
             break; \
         case 2: \
-            src = HOST_TO_BE_16(src); \
+            src = (type)HOST_TO_BE_16((UINT16)src); \
             break; \
         case 4: \
-            src = HOST_TO_BE_32(src); \
+            src = (type)HOST_TO_BE_32((UINT32)src); \
             break; \
         case 8: \
-            src = HOST_TO_BE_64(src); \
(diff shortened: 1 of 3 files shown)


JuergenReppSIT (commented):
@williamcroberts Is there a reason to add the UINT cast only to case 2?

williamcroberts:
> @williamcroberts Is there a reason to add the UINT cast only to case 2?

size_t is 32 bits on the platform I am building so truncation only occur when going from a UINT32 to a UINT16. Likely would need the same for UINT64 to UINT32 for 64 bit platforms but this placates the compiler.

I'll probably add `-WConversion ` to the specific libraries at some point so we catch them here as well as `-Wunreachable-code`, but I am reading that GCC deprecated that option.

JuergenReppSIT:
> size_t is 32 bits on the platform I am building so truncation only occur when going from a UINT32 to a UINT16. Likely would need the same for UINT64 to UINT32 for 64 bit platforms but this placates the compiler.

@williamcroberts so would something speak against adding the cast in all three cases?

williamcroberts:
> > size_t is 32 bits on the platform I am building so truncation only occur when going from a UINT32 to a UINT16. Likely would need the same for UINT64 to UINT32 for 64 bit platforms but this placates the compiler.
> 
> @williamcroberts so would something speak against adding the cast in all three cases?

I would actually like to see the compiler complain just to verify my understanding before adding it.

JuergenReppSIT (approved):

PR #2409, opened 2022-09-08
test: fix buffer too small in sys-session-util.c

williamcroberts:
When the values being concatenated overflowed a TPM2B_MAX_BUFFER, which
is bounded on 1024 bytes, TSS2_SYS_RC_INSUFFICIENT_BUFFER was returned
from tpm_calc_phash due to a bound check in ConcatSizedByteBuffer.

Calculating the phash requires a lot of space becuase it could require 3
names which could be 512 bytes (SHA512 for instance) each, which
overflows the space available. Bump an internal buffer to 4096 for this
calculation, which should be plenty:
1. Command Code: 4 bytes
2. name1: Max 64 bytes
3. name2: Max 64 bytes
4. name3: Max 64 bytes
5. Remaining serialized paramters (command specific)

Items 1 through 4 yeild a size of 196 bytes which yeilds a remaining
3900 bytes bytes for the rest of the paramters. After that, the largest
commands seem to take a TPM2B_MAX_NV_BUFFER (2048 bytes) and then some
small scalar values.

Items 1 through 4 yeild a size of 1540 bytes which yeilds a remaining 2556 bytes bytes for the rest of the paramters. After that, the largest commands seem to take a TPM2B_MAX_NV_BUFFER (2048 bytes) and then some small scalar values.

### DIFF
--- a/test/integration/sys-session-util.c
+++ b/test/integration/sys-session-util.c
@@ -19,6 +19,12 @@
 
 static SESSION *sessions = NULL;
 
+typedef struct TPM2B_BIG_BUFFER TPM2B_BIG_BUFFER;
+struct TPM2B_BIG_BUFFER {
+    UINT16 size;
+    BYTE buffer[TPM2_MAX_DIGEST_BUFFER * 4];
+};
+
 SESSION *
 get_session(TPMI_SH_AUTH_SESSION hndl)
 {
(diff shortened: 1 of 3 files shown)


williamcroberts:
I'm not sure how to trigger a test that will cause this to fail, maybe use objects with SHA384 names or something? it seems to be specific to certain environments.

williamcroberts:
> I'm not sure how to trigger a test that will cause this to fail, maybe use objects with SHA384 names or something? it seems to be specific to certain environments.

Maybe NV because that buffer is 2048?

williamcroberts:
Fix for oss-fuzz here: https://github.com/google/oss-fuzz/pull/8445

PR #2414, opened 2022-09-17
tss2: update defines for spec version to 1.38

williamcroberts:
It was set for 1.26, which AFAICT, was never released. Set this for the implemented version of 1.38. FYI, the TSS supports a small smattering of the 1.59 constructs currently.


williamcroberts:
backported: https://github.com/tpm2-software/tpm2-tss/pull/2415

PR #2416, opened 2022-09-17
rc: add function to decode the bitfields in TSS2_RC

whooo:
This should enable developers to extract response code information and provide better logging/error messages based on application specific context

### DIFF
--- a/include/tss2/tss2_rc.h
+++ b/include/tss2/tss2_rc.h
@@ -17,6 +17,17 @@ const char *Tss2_RC_Decode(TSS2_RC rc);
 
 TSS2_RC_HANDLER Tss2_RC_SetHandler(uint8_t layer, const char *name, TSS2_RC_HANDLER handler);
 
+typedef struct TSS2_RC_INFO TSS2_RC_INFO;
+struct TSS2_RC_INFO {
+    UINT8 layer;
+    TSS2_RC error;
+    UINT8 parameter;
+    UINT8 session;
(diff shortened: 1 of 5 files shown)


williamcroberts:
@whooo I noticed one thing I just thought of, how do folks know if they should interrogate the parameter,session,handle area versus the non-format 1 type where it's just error? Should we include the format or just expect them to check that those 3 fields are 0? We should probably memset 0 the structure when it comes in no matter what.

whooo:
> @whooo I noticed one thing I just thought of, how do folks know if they should interrogate the parameter,session,handle area versus the non-format 1 type where it's just error? Should we include the format or just expect them to check that those 3 fields are 0? We should probably memset 0 the structure when it comes in no matter what.

A format field would be good so users don't need to check all the fields, and memset of course, I'll create a PR for it.
Having a function to get the string of the error (instead of the full `tpm:handle(..):...` string) might be a good idea as well

williamcroberts:
> > @whooo I noticed one thing I just thought of, how do folks know if they should interrogate the parameter,session,handle area versus the non-format 1 type where it's just error? Should we include the format or just expect them to check that those 3 fields are 0? We should probably memset 0 the structure when it comes in no matter what.
> 
> A format field would be good so users don't need to check all the fields, and memset of course, I'll create a PR for it. Having a function to get the string of the error (instead of the full `tpm:handle(..):...` string) might be a good idea as well

+1

PR #2432, opened 2022-10-06
fapi: Skip policy check if PCR value is not expected

rosslagerwall:
When running the testcase against a real device, PCR 8 might have been extended (typically by GRUB) causing the test to fail due to unexpected preconditions. Skip the policy check in this case.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>


JuergenReppSIT (approved):

JuergenReppSIT:
@rosslagerwall That makes sense. Thank you!

PR #2435, opened 2022-10-12
FAPI: Fix usage of second profile in path.

JuergenReppSIT:
The usage of a second profile in a path was not possible because allways the default profile was used.
So after a second provisioning with a second fapi config file this profile could not be uses in fapi pathnames.


williamcroberts:
See comments on #2436

williamcroberts (approved):

PR #2436, opened 2022-10-12
FAPI: Fix usage of second profile in path 3.2.x

JuergenReppSIT:
The usage of a second profile in a path was not possible because allways the default profile was used.
So after a second provisioning with a second fapi config file this profile could not be uses in fapi pathnames.

### DIFF
--- a/src/tss2-fapi/ifapi_profiles.c
+++ b/src/tss2-fapi/ifapi_profiles.c
@@ -250,7 +250,7 @@ ifapi_profiles_get(
     size_t len;
 
     /* if no name or nor profile prefix is given, use the default profile */
-    if (!name || strncmp(name, "P_", 2) != 0 || strncmp(name, "/P_", 2) != 0) {
+    if (!name || !(strncmp(name, "P_", 2) == 0 || strncmp(name, "/P_", 3) == 0)) {
         *profile = &profiles->default_profile;
         return TSS2_RC_SUCCESS;
     }


src/tss2-fapi/ifapi_profiles.c
```diff
@@ -250,7 +250,7 @@ ifapi_profiles_get(
     size_t len;
 
     /* if no name or nor profile prefix is given, use the default profile */
-    if (!name || strncmp(name, "P_", 2) != 0 || strncmp(name, "/P_", 2) != 0) {
+    if (!name || !(strncmp(name, "P_", 2) == 0 || strncmp(name, "/P_", 2) == 0)) {
```
williamcroberts, line 253: @JuergenReppSIT is my breakdown of this logic correct?
if name OR does not start with `P_` OR starts with `/P`.

The second strncmp has 3 characters in the comparison string but has a length of 2 characters set.

Don't we just want:
```C
 if (!name || (!strncmp(name, "P_", 2) && !strncmp(name, "/P_", 3))) {
```
Then I think we can apply demorgans and get:
```C
 if (!name || !(strncmp(name, "P_", 2) || strncmp(name, "/P_", 3))) {
```

JuergenReppSIT: @williamcroberts yes I will fix it and we should use the version with == 0:
```
(gdb) set name="XXXX"
(gdb) p  (!name || !(strncmp(name, "P_", 2) == 0 || strncmp(name, "/P_", 3) == 0))
$36 = 1
(gdb) p   (!name || !(strncmp(name, "P_", 2) || strncmp(name, "/P_", 3)))
$37 = 0
```

Issue #2446, opened 2022-10-31
Installation of tpm-event-logging.service breaks non-root installation


whooo:
During a make install from the master branch the installation fails as part of make install tries to add dist/tpm-event-logging.service directly to /lib/systemd/system event when configure was passed a prefix (see https://github.com/tpm2-software/tpm2-pytss/actions/runs/3364252134/jobs/5578387782 ).
It might also fail when trying to enable the service, even when running as root, (see https://github.com/tpm2-software/tpm2-pytss/actions/runs/3364184969/jobs/5578244407 ).

One alternative might be to add both `/sys/kernel/security/tpm[0-9]/binary_bios_measurements` and `/sys/kernel/security/ima/binary_runtime_measurements` to the tmpfiles configuration file tpm2-tss-fapi.conf and let that handle the permissions instead.

PR #2450, opened 2022-11-01
event log permissions

whooo:
Use tpm2-tss-fapi.conf to set permissions for UEFI and IMA events logs in sysfs.
Works well on my laptop.

Fixes https://github.com/tpm2-software/tpm2-tss/issues/2446


JuergenReppSIT:
> Use tpm2-tss-fapi.conf to set permissions for UEFI and IMA events logs in sysfs.
> Works well on my laptop.
I wanted the group assignment to remain intact over the boot cycle. Therefore during experimenting I did copy the described tmp file config to /lib/tmpfile.d/. After the reboot the group assignment was only set for the ima files. Therefore I decided to create the one shot service. With your solution after the reboot the files are still owned by root.root.
So the tmp file config solution works only for `make install`.

williamcroberts:
This seems related to:
  - https://github.com/tpm2-software/tpm2-tss/issues/2451

However, I see the fix and keeping the oneshot service as we need something to run on every boot. I guess we gotta figure out why this fails in our ci

whooo:
> > Use tpm2-tss-fapi.conf to set permissions for UEFI and IMA events logs in sysfs.
> > Works well on my laptop.
> 
> @whooo I wanted the group assignment to remain intact over the boot cycle. Therefore during experimenting I did copy the described tmp file config to /lib/tmpfile.d/. After the reboot the group assignment was only set for the ima files. Therefore I decided to create the one shot service (Did work without problems on Ubuntu 20.04). With your solution after the reboot the files are still owned by root.root. So the tmp file config solution works only for `make install`.

On the same machine, do you have another file named tpm2-tss-fapi.conf under /etc/tmpfiles.d/ without the event log paths?

JuergenReppSIT:
> On the same machine, do you have another file named tpm2-tss-fapi.conf under /etc/tmpfiles.d/ without the event log paths?

yes there was another file. Now it works also after reboot.
@whooo  Thank you!

JuergenReppSIT (approved):

PR #2452, opened 2022-11-01
FAPI: Fix policy secret.

### LINKED ISSUE #2448: FAPI: Policy secret fails when secret object is a key
Only hierarchies and nv objects can currently be used in PolicySecret.
If a key is used a TPM error is returned.

JuergenReppSIT:
The authorization did not work when key objects were used as secret. The current object to be authorized was overwritten by the secret object. To fix this problem backups of the objects to be authorized were stored in the policy stack. The loaded secret key is  flushed after policy execution if it's a transient key.
Fixes #2448.

### DIFF
--- a/src/tss2-fapi/ifapi_policy_callbacks.c
+++ b/src/tss2-fapi/ifapi_policy_callbacks.c
@@ -484,6 +484,7 @@ ifapi_policyeval_cbauth(
         next_case = false;
         switch (cb_ctx->cb_state) {
         statecase(cb_ctx->cb_state, POL_CB_EXECUTE_INIT);
+            cb_ctx->flush_handle = false;
             cb_ctx->auth_index = ESYS_TR_NONE;
             /* Search object with name in keystore. */
             r = ifapi_keystore_search_obj(&fapi_ctx->keystore, &fapi_ctx->io,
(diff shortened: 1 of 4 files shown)


williamcroberts:
@whooo this may break our bindings....

include/tss2/tss2_policy.h
```diff
@@ -110,6 +110,7 @@ typedef TSS2_RC (*TSS2_POLICY_CB_EXEC_AUTH) (
     TPM2B_NAME *name,
     ESYS_TR *object_handle,
     ESYS_TR *auth_handle,
+    bool *flush_handle,
```
williamcroberts, line 113: We add this here but then don't actually make use of this in the generic policy callbacks and only in the IFAPI portions of the code. Shouldn't this be in either the userdata pointer within the ifapi_policyeval_cbauth (which is a FAPI_CONTEXT) or within one of the other context structures:
```C
    IFAPI_POLICY_EXEC_CTX *current_policy;
    IFAPI_POLICY_EXEC_CB_CTX *cb_ctx;
```

JuergenReppSIT: Yes that's true. I will move `flush_handle` to the context. I wanted to add another parameter  `auth_object` to the callback to enable the usage of the callback for  PolicySigned with symmetric keys. Here only the key has to be loaded and the authorization is only needed for parent objects. If TSS2_POLICY_CB_EXEC_AUTH should not be changed I will add a new callback TSS2_POLICY_CB_EXEC_LOAD. Would this be OK or do you see any problems?

williamcroberts: > Yes that's true. I will move `flush_handle` to the context. I wanted to add another parameter `auth_object` to the callback to enable the usage of the callback for PolicySigned with symmetric keys. Here only the key has to be loaded and the authorization is only needed for parent objects. If TSS2_POLICY_CB_EXEC_AUTH should not be changed I will add a new callback TSS2_POLICY_CB_EXEC_LOAD. Would this be OK or do you see any problems?

That seems fine

williamcroberts (approved):

include/tss2/tss2_policy.h
williamcroberts: > > Yes that's true. I will move `flush_handle` to the context. I wanted to add another parameter `auth_object` to the callback to enable the usage of the callback for PolicySigned with symmetric keys. Here only the key has to be loaded and the authorization is only needed for parent objects. If TSS2_POLICY_CB_EXEC_AUTH should not be changed I will add a new callback TSS2_POLICY_CB_EXEC_LOAD. Would this be OK or do you see any problems?
> 
> That seems fine

[This](https://github.com/tpm2-software/tpm2-tss/commit/067dc2a028e131f11f40e37d072c4219e840a5ac) patch seems simpler than whats described and it fixes the problem? Do you have a test?

JuergenReppSIT: > [067dc2a](https://github.com/tpm2-software/tpm2-tss/commit/067dc2a028e131f11f40e37d072c4219e840a5ac) patch seems simpler than whats described and it fixes the problem? Do you have a test?

I have tested PolicySecret with a key by using the command line tools. I will backport the PR and can create an integration test afterwards.

PR #2453, opened 2022-11-01
types: fix define for TPM2_HR_NV_AC

whooo:
Missing TPM2_ prefix for HR_SHIFT, so add it

Fixes tpm2-pytss builds against tpm2-tss master


williamcroberts (approved):

williamcroberts:
It took me a second to understand why the tpm2-tss build didn't fail, duh, it's pre-processor and wasn't used and thus not expanded.

PR #2455, opened 2022-11-02
Fuzz additions

williamcroberts:
- Fix fuzzing targets still being built with CPP, no need C is fine.
- tcti spi helper, make the structure of fn pointers a pointer and not pass by value
- tcti spi helper check that size is to a big enough block of memory
- fuzz transfer interface within the tcti spi framework


Makefile-fuzz.am
```diff
@@ -46,5 +46,11 @@ FUZZ = $(check_PROGRAMS)
 
 DISTCLEANFILES += Makefile-fuzz-generated.am
 
+# Non Autogenerated Fuzz Tests
+noinst_PROGRAMS += test/fuzz/tcti-spi-helper-fuzz-test
+test_fuzz_tcti_spi_helper_fuzz_test_CFLAGS         = -I bill $(FUZZ_CFLAGS) $(FUZZ_CFLAGS) $(AM_CFLAGS)
```
JuergenReppSIT, line 51: @williamcroberts Can -I bill be removed?

williamcroberts: > @williamcroberts Can -I bill be removed?

Hahaha debugging the makefile, but I think it should stay 😂

PR #2460, opened 2022-11-03
3.2 fix pcr setauthvalue

williamcroberts:


JuergenReppSIT (changes requested):
We have an integration test for Esys_PCR_SetAuthValue and Esys_PCR_SetAuthPolicy. I thin this integration test should be moved from optional to destructive for physical TPMs.

williamcroberts:
> We have an integration test for Esys_PCR_SetAuthValue and Esys_PCR_SetAuthPolicy.
Yeah I forgot to cherrypick the test patch back.

> I think this integration test should be moved from optional to destructive for physical TPMs.
I don't know what this means as far as what to do. But the existing tests are already twiddling those values, we're those tests just set up wrong?

JuergenReppSIT:
> > I don't know what this means as far as what to do. But the existing tests are already twiddling those values, we're those tests just set up wrong?

Currently after executing the test  esys-pcr-auth-value.int.c  with a physical TPM a tpm_clear is needed afterwards if the TPM has these commands. I can create a separate PR to move this test to the destructive tests.

williamcroberts:
> > > I don't know what this means as far as what to do. But the existing tests are already twiddling those values, we're those tests just set up wrong?
> 
> Currently after executing the test esys-pcr-auth-value.int.c with a physical TPM a tpm_clear is needed afterwards if the TPM has these commands. I can create a separate PR to move this test to the destructive tests.

Ok I see those, ill merge this and then pull your PRs in

Issue #2454, opened 2022-11-02
CEL events for PCR 0 is included when Fapi_PcrRead is called for other PCR


whooo:
In some tpm2-pytss tests we call Fapi_PcrRead with a PCR > 0, but with latest master all CEL events for PCR 0 are included in the response

whooo:
I should mention that this affects Fapi_Quote as well

williamcroberts:
@JuergenReppSIT I have no idea what to do with this, could you help out?

JuergenReppSIT:
Only events which do not change PCR 0  (e.g. CEL management events) are included if PCR 0 is not in the PCR list. They are related to the firmware events.
@AndreasFuchsTPM If the PCR selection produces a list without firmware events, should we omit these events?

whooo:
I see the EV_NO_ACTION firmware log header as well unless I disable the firmware log parsing

JuergenReppSIT:
yes that's the SPECID event which also is not extended to PCR0.
@AndreasFuchsTPM @whooo @williamcroberts or should we omit all these events if PCR0 is not selected?

williamcroberts:
> @AndreasFuchsTPM @whooo @williamcroberts or should we omit all these events if PCR0 is not selected?

I am not familiar enough with that API to have any reasonable answer, im assuming we're talking about the log file that comes in through the `char **pcrLog` argument?

whooo:
I don't think the SpecID and firmware end events should be included, unless PCR 0 is requested as they are only used when looking at at firmware events.
Perhaps the CEL version event should only be included if the response contains other CEL events? If an application extends for example PCR 23 with it's own data and then requests the events for that does the version event make sense?

PR #2462, opened 2022-11-07
FAPI: Fix usage of management events in event list.

JuergenReppSIT:
CEL management and advisory only events were added to the event list although PCR0 was not selected.
Now these events will only be added if PCR0 is selected. Fixes: #2454

### DIFF
--- a/src/tss2-fapi/ifapi_json_eventlog_serialize.c
+++ b/src/tss2-fapi/ifapi_json_eventlog_serialize.c
@@ -29,6 +29,22 @@
 #include "util/log.h"
 #include "util/aux_util.h"
 
+bool ifapi_pcr_used(uint32_t pcr, const uint32_t *pcr_list,  size_t pcr_list_size)
+{
+    size_t i;
+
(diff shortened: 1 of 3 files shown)


src/tss2-fapi/ifapi_json_eventlog_serialize.c
```diff
@@ -795,10 +791,24 @@ bool ifapi_json_TCG_SPECID_EVENT_cb(
...
         return true;
 
+    if (cb_data->pcr_list_size) {
+        for (i = 0; i < cb_data->pcr_list_size; i++) {
+            if (cb_data->pcr_list[i] == 0)
+                use_pcr0 = true;
+        }
+    } else {
+        use_pcr0 = true;
+    }
```
williamcroberts, line 807: This block of code is repeated, should we make it a function?

JuergenReppSIT: yes that would be better. I will add a function.

williamcroberts (approved):

PR #2473, opened 2022-11-08
3.2.x update mailing list

williamcroberts:


williamcroberts:
Darn this missed 3.2.1-rc0, I forgot to tag it on the milestone. Since this is purely a doc change I think we can take this without needing another round of RC.

PR #2475, opened 2022-11-09
FAPI: Allow keyedhash keys in PolicySigned.

### LINKED ISSUE #2445: FAPI Feature Request: Enable PolicySigned with a KEYEDHASH object.
My use case are interactions with resource-constrained devices (e.g., MCUs in an IoT network) that cannot afford to use public key cryptography. 

Implementation idea:
Because the KEYEDHASH has no public key, the prototype TSS2_RC (*Fapi_CB_Sign) should allow object name (e.g., path as hint) as parameter publicKey instead of the public PEM key. The verification of the returned HMAC should work without problems with TPM2_VerfifySignature with the symmetric key stored in the keystore.

JuergenReppSIT:
* The attribute sign is now possible for sealed objects, to enable the usage of the sealed object as HMAC key in policy signed.
* Instead of the base64 encoded public key now the name of the sealed object will be passed (base64 encoded) to the signature callback
* An integration test is added.

Fixes #2445


src/tss2-fapi/ifapi_policy_callbacks.c
```diff
@@ -495,7 +495,7 @@ ifapi_policyeval_cbauth(
             SAFE_FREE(cb_ctx->object_path);
             fallthrough;
 
-            statecase(cb_ctx->cb_state, POL_CB_READ_OBJECT);
+        statecase(cb_ctx->cb_state, POL_CB_READ_OBJECT);
```
williamcroberts, line 498: put this hunk on it's own patch

williamcroberts (approved):

williamcroberts (changes requested):
just split that hunk out... I was forced by GH to put a comment here... so let me say split that hunk out for the fiftieth time like a jerk :-p

JuergenReppSIT:
> just split that hunk out... I was forced by GH to put a comment here... so let me say split that hunk out for the fiftieth time like a jerk :-p

@williamcroberts Should I create a second commit, just with the fix of indentation? Would that solve your problem with gh?

JuergenReppSIT:
@williamcroberts I have added a second commit. Can the PR be merged now?

williamcroberts (approved):

PR #2478, opened 2022-11-10
Generalize GetCapability for TPM2_CAP_VENDOR

### LINKED ISSUE #2449: SAPI/ESYS_GetCapability / TPMU_CAPABILITY unmarshalling for TPM2_CAP_VENDOR_PROPERTY harded to TPML_INTEL_PTT_PROPERTY
As I just learned, whenever TPM2_GetCapability is called for TPM2_CAP_VENDOR_PROPERTY the response is hardcoded to decode a TPML_INTEL_PTT_PROPERTY in the response.
If there was a second vendor who wanted to have their vendor-specific capabilities managed by tpm2-tss, that's currently not possible.

These are the relevant code paths:
https://github.com/tpm2-software/tpm2-tss/blob/9288970a3e657cdee85d08d3813199ec864de3ad/src/tss2-sys/api/Tss2_Sys_GetCapability.c#L78-L81
https://github.com/tpm2-software/tpm2-tss/blob/95b8a529af6671f464b67116766435960dfe7f03/src/tss2-mu/tpmu-types.c#L470-L482
https://github.com/tpm2-software/tpm2-tss/blob/8e64f13a1ecb95fead3b1b6d260d9a2d83008b2e/include/tss2/tss2_tpm2_types.h#L1222-L1237

The problem is now that I don't know how we could change the API in order to support vendor-specific capabilities for multiple vendors.

The only idea I currently have is to have change the member of TPMU_CAPABILITIES to something like TPMT_PROPERTIES_VENDOR which in turn would have the UINT32 Vendor String as selector and another union for the different vendors.
For Esys, ESYS could go ahead and query the Vendor string. For SAPI this would be less good, because we cannot make two requests to this.
Maybe a third entry to the TPMT_PROPERTIES_VENDOR could be a TPM2B with the plain bytes if the selector is empty and SAPI would only fill this in whereas ESYS could then further decode to each of the vendor string.

AndreasFuchsTPM:
Fixes the PTT hardcoding issue.

Unfortunately, TPMS_CAPABILTIY and TPMU_CAPABILITY do not have a size field.
I added the TPM2B for vendor data as agreed.
But for the size-calculation I had to just use up the remainder of the response buffer during Tss2_Sys_GetCapability.
This is kind of ugly.

Also, this makes it impossible to support the new TPM2B in the usual Unmarshalling functions of TPMU_CAPABILITIES, since we don't know in general if the Capability really goes until the end of the buffer passed into the unmarshal. Thus, I intentionally did not include it there.

Also, I rolled the soname for tss2-sys, tss2-esys and tss2-mu accordingly.

I also added a soname roll for tss2-fapi which was previously forgotten after the change to eventlog, I guess.

P.S. This is untested for the moment unfortunately since I have no possibility to do so right now (traveling).

Fixes: #2449


src/tss2-sys/api/Tss2_Sys_GetCapability.c
```diff
@@ -75,7 +78,31 @@ TSS2_RC Tss2_Sys_GetCapability_Complete(
...
+       if this is a vendor cap, we fill a tpm2b inside the struct buffer.*/
+    rval = Tss2_MU_UINT32_Unmarshal(&ctx->cmdBuffer[ctx->nextData],
+                                  ctx->maxCmdSize - ctx->nextData,
+                                  NULL,
+                                  &cap);
+    if (rval)
+        return rval;
+
+    /* Size the capabilityData is the end of response we can copy the remainder
+       of the response in cas of cap_vendor */
```
williamcroberts, line 92: spelling: s/cas/case

src/tss2-sys/api/Tss2_Sys_GetCapability.c
```diff
@@ -75,7 +78,31 @@ TSS2_RC Tss2_Sys_GetCapability_Complete(
...
+            return TSS2_MU_RC_BAD_SIZE;
+
+        capabilityData->capability = cap;
+        capabilityData->data.vendor.size = ctx->maxCmdSize - ctx->nextData;
+        memcpy(&capabilityData->data.vendor.buffer[0],
+               &ctx->cmdBuffer[ctx->nextData],
+               capabilityData->data.vendor.size);
+        ctx->nextData = ctx->maxCmdSize;
+        return TSS2_RC_SUCCESS;
+    } else
```
williamcroberts, line 104: No need for the else with return. Even though we lose the tail call optimization, I would be in favor of the flow being:

```C
if (cap != TPM2_CAP_VENDOR_PROPERTY)
        return Tss2_MU_TPMS_CAPABILITY_DATA_Unmarshal(ctx->cmdBuffer,
                                                  ctx->maxCmdSize,
                                                  &ctx->nextData,
                                                  capabilityData);

/* now do complicated vendor stuff */
...
```

src/tss2-sys/api/Tss2_Sys_GetCapability.c
```diff
@@ -75,7 +78,31 @@ TSS2_RC Tss2_Sys_GetCapability_Complete(
...
+        return rval;
+
+    /* Size the capabilityData is the end of response we can copy the remainder
+       of the response in cas of cap_vendor */
+    if (cap == TPM2_CAP_VENDOR_PROPERTY) {
+        if (ctx->maxCmdSize - ctx->nextData > sizeof(capabilityData->data.vendor.buffer))
+            return TSS2_MU_RC_BAD_SIZE;
+
+        capabilityData->capability = cap;
+        capabilityData->data.vendor.size = ctx->maxCmdSize - ctx->nextData;
```
williamcroberts, line 98: I think we need to subtract sizeof(cap) to account for that we've marshaled this out.

williamcroberts:
> Unfortunately, TPMS_CAPABILTIY and TPMU_CAPABILITY do not have a size field. I added the TPM2B for vendor data as agreed. But for the size-calculation I had to just use up the remainder of the response buffer during Tss2_Sys_GetCapability. This is kind of ugly.

Yeah that is... but I think saying that its too big isn't an issue. So long as folks go to unmarshal the expected data structure they don't expect to consume all the bytes. We would want to document this well.

williamcroberts:
I have this over on https://github.com/tpm2-software/tpm2-tss/pull/2488. I'm going to close this to avoid confusion

PR #2484, opened 2022-11-16
[RFC] Add tests for vendor

williamcroberts:
@AndreasFuchsTPM for completeness wanted to show the state of the current master:

Unmarshaling a TPMU_CAPABILITIES fails for the last field, which happened to be intelPttProperty. Not sure about marshaling, seems to suffer from the same bug. The bug is that @tstruk can't count :-p. The magic macros work for up to 11 fields, but their was 12. The direct functions Tss2_MU_TPML_INTEL_PTT_PROPERTY_Unmarshal seem to work.

So right now their was no way to call GetCap and get a valid unmarshalled intelPttProperty.

I think we want to merge this, as it keeps it "fixed" for the vendor field even though we marshal that by hand. 

I have some code on another branch I need to clean up, but it does this all with the vendor buffers, and Ill make the tests backwards compat and cherry-pick it before the vendor tpm2b change so we can ensure it's backwards compat.

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -31,8 +31,6 @@ check-programs: $(check_PROGRAMS)
 check_PROGRAMS = $(TESTS_UNIT) $(TESTS_INTEGRATION)
 TESTS = $(TESTS_UNIT) $(TESTS_INTEGRATION)
 
-XFAIL_TESTS =
-
 if UNIT
 check_PROGRAMS += test/helper/tpm_cmd_tcti_dummy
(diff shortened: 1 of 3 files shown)


AndreasFuchsTPM:
test/unit/esys-vendor.c is missing

williamcroberts:
> test/unit/esys-vendor.c is missing

Haha of course it is.

williamcroberts:
Closing this was just for discussion, final PR heading up soon

PR #2487, opened 2022-11-16
FAPI: Fix double free if keystore is corrupted.

JuergenReppSIT:
If the keystore contains invalid JSON data for an object a double free during the object cleanup occurs.
The corresponding value in the object is now set to NULL after the first free.

### DIFF
--- a/src/tss2-fapi/ifapi_keystore.c
+++ b/src/tss2-fapi/ifapi_keystore.c
@@ -644,6 +644,7 @@ ifapi_keystore_load_finish(
     if (jso)
         json_object_put(jso);
     LOG_TRACE("Return %x", r);
+    object->rel_path = NULL;
     SAFE_FREE(keystore->rel_path);
     return r;
 }


src/tss2-fapi/ifapi_keystore.c
```diff
@@ -644,6 +644,7 @@ ifapi_keystore_load_finish(
     if (jso)
         json_object_put(jso);
     LOG_TRACE("Return %x", r);
+    object->rel_path = NULL;
     SAFE_FREE(keystore->rel_path);
```
AndreasFuchsTPM, line 648: This is dead code now, so we should remove it.

AndreasFuchsTPM (commented):
This would also need to be backported, right ?

src/tss2-fapi/ifapi_keystore.c
williamcroberts: LGTM on line 632 object->_rel_path is assigned the value of keystore->rel_path (pointer) for the purpose of calling ifapi_json_IFAPI_OBJECT_deserialize. Which if that fails we free the keystore->rel_path and leave object->_rel_path dangling which is then freed later in ifapi_cleanup_ifapi_object.

williamcroberts (approved):
I'm going to backport this to the 3.2.x branch, looks like we will have an rc1.

src/tss2-fapi/ifapi_keystore.c
AndreasFuchsTPM: > > This is dead code now, so we should remove it.
> 
> @AndreasFuchsTPM Why is there dead code?

my mistake. I misread the code

PR #2500, opened 2022-11-24
test/fapi: test Fapi_SetBranchCB NULL

williamcroberts:
Somewhere between 3.0.0 and 3.1.0 we fixed double free but, but we did not maintain commit atomicity and therfore my efforts to bisect were thwarted. I checked a few patches by hand and couldn't find the source of the fix. However, add a regression test.


williamcroberts:
itrs between 3.0.1 (works) and 3.0.0 (broken)

williamcroberts:
Fixed in e9be3981bc20 FAPI: Fix policy execution. Which is on 3.0.x
660e444, backports for 2.4.3+
f695b9ef 3.1.0+

Finally got to the bottom of it. Was preventing tests on pytss.

JuergenReppSIT (approved):

PR #2502, opened 2022-11-28
esys: add SM4 algorithm support.

mayuanchenma:
add sm4 algorithm support.

### DIFF
--- a/src/tss2-esys/esys_crypto_ossl.c
+++ b/src/tss2-esys/esys_crypto_ossl.c
@@ -1212,6 +1212,164 @@ iesys_cryptossl_sym_aes_decrypt(uint8_t * key,
     return r;
 }
 
+#if HAVE_EVP_SM4_CFB && !defined(OPENSSL_NO_SM4)
+/** Encrypt data with SM4.
+ *
+ * @param[in] key key used for SM4.
+ * @param[in] tpm_sym_alg SM4 type in TSS2 notation (must be TPM2_ALG_SM4).
+ * @param[in] key_bits Key size in bits.
+ * @param[in] tpm_mode Block cipher mode of opertion in TSS2 notation (CFB).
+ *            For parameter encryption only CFB can be used.
+ * @param[in,out] buffer Data to be encrypted. The encrypted date will be stored
+ *                in this buffer.
+ * @param[in] buffer_size size of data to be encrypted.
+ * @param[in] iv The initialization vector.
+ * @retval TSS2_RC_SUCCESS on success, or TSS2_ESYS_RC_BAD_VALUE and
+ * @retval TSS2_ESYS_RC_BAD_REFERENCE for invalid parameters,
+ * @retval TSS2_ESYS_RC_GENERAL_FAILURE for errors of the crypto library.
+ */
+TSS2_RC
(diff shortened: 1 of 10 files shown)


src/tss2-esys/esys_crypto_ossl.h
```diff
@@ -109,6 +109,28 @@ TSS2_RC iesys_cryptossl_sym_aes_decrypt(
     uint8_t *iv,
     void *userdata);
 
+#if HAVE_EVP_SM4_CFB && !defined(OPENSSL_NO_SM4)
```
williamcroberts, line 112: Do we really need the check with OPENSSL_NO_SM4, wont, the check to determine HAVE_EVP_SM4_CFB be undefined if OpenSSL is built without SM4?

src/tss2-esys/esys_crypto_ossl.c
```diff
@@ -1212,6 +1212,164 @@ iesys_cryptossl_sym_aes_decrypt(uint8_t * key,
...
+
+    if (tpm_sym_alg != TPM2_ALG_SM4) {
+        goto_error(r, TSS2_ESYS_RC_BAD_VALUE,
+                   "SM4 decrypt called with wrong algorithm.", cleanup);
+    }
+
+    if (key_bits == 128 && tpm_mode == TPM2_ALG_CFB)
+        cipher_alg = EVP_sm4_cfb128();
+    else {
+
```
williamcroberts, line 1339: extra newline

williamcroberts (changes requested):
Overall looks good, i'd like to see a test that includes an explicit usageof SM4 and skipped if the TPM doesn't support it. We should probably look at building swtpm on the test images with SM4 support as well.

src/tss2-esys/esys_crypto_ossl.h
mayuanchenma: the same as SM3 check. @williamcroberts

mayuanchenma:
@williamcroberts

src/tss2-esys/esys_iutil.c
```diff
@@ -866,7 +928,7 @@ iesys_decrypt_param(ESYS_CONTEXT * esys_context)
         return_if_error(r, "Setting plaintext");
     } else {
         return_error(TSS2_ESYS_RC_BAD_VALUE,
-                     "Invalid symmetric algorithm (should be XOR or AES)");
+                     "Invalid symmetric algorithm (should be XOR or AES or SM4)");
```
williamcroberts, line 931: Change the wording here to:
"Should be XOR, AES, or SM4"

src/tss2-esys/esys_crypto_ossl.c
williamcroberts: Still need to remove this newline

williamcroberts (changes requested):

williamcroberts:
> @williamcroberts

Fix those issues for the comments I left and assuming the CI is happy we can pull it in after 4.0.0 release so it will be for 4.1.0+

williamcroberts:
I think you'll also want to rebase this ontop of current master:
```bash
git pull --rebase origin master
```

williamcroberts (approved):

PR #2506, opened 2022-12-04
FAPI: Fix deviation from specification in Fapi_GetDescription.

JuergenReppSIT:
If no description exists Fapi_GetDescription returns NULL. The FAPI specification states:
If no description is present, description SHALL be set to an empty string.


williamcroberts (approved):

williamcroberts:
The question now is do I follow the letter of the release process and conduct another RC cycle.

williamcroberts:
backported here:
  - https://github.com/tpm2-software/tpm2-tss/pull/2511

PR #2508, opened 2022-12-04
FAPI: Fix return value if policy does not exist.

JuergenReppSIT:
FAPI returns FAPI_RC_IO_ERROR if a policy does not exist in keystore. The Spec requires the return value FAPI_RC_BAD_PATH.

### DIFF
--- a/src/tss2-fapi/ifapi_policy_store.c
+++ b/src/tss2-fapi/ifapi_policy_store.c
@@ -166,6 +166,10 @@ ifapi_policy_store_load_async(
         }
     }
 
+    if (!ifapi_io_path_exists(abs_path)) {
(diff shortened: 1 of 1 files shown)


williamcroberts:
@JuergenReppSIT should we have pulled this for 4.0?

williamcroberts (approved):

JuergenReppSIT:
> @JuergenReppSIT should we have pulled this for 4.0?

It would be better. @AndreasFuchsTPM what would you say?

williamcroberts:
@JuergenReppSIT will pull this for next RC cycle

JuergenReppSIT:
> @JuergenReppSIT will pull this for next RC cycle

@williamcroberts @AndreasFuchsTPM I'm not sure whether we really need an extra rc cycle for this PR.
Perhaps we can add the PR later when the FAPI Spec is updated?

Issue #2515, opened 2022-12-10
ESAPI: extra nonceTPM parameter


IKhanov:
According to the latest version of [ESAPI specification](https://trustedcomputinggroup.org/wp-content/uploads/TSS_ESAPI_v1p0_r14_pub10012021.pdf) there's no `nonceTPM` parameter for `Esys_*` commands except `Esys_TRSess_GetNonceTPM`:

1. For `Esys_PolicySigned`
```
TSS2_RC
Esys_PolicySigned(
    ESYS_CONTEXT *esysContext,
    ESYS_TR authObject,
    ESYS_TR policySession,
    ESYS_TR shandle1,
    ESYS_TR shandle2,
    ESYS_TR shandle3,
    const TPM2B_NONCE *nonceTPM,
    const TPM2B_DIGEST *cpHashA,
    const TPM2B_NONCE *policyRef,
    INT32 expiration,
    const TPMT_SIGNATURE *auth,
    TPM2B_TIMEOUT **timeout,
    TPMT_TK_AUTH **policyTicket)
{
...
```

2. For `Esys_PolicySecret`:
```
TSS2_RC
Esys_PolicySecret(
    ESYS_CONTEXT *esysContext,
    ESYS_TR authHandle,
    ESYS_TR policySession,
    ESYS_TR shandle1,
    ESYS_TR shandle2,
    ESYS_TR shandle3,
    const TPM2B_NONCE *nonceTPM,
    const TPM2B_DIGEST *cpHashA,
    const TPM2B_NONCE *policyRef,
    INT32 expiration,
    TPM2B_TIMEOUT **timeout,
    TPMT_TK_AUTH **policyTicket)
{
...
```

`const TPM2B_NONCE *nonceTPM` should not be there. Please, correct me if I'm wrong.

AndreasFuchsTPM:
I'm thinking right now, if we added the nonceTPM for a good reason here to be honest.

I think the idea is to get the nonceTPM from the session, then do you calculation and then add it here in the call.
Maybe @JuergenReppSIT still remembers.

Also @idesai I think you asked for ESYS_TR_GetTPMNone in the first place for his use case ? Do you know ?

AndreasFuchsTPM:
@IKhanov Also please note, that in the ESAPI spec, Sections 9 and 10 are normative. Section 11 is only informative.
If I see things correctly here, then Section 9+10 contradict Section 11 here.

And tpm2-tss complies with Section 9+10.

[...]

PR #2516, opened 2022-12-10
ESAPI: remove nonceTPM parameter according to ESAPI specification

IKhanov:
See #2515.

Signed-off-by: Ilya Khanov <ilya.khanovvv@gmail.com>


williamcroberts (approved):

williamcroberts:
This will bump soname to 1.0.0

williamcroberts:
> @AndreasFuchsTPM not sure if we want to drop this parameter or support NULL and update the spec. This would have impact on systemd.

Actually this doesn't affect systemd, I'm they dont use these APIs

JuergenReppSIT (approved):

AndreasFuchsTPM:
See my comment https://github.com/tpm2-software/tpm2-tss/issues/2515#issuecomment-1348871765

I would be against a merge and a quick update to the TCG specs (removal of headers).

williamcroberts (changes requested):
After the lengthy discussion on the issue ticket it was decided we wanted the caller to retain control of this input parameter

PR #2522, opened 2022-12-14
FAPI: Improve documentation fapi-profile.

JuergenReppSIT:
An example for usage of the PCR bank SHA1 was added.


JuergenReppSIT:
I have added the PR to milestone 4.0 because the profile parameter `session_symmetric` was not mentioned.

williamcroberts (approved):

PR #2525, opened 2022-12-16
FAPI: Add http return code handling

AndreasFuchsTPM:
Add handling of the http return code to the downloading of the EK intermediate CA certificates.

Signed-off-by: Andreas Fuchs <andreas.fuchs@infineon.com>

### DIFF
--- a/src/tss2-fapi/ifapi_curl.c
+++ b/src/tss2-fapi/ifapi_curl.c
@@ -356,6 +356,7 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
                           size_t *buffer_size) {
     int ret = -1;
     struct CurlBufferStruct curl_buffer = { .size = 0, .buffer = NULL };
+    long http_code;
 #ifdef CURLU_ALLOW_SPACE
     CURLU *urlp = NULL;
 #endif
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/ifapi_curl.c
```diff
@@ -435,6 +436,12 @@ ifapi_get_curl_buffer(unsigned char * url, unsigned char ** buffer,
         goto out_easy_cleanup;
     }
 
+    curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &http_code);
+    if (http_code >= 400) {
```
williamcroberts, line 440: `curl_easy_perform` returns `CURLE_HTTP_RETURNED_ERROR` if the error code is `>= 400`, I don't see how this check adds anything?

See:
  - https://curl.se/libcurl/c/curl_easy_perform.html
  - https://curl.se/libcurl/c/libcurl-errors.html

williamcroberts (changes requested):

src/tss2-fapi/ifapi_curl.c
AndreasFuchsTPM: Hmmm... That's weird.
I have a system here where the CA is not online and I get this:
```
ERROR:fapi:src/tss2-fapi/ifapi_curl.c:193:ifapi_curl_verify_ek_cert() Failed to convert intermediate certificate to X509 format. (size=1245): (then the segfault)
```

AFAIK this should not happen if curl_easy_perform() returns an error.

With the patch, it does work.

JuergenReppSIT (approved):
I could also observe this behaviour with libcurl4-openssl-dev:amd64. An error is only returned if CURLOPT_FAILONERROR is set. So the PR should be merged.

src/tss2-fapi/ifapi_curl.c
williamcroberts: Sounds like a bug in curl, but sometimes we just gotta code around those :-p

williamcroberts (approved):

williamcroberts:
> I could also observe this behaviour with libcurl4-openssl-dev:amd64. An error is only returned if CURLOPT_FAILONERROR is set. So the PR should be merged.

Should we have set that option instead?

JuergenReppSIT:
> Should we have set that option instead?

imho the version from the pr is ok and it's not necessary.

PR #2526, opened 2022-12-16
3.2.x: add NULL check on doLogBlob

williamcroberts:


williamcroberts:
This is optional to backport, ill let others decide.

JuergenReppSIT:
> This is optional to backport, ill let others decide.

It's a good idea. I would vote for the backport because it it prevents further mistakes.

PR #2535, opened 2023-01-05
ESYS: Fix wrong typename for sha512 on mbedtls

babyraging:


williamcroberts (approved):

JuergenReppSIT:
@AndreasFuchsTPM Is the backport needed? 3.2.x does not have this bug.

PR #2549, opened 2023-01-24
tcti_spi_helper_transmit: ensure FIFO is accessed only after TPM repo…

jferreir:
tcti_spi_helper_transmit: ensure TPM FIFO is only accessed after TPM reports commandReady is set to 1 in the TPM_STS register. 

The current implementation is accessing the FIFO with the commandReady not set to 1 which leads to problems that manifest as unrelated SPI timings issues.
image


JuergenReppSIT:
@AndreasFuchsTPM backport? We do not have the tcti spi helpers in 3.2.x.

wxleong (approved):

AndreasFuchsTPM (approved):

PR #2566, opened 2023-02-13
FAPI: Support download of AMD certificates

JuergenReppSIT:
* For AMD TPMs the EK certificates will be downloaded during provisioning.
* The corresponding unit test is extended.
    
Signed-off-by: Juergen Repp <juergen_repp@web.de>

PR #2579, opened 2023-03-03
Fixed the error in the return value of DIO_READ.

jinlun145:
Function 'BIO_read' return 'int'
if BIO_read return -1
(r = BIO_read(bio, binary_data, unescape_len)) <= 0    is  false
It's not up to expectations

Signed-off-by: jinlun145 <869793317@qq.com>


JuergenReppSIT:
@jinlun145 Thank you for the PR.  The file ifapi_get_intl_cert.c has been renamed in PR https://github.com/tpm2-software/tpm2-tss/pull/2566 which enables the download of AMD certificates. I have added your changes to the file ifapi_get_web_cert.c. So the PR can be closed.

PR #2618, opened 2023-05-15
TCTI: Implement tcti-i2c-helper and tcti-i2c-ftdi

wxleong:
Implement the tcti modules `tcti-i2c-ftdi` and `tcti-i2c-helper`:
- **`tcti-i2c-ftdi`:** The I2C TCTI FTDI can be used for communication with an I2C-based TPM over the FTDI MPSSE USB to I2C bridge. The FTDI module utilizes the `tcti-i2c-helper` library for handling the PTP I2C protocol and the `libftdi-dev` library for handling the USB to I2C communication.
- **`tcti-i2c-helper`:** The I2C TCTI helper can be used for TPM communication over I2C e.g. in embedded systems. It uses user supplied methods for I2C and timing operations in order to be platform independent. These methods are supplied to `Tss2_Tcti_I2c_Helper_Init` via the `TSS2_TCTI_I2C_HELPER_PLATFORM` struct. `tcti-i2c-ftdi` is an example implementation that utilizes the I2C TCTI helper.

### DIFF
--- a/include/tss2/tss2_tcti_i2c_ftdi.h
+++ b/include/tss2/tss2_tcti_i2c_ftdi.h
@@ -1,0 +1,42 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (c) 2023 Infineon Technologies AG
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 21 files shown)


include/tss2/tss2_tcti_i2c_ftdi.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2023 Infineon Technologies AG
```
joholl, line 2: Use SPDX Licence header. Note that the license you provided is not BSD-2.

/*
 * Copyright (c) 2023 Infineon Technologies AG
 * All rights reserved.
 */

wxleong: fixed all license header.

AndreasFuchsTPM:
@wxleong @joholl Should we include CRC checking and retransmission right now from the start or do we want to add this only in a later version ?
Given problems seen on Linux kernel with this, maybe it would make sense.

joholl:
@AndreasFuchsTPM @wxleong CRC checking is a nice feature, but not crucial as a first patch, imho. What I find important tho, is taking guard times into account. At the very least, we have to wait the default of 250µs (although that is technically not spec compliant).

On top of that, we should not duplicate the TIS driver code, which is now in both tcti-i2c-helper and tcti-spi-helper. That code is a beast, anyway. Encapsulating that (and coming up with a sensible API) might not be trivial. A glance at how the Linux kernel does it certainly helps.

@AndreasFuchsTPM We could also merge with TIS logic baked in (as it is in tcti-spi-helper) and pull out common code as part of later work, what do you think?

wxleong:
@joholl @AndreasFuchsTPM  CRC and guard times are now implemented. We can deal with the common code in a separate PR.

joholl:
On first glance, the CRC and guard time implementation looks good. I did not have a look on the I2C layer, so just make sure you are not using the "Repeated Start" feature of I2C. If you do use that, there is another (imho nonsensical) guard time.

wxleong:
@joholl Indeed, the "Repeated Start" feature is not utilized here.

AndreasFuchsTPM (approved):

PR #2632, opened 2023-06-14
Simplify windows build

philippun1:
This MR simplifies the windows build with these updates:

- add a openssl.props to replace the openssl path magic strings
- update the platform toolset to ClangCL, which aligns it with the tpm2-provider project. (Also v141_clang_c2 is experimental)
- generalize the windows target platform to 10.0 (which will make it pick the latest installed one)
- update the VS version in the .sln file to 17 (the same as in the tpm2-provider project)
- update image to VS 2019 in .appveyor.yml
- update INSTALL.md to reflect ClangCL / VS 2019 changes

No changes to actual source code are made.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@ version: '{build}'
 pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
-image: Visual Studio 2017
+image: Visual Studio 2019
 configuration:
   - Debug
   - Release
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -90,7 +90,7 @@ Building should be as simple as loading the tpm2-tss solution (tpm2-tss.sln) wit
 a compatible and properly configured version of Visual Studio 2019+ and pressing the 'build' button.
 
 ### References
-Visual Studio 2017 with "Clang for Windows": https://blogs.msdn.microsoft.com/vcblog/2017/03/07/use-any-c-compiler-with-visual-studio/
+Visual Studio 2019+ with "Clang for Windows": https://learn.microsoft.com/en-us/cpp/build/clang-support-msbuild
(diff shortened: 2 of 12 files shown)


Superhepper:
This works nicely! I tested it with Visual Studio 2022 on Windows 11. If you do not mind I would like to add a suggestion to this.

I would like to see a post build step that creates the "VERSION" file. I did an experiment and added a 'Post-Build Event' to the mu project which just executed the following ```call git describe --tags --always --dirty> "$(SolutionDir)VERSION"````. By doing this the solution will behave similar to the VERSION file is created in ```bootstrap```. And by doing this as Post-Build it should still be possible to build everything even though git is not installed.


And in a future PR it would really nice to see if the version information could be built into LIBs and DLLs them self's using resource files.

Superhepper:
I think you will need to add the following lines:
```
environment:
  matrix:
    - toolchain: clang
 ```
 to the ```.appveyor.yml``` file in order for the tests to work. But I can be wrong.

philippun1:
@Superhepper Updating the image to VS 2019 did the trick.
I also removed the .gitignore. It is only being generated with the makefile though, so it will not be generated if compiling under Windows.

Superhepper (approved):

Superhepper:
I think you might also need to update the documentation in INSTALL.md

AndreasFuchsTPM:
In general this looks good to me and I highly appreciate the work.
One note though:
It seems like not all commits will pass CI, since the Appveyor is updated in a later commit than the former. I might be wrong here.

Just in general, I want to make sure that each and every commit would be able to pass CI.

philippun1:
@AndreasFuchsTPM I rebased onto latest master, did all the tests run now?

philippun1:
@Superhepper Thanks for the clarification. I squashed the appveyor changes into the first commit.

Superhepper (approved):

AndreasFuchsTPM (approved):

PR #2636, opened 2023-06-26
fix  Uninitialized variable   TPMS_POLICY *policy

bearxy123:
fix  Uninitialized variable   TPMS_POLICY *policy


JuergenReppSIT:
The variable policy is initialized in the if statement after the declarations. So there in no need to initialize it in the declaration.

bearxy123:
> The variable policy is initialized in the if statement after the declarations. So there in no need to initialize it in the declaration.

PR #2641, opened 2023-07-01
esys tests: Fix layer check for TPM2_RC_COMMAND_CODE.

JuergenReppSIT:
Two integration tests failed for tests with /dev/tpmrm0 because the layer of the return code was not checked.

### DIFF
--- a/test/integration/esys-act-set-timeout.int.c
+++ b/test/integration/esys-act-set-timeout.int.c
@@ -75,7 +75,9 @@ test_esys_act_set_timeout(ESYS_CONTEXT * esys_context)
         }
     }
     /* If the TPM doesn't support it return skip */
-    if (r == TPM2_RC_COMMAND_CODE)
+    if ((r == TPM2_RC_COMMAND_CODE) ||
+        (r == (TPM2_RC_COMMAND_CODE | TSS2_RESMGR_RC_LAYER)) ||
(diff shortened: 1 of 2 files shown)


joholl:
Looks good and fixes the issue.

I did a quick search and found a few more potentially problematic checks for `TPM2_RC_COMMAND_CODE`.

This one library code. I'm not sure if there is some FAPI magic or if this is a bug.
https://github.com/tpm2-software/tpm2-tss/blob/11d9d056fb3b559646eaa88995a4db822fa8af3b/test/integration/fapi-key-create-policies-sign.int.c#L165-L171

These are not executed when testing against the device. IMO we don't have to fix them:

https://github.com/tpm2-software/tpm2-tss/blob/11d9d056fb3b559646eaa88995a4db822fa8af3b/test/integration/sys-encrypt-decrypt-2.int.c#L55

https://github.com/tpm2-software/tpm2-tss/blob/11d9d056fb3b559646eaa88995a4db822fa8af3b/test/integration/sys-encrypt-decrypt.int.c#L59

https://github.com/tpm2-software/tpm2-tss/blob/11d9d056fb3b559646eaa88995a4db822fa8af3b/test/integration/sys-policy-authorizeNV.int.c#L232

joholl (approved):

JuergenReppSIT:
> This one is library code. I'm not sure if there is some FAPI magic or if this is a bug.

In several integration tests the TSS internal macro number_rc was added which normally should not be part of an integration test. To fix the bug I will create a PR.

PR #2642, opened 2023-07-03
fapi tests: Fix layer check for TPM2_RC_COMMAND_CODE.

JuergenReppSIT:
A bit mask for the layer of the return code was not used in one case of the  integration test fapi-key-create-policies-sign.int.c


test/integration/fapi-key-create-policies-sign.int.c
```diff
@@ -165,7 +165,7 @@ test_fapi_key_create_policies_sign(FAPI_CONTEXT *context)
     if (number_rc(r) == TPM2_RC_PP) {
```
joholl, line 165: Sorry to be like that - but that seems to be a bug, too...

joholl (approved):

PR #2646, opened 2023-07-08
README: replace wall of text with a table

joholl:
I stripped the first section down a lot to make it as clean as possible. I stripped some information that is crucial (because only what is crucial should be at the beginning of the README).

I might also add a `doc/tcti.md` in a later PR because this has developed a lot and could be documented better.

If you wonder about the non-breaking space in the table header: those are a hack for controlling the column width. Otherwise the spec column would take over.


JuergenReppSIT:
@joholl yes you are right about the async calls and with the new tcti.md it will be very good.

JuergenReppSIT (approved):

PR #2647, opened 2023-07-08
doc: add markdown file about TCTIs

joholl:
Depends on #2646.


joholl:
@JuergenReppSIT Ok, this is done now.

The only thing I could not verify is calling something like `tpm2_getrandom --tcti="cmd:ssh 192.168.178.123 tpm2_send --tcti='device:/dev/tpm0'"`to verify the ssh example.

JuergenReppSIT:
> @JuergenReppSIT Ok, this is done now.
> 
> The only thing I could not verify is calling something like `tpm2_getrandom --tcti="cmd:ssh 192.168.178.123 tpm2_send --tcti='device:/dev/tpm0'"`to verify the ssh example.

@joholl  I could verify it with an 192.168.... address after i changed the group of the remote /dev/tpm0 to tss. With /dev/tpmrm0 it worked at once. So perhaps it would be better to use /dev/tpmrm0 in the examples to avoid problems if someone tries it out?

joholl:
@JuergenReppSIT done :)

Akuli:
I am new to TPM and I think this doc is awesome :)

JuergenReppSIT (approved):
Very good, thank you!

PR #2652, opened 2023-07-11
TCTI-SPI-LT2GO: Change VID and PID to final values

PaulKissinger:
Change VID and PID to final values

Signed-off-by: Paul Kissinger <paul.kissinger@letstrust.de>


joholl (approved):
LGTM

Issue #2664, opened 2023-07-18
Esys_TR_SetAuth should strip trailing 0's


williamcroberts:
When the TPM calculates an HMAC, the trailing zeros are removed but tpm2-tss function Esys_TR_SetAuth doesn't remove them which causes the HMACs to be different. It should strip trailing zeros, this is causing a bug in systemd:
   - https://github.com/systemd/systemd/issues/27716

JuergenReppSIT:
@williamcroberts Before creating the PR I tested create primary with the following inSensitive:
```
  TPM2B_AUTH authValuePrimary = {
        .size = 6,
        .buffer = {1, 2, 3, 4, 5, 0}
    };

    TPM2B_SENSITIVE_CREATE inSensitivePrimary = {
        .size = 0,
        .sensitive = {
            .userAuth = authValuePrimary,
            .data = {
                 .size = 0,
                 .buffer = {0},
             },
        },
    };
```
and it worked with the new Esys_TR_SetAuth.

williamcroberts:
We need to update some tests, left comments on your commit

JuergenReppSIT:
> We need to update some tests, left comments on your commit

I will update the integration tests to check the new Esys_TR_SetAuth. 

> Probably also need to do this in the in-sensitive portions, I think ESAPI remembers those

Yes ESAPI remembers those. ESAPI did use the HMAC key with the trailing zero for the HMAC computation
if Esys_TR_SetAuth was not called.  After a call of Esys_TR_SetAuth the HMAC key without the trailing is used.
In both cases the same HMAC was computed  by ESAPI (zero padding of the key for HMAC computation?)
If ESYS_TR_PASSWORD is used instead of a session also both cases work.
I wonder therefore what exactly was the error caused by ESYS_TR_SetAuth with a trailing zero?

JuergenReppSIT:
I have added an integration test which shows that there is no difference between an auth value with and without a trailing zero.
The error in systemd seems to be caused by an HMAC error in StartAuthSession:
```
71.581486] testsuite-70.sh[4385]: WARNING:esys:src/tss2-esys/api/Esys_StartAuthSession.c:391:Esys_StartAuthSession_Finish() Received TPM Error
[   71.582210] testsuite-70.sh[4385]: ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:136:Esys_StartAuthSession() Esys Finish ErrorCode (0x0000098e)
```
[...]

Issue #2660, opened 2023-07-13
Is FAPI supposed to work without `HAVE_CONFIG_H`?


Akuli:
In a project I am working on, `#include <config.h>` tends to pick up whatever garbage is already in the include path and cannot be used. So I added a file that `#undef`s `HAVE_CONFIG_H`.

Turns out that wasn't enough, and I also had to include `<ctype.h>` in my new file:

https://github.com/tpm2-software/tpm2-tss/blob/74e8294ca7b0280957a518c1bab0266510962870/src/tss2-fapi/ifapi_keystore.c#L7-L11

Are the standard-library includes supposed to be inside the ifdef?

JuergenReppSIT:
> Are the standard-library includes supposed to be inside the ifdef

No I will fix it.

PR #2665, opened 2023-07-18
esys: remove trailing zeros in auth value.

JuergenReppSIT:
When TPM Calulates the HMAC trailing zeros are rmoved. Therfore the
trailing zeros are removed in Esys_TR_SetAuth, Esys_Create, Esys_CreateLoded,
Esys_NV_DefineSpace, and Esys_CreatePrimary.
The auth value is also adapted in the ChangeAuth programs

Addresses  #2664

### DIFF
--- a/src/tss2-esys/api/Esys_HierarchyChangeAuth.c
+++ b/src/tss2-esys/api/Esys_HierarchyChangeAuth.c
@@ -26,11 +26,13 @@ static void store_input_parameters (
     const TPM2B_AUTH *newAuth)
 {
     esysContext->in.HierarchyChangeAuth.authHandle = authHandle;
-    if (newAuth == NULL)
+    if (newAuth == NULL) {
         memset(&esysContext->in.HierarchyChangeAuth.newAuth, 0,
                 sizeof(esysContext->in.HierarchyChangeAuth.newAuth));
-    else
+    } else {
         esysContext->in.HierarchyChangeAuth.newAuth = *newAuth;
+        iesys_strip_trailing_zeros(&esysContext->in.HierarchyChangeAuth.newAuth);
+    }
 }
 
 /** One-Call function for TPM2_HierarchyChangeAuth
@@ -175,7 +177,9 @@ Esys_HierarchyChangeAuth_Async(
     /* Check input parameters */
     r = check_session_feasibility(shandle1, shandle2, shandle3, 1);
     return_state_if_error(r, _ESYS_STATE_INIT, "Check session usage");
+
     store_input_parameters(esysContext, authHandle, newAuth);
+    iesys_strip_trailing_zeros(&esysContext->in.HierarchyChangeAuth.newAuth);
 
     /* Retrieve the metadata objects for provided handles */
     r = esys_GetResourceObject(esysContext, authHandle, &authHandleNode);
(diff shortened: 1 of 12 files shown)


williamcroberts:
@JuergenReppSIT we updated the bug tracker #2660 with complete trigger examples, any way you could respin this to get it to fail? I backed your patch out and ran the test and couldn't get a failure. Perhaps I didn't do something properly locally. Don't we also need to handle the auth values when we create objects (Esys_Create, Esys_CreateLoaded, Esys_CreatePrimary, Esys_NV_DefineSpace) since they get remembered?

JuergenReppSIT:
@williamcroberts sorry I forgot, that Esys_Create does not create an esys object. The object will be created by Esys_Load. So the call of Esys_TR_SetAuth is needed for keys. 
I just added the zero stripping to the function which computes the hash for long password. The integration tests work. 
I will remove the draft flag after the following additional changes:
* The integration test is extended with the example which did produce the failure.
* The functions Esys_ChangeAuth, Esys_NV_ChangeAuth , and Esys_ObjectChangeAuth. do not call the function which adapts long passwords.

src/tss2-esys/api/Esys_HierarchyChangeAuth.c
```diff
@@ -175,7 +175,9 @@ Esys_HierarchyChangeAuth_Async(
     /* Check input parameters */
     r = check_session_feasibility(shandle1, shandle2, shandle3, 1);
     return_state_if_error(r, _ESYS_STATE_INIT, "Check session usage");
+
     store_input_parameters(esysContext, authHandle, newAuth);
+    iesys_strip_trailing_zeros(&esysContext->in.HierarchyChangeAuth.newAuth);
```
williamcroberts, line 180: any reason to not do this in store_input_parameters?

src/tss2-esys/api/Esys_ObjectChangeAuth.c
```diff
@@ -163,12 +179,22 @@ Esys_ObjectChangeAuth_Async(
...
 
+    store_input_parameters(esysContext, objectHandle, newAuth);
+
     /* Retrieve the metadata objects for provided handles */
     r = esys_GetResourceObject(esysContext, objectHandle, &objectHandleNode);
     return_state_if_error(r, _ESYS_STATE_INIT, "objectHandle unknown.");
     r = esys_GetResourceObject(esysContext, parentHandle, &parentHandleNode);
     return_state_if_error(r, _ESYS_STATE_INIT, "parentHandle unknown.");
 
+    if (objectHandleNode->rsrc.rsrcType == IESYSC_KEY_RSRC) {
```
williamcroberts, line 190: what about NV objects?

src/tss2-esys/api/Esys_HierarchyChangeAuth.c
JuergenReppSIT: A good hint. The spec states:

This command does not change the authorization of the TPM-resident object on which it operates.
Therefore, the old authValue (of the TPM-resident object) is used when generating the response HMAC
key if required. 

So it makes no sense to store the auth value in the esys object. I will remove this change.

src/tss2-esys/api/Esys_ObjectChangeAuth.c
JuergenReppSIT: The spec states:
NOTE 3 If an NV Index is to have a new authorization, it is done with TPM2_NV_ChangeAuth().

src/tss2-esys/api/Esys_HierarchyChangeAuth.c
cplappert: I think @williamcroberts wants the "iesys_strip_trailing_zeros" command to be implemented in the "store_input_parameters" function. This makes also sense to me. 
@JuergenReppSIT: Could you do this?

AndreasFuchsTPM (approved):

PR #2673, opened 2023-08-04
Do not consider failures to write files in /sys hard errors

DaanDeMeyer:
systemd-tmpfiles can run in containers, chroots, ... where writing to /sys will fail, so let's suffix these lines with "-" to avoid considering these cases hard errors.


AndreasFuchsTPM (approved):

AndreasFuchsTPM (changes requested):
Please add a "Signed-off-by:" tag

DaanDeMeyer:
Added signed off tag

AndreasFuchsTPM (approved):

Issue #2531, opened 2023-01-04
s390x test failures


nekopsykose:
when running check for 4.0.0 on s390x on alpine linux:  

<details>
<summary>test log</summary>

```
============================================================================
Testsuite summary for tpm2-tss 4.0.0
============================================================================
# TOTAL: 55
# PASS:  52
# SKIP:  0
# XFAIL: 0
# FAIL:  3
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log
Please report to https://github.com/tpm2-software/tpm2-tss/issues
============================================================================
make[2]: *** [Makefile:32811: test-suite.log] Error 1
make[2]: Leaving directory '/home/buildozer/aports/community/tpm2-tss/src/tpm2-tss-4.0.0'
make[1]: *** [Makefile:32919: check-TESTS] Error 2
make[1]: Leaving directory '/home/buildozer/aports/community/tpm2-tss/src/tpm2-tss-4.0.0'
make: *** [Makefile:33539: check-am] Error 2
======================================
   tpm2-tss 4.0.0: ./test-suite.log
======================================

# TOTAL: 55
# PASS:  52
# SKIP:  0
# XFAIL: 0
# FAIL:  3
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: test/unit/tcti-spi-helper
===============================

[==========] Running 4 test(s).
[ RUN      ] tcti_spi_no_wait_state_success_test
ERROR:tcti:src/tss2-tcti/tcti-spi-helper.c:732:Tss2_Tcti_Spi_Helper_Init() Failed waiting for TPM to become ready 
[  ERROR   ] --- 0xa0009 != 0
[   LINE   ] --- test/unit/tcti-spi-helper.c:228: error: Failure!
[  FAILED  ] tcti_spi_no_wait_state_success_test
[ RUN      ] tcti_spi_with_wait_state_success_test
ERROR:tcti:src/tss2-tcti/tcti-spi-helper.c:732:Tss2_Tcti_Spi_Helper_Init() Failed waiting for TPM to become ready 
[  ERROR   ] --- 0xa0009 != 0
[   LINE   ] --- test/unit/tcti-spi-helper.c:257: error: Failure!
[  FAILED  ] tcti_spi_with_wait_state_success_test
[ RUN      ] tcti_spi_with_bad_callbacks_test
[...]

PR #2675, opened 2023-08-05
tss2-tcti: tcti-libtpms: fix test failure on big-endian platform

grandpaul:
Due to tcti_libtpms->response_len and tcti_libtpms->response_buffer_len are size_t. We cannot convert the (size_t *) to (uint32_t *) on big-endian platforms. Thus we create temp uint32_t variables. Make the call and then assign it back to size_t variables.

This commit partially fix the test failure on big-endian platforms.


bgermann:
This fixes #2531 in part.

AndreasFuchsTPM (approved):

PR #2686, opened 2023-08-22
tcti-spi-helper: uint32_t is conversion to a void* buffer fails on big endian

slyon:
We need to convert it back to the correct endianess.

Partly fixes: https://github.com/tpm2-software/tpm2-tss/issues/2531

### DIFF
--- a/src/tss2-tcti/tcti-spi-helper.c
+++ b/src/tss2-tcti/tcti-spi-helper.c
@@ -2,6 +2,8 @@
 /*
  * Copyright 2020 Fraunhofer SIT. All rights reserved.
  */
+#define _BSD_SOURCE
+#include <endian.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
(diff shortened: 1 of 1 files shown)


wxleong (changes requested):
- Please avoid using `le32toh()` directly; instead, use `LE_TO_HOST_32()` from https://github.com/tpm2-software/tpm2-tss/blob/master/src/util/tss2_endian.h.
- The `spi_tpm_helper_write_sts_reg()` function also requires endianness correction.
- `did_vid` check requires correction too.

You may refer to:
https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-tcti/tcti-i2c-helper.c#L349
https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-tcti/tcti-i2c-helper.c#L356
https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-tcti/tcti-i2c-helper.c#L785

JuergenReppSIT:
@slyon Thank you for the PR. I have added your change with the requested changes to a commit in #2716, and added you as Co-Author to this commit. So we can close the PR.

PR #2696, opened 2023-10-16
tcti: Add '/dev/tcm0' to default conf

jouyouyun:
A standard similar to TPM has been released in China, called TCM(Trusted Cryptography Module), and its device path is /dev/tcm0. 

The TCM standard is compatible with TPM, and TSS can be used to manage its device path.


AndreasFuchsTPM (changes requested):
There is still something not working with the unit tests.
Please fix that.

Also please add a signed-of-by tag to the commit message.

Otherwise looks good, and could be merged for the next release planned before end of the year.

jouyouyun:
Ok, will update the commit after the unit tests pass.

AndreasFuchsTPM (approved):

PR #2701, opened 2023-11-02
tcti-libtpms: fix big endian response size.

JuergenReppSIT:
size_t was used for the sizes determined by the TPMLIB_Process call. TPMLIB_Process uses uint32_t values. This did work on little endian architectures. On big endian architectures wrong sizes were computed by the tcti receive function. Now uint32_t parameters are used for the TPMLIB_Process call.


JuergenReppSIT:
was already fixed by #2675.

Issue #2709, opened 2023-11-16
Creating a primary key with the FAPI with an empty "unique" field


ahllamas:
I'm trying to create the keys based on the TCG specification: https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf 

When creating a primary signing key under the endorsement hierarchy the unique field gets populated automatically. Is there any way to disable this behavior?

JuergenReppSIT:
Sorry I was to fast and didn't read your question correctly.
The unique field is initialized according to EK credential profile for the EK (`/HE/EK`). If you create a primary signing key in the endorsement hierarchy (e.g. `/HE/mykey`) with the attributes `"user,sign"` this initialization is not done.

ahllamas:
I tried doing this and still the unique field being populated for both "restricted, sign, user" and "user, sign". So, the FAPI assumes that any primary key under HE is an EK?

JuergenReppSIT:
I have created a primary signing key with FAPI:
`tss2_createkey -p /P_RSA2048SHA256/HE/mykey -t "user,sign"`
and with the tpm tools:
`tpm2_createprimary -Grsa2048 -C e -c prim.ctx -a "fixedtpm|fixedparent|sensitivedataorigin|userwithauth|sign`
In both cases the same public key is produced. The unique field is empty when Esys_CreatPrimary is called.
But if the key has to be recreated (e.g. for `tss2_sign` the key is incorrectly under the storage hierarchy. 
I will create a PR to fix this problem.

PR #2711, opened 2023-11-21
FAPI: Fix usage of endorsement handle

JuergenReppSIT:
In several cases the wrong handle TPM2_RH_EK was used instead of TPM2_RH_ENDORSEMENT.
This caused a wrong recreation of keys (except the EK) under the endorsement hierarchy.
Addresses: #2709

### DIFF
--- a/src/tss2-fapi/fapi_util.c
+++ b/src/tss2-fapi/fapi_util.c
@@ -944,9 +944,12 @@ ifapi_load_primary_finish(FAPI_CONTEXT *context, ESYS_TR *handle)
         /* Check whether a persistent key was loaded.
            In this case the handle has already been set. */
         if (pkey_object->public.handle != ESYS_TR_NONE) {
-            if (pkey->creationTicket.hierarchy == TPM2_RH_EK) {
+            if (pkey->creationTicket.hierarchy == TPM2_RH_ENDORSEMENT &&
+                strcmp("/EK",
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/fapi_util.c
```diff
@@ -944,17 +944,20 @@ ifapi_load_primary_finish(FAPI_CONTEXT *context, ESYS_TR *handle)
...
-            if (pkey->creationTicket.hierarchy == TPM2_RH_EK) {
+            if (pkey->creationTicket.hierarchy == TPM2_RH_ENDORSEMENT &&
+                strcmp("/EK",
+                       &pkey_object->rel_path[strlen(pkey_object->rel_path)-3]) == 0) {
                 context->ek_persistent = true;
-            } else {
+            } else if (strcmp("/SRK",
+                              &pkey_object->rel_path[strlen(pkey_object->rel_path)-4]) == 0) {
                 context->srk_persistent = true;
             }
```
AndreasFuchsTPM, line 954: Is the no else possible after these two cases ?

JuergenReppSIT: Yes it's possible. `ek_persistent`and `srk_persistent`are only used in the provisioning cleanup for the EK and SRK and in the cleanup for sessions where the SRK is used as TPM key.

AndreasFuchsTPM (approved):

PR #2715, opened 2023-11-27
TCTI: Add tcti-spidev

AndreasFuchsTPM:
Add a new TCTI module that can talk to TPMs connected via spidev to the host.

### DIFF
--- a/src/tss2-tcti/tcti-spidev.c
+++ b/src/tss2-tcti/tcti-spidev.c
@@ -1,0 +1,214 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright 2020 Peter Huewe
+ */
+#include <errno.h>
+#include <fcntl.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/select.h>
+#include <sys/time.h>
+#include <sys/ioctl.h>
+#include <linux/ioctl.h>
+#include <unistd.h>
+#include <assert.h>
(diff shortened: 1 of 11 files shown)


doc/tcti-spidev.md
```diff
@@ -0,0 +1,48 @@
+# SPI TCTI LTT2GO
```
joholl, line 1: The filename `tcti-spidev.md` is a little misleading since this is about `tcti-spi-ltt2go`.

src/tss2-tcti/tcti-spidev.c
```diff
@@ -0,0 +1,202 @@
...
+        return Tss2_Tcti_Spi_Helper_Init (NULL, size, NULL);
+    }
+
+    /* Create required platform user data */
+    PLATFORM_USERDATA *platform_data = calloc (1, sizeof (PLATFORM_USERDATA));
+    if (platform_data == NULL) {
+        return TSS2_BASE_RC_MEMORY;
+    }
+
+    platform_data->fd = open(config, O_RDWR);
```
joholl, line 169: The TCTI config is documented nowhere. We probably want to add a man page. We might want to add it to `doc/tcti.md`, as well.

doc/tcti-spidev.md
AndreasFuchsTPM: I did not update this. But IMHO there is almost no value to this file anyways.
I did add a manpage now.

src/tss2-tcti/tcti-spidev.c
AndreasFuchsTPM: Added manpage and section to tcti.md.

wxleong (approved):

PR #2716, opened 2023-11-28
FAPI TEST: Rework the execution of FAPI integration tests

JuergenReppSIT:
* The of usage of make check on big endian platforms is now possible.
* The tests executed by fint-log-compiler.sh were integrated into main-fapi.c
* A CA script for initializing the CA before make check was added.
* The EK, the EK certificate and the EK fingerprint are now created in main-fapi.c


src/tss2-tcti/tcti-libtpms.c
```diff
@@ -384,6 +384,8 @@ tcti_libtpms_transmit(
     tpm_header_t header;
     TSS2_RC rc;
     TPM_RESULT ret;
+    uint32_t resp_size;
```
joholl, line 387: This change is already part of master, so ACK.

test/integration/test-common.c
```diff
@@ -582,3 +580,14 @@ test_esys_teardown(TSS2_TEST_ESYS_CONTEXT *test_ctx)
         free(test_ctx);
     }
 }
+
+void
+test_fapi_esys_teardown(TSS2_TEST_ESYS_CONTEXT *test_ctx)
```
joholl, line 585: I'm a little confused by this. We already have  `test_esys_teardown()` and `test_fapi_teardown()`. This new function `test_fapi_esys_teardown()` seems to be never called. Maybe you added this accidentally?

test/integration/main-fapi.c
```diff
@@ -530,20 +1411,67 @@ main(int argc, char *argv[])
...
+    test_esys_teardown(test_esys_ctx);
+
     ret = test_fapi_setup(&test_ctx);
     if (ret != 0) {
         goto error;
     }
 
 #if !defined(FAPI_NONTPM) && !defined(DLOPEN)
-    ret = test_fapi_checks_pre(test_ctx);
+        ret = test_fapi_checks_pre(test_ctx);
```
joholl, line 1468: Wrong indentation

JuergenReppSIT:
@joholl I have fixed most of your remarks. We could also completely switch to libtpms as default. It  also would be possible to  remove the int log compile scripts with the disadvantage that make check could not be started with the tpm servers.
@AndreasFuchsTPM Should we do that?

AndreasFuchsTPM:
> @AndreasFuchsTPM Should we do that?

I would be in favor of removing all scripts surrounding the tests.

Maybe a quick idea: Let's keep the capability for this release and drop it on the next release ?

JuergenReppSIT:
> Maybe a quick idea: Let's keep the capability for this release and drop it on the next release ?

That sounds reasonable.

joholl:
@JuergenReppSIT Cool!

About getting rid of mssim/swtpm: I'd prefer if we still supported testing against mssim/swtpm. This is because it might be more convenient in some contexts and it might have some upsides (testing against remote etc.).

Maybe related: There is a way to reset all three simulators (see below). Just spitballing: Maybe we want to have a future setup where the simulator is started once and then tested against?

@AndreasFuchsTPM Maybe we even want to standardize the TCTI reset functionality in the TCTI spec (would mean a v3 of the context).

https://github.com/tpm2-software/tpm2-tss/blob/182b0273e0089c533eec5e599ad0b9267795150f/test/tpmclient/tpmclient.int.c#L176-L205

AndreasFuchsTPM (approved):
So is this ready to merge ?
@JuergenReppSIT @joholl all agree ?

joholl:
Ready to merge from my perspective :)

AndreasFuchsTPM (approved):

Issue #2730, opened 2023-12-20
libtpms: FreeBSD does not support mremap()


joholl:
FreeBSD does not support `mremap()`, which is called in `tcti_libtpms_map_state_file()`.

Currently, we open the state file, map it into memory, and then close it again. We can then continue working with the memory. If the memory is too small, we can use `mremap()` to enlarge it (without having to open the file again).

Two possible solutions:

### A) We remove the state file feature for FreeBSD

If tcti-libtpms uses memory-only (without state file), then we never call `mremap()`. We could remove the state file feature for FreeBSD via `#ifdef`. For this to work we must be able to run out tests completely stateless :)

### B) Replacing `mremap()` with `munmap()` and `mmap()`.

This is unfortunately a bit unpleasant. As an alternative to `mremap()`, we could use `munmap()` and `tcti_libtpms_map_state_file(min_size)` instead. Unfortunately, this openes, maps, and then closes the file again.

This might get a bit ugly in the unit tests.

### For Reference

https://github.com/tpm2-software/tpm2-tss/blob/166a5538edb4eafaf64c7a45a63ab1a20ac30361/src/tss2-tcti/tcti-libtpms.c#L47-L55

@JuergenReppSIT What do you think?

JuergenReppSIT:
> @JuergenReppSIT What do you think?

I think there are two FAPI integration tests which would cause problems with version A) But they could be skipped  for netBSD ;-)

PR #2734, opened 2023-12-22
tcti-libtpms: remove state file support for FreeBSD

joholl:
On FreeBSD, `mremap()` is a stub returning `-1`/`ENOMEM`. Return `TSS2_TCTI_RC_BAD_VALUE` if tcti-libtpms conf is non-empty.

Fixes: #2730


joholl:
@JuergenReppSIT This is only tested locally (on Linux). Surely, our FreeBSD job is not testing against libtpms (yet). Any idea how I can convince our ci to run test-wise against libtpms for FreeBSD?

AndreasFuchsTPM (approved):

PR #2749, opened 2024-01-14
[RFC] tcti/py: initial commit

williamcroberts:
Implmenet a TCTI module that will allow folks to write TCTIs in Python3. This will allow for more crazy and complex TCTI scenarios without having to re-write everyones stacks to use tpm2-pytss.

TODO:
1. Check error paths
2. Add a test
3. doc cleanups and add examples

Example Commandline:
```bash
PYTHONPATH=$HOME tpm2_getcap --verbose --tcti=$(realpath ./src/tss2-tcti/.libs/libtss2-tcti-py.so.0.0.0):pytcti:$TPM2TOOLS_TCTI properties-fixed
```

Example TCTI:
For the ultimate in mind-bending: C -> Python -> C and back for efficiency :-p.

```python3
from tpm2_pytss import TCTILdr

class MyPyTCTI(object):
    def __init__(self, args: str):
        c = args.split(":", maxsplit=1)
        print(f"PYTHON: Initializing TCTI Ldr with mod: {c[0]} args: {c[1]}")
        self._tcti = TCTILdr(c[0], c[1])

    @property
    def magic(self):
         # Optional Method
         print("PYHTON magic")
         return 42

    def receive(self, timeout: int) -> bytes:
        print("PYTHON receive")
        return self._tcti.receive(timeout=timeout)

    def transmit(self, data:bytes):
        print("PYTHON transmit")
        self._tcti.transmit(data)

def tcti_init(args: str) -> MyPyTCTI:
    print(f"PYTHON tcti_init called with: {args}")
    return MyPyTCTI(args)
```


williamcroberts:
Ahh I'm not supposed to link explicitly to libpython I guess: https://peps.python.org/pep-0513/#libpythonx-y-so-1

joholl:
Woah, exciting! I was working on something similar, a [Rust framework ](https://github.com/joholl/tpm2-tss-rs/blob/main/tss2-tcti-foobar/src/lib.rs) for writing TCTIs:

```rust
pub mod lib {
    use tss2_tcti::define_api_symbols;
    use tss2_tcti::tcti::error::TctiError;
    use tss2_tcti::tcti::tcti::{Api, Info, State, Tcti, TctiLib};
    use tss2_tcti::tctildr::tcti_loader::TctiLoader;
    use tss2_tcti_sys::tpm2_tss;

    #[repr(C)]
    #[derive(Debug)]
    pub struct TctiFoobar {
        api: Api,
        state: State,
        child_tcti: Option<TctiLoader>,
    }

    impl TctiLib for TctiFoobar {
        const INFO: Info<'static> = Info {
            name: b"tpm2_tcti-foobar\0",
            description: b"A demo tcti written in Rust.\0",
            config_help: b"Child TCTI config string as expected by TctiLdr.\0",
        };
        const MAGIC: u64 = 0x44a50b8745675fe5;

        fn new(conf: &str) -> Result<Self, TctiError> {
            let mut tcti = Self {
                api: Self::get_api_static(),
                state: State::NotInitialized,
                child_tcti: None,
            };

            tcti.init(conf)?;

            Ok(tcti)
        }

        fn init_inner(&mut self, conf: &str) -> Result<(), TctiError> {
            self.api = TctiFoobar::get_api_static();
            self.child_tcti = Some(<TctiLoader as Tcti>::new(conf)?);
            self.state = State::Transmit;
            Ok(())
        }

        fn transmit_inner(&mut self, command: &[u8]) -> Result<(), TctiError> {
            println!("TX: {:?}", command);

            if let Some(child_tcti) = self.child_tcti.as_mut() {
                return child_tcti.transmit(command);
            }

            todo!()
        }

        fn receive_inner(&mut self) -> Result<Vec<u8>, TctiError> {
            if let Some(child_tcti) = self.child_tcti.as_mut() {
                return child_tcti.receive();
            }

            todo!()
        }

        fn get_state(&self) -> Option<State> {
            Some(self.state)
        }
        fn set_state(&mut self, state: State) {
            self.state = state;
        }
    }

    define_api_symbols!(TctiFoobar);
}
```

src/tss2-tcti/tcti-py.c
```diff
@@ -0,0 +1,526 @@
...
+
+    /* Increment the reference count on PYTHON init */
+    pthread_mutex_lock(&_global.lock);
+    size_t new_cnt = 0;
+    bool overflow = __builtin_add_overflow(_global.refcnt, 1, &new_cnt);
+    if (!overflow) {
+        _global.refcnt = new_cnt;
+        if (new_cnt == 1) {
+            /* See: https://stackoverflow.com/questions/60719987/embedding-python-which-uses-numpy-in-c-doesnt-work-in-library-dynamically-loa */
+            _global.dlhandle = dlopen ("libpython3.so", RTLD_LAZY|RTLD_GLOBAL);
```
joholl, line 440: Why do we not link `libpython3.so` as a shared library? That would lead to compile-time errors if the libpython dependency is not installed (instead of runtime errors). Also, it would simplify the code.

billatarm: We actually do link it, but if other bindings are in the link chain, like CFFI, it will resolve the symbols for that and leave some of the symbols unresolved, so we add GLOBAL and LAZY to force the resolution of undefined symbols at run time.

That comment above is what led me down this path, it seems to work as expected. I didn't find any other way to work around it.

williamcroberts: We actually do link it, but if other bindings are in the link chain, like CFFI, it will resolve the symbols for that and leave some of the symbols unresolved, so we add GLOBAL and LAZY to force the resolution of undefined symbols at run time.

That comment above is what led me down this path, it seems to work as expected. I didn't find any other way to work around it.

whooo (changes requested):

src/tss2-tcti/tcti-py.c
joholl: Hm, I'm not sure I have fully understood the issue. You mean because the python module uses tctildr?
Another thought: we need to get the pkg-config library name (for me "libpython3.10.so") via define, I think. For me, I had to change the hardcoded shared library name.

joholl:
Works for me!

```bash
LD_LIBRARY_PATH=:../tpm2-tss/src/tss2-tcti/.libs PYTHONPATH=/path/to/python/script/ TSS2_LOG=tcti+trace tpm2 startup -c --tcti="py:foobar:libtpms"
```

When my python script errors during init (e.g. because call it with args=None which your example script cannot handle), I get a seg fault, probably because we call `tcti_py_finalize()` on the not-yet-initialized context.

joholl:
Also, I made [tpmstream](https://github.com/joholl/tpmstream) (master branch, not released yet) a tcti like in your example. So you can do the following now:

```bash
LD_LIBRARY_PATH=:../tpm2-tss/src/tss2-tcti/.libs  tpm2 startup -c --tcti="py:tpmstream:libtpms"
```

src/tss2-tcti/tctildr-dl.c
```diff
@@ -112,7 +112,7 @@ handle_from_name(const char *file,
         if (*handle != NULL) {
             return TSS2_RC_SUCCESS;
         } else {
-            LOG_DEBUG("Could not load TCTI file \"%s\": %s", file, dlerror());
```
joholl, line 115: We don't want to change this to an error. Firstly, it is not fatal. Secondly, it is expected that different tcti name variants are not found while tctildr tries different variants (`<child_name>`, `libtss2-tcti-<child_name>.so.0`, ...)

williamcroberts: whoops, that was unattended, I thought I cleaned up all my cruft

src/tss2-tcti/tcti-py.c
williamcroberts: > Hm, I'm not sure I have fully understood the issue. You mean because the python module uses tctildr? Another 

No it has to do with symbol resolution and the dynamic linker loader, not anything tss2 specific.

williamcroberts: > we need to get the pkg-config library name (for me "libpython3.10.so") via define

Did that in the newest spin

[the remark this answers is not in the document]
williamcroberts: Should work now, I parse out the library name, the proper package to have is python3-embed, figuring that out for pkg-config helped a bit to simplify things.

src/tss2-tcti/tcti-py.c
```diff
@@ -0,0 +1,526 @@
...
+    }
+
+    tcti_common->state = TCTI_STATE_RECEIVE;
+    rc = TSS2_RC_SUCCESS;
+out:
+    return rc;
+
+py_error:
+    PyErr_Print();
+    SAFE_DECREF(py_buf);
```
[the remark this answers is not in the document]
williamcroberts: It would take a lot of work to plumb everything I think, far more work than it's worth IMO. Unless I am missing something.

src/tss2-tcti/tcti-py.h
```diff
@@ -0,0 +1,36 @@
...
+struct TSS2_TCTI_PY_CONTEXT {
+    TSS2_TCTI_COMMON_CONTEXT common;
+    PyObject *py_tcti;
+    PyObject *cur_response;
+    struct {
+        /* no finalize as __del__ will be called for the TCTI instance */
+        PyObject *transmit;
+        PyObject *receive;
+        PyObject *make_sticky;
+    } methods;
```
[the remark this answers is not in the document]
williamcroberts: Well I dropped these do to a weird bug, but I spent all day figuring it out, It was all from not clearing the error handler when get_attr() in the C code failed. Later I would get errors when calling the next Py_<function> but it was related to something N calls ago :-p. It was wierd that when you called transmit, it would say, no attribute named 'make_sticky'.

williamcroberts:
> Works for me!
> 
> ```shell
> LD_LIBRARY_PATH=:../tpm2-tss/src/tss2-tcti/.libs PYTHONPATH=/path/to/python/script/ TSS2_LOG=tcti+trace tpm2 startup -c --tcti="py:foobar:libtpms"
> ```
> 
> When my python script errors during init (e.g. because I call it with args=None which the example script cannot handle), I get a seg fault, probably because we call `tcti_py_finalize()` on the not-yet-initialized context.

Fixed that, it was from calling Py_Finalize() followed by a Py_XDECREF()

williamcroberts:
Building Fedora-34 with python3-devel, so once that is published I can add a test to the ci:
  - https://github.com/tpm2-software/tpm2-software-container/pull/82

PR #2765, opened 2024-02-03
unit tests: fix pointer casts on 32 bit systems.

JuergenReppSIT:
Conversions of max uint64 numbers to pointers did produce compile errors on 32bit systems. A cast to size_t is now added before the casting to a pointer.

### DIFF
--- a/test/unit/esys-vendor.c
+++ b/test/unit/esys-vendor.c
@@ -128,7 +128,8 @@ static TSS2_RC tcti_fake_recv(
     UNUSED(tctiContext);
     UNUSED(timeout);
 
-    const char *id = (const char*)mock();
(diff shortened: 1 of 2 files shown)


AndreasFuchsTPM:
What exactly was the error message ?
Because this seems kind of ugly...

JuergenReppSIT:
> What exactly was the error message ? Because this seems kind of ugly...
the error message was:
```
test/unit/tcti-spidev.c:111:23: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
  111 |     uint8_t *tx_buf = (uint8_t *) tr->tx_buf;
  
```
`tx_buf` had the type `__u64`. Yes it's ugly. Do you have a better suggestion?

AndreasFuchsTPM (approved):

PR #2789, opened 2024-03-18
FAPI: Fix compiler warnings

jasonbking:
On some platforms, compiler warnings are generated due to not include <strings.h> (note plural) when using index(3) as well as not utilizing the return value to BIO_flush().

### DIFF
--- a/src/tss2-fapi/fapi_crypto.c
+++ b/src/tss2-fapi/fapi_crypto.c
@@ -2090,7 +2090,7 @@ ifapi_base64encode(uint8_t *buffer, size_t buffer_size, char** b64_data) {
                    cleanup);
     }
 
-    BIO_flush(bio);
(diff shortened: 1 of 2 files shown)


AndreasFuchsTPM (changes requested):
Now that we have the CI up and running again...
@jasonbking Would you mind rebasing to latest master and force-pushing ?

AndreasFuchsTPM (changes requested):
This somehow seems to mess up the BSD build.
We see all failures now for FAPI.
But I'm puzzled, as to why this is...

JuergenReppSIT:
@AndreasFuchsTPM after starting the rerun everything worked. Also in my local BSD vm I could run the test several times without problems.

AndreasFuchsTPM (approved):

Issue #2799, opened 2024-03-26
Use of TPM2_HR_PERSISTENT triggers undefined behaviour


polarina:
Take for example this program:
```c
#include <tss2/tss2_tpm2_types.h>

int main() {
	if (TPM2_HR_PERSISTENT) {
		return 0;
	}

	return 1;
}
```

Compile:
```sh
gcc -fsanitize=undefined ubsan.c
```

Execute:
```
$ ./a.out 
ubsan.c:4:6: runtime error: left shift of 129 by 24 places cannot be represented in type 'int'
```

AndreasFuchsTPM:
I guess we are casting wrongly.

Currently we have
```
#define TPM2_HR_PERSISTENT        ((TPM2_HC) (TPM2_HT_PERSISTENT << TPM2_HR_SHIFT))
```

I guess this should be
```
#define TPM2_HR_PERSISTENT        (((TPM2_HC) TPM2_HT_PERSISTENT) << TPM2_HR_SHIFT)
```

Would you agree ?

polarina:
> Would you agree ?

Yes, that would fix the problem. Both arguments to `<<` would then be of type `uint32_t` which vibes with the C integer promotion rules.

PR #2801, opened 2024-04-09
All: Fix left-shifts of signed vs unsigned constants

AndreasFuchsTPM:
In order to silence -fsanitize=undefined all left shifts of constants are now cast before shifting, in order to avoid undefined behavior, if the target variable is unsigned.

Fixes: #2799


src/tss2-fapi/ifapi_helpers.c
```diff
@@ -2393,10 +2393,10 @@ ifapi_filter_pcr_selection_by_index(
     UINT8 selection[] = { 0, 0, 0, 0 };
 
     for (i = 0; i < pcr_count; i++) {
-        selection[0] |= (1 << pcr_index[i]) % 256;
-        selection[1] |= (1 << (pcr_index[i] - 8)) % 256;
-        selection[2] |= (1 << (pcr_index[i] - 16)) % 256;
-        selection[3] |= (1 << (pcr_index[i] - 24)) % 256;
+        selection[0] |= (((UINT32)1) << pcr_index[i]) % 256;
```
JuergenReppSIT, line 2396: Should not be used here UINT8?

JuergenReppSIT (changes requested):
see comment.

src/tss2-fapi/ifapi_helpers.c
AndreasFuchsTPM: No, because we want to shift for up to 24 bits to the left. So we need 32bit.
Then we cast it down to BYTE using modulo.

JuergenReppSIT: @AndreasFuchsTPM yes that's true. There are still some cases  in the test files and one in mpsse.c. 
`grep -R  "1 << [^0-9]" .`
Should they also be changed?

wxleong: The mpsse.c file should be ok for the time being. In ReadBits() and WriteBits(), the code restricts the max left shift to 8 bits. Regarding PinState(), the max left shift is up to 12+4 bits (currently only 12 GPIOs are defined by libmpsse: https://github.com/devttys0/libmpsse/blob/master/docs/README.GPIO).

wxleong:
Reference: https://github.com/AGHABEY/Books/blob/master/c-in-a-nutshell-o-reilly-peter-prinz-tony-crawford.pdf

Considering the integer promotion rules, operations like `((uint8_t)1 << x)` are ineffective/not necessary. This is due to the integer promotion: 
- The operands of the shift operators must be integers. Before the actual bit-shift, the integer promotions are performed on both operands.
- Any operand whose type ranks lower than int is automatically converted to the type int. 

Type ranks:
- The standard integer types are ranked in the order:
  _Bool < char < short < int < long < long long
- Each signed integer type has the same rank as the corresponding unsigned

If `x` is controlled and restricted to be less than `(sizeof(int) * 8 - 1)`, then no fix is required. Otherwise, the correct implementation should be: `(__output_datatype__)((unsigned int)1 << x)`. Using `unsigned int` explicitly prevents the integer promotion operation from occurring, as `unsigned int` is of the same type rank as `int`.

You can experiment with:
```
#include <stdint.h>
int main() {
    uint8_t var1 = (uint8_t)(((uint8_t)1) << 31); // Error
    int8_t var2 = ((int8_t)1) << 7;
    int var3 = 1 << 31; // Error
    int var4 = ((unsigned int)1) << 31;
    return 0;
}
```
```
$ gcc -fsanitize=undefined test.c`
$ ./a.out
test.c:3:43: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
test.c:5:33: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
```

src/tss2-fapi/ifapi_helpers.c
JuergenReppSIT: I checked the remaining shifts in the test files (displayed by grep). They also should be ok.

wxleong (approved):

JuergenReppSIT (approved):

PR #2824, opened 2024-05-03
FAPI: Add check whether auth values exist for hierarchies.

JuergenReppSIT:
Currently FAPI provisioning tries to create the EK and SRK with the NULL auth value for the hierarchies.
Now first the corresponding flag in  TPM2_CAP_TPM_PROPERTIES with the property TPM2_PT_PERMANENT is checked. If an auth value is used for the hierarchy the auth value callback will be called. The "retry" code in the BAD_AUTH case is removed.


AndreasFuchsTPM:
Question: When you set an authValue and then you set the authValue back to NULL, does the authSet capability turn back to "false" ?

JuergenReppSIT:
> Question: When you set an authValue and then you set the authValue back to NULL, does the authSet capability turn back to "false" ?
yes. "with_auth" in the keystore is set to "NO" and the authSet capability is set to false.

AndreasFuchsTPM (approved):

PR #2825, opened 2024-05-03
int-log-compiler-common.sh: Remove dependency on which

salahcoronya:
See https://lwn.net/Articles/874049/

### DIFF
--- a/script/int-log-compiler-common.sh
+++ b/script/int-log-compiler-common.sh
@@ -55,12 +55,12 @@ sanity_test ()
         exit 1
     fi
 
-    if [ -z "$(which ${sock_tool})" ]; then
+    if ! command -v ${sock_tool} > /dev/null ; then
         echo "ss not on PATH; exiting"
         exit 1
     fi
 
-    if [[ ! -z ${simulator_bin} && -z "$(which ${simulator_bin})" ]]; then
+    if [[ ! -z ${simulator_bin} && -z "$(command -v ${simulator_bin})" ]]; then
         echo "${simulator_bin} not on PATH; exiting"
         exit 1
     fi


AndreasFuchsTPM:
Why use `type -P` ?
The lwn article mentions `command -v`.

thesamesam:
If the file relies on bash already, `type -P` is mildly superior as it will ignore functions defined in the user's environment. I haven't checked if that's the case here.

AndreasFuchsTPM:
Given the comment in https://github.com/tpm2-software/tpm2-tss/pull/2830 I'd prefer `command -v` to have as few bash-specifics as possible in favor of posix.
Could you update the PR for me to merge ?

eli-schwartz:
> Given the comment in #2830 I'd prefer `command -v` to have as few bash-specifics as possible in favor of posix. Could you update the PR for me to merge ?

I'm not attached to the use of type -P or bash, but just as a devil's advocate I would like to point out that you can have a policy that your tests use bash, and that's a project choice.

configure.ac *has* to be written in POSIX sh because the autotools don't let you choose the script interpreter shebang for your ./configure script -- it is always /bin/sh, never bash.

script/int-log-compiler-common.sh
```diff
@@ -55,12 +55,12 @@ sanity_test ()
         exit 1
     fi
 
-    if [ -z "$(which ${sock_tool})" ]; then
+    if [ -z "$(command -v ${sock_tool})" ]; then
```
eli-schwartz, line 58: You can just check `if ! command -v XXX >/dev/null; then`

AndreasFuchsTPM (approved):

PR #2830, opened 2024-05-07
Remove bash syntax

SoapGentoo:
* `==` is bash and not POSIX Bug: https://bugs.gentoo.org/931239


AndreasFuchsTPM (approved):
Do you need a 4.1.2 for this ?

SoapGentoo:
4.1.2, but only if we get the `command -v`/`which` fix in too

SoapGentoo:
sure, you can ping me during the rc phase and I will test it to check that the bashism/`which` things are fixed. In general, you can ping me for build system related issues, seeing how Gentoo is more likely to run into them than other distros.

AndreasFuchsTPM:
@SoapGentoo 4.1.2-rc0 is tagged.
Let me know if there are issues remaining...

SoapGentoo:
@AndreasFuchsTPM thanks, just tested it with `dash` and without `which` present and all tests have passed!

PR #2836, opened 2024-05-10
ESYS: StartAuthSession bind auth trailing zeroes

throwException:
When StartAuthSession is called with a bind entity with a auth value containing trailing zeroes, the HMAC or policy session computation of ESYS does not match the computation on the TPM2.

The fix is to remove trailing zeroes from the auth value according to the specification (TPM2 Architecture, 19.6.5, Note 2) before computation of the session key.

The fixed bug is especially tricky as a randomly generated auth value of the bind object can cause HMAC or policy session to fail occassionally.

### DIFF
--- a/src/tss2-esys/api/Esys_StartAuthSession.c
+++ b/src/tss2-esys/api/Esys_StartAuthSession.c
@@ -458,8 +458,23 @@ Esys_StartAuthSession_Finish(
             size_t secret_size = 0;
             if (tpmKey != ESYS_TR_NONE)
                 secret_size += keyHash_size;
-            if (bind != ESYS_TR_NONE && bindNode != NULL)
-                secret_size += bindNode->auth.size;
+            TPM2B_DIGEST bindNodeAuth = {
+                .buffer = { },
+                .size = 0,
+            };
+            if (bind != ESYS_TR_NONE && bindNode != NULL) {
+                /*
+                 * TPM2.0 Architecture 19.6.5 Note 2
+                 *
+                 * Remove tailing zeroes from the auth value
+                 */
+                bindNodeAuth.size = bindNode->auth.size;
+                while ((bindNodeAuth.size > 0) &&
+                       (bindNode->auth.buffer[bindNodeAuth.size - 1] == 0x00))
+                    bindNodeAuth.size--;
+                memcpy(&bindNodeAuth.buffer[0], &bindNode->auth.buffer[0], bindNodeAuth.size);
+                secret_size += bindNodeAuth.size;
+            }
             /*
              * A non null pointer for secret is required by the subsequent functions,
              * hence a malloc is called with size 1 if secret_size is zero.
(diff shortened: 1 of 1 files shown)


tomoveu:
good find @throwException

src/tss2-esys/api/Esys_StartAuthSession.c
```diff
@@ -469,9 +469,20 @@ Esys_StartAuthSession_Finish(
...
-                 && bindNode->auth.size > 0)
-                memcpy(&secret[0], &bindNode->auth.buffer[0], bindNode->auth.size);
+
+            if  (bind != ESYS_TR_NONE && bindNode != NULL) {
+                /*
+                 * TPM2.0 Architecture 19.6.5 Note 2
+                 *
+                 * Remove tailing zeroes from the auth value
+                 */
+                while ((bindNode->auth.size > 0) &&
```
tomoveu, line 479: I think the auth.size > 0 check needs to be part of the original if statement and then use a working variable for the removal of trailing zeroes. It is more steps, but safer in my opinion.

AndreasFuchsTPM:
I am wondering, why
https://github.com/tpm2-software/tpm2-tss/blob/80f87339fb73fd50b581b5927711e45e64279348/src/tss2-esys/esys_tr.c#L510
and
https://github.com/tpm2-software/tpm2-tss/blob/a19ac4c61c3f3bbdfc61cf01040a30729d6ba441/src/tss2-esys/esys_iutil.c#L1734
are not effective in this case of bind ?

The idea was and is that the auth values carried in the metadata are always free of trailing zeros to begin with.
Thus I don't like the fix at this place but rather want to find out the call path that circumvented the previous cases.

src/tss2-esys/api/Esys_StartAuthSession.c
throwException: @tomoveu You are right, the `bindNode->auth` shouldn't be changed. But since this is passed to `iesys_compute_bound_entity` later, the `TPM2B_DIGEST` structure needs to be copied into a working variable I name `bindNodeAuth`.

The computation of `secret_size` was off in the original merge request, which I corrected by moving the removal of trailing zeros.

Finally, some lines below can be simplied since `bindNodeAuth.size` is always valid and 0 if `bindNode == NULL`.

throwException:
> The idea was and is that the auth values carried in the metadata are always free of trailing zeros to begin with.
> Thus I don't like the fix at this place but rather want to find out the call path that circumvented the previous cases.

I think it's because the auth isn't set for authentication, but taken from an entity obtained by `esys_GetResourceObject` in `Esys_StartAuthSession` to compute the session secret.

throwException:
@AndreasFuchsTPM I'm sorry, this bug was already fixed in 4.0.2 by commit d3bcce8c7ed13d7eef6a0fbcb536fcf217804289.

AndreasFuchsTPM:
No worries, PRs are always welcome !

PR #2837, opened 2024-05-13
iwyu: add map file to correctly and deterministicly lint includes

joholl:
Firstly, add json-c include flags to some tests to get rid of the leading `json-c/` when including.

Secondly, with map files, we can control how iwyu deals with facade headers (like json-c) and where to get certain symbols from. Add a map file and have iwyu be more deterministic about its changes.

I also changed the documentation a bit, including fixing a typo (`-Xiwyu`), adding `-j`, piping directly (`fix_includes` would sometimes not apply all changes) etc.

Note that depending on the system, some files might be compiled or not compiled (or compiled differently). E.g. on my system I had libtpms installed, so tcti-libtpms was linted additionally.


joholl:
This is ready for review now.

In the future, we probably want to add a check to our ci.

PR #2846, opened 2024-05-18
TCTI: Fix leak produced in Tss2_TctiLdr_Initialize_Ex

### LINKED ISSUE #2842: Memory leak in error path of Esys_Initialize
`Esys_Initialize` initializes the TCTI:

https://github.com/tpm2-software/tpm2-tss/blob/9f336a5064a0295608ca0e7a4d70c4db18664a37/src/tss2-esys/esys_context.c#L71-L75

The function `Tss2_TctiLdr_Initialize` (via `Tss2_TctiLdr_Initialize_Ex`) allocates memory that is not implicitly freed on failure:

https://github.com/tpm2-software/tpm2-tss/blob/9f336a5064a0295608ca0e7a4d70c4db18664a37/src/tss2-tcti/tctildr.c#L521-L528

When `Tss2_TctiLdr_Initialize` returns after a failure, the `cleanup_return` in `Esys_Initialize` is then executed:

https://github.com/tpm2-software/tpm2-tss/blob/9f336a5064a0295608ca0e7a4d70c4db18664a37/src/tss2-esys/esys_context.c#L95-L98

However, after the execution of `Tss2_Tcti_Finalize` there is a missing call to `free` unlike what is done in the error path of `Tss2_TctiLdr_Initialize_Ex`:

https://github.com/tpm2-software/tpm2-tss/blob/9f336a5064a0295608ca0e7a4d70c4db18664a37/src/tss2-tcti/tctildr.c#L530-L535

JuergenReppSIT:
The return code of tctildr_init_context_data was not checked in Tss2_TctiLdr_Initialize_Ex. The cleanup part of this function was not executed and so a leak was produced.
Fixes #2842

### DIFF
--- a/src/tss2-tcti/tctildr.c
+++ b/src/tss2-tcti/tctildr.c
@@ -525,7 +525,10 @@ Tss2_TctiLdr_Initialize_Ex (const char *name,
     }
 
     *tctiContext = (TSS2_TCTI_CONTEXT *) ldr_ctx;
-    return tctildr_init_context_data(*tctiContext, local_name, local_conf);
+    rc = tctildr_init_context_data(*tctiContext, local_name, local_conf);
+    if (rc == TSS2_RC_SUCCESS) {
+        return rc;
+    }
 
 err:
     if (*tctiContext != NULL) {


joholl (approved):
Looks good!

PR #2847, opened 2024-05-25
Add first .clang-tidy file

joholl:
Add a `.clang-tidy` file with some of the less opinionated check. Fix the linting issues (one commit per check).

For linting, I don't call clang-tidy directly, but the official python script for multi-core support. You can add `-fix` for automatic fixing.

 ```sh
make -j compile_commands.json
 run-clang-tidy-15 -header-filter $PWD $(find src include -name '*.c' -o -name '*.h') | ansi2txt | tee clang-tidy.log
```

For now, I focused on `src/`, but not `test/`

Fixes: #2826

### CODE REVIEW

**Comment by joholl** [comment]
@JuergenReppSIT @AndreasFuchsTPM cirrus keeps failing on me, but otherwise the CI is happy. What do you think of this?

File: include/tss2/tss2_esys.h [hunk 1]
```diff
@@ -3796,7 +3796,7 @@ Esys_Vendor_TCG_Test_Finish(
  */
 void
 Esys_Free(
-    void *__ptr);
+    void *_ptr);
```
AndreasFuchsTPM, line 3799: Why is this ?

include/tss2/tss2_sys.h
```diff
@@ -23,7 +23,7 @@ extern "C" {
 #endif
 
 /* SAPI context blob */
-typedef struct _TSS2_SYS_OPAQUE_CONTEXT_BLOB TSS2_SYS_CONTEXT;
+typedef struct TSS2_SYS_OPAQUE_CONTEXT_BLOB TSS2_SYS_CONTEXT;
```
AndreasFuchsTPM, line 26: This one is definitively weird...

include/tss2/tss2_tpm2_types.h
```diff
@@ -1877,9 +1877,9 @@ struct TPM2B_SENSITIVE {
     TPMT_SENSITIVE sensitiveArea;
 };
 
-/* Definition of _PRIVATE Structure <> */
-typedef struct _PRIVATE _PRIVATE;
-struct _PRIVATE {
+/* Definition of PRIVATE Structure <> */
+typedef struct PRIVATE PRIVATE;
+struct PRIVATE {
```
AndreasFuchsTPM, line 1882: Also weird.

AndreasFuchsTPM (changes requested):
Seams to dislike leading `_`
This is especially bad in the cases of API incompatibilities.

joholl:
@AndreasFuchsTPM TLDR; it is undefined behavior since those identifiers are reserved (for stdlib and compiler i guess).
 
Here's what [the C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf#page=200) says (section 7.1.3):

>  * All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.   
>  * All identifiers that begin with an underscore are always reserved for use as identifiers with file scope in both the ordinary and tag name spaces.

And further:

> If the program declares or defines an identifier in a context in which it is reserved [...] the **behavior is undefined**.

Clang-tidy checks that via rule [bugprone-reserved-identifier](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html).

## `_PRIVATE`

`_PRIVATE` might not really be a "private" type, but just a utility type for defining `TPM2B_PRIVATE`. While not useful to the user either, we could make it public (but we should mangle the name somehow).

```c
struct TPM2B_PRIVATE {
    UINT16 size;
    BYTE buffer[sizeof(PRIVATE)];
};
```

If we really wanted to make it private, we would need to jump through hoops because we have a `sizeof(_PRIVATE)` in our public header (e.g. create a configure-time macro which provides the size of PRIVATE and use that instead).

Suggestion: rename it `TPM_OPAQUE_PRIVATE` (and adding docs telling users it is private). Technically, this is a breaking API change. I would argue, using this type as a user is fishy in any case. In my understanding this is supposed to be an opaque blob. We also don't use it in the TSS (except as an argument to `sizeof()` in the json (de)serialization code.

An alternative is leaving it as-is and technically introducing undefined behavior in every user's namespace.

## `_TSS2_SYS_OPAQUE_CONTEXT_BLOB`

So this is really a private type. I'd argue the `OPAQUE`  portion of the name already makes it clear what this is.

Suggestion: Rename to `TSS2_SYS_OPAQUE_CONTEXT_BLOB` and writing a docstring telling users to not use it.

## `__ptr`

It is technically legal to have a leading underscore (as `_foo`) in a limited scope, so e.g. as a parameter (but not on file scope). We probably just want to rename that to `ptr`. This has no API implications anyway.

Suggestion: rename to `ptr`

include/tss2/tss2_tpm2_types.h
joholl: See my comment.

include/tss2/tss2_sys.h
joholl: See my comment.

include/tss2/tss2_esys.h
joholl: See my comment.

AndreasFuchsTPM:
Understood.
`__ptr -> ptr` no problem. Agreed.

For the rest: I guess they make sense, BUT we'd need to roll 3.0 for this.
Thankfully, it does not change ABI but only API, so we don't have to bump the .so-Version.

The big question is: Is it worth to go to 3.0 over this ?

JuergenReppSIT:
> The big question is: Is it worth to go to 3.0 over this ?

Maybe we could  first split the PR and extract the "real" bugfxes?

joholl:
@JuergenReppSIT Alright, I split the changes. The API braking changes are currently on [joholl/tpm2-tss@clang_tidy_api_changes](https://github.com/joholl/tpm2-tss/tree/clang_tidy_api_changes). I'm going to submit a PR once this is merged. We could roll that out when we roll a new major version anyway due to another API change (@AndreasFuchsTPM I assume you meant 5.0).

Any more feedback?

AndreasFuchsTPM:
@joholl Scan-Build complains about
```
../src/tss2-fapi/api/Fapi_GetTpmBlobs.c:221:32: warning: Result of 'malloc' is converted to a pointer of type 'uint8_t', which is incompatible with sizeof operand type 'TPM2B_PUBLIC'
                *tpm2bPublic = malloc(sizeof(TPM2B_PUBLIC));
                               ^~~~~~ ~~~~~~~~~~~~~~~~~~~~
```

AndreasFuchsTPM (approved):

PR #2857, opened 2024-06-26
FAPI: Fix usage of external PEM keys for PolicyAuthorize. 4.1.x

JuergenReppSIT:
* PolicyAuthorize with an external imported PEM key did not work if the default nameAlg for the imported key (sha1) was not equal to the default nameAlg in the current fapi profile. The nameAlg from the profile is now used for the imported key.
* To prevent a possible double free after cleanup of a policy the NULL pointer is set for the reference to this policy.


AndreasFuchsTPM:
@JuergenReppSIT Do you remember, why we did not merge this ?
It was originally assigned to milestone 4.1.4 but we totally did not release it...

JuergenReppSIT:
> @JuergenReppSIT Do you remember, why we did not merge this ? It was originally assigned to milestone 4.1.4 but we totally did not release it...

No, I can't remember. It actually should have been merged. The fix is ​​included in 4.2.

PR #2871, opened 2024-07-21
Fix tests with a physical tpm

JuergenReppSIT:
* policy: Add check for sha384 in test policy execute.
* FAPI: Fix tests with physical TPM using SHA384.
* tcti mssim: Fix wrong cast on 32 bit systems.
    
Fixed: #2873


joholl:
* policy/FAPI: LGTM
* mssim: What about the following two calls? They are even in the same function.

https://github.com/tpm2-software/tpm2-tss/blob/3f5a4568fda6a7ebc412350918b1b325c8678bba/src/tss2-tcti/tcti-mssim.c#L366

https://github.com/tpm2-software/tpm2-tss/blob/3f5a4568fda6a7ebc412350918b1b325c8678bba/src/tss2-tcti/tcti-mssim.c#L424

joholl (changes requested):
Please fix also the other two calls of `socket_recv_buf()` in tcti-mssim.

tcti-swtpm is fine.

joholl (approved):

joholl:
@JuergenReppSIT Thanks for fixing this (including the things I broke)! Looks good.

PR #2875, opened 2024-07-28
Remote trailing slash from bootstrap

Firstyear:
With autoconf 2.72, the trailing / in bootstrap causes cryptic build failures to occur on MacOS. Removing these / allows autoconf to proceed.

And example of the error is:

```
# ./bootstrap
aminclude_static.am:63: warning: GITIGNOREFILES was already defined in condition TRUE, which includes condition AUTOCONF_CODE_COVERAGE_2019_01_06 and CODE_COVERAGE_ENABLED ...
Makefile.am:56:   'aminclude_static.am' included from here
Makefile.am:52: ... 'GITIGNOREFILES' previously defined here
Makefile.am: error: object 'src/tss2-esys/libtss2_esys_la-esys_crypto_mbed.lo' created by 'src/tss2-esys/esys_crypto_mbed.c' and 'src/tss2-esys//esys_crypto_mbed.c'
Makefile.am: error: object 'src/tss2-esys/libtss2_esys_la-esys_crypto_ossl.lo' created by 'src/tss2-esys/esys_crypto_ossl.c' and 'src/tss2-esys//esys_crypto_ossl.c'
autoreconf: error: automake failed with exit status: 1
```

Note the double `//` in `src/tss2-esys//esys_crypto_mbed.c`

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -56,16 +56,16 @@ echo "Generating file lists: ${VARS_FILE}"
   src_listvar "src/util-io" "*.h" "UTIL_IO_H"
   printf "UTIL_IO_SRC = \$(UTIL_IO_C) \$(UTIL_IO_H)\n"
 
-  src_listvar "src/tss2-sys/" "*.c" "TSS2_SYS_C"
-  src_listvar "src/tss2-sys/" "*.h" "TSS2_SYS_H"
+  src_listvar "src/tss2-sys" "*.c" "TSS2_SYS_C"
+  src_listvar "src/tss2-sys" "*.h" "TSS2_SYS_H"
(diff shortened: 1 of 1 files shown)


Firstyear:
Relates https://github.com/parallaxsecond/rust-tss-esapi/pull/523

Firstyear:
> If that's the output, then it sounds like `find` does not work correctly on MacOS.

Could be a BSD find thing? 

> Is the output really different for `find ./` vs `find .` ?

```
[william@maxixe 11:55] /tmp/testfoo # find .
.
[william@maxixe 11:55] /tmp/testfoo # find ./
./
```

Seems so.

AndreasFuchsTPM (approved):

Issue #2878, opened 2024-07-31
Fapi_Import crashes with SIGSEGV when importing a policy that contains "null" element


dEajL3kA:
For example, importing this policy with `Fapi_Import()` triggers a SIGSEGV for me:
```
{
  "description": "Description pol_duplicate",
  "policy": [
    {
      "type": "POLICYDUPLICATIONSELECT",
      "newParentPath": null
    }
  ]
}
```

Stack trace:
```
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:76
#1  0x00007f4d5ca888de in __GI___strdup (s=0x0) at ./string/strdup.c:41
#2  0x00007f4d5cd22a11 in ?? () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
#3  0x00007f4d5cd3e765 in ?? () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
#4  0x00007f4d5cd419fc in ?? () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
#5  0x00007f4d5cd40cc4 in ?? () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
#6  0x00007f4d5cd41d25 in ?? () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
#7  0x00007f4d5cced0b9 in Fapi_Import_Async () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
#8  0x00007f4d5ccef114 in Fapi_Import () from /lib/x86_64-linux-gnu/libtss2-fapi.so.1
[...]
#19 0x00007f4d5ca73134 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
```

Tested on Debian Bookworm as well as on Sid/Trixie, both with the "built-in" version of `libtss2-fapi1`.

I know that this is an *invalid* input, but it shouldn't cause the FAPI to crash, I suppose 😏

dEajL3kA:
Closing, as fix seems to be in the way. Thanks.

PR #2880, opened 2024-07-31
FAPI: Fix segfault if json field is null.

JuergenReppSIT:
The function json_object_object_get_ex does not create a json object for the parameter value in the case "key":null. This caused a segfault in json deserialization.
Fixes: #2878

### DIFF
--- a/src/tss2-fapi/tpm_json_deserialize.c
+++ b/src/tss2-fapi/tpm_json_deserialize.c
@@ -253,13 +253,25 @@ ifapi_get_sub_object(json_object *jso, char *name, json_object **sub_jso)
 {
     int i;
     if (json_object_object_get_ex(jso, name, sub_jso)) {
-        return true;
+        if (*sub_jso) {
+            return true;
+        } else {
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/tpm_json_deserialize.c
dEajL3kA, line 1:
```
if (*sub_jso) {
    return true;
} else {
    return false;
}
```

could be simplified to:
```
return (*sub_jso);
```

AndreasFuchsTPM: This could give some static code analysis alarms.
```
return !!(*sub_jso);
```
might however work.
But I also don't mind the explicit version.

JuergenReppSIT: Nice solution `return !!(*sub_jso);` but i would prefer the explicit version. @AndreasFuchsTPM could you please approve the PR.

AndreasFuchsTPM (approved):

PR #2899, opened 2024-09-24
Refactor and Update the tcti-x-helper

### LINKED ISSUE #2620: Improve Robustness of tcti-spi-helper
Currently, tcti-spi-helper will just use whatever values it finds for the TIS registers without thinking about them. On a disconnected line or unresponsive TPM this is all 0xffffff...

Some bits on some registers are however known to always be 0 or to always to 1.

The tcti-spi-helper should check these known values. If they differ from then a communication error is detected and can be reported and handled accordingly.

### LINKED ISSUE #2891: SPI-Helper: Checking expect bit requires checking stsValid as well
If stsValid is clear, then the expect bit information is BS. Thus, we need to read the status register until we stsValid turns set. Only then can we evaluate expect.

https://github.com/tpm2-software/tpm2-tss/blob/c641c77769a6953fea8ec986d22c754757426f42/src/tss2-tcti/tcti-spi-helper.c#L385

wxleong:
- Moved common transmit/receive FIFO protocol code to tcti-helper-common.
- Refined the FIFO protocol implementation.
- Updated unit test implementation so that changes to the FIFO protocol no longer require modifications to all dependent unit tests.

Resolves #2891, resolves #2620


src/tss2-tcti/tcti-helper-common.c
```diff
@@ -0,0 +1,761 @@
...
+        return rc;
+    }
+
+    /* Send all but the last byte in the FIFO */
+    rc = tcti_helper_common_fifo_transfer (ctx, (void *)(cmd_buf + 1), size - 2, TCTI_HELPER_COMMON_FIFO_TRANSMIT_2);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOG_ERROR ("Failed to send the command");
+        return rc;
+    }
+
```
AndreasFuchsTPM, line 630: Here we should again check for STS_VALID | EXPECT

src/tss2-tcti/tcti-helper-common.c
```diff
@@ -0,0 +1,761 @@
...
+    /* Send the last byte */
+    rc = tcti_helper_common_fifo_transfer (ctx, (void *)(cmd_buf + size - 1), 1, TCTI_HELPER_COMMON_FIFO_TRANSMIT_2);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOG_ERROR ("Failed to send the command");
+        return rc;
+    }
+
+    /* Await TPM readiness */
+    expected_status_bits = TCTI_HELPER_COMMON_TPM_STS_VALID | TCTI_HELPER_COMMON_TPM_STS_DATA_EXPECT;
```
AndreasFuchsTPM, line 639: Here, the expected status should be STS_VALID | !EXPECT

I wonder why this still works ?

src/tss2-tcti/tcti-helper-common.c
```diff
@@ -0,0 +1,761 @@
...
+    uint16_t vendor_id, device_id;
+    vendor_id = did_vid & 0xffff;
+    device_id = did_vid >> 16;
+    LOG_DEBUG ("Connected to TPM with vid:did:rid of 0x%04" PRIx16 ":%04" PRIx16 ":%02" PRIx8, vendor_id, device_id, rid);
+#endif
+
+    return TSS2_RC_SUCCESS;
+}
+
+TSS2_RC Tcti_Helper_Common_Transmit (TCTI_HELPER_COMMON_CONTEXT *ctx, size_t size, const uint8_t *cmd_buf)
```
AndreasFuchsTPM, line 579: We are missing the retry-logic in this function.
It always goes directly to "return error"

PR #2905, opened 2024-10-17
Add udev rules to create tcm devices

chench246:
TCM is a Chinese standard and is compatible with TPM. Therefore, you also need to add udev rules to create the TCM device node.


chench246:
@AndreasFuchsTPM

williamcroberts:
As well as @JuergenReppSIT since hes the main maintainer now!

Issue #2674, opened 2023-08-04
TCTI state transitions are inconsistent and unpredictable


fergus-dall:
Looking through the TCTI code, I have noticed a couple of inconsistencies.

## Cancellation

The `tcti-mssim` and `tcti-pcap` TCTIs will transition directly to the `TCTI_STATE_TRANSMIT` state after a successful call to `cancel`. This contradicts the specification which states that another call to `receive` is required before another command can be transmitted. This means there's no way for the caller to get the response buffer for the cancelled command, which is required to tell if the command actually executed or not. Higher layers in the stack will likely also hit unexpected errors if a command using these TCTIs is cancelled. For example, I think the ESYS library will enter an unrecoverable internal error state if the caller does
```
Esys_{Command}_Async(context, ...);
Esys_GetTcti(context, &tcti);
tcti->cancel(tcti);
Esys_{Command}_Finish(context, ...);
```
which I believe is the intended pattern for cancelling a command. Note that skipping the `Esys_{Command}_Finish` call leaves the ESYS context in the `_ESYS_STATE_SENT` state where new commands cannot be issued, which is effectively the same as the unrecoverable error state for this purpose.

All other TCTIs in this repo appear to act correctly in this regard. This looks like it's straight forward to fix, but I don't know much about the lower level interface they're using. Given the above, I would guess no one is relying on this behaviour.

## Receive errors

Some TCTIs will also transition to the `TCTI_STATE_TRANSMIT` state on certain errors in the `receive` call. The spec doesn't explicitly comment on this either way, and neither do the man pages or installed headers, but there is a comment in internal `tcti-common.h` header that claims this occurs for all return codes other than `TRY_AGAIN`, `INSUFFICIENT_BUFFER`, `BAD_CONTEXT`, `BAD_REFERENCE`, `BAD_VALUE`, and `BAD_SEQUENCE`, but the behaviour of some TCTIs diverges from this. For example:
- `tcti-cmd` transitions when returning `INSUFFICIENT_BUFFER`.
- `tcti-device` does not transition when returning `IO_ERROR`. It is also ambiguous when returning unmarshaling errors and `GENERAL_FAILURE`, transitioning on some code paths but not others.
- `tcti-i2c-helper` only transitions when returning `SUCCESS`, even though there are paths that return `IO_ERROR` and lower-layer errors that don't document any restrictions on the errors they should return.
[...]

PR #2929, opened 2024-12-19
reset tcti_common->state in tcti_device_receive when return TSS2_TCTI…

xcfxr:
when IO device has something wrong, It needs to reset tcti_common->state to TCTI_STATE_TRANSMIT. Or when another message was sent to tcti_device(E.g. tpm2_getcap) it will fail in tcti_common_transmit_checks.


AndreasFuchsTPM:
Hmmm... Just always resetting state in the tcti is probably not good, if the kernel driver context does not also go back into reseted state.
We would need to look at and document, which error codes lead the kernel driver to go back into ready for the opened fd and then we can implement this behavior for said specific error codes.

joholl:
Related issue: #2674

Issue #2930, opened 2025-01-08
Support of CpHashes from Esys Layer


AndreasFuchsTPM:
Proposal to add

```C
Esys_GetCpHash(ESYS_CONTEXT* ctx, TPM2_ALG_HASH hashAlg, uint8_t **cpHash, size_t *cpHash_size); // Only to be called after Esys_*_Prepare() and before Esys_Execute/_Async() or Esys_Abort()
Esys_Abort(ESYS_CONTEXT ctx) // Can be called instead of Esys_Execute/_Async()
Esys_GetRpHash(ESYS_CONTEXT* ctx, TPM2_ALG_HASH hashAlg, uint8_t **rpHash, size_t *rpHash_size); // Only to be called after Esys_Execute/_Finish()
```

As a followup we could switch the tools over to this approach and clean up `tpm2-tools/lib/tpm2.c`.

PR #2932, opened 2025-01-16
ESYS: Support of Cp and RpHashes from Esys Layer

JuergenReppSIT:
* The ESYS API is extended with the functions:
   Esys_GetCpHash, Esys_GetRpHash, and Esys_Abort.
   The cp hash can computed after the async call of a function. The rp hash
   after the finish call. If only the async call is executed to to compute the cp hash
   Esys_Abort has to be called to enable the execution of further ESYS commands.
* The function Tss2_Sys_Abort to reset the SYS sate is added.
*  The state handling of Tss2_Sys_GetCpBuffer is changed.
* An integration test is added.
* A tcti module which can only used for tcti initialization is added.

Addresses: #2930.

### DIFF
--- a/src/tss2-esys/esys_cp_rp_hash.c
+++ b/src/tss2-esys/esys_cp_rp_hash.c
@@ -1,0 +1,113 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*******************************************************************************
+ * Copyright 2025, Juergen Repp
+ * All rights reserved.
+ ******************************************************************************/
+
+#ifdef HAVE_CONFIG_H
+#include "config.h" // IWYU pragma: keep
+#endif
+
+#include <inttypes.h>         // for PRIx32, uint8_t, SIZE_MAX, int32_t
+#include <stdbool.h>          // for bool, false, true
+#include <stdlib.h>           // for NULL, malloc, size_t, calloc
+#include <string.h>           // for memcmp
+
+#include "esys_int.h"         // for RSRC_NODE_T, ESYS_CONTEXT, _ESYS_ASSERT...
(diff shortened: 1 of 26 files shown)


src/tss2-esys/esys_cp_rp_hash.c
```diff
@@ -0,0 +1,113 @@
...
+    TSS2_RC r;
+
+    return_if_null(esys_ctx, "ESYS context is NULL",
+                   TSS2_ESYS_RC_BAD_REFERENCE);
+    esys_ctx->state = ESYS_STATE_INIT;
+
+    r = Esys_GetSysContext(esys_ctx, &sys_ctx);
+    return_if_error(r, "Could not get Sys context");
+
+    return Tss2_Sys_Abort(sys_ctx);
```
AndreasFuchsTPM, line 112: I would first execute Tss2_Sys_Abort and only if that succeeds we set the state to ESYS_STATE_INIT.

src/tss2-tcti/tcti-null.c
```diff
@@ -0,0 +1,173 @@
...
+    return &tcti_null->common;
+}
+
+TSS2_RC
+tcti_null_transmit (
+    TSS2_TCTI_CONTEXT *tcti_ctx,
+    size_t size,
+    const uint8_t *cmd_buf)
+{
+    UNUSED(tcti_ctx);
```
AndreasFuchsTPM, line 60: I would add a LOG_WARNING to each of these _null_ calls so that the user knows something is weird here.

AndreasFuchsTPM (changes requested):
Very few comments but LGTM otherwise

AndreasFuchsTPM (approved):

PR #2933, opened 2025-01-17
Add support to use TSS with Zephyr

cplappert:
WIP to prepare TSS for Zephyr integration.

@drandreas: I would like to have your opinion on this. I noticed that you [introduced Zephyr](https://github.com/tpm2-software/tpm2-tss/commit/7482d46f1fa017ee370aa132de10f6012dff588c) some time ago. This commit overwrites the void* TSS2_TCTI_POLL_HANDLE, since it was not working for me. Would this be a problem from your side?

### DIFF
--- a/doc/tss-for-zephyr.md
+++ b/doc/tss-for-zephyr.md
@@ -1,0 +1,321 @@
+# TSS for Zephyr Example with Remote TPM
+## Compiling the TSS for Zephyr in QEMU and connect to Hostside TPM Simulator with Remote TCTI
+
+
+This example shows how to compile the TSS for an application running in [Zephyr RTOS](https://zephyrproject.org/).
+The app is executed on QEMU and uses the remote TCTI to connect to a hostside Software TPM.
+
+## Building the TSS
+
+### 1. Adding TSS and mbedTLS to the Zephyr project
+
+To integrate the TSS for Zephyr, the following files needs to be adjusted.
+
+#### west.yml
+
+```yaml
+manifest:
+    remotes:
+        [...]
+        - name: tpm2-software
+            url-base: https://github.com/tpm2-software
+
+    projects:
+        [...]
+        - name: tpm2-tss
+          remote: tpm2-software
+          revision: master
+          path: my_zephyr_app/lib/tpm2-tss
+
+        - name: mbedtls
+          remote: zephyrproject-rtos
+          revision: v3.0.0
+          path: modules/crypto/mbedtls
+```
+
+#### prj.conf
+
+```yaml
+[...]
+# Enable Basic POSIX and socket support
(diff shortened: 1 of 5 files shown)


drandreas:
Hi @cplappert thank you for reaching out. It has been a while since did those changes. I also changed to a different team so I'm not up-to-date on the Zephyr integration.

I've reached out to a colleague who is still working on the project. I received the following response (they still use the tpm-tss Zephyr integration): "
The patch is fine in my opinion. I don't understand its description in doc/tss-for-zephyr.md and it seems too complicated to me.

`TSS2_TCTI_POLL_HANDLE` simply has to be defined. It is only used as a parameter for `get_poll_handles` which only returns `TSS2_TCTI_RC_NOT_IMPLEMENTED`.

It is somewhat unattractive that `UNUSED` is now defined twice with the `#include <zephyr/posix/poll.h>` and the following warning is issued very often:
`
.../modules/lib/tpm2-tss/import/src/util/aux_util.h:33: warning: "UNUSED" redefined
   33 | #define UNUSED(x) (void)(x)
`
"

cplappert:
Hi @drandreas,

thank you for your feedback and reaching out to your collegue.

If I understand correctly overwriting the TSS2_TCTI_POLL_HANDLE is no issue on your side.

Regarding the `UNUSED` definition, I do not get a warning on my side. 
I also do no see an `UNUSED` definition in [`#include <zephyr/posix/poll.h>`](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/posix/poll.h).

If I grep for `UNUSED(x)` in the Zephyr repository, I only get two definitions with another name:

```console
zephyr$ grep -r "UNUSED(x)"
scripts/native_simulator/common/src/include/nsi_utils.h:#define NSI_ARG_UNUSED(x) (void)(x)
include/zephyr/toolchain/gcc.h:#define ARG_UNUSED(x) (void)(x)
```

Maybe we have different versions? I use both main branches from Zephyr and TSS.
Anyway, I could just add a guard for the `UNUSED` definition (`#ifndef UNUSED`) in [tpm2tss/src/util/aux_util.h](https://github.com/tpm2-software/tpm2-tss/blob/master/src/util/aux_util.h#L35).


Regarding `doc/tss-for-zephyr.md`: Could you provide more details on how a simpler integration of TSS into Zephyr could look like?

cplappert:
FYI: 
- I added the guard for the UNUSED definition 
- I updated the doc description with a working example for a Qemu-based application communicating with a hostside TPM simulator using Remote TCTI.  In future, we can simplify the description further if possible.

JuergenReppSIT (approved):

cplappert:
Thank you for the follow-up. Luckily, we have already integrated the guard for `UNUSED`. The merged PR should then cause no additional issues for Zephyr project.

PR #2976, opened 2025-07-23
FAPI: Add ECC encryption/decryption to Fapi_Encrypt/Decrypt

JuergenReppSIT:
The cipher produced by Fapi_Encrypt contains the marshaled data for the c1, c2, and c3 parameter described in the TPM Spec 1.83. TPM2B_ECC_POINT c1;
TPM2B_MAX_BUFFER c2;
TPM2B_DIGEST c3;


JuergenReppSIT:
> Would it also be possible to rebase this PR onto the current main branch? It does cause a lot of non-trivial merge conflicts when I try to do so after commit [a1c863e](https://github.com/tpm2-software/tpm2-tss/commit/a1c863e29c32ade602a092a49fe133b09127220f)

I will rebase the PR when PR #3034 is merged.

PR #2995, opened 2025-09-18
tcti: add interface config option for spi-ftdi.

mattifinder:
Allow users to specify which interface they want to use from the chip.
To not break previous use cases, unspecified interfaces default to A.

This was necessary for me to use this programmer with my TIGARD board, since it uses interface B for communicating via SPI.


AndreasFuchsTPM:
I don't know why CI failed and I also cannot restart it.
Could you maybe change a little something and force-push ?

PR #3008, opened 2025-10-30
Fix a bunch of CodeQL warning

AndreasFuchsTPM:
This way, we can have a better overview if there is anything actually important in there.


src/tss2-esys/esys_iutil.c
JuergenReppSIT: why is SAFE_FREE replaced with free several times?

src/tss2-tcti/tcti-libtpms.c
JuergenReppSIT: why is the check removed?

src/tss2-esys/esys_free.c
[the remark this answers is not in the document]
AndreasFuchsTPM: It does "not need" to be checked, because free(NULL) is allowed.
And CodeQL gives and error on this.

JuergenReppSIT (approved):

PR #3021, opened 2025-11-10
Resume upgrade after power interruption #2984

### LINKED ISSUE #2984: Resume upgrade after power interruption
Hello, 
In the case of resuming a firmware upgrade after a power interruption, the TPM should respond to any command with TPM_RC_UPGRADE, as specified in the TCG documentation.
However, when issuing the TPM2_Startup command after the power interruption, the call returns an error. This is because TPM_RC_UPGRADE is not handled as an expected response in the current implementation. 
As shown in the code excerpt below (from Esys_Startup.c), any returned code other than TSS2_RC_SUCCESS and TPM2_RC_INITIALIZE leads to the command being treated as an error:

if (iesys_tpm_error(r) && r != TPM2_RC_INITIALIZE) {
        LOG_WARNING("Received TPM Error");
        esysContext->state = _ESYS_STATE_INIT;
        return r;
    } else if (r != TSS2_RC_SUCCESS && r != TPM2_RC_INITIALIZE) {
        LOG_ERROR("Received a non-TPM Error");
        esysContext->state = _ESYS_STATE_INTERNALERROR;
        return r;
    }

Should TPM_RC_UPGRADE be considered a valid and expected return code for TPM2_Startup in this context? Or should it be handled differently? 

Here is a console log showing the error:
Image

Meriem25-tech:
Return TPM_RC_UPGRADE in Esys_Startup and log an informational message when the TPM is in upgrade mode.
Skip calling Tss2_Sys_Startup_Complete when TPM_RC_UPGRADE is returned.
Fixes the error encountered when issuing the Startup command after resuming a firmware upgrade.
Fixes #2984.

### DIFF
--- a/src/tss2-esys/api/Esys_Startup.c
+++ b/src/tss2-esys/api/Esys_Startup.c
@@ -5,11 +5,11 @@
  ******************************************************************************/
 
 #ifdef HAVE_CONFIG_H
-#include "config.h" // IWYU pragma: keep
+#include "config.h"  // IWYU pragma: keep
 #endif
 
-#include <inttypes.h>         // for PRIx32, PRIx16, int32_t
-#include <stddef.h>           // for NULL
+#include <inttypes.h>  // for PRIx32, PRIx16, int32_t
+#include <stddef.h>    // for NULL
 
 #include "esys_int.h"         // for ESYS_CONTEXT, _ESYS_STATE_INTERNALERROR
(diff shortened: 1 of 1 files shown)


Meriem25-tech:
Hello, I opened a new pull request #3030 that replaces this one. 
Could you please close this PR? Thank you.

PR #3046, opened 2026-01-23
Initial PQC work

tdjCisco:
- Created new configure option --enable-pqc for new PQC TPM Commands
- Check for openssl 3.5 or greater when the option is used.


include/tss2/tss2_tpm2_types.h
```diff
@@ -15,8 +15,8 @@
 
 #include <stdint.h>
 
-#define TPM2_MAX_COMMAND_SIZE  4096 /* maximum size of a command */
-#define TPM2_MAX_RESPONSE_SIZE 4096 /* maximum size of a response */
+#define TPM2_MAX_COMMAND_SIZE  8192 /* maximum size of a command */
```
AndreasFuchsTPM, line 18: Could we somehow evaluate what the exact number of a maximum buffer size would be ?
Is 8192 too large or too small ?
I think we really need to know a validated value here.

test/integration/esys-pqc-signdigest.int.c
```diff
@@ -0,0 +1,383 @@
...
+
+    memcpy(mu_digest.buffer, mu, 64);
+
+    /* Set up the context string for SignDigest / VerifyDigestSignature */
+    sig_ctx.size = (UINT16)strlen(ctx_str);
+    memcpy(sig_ctx.buffer, ctx_str, sig_ctx.size);
+
+    /* ---- SignDigest --------------------------------------------------- */
+    r = Esys_SignDigest(esys_context, mldsa_handle, ESYS_TR_PASSWORD, ESYS_TR_NONE, ESYS_TR_NONE,
+                        &sig_ctx, &mu_digest, &hashcheck, &signature);
```
lcharpentier-stm, line 332: It is not needed to provide the context to SignDigest, the context is already included in the external MU.
The device receiving this context will probably ignore it or respond an error.

AndreasFuchsTPM:
TODOs before review:
- FAPI_Sign() missing ML_DSA support
- Tests for FAPI_Sign()
- Cleanup of commit history
  - 1 commit for Sys and Types
  - 1 commit for Esys
  - 1 commit for Fapi 
- rebase upon latest master branch

PR #3068, opened 2026-04-02
FAPI: fix TOCTOU race in ifapi_io_read_async

hyperfinitism:
This PR removes `fclose` + `fopen` in `ifapi_io_read_async` to avoid the TOCTOU risk. It also replaces `fseek` + `ftell` logic with a single `statbuf.st_size`.

### DIFF
--- a/src/tss2-fapi/ifapi_io.c
+++ b/src/tss2-fapi/ifapi_io.c
@@ -141,24 +141,8 @@ ifapi_io_read_async(struct IFAPI_IO *io, const char *filename) {
         return TSS2_FAPI_RC_IO_ERROR;
     }
 
-    if (fseek(io->stream, 0L, SEEK_END) == -1) {
-        LOG_ERROR("fseek failed for \"%s\".", filename);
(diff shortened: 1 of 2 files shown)


hyperfinitism:
Pushed a test fix in the latest commit: updated mock expectations in `check_io_read_async` to match the new syscall sequence.

On a side note — this test is tightly coupled to the implementation's internal syscall ordering via `--wrap` mocks. It essentially replays the exact control flow rather than testing behavioral contracts (e.g. "return `IO_ERROR` when the file cannot be read"). As a result, any refactor to the internals (like this TOCTOU fix) requires a mechanical mock update even though the external behavior is unchanged.

hyperfinitism:
Apologies for the CI runs timing out, and they ran for an hour until hitting the rate limit. I have since run the CI on my fork and confirmed that all jobs pass green across *all* environments; it appears that the tests were previously getting stuck in some environments.

JuergenReppSIT (approved):

PR #3070, opened 2026-04-04
ifapi_profiles: proper OOM handling for profile realloc

hyperfinitism:
This PR adds missing NULL checks after `realloc` / `reallocarray` in `ifapi_profiles_initialize_async()`.

The current code assumes that shrinking allocations cannot fail and therefore can omit OOM checks. However, the C/POSIX specification does not guarantee that `realloc` will succeed when reducing the allocated size.

### DIFF
--- a/src/tss2-fapi/ifapi_profiles.c
+++ b/src/tss2-fapi/ifapi_profiles.c
@@ -110,18 +110,24 @@ ifapi_profiles_initialize_async(IFAPI_PROFILES *profiles,
         return TSS2_FAPI_RC_BAD_VALUE;
     }
 #ifdef HAVE_REALLOCARRAY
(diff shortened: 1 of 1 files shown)


hyperfinitism:
I have addressed lint error (`clang-tidy`) and also applied `clang-format` to all code.

JuergenReppSIT:
@hyperfinitism for clang-format you should use:
See: https://github.com/tpm2-software/tpm2-tss/?tab=contributing-ov-file#readme

hyperfinitism:
Reverted `clang-format` changes and applied again inside `ghcr.io/tpm2-software/ubuntu-24.04`.

JuergenReppSIT (approved):

PR #3077, opened 2026-04-28
FAPI: validate eventlog size before first dereference in parse_eventlog

### LINKED ISSUE #3076: `parse_eventlog`: missing length check before reading `event->eventType`
### Bug

`parse_eventlog` dereferences `event->eventType` (offset 4 of `TCG_EVENT`) without first validating that the input buffer is at least `sizeof(TCG_EVENT)` bytes. The function takes a `(BYTE *eventlog, size_t size)` pair, so the size argument exists for a reason; it just is not consulted before the first field read. Any caller that passes a buffer shorter than 8 bytes hits a heap OOB read on the first dereference, including the empty-buffer case.

```c
// src/tss2-fapi/ifapi_eventlog_system.c, around line 692
bool
parse_eventlog(tpm2_eventlog_context *ctx, BYTE const *eventlog, size_t size) {
    TCG_EVENT_HEADER2 *next;
    TCG_EVENT         *event = (TCG_EVENT *)eventlog;
    bool               ret;

    if (event->eventType == EV_NO_ACTION) {     // OOB read when size < sizeof(TCG_EVENT)
        ...
```

### Reachability

Public `Fapi_GetEventLog` doesn't reach this today: `file_to_buffer` always does `calloc(1, UINT16_MAX)` before reading the file, so even an empty file is backed by a 65535-byte allocation and the offset-4 read lands on zero padding. Live only when something passes `ifapi_tcg_eventlog_serialize` a tight buffer directly.

Still worth fixing in `parse_eventlog` itself: it takes a `(buffer, size)` pair and ignores `size` until later. Validating it up front is more robust than relying on every caller to over-allocate.

### Reproduction

Direct call into `ifapi_tcg_eventlog_serialize` with a 4-byte allocation under ASAN:

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 4
    #0 parse_eventlog                ifapi_eventlog_system.c:698
    #1 ifapi_tcg_eventlog_serialize  ifapi_json_eventlog_serialize.c:921
located 0 bytes after 4-byte region
```

PoC source available on request.

### Suggested fix

```c
if (size < sizeof(TCG_EVENT)) {
    LOG_ERROR("eventlog too small for first TCG_EVENT (%zu < %zu)",
              size, sizeof(TCG_EVENT));
    return false;
}
```

Happy to send a PR.

MarkLee131:
Fix https://github.com/tpm2-software/tpm2-tss/issues/3076:
parse_eventlog dereferences event->eventType (offset 4 of TCG_EVENT) before validating that the input buffer is at least sizeof(TCG_EVENT) bytes. Any caller that passes a buffer shorter than 8 bytes triggers a heap OOB read on the first field access, including the empty-buffer case.

Public Fapi_GetEventLog is currently safe because file_to_buffer always calloc(1, UINT16_MAX) and the offset-4 read lands on zero padding, but the function-level invariant is wrong: parse_eventlog accepts a (buffer, size) pair and uses size correctly later (for foreach_event2, specid_event), only the first dereference skips the precondition check.

Add the up-front size check, matching the bound-check convention already used throughout this file (lines 117, 124, 232, 250, 256, 269, 280, ...).

### DIFF
--- a/src/tss2-fapi/ifapi_eventlog_system.c
+++ b/src/tss2-fapi/ifapi_eventlog_system.c
@@ -695,6 +695,14 @@ parse_eventlog(tpm2_eventlog_context *ctx, BYTE const *eventlog, size_t size) {
     TCG_EVENT         *event = (TCG_EVENT *)eventlog;
     bool               ret;
 
+    if (size == 0) {
+        return true;
+    }
+    if (size < sizeof(*event)) {
+        LOG_ERROR("eventlog too small for first TCG_EVENT (%zu < %zu)", size, sizeof(*event));
+        return false;
+    }
+
     if (event->eventType == EV_NO_ACTION) {
         ret = specid_event(event, size, &next);
         if (!ret) {


JuergenReppSIT:
@MarkLee131 Thank you for the PR. Could you please sign the commits.
For simplicity, we use clang-format for all files (also as part of the CI). Thus before commiting any code, we recommend you run

This ensures that the code will not break during your pull request.

MarkLee131:
Hi @JuergenReppSIT, I have completed the requested updates, and all CI checks are now passing.
Please let me know if there is anything else I can do to help move this PR forward. Thanks! :)

JuergenReppSIT (approved):

PR #3078, opened 2026-04-28
FAPI: validate EventSize before reading Event[] in check_event_string

### LINKED ISSUE #3075: `check_event_string`: `strncmp` reads `strlen(string)` without validating `EventSize`
### Bug

`check_event_string` reads `strlen(string)` bytes from `in->Event[]` via `strncmp` before checking that `in->EventSize >= strlen(string)`. The `EventSize` check happens later, after the comparison has already run. Any `TCG_EVENT2` whose `EventSize` is smaller than the expected fixed string causes `strncmp` to read past the event payload.

```c
// src/tss2-fapi/ifapi_json_eventlog_serialize.c, around line 273
static TSS2_RC
check_event_string(const TCG_EVENT2 *in, const char *string) {
    if (strncmp((const char *)&in->Event[0], string, strlen(string)) != 0) {
        return_error2(TSS2_FAPI_RC_BAD_VALUE, "Invalid event string. %s expected", string);
    }
    if (in->EventSize == strlen(string)) {
        return TSS2_RC_SUCCESS;
    }
    if (in->EventSize == strlen(string) + 1 && in->Event[strlen(string)] == '\0') {
        return TSS2_RC_SUCCESS;
    }
    return_error2(TSS2_FAPI_RC_BAD_VALUE, "Invalid event string. %s expected", string);
}
```

The two callers pass fixed strings:

```c
// line 348-356
case EV_EFI_HCRTM_EVENT:
    r = check_event_string(in, "HCRTM");                  // 5 bytes
case EV_OMIT_BOOT_DEVICE_EVENTS:
    r = check_event_string(in, "BOOT ATTEMPTS OMITTED");  // 21 bytes
```

So an HCRTM event with `EventSize = 4` causes `strncmp` to read 5 bytes from a 4-byte event payload. The `EventSize` field literally tells the function the payload is 4 bytes; the function ignores it during the comparison and then validates it afterwards.

### Reachability

Public `Fapi_GetEventLog` doesn't reach this today: `parse_event2` checks `buf_size >= event_size + EventSize`, and `file_to_buffer`'s `calloc(1, UINT16_MAX)` over-allocates so the extra bytes `strncmp` reads land on zero padding. Both layers have to hold for the OOB to stay masked.

Still worth fixing in `check_event_string` itself: validating `EventSize` after using `Event[]` is just the wrong order.

### Reproduction

Direct call into `ifapi_json_TCG_EVENT2_serialize` with `EV_EFI_HCRTM_EVENT` and a `TCG_EVENT2` allocated as `sizeof(TCG_EVENT2) + 4`:

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 5
    #0 strncmp
    #1 check_event_string             ifapi_json_eventlog_serialize.c:275
    #2 ifapi_json_TCG_EVENT2_serialize ifapi_json_eventlog_serialize.c:350
located 0 bytes after 8-byte region
```

PoC source available on request.

### Suggested fix

Validate `EventSize` first, and switch to `memcmp` since the comparison is bounded:

```c
size_t expected = strlen(string);
if (in->EventSize < expected) {
    return_error2(TSS2_FAPI_RC_BAD_VALUE, "EventSize %u too small for %s",
                  in->EventSize, string);
}
if (memcmp(in->Event, string, expected) != 0) {
    return_error2(TSS2_FAPI_RC_BAD_VALUE, "Invalid event string. %s expected", string);
}
```

Happy to send a PR.

MarkLee131:
Fix https://github.com/tpm2-software/tpm2-tss/issues/3075:
check_event_string compares in->Event[] against a fixed string with strncmp(in->Event, string, strlen(string)) before validating that in->EventSize >= strlen(string). When EventSize is smaller than the expected fixed string, the comparison reads past the event payload. Two callers pass fixed strings: EV_EFI_HCRTM_EVENT ("HCRTM", 5 bytes) and EV_OMIT_BOOT_DEVICE_EVENTS ("BOOT ATTEMPTS OMITTED", 21 bytes).

Public Fapi_GetEventLog has two layered caller-side mitigations on this path: parse_event2 enforces buf_size >= event_size + EventSize, and file_to_buffer over-allocates the file buffer to calloc(1, UINT16_MAX). Together they mask the OOB on stock builds for the HCRTM case. The BOOT ATTEMPTS OMITTED case (21-byte string) needs up to 17 bytes of trailing slack and can OOB on file sizes near UINT16_MAX*n boundaries, so the bug is reachable through public API on narrow preconditions and is straightforwardly latent on direct callers.

Validate EventSize first, and switch to memcmp now that the comparison is bounded. Same shape as the EV_EFI_HCRTM_EVENT check at ifapi_eventlog_system.c:208 (parse_event2body), which already validates EventSize == len before strncmp via short-circuit evaluation.

### DIFF
--- a/src/tss2-fapi/ifapi_json_eventlog_serialize.c
+++ b/src/tss2-fapi/ifapi_json_eventlog_serialize.c
@@ -272,14 +272,19 @@ ifapi_json_TCG_DIGEST2_cb(const TCG_DIGEST2 *in, size_t size, void *data) {
  */
 static TSS2_RC
 check_event_string(const TCG_EVENT2 *in, const char *string) {
-    if (strncmp((const char *)&in->Event[0], string, strlen(string)) != 0) {
+    size_t expected = strlen(string);
+    if (in->EventSize < expected) {
+        return_error2(TSS2_FAPI_RC_BAD_VALUE, "EventSize %u too small for %s", in->EventSize,
+                      string);
+    }
+    if (memcmp(in->Event, string, expected) != 0) {
         return_error2(TSS2_FAPI_RC_BAD_VALUE, "Invalid event string. %s expected", string);
     }
-    if (in->EventSize == strlen(string)) {
+    if (in->EventSize == expected) {
         /* String without NULL terminator */
         return TSS2_RC_SUCCESS;
     }
-    if (in->EventSize == strlen(string) + 1 && in->Event[strlen(string)] == '\0') {
+    if (in->EventSize == expected + 1 && in->Event[expected] == '\0') {
         /* String with NULL terminator */
         return TSS2_RC_SUCCESS;
     }


MarkLee131:
Hi @JuergenReppSIT, please let me know if there is anything else I can do to help move this PR forward. Thanks! :)

JuergenReppSIT (approved):

PR #3079, opened 2026-04-29
Update OpenSSL API compatibility

shoracek:
This change allows the code to be compiled against OpenSSL 4.

Replace direct access to ASN1_STRING data members with the ASN1_STRING_get0_data() accessor function, as modern OpenSSL versions use opaque structures. Additionally, add const qualifiers to X509_NAME pointers to ensure const-correctness with newer OpenSSL APIs.


JuergenReppSIT:
@shoracek Thank you for the PR.
For simplicity, we use clang-format for all files (also as part of the CI). Thus before commiting any code, we recommend you run

This ensures that the code will not break during your pull request.

shoracek:
Done. I thought that it would be better to keep the existing formatting of the line, but I'll keep that in mind for next time.

AndreasFuchsTPM (approved):

PR #3084, opened 2026-05-06
build: add CMake (Windows only)

asamy:
This allows out-of-tree building and some automation for external projects, specifically those that use it as a static library. Unfortunately the MSVC projects that are provided aren't flexible, specifically ConfigurationType and RuntimeLibrary aren't configurable via the command line due to the way they are specified, it was also difficult to configure other things such as toolset etc.

This is only limited to Windows. I am not setup for Linux/Mac.


AndreasFuchsTPM:
I'd really like to see CI for this before merging.

asamy:
Done.

I also recommend that you adjust repo settings to allow for NuGet package cache (vcpkgs) so that the build just fetches the prebuilt libs from there instead of rebuilding each time which usually takes some time. The NuGet fetch part is done in the CI file.

asamy:
Hi, is this being considered now that CI is in as requested? I would rather not maintain an own fork just for this small build thing.

Thanks

PR #3093, opened 2026-05-28
FAPI: Fix json deserialization

JuergenReppSIT:
* A memory leak, which occurs if the conversion of an hex string to an byte array fails is fixed.
* Due to complaints from the code checker, the `get-number` function has been modified to always return false if errors occur or when an empty string is passed.

### DIFF
--- a/src/tss2-fapi/tpm_json_deserialize.c
+++ b/src/tss2-fapi/tpm_json_deserialize.c
@@ -211,23 +211,41 @@ ifapi_json_byte_deserialize(json_object *jso, UINT32 max, BYTE *out, UINT16 *out
  * @retval true if token represents a number
  * @retval false if token does not represent a number.
  */
+#include <inttypes.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
 static bool
 get_number(const char *token, int64_t *num) {
-    int itoken = 0;
     int pos = 0;
-    if (!token) {
+    int ret;
+
+    if (token == NULL || num == NULL) {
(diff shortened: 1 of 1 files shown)


src/tss2-fapi/tpm_json_deserialize.c
```diff
@@ -211,23 +211,41 @@ ifapi_json_byte_deserialize(json_object *jso, UINT32 max, BYTE *out, UINT16 *out
...
+        if (unum > INT64_MAX) {
+            return false;
+        }
+        *num = (int64_t)unum;
+        return (size_t)pos == strlen(token + 2);
     } else {
-        sscanf(&token[itoken], "%" PRId64 "%n", num, &pos);
+        int64_t snum = 0;
+
+        ret = sscanf(token, "%" SCNd64 "%n", &snum, &pos);
```
moritzbuhl, line 242: ~I think we should get rid of %n while touching this.~
~pos should be the same value as ret.~

~%n is a terrible gadget for binary exploitation and I am not aware of any use case where it is a necessary evil.~
sorry, in sscanf it behaves fine, in printf it is bad. I got this wrong.

src/tss2-fapi/tpm_json_deserialize.c
moritzbuhl, line 383: if strlen(hex_string) is odd, dividing by two should truncates one byte then?
Or does hex_string start with 0x... and then we potentially allocate one byte too much?

moritzbuhl (approved):

Issue #2889, opened 2024-08-15
tpm2 createprimary fails on Ubuntu 22.04 LTS When FIPS is enabled


galroty:
Hey,
I run on ubuntu 22 and after enabling fips (using Ubuntu pro, enabled fips-preview) tpm2 createprimary fails. before enabling fips everything worked fine.
`user@user:~$ uname -a
Linux user 5.15.0-73-fips #80+fips1-Ubuntu SMP Thu Jun 1 20:57:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
`
`user@user:~$ sudo tpm2 createprimary 
[sudo] password for user: 
ERROR:esys_crypto:src/tss2-esys/esys_crypto_ossl.c:412:iesys_cryptossl_hmac_start() ErrorCode (0x00070001) DigestSignInit 
ERROR:esys_crypto:src/tss2-esys/esys_crypto.c:185:iesys_crypto_authHmac() Error ErrorCode (0x00070001) 
ERROR:esys:src/tss2-esys/esys_iutil.c:1244:iesys_compute_hmac() HMAC error ErrorCode (0x00070001) 
ERROR:esys:src/tss2-esys/esys_iutil.c:1354:iesys_gen_auths() Error while computing hmacs ErrorCode (0x00070001) 
ERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:244:Esys_CreatePrimary_Async() Error in computation of auth values ErrorCode (0x00070001) 
ERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:110:Esys_CreatePrimary() Error in async function ErrorCode (0x00070001) 
ERROR: Esys_CreatePrimary(0x70001) - esapi:Catch all for all errors not otherwise specified
ERROR: Unable to run createprimary
`
`user@user:~$ openssl version
OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)
user@user:~$ openssl list -providers
Providers:
  base
    name: OpenSSL Base Provider
    version: 3.0.2
    status: active
  fips
    name: Ubuntu 22.04 OpenSSL Cryptographic Module
    version: 3.0.5-0ubuntu0.1+Fips2.1
    status: active
user@user:~$ 
[0] 0:bash  1:`
I also tried building and installing tpm2-tss and tpm2-tools from master but it didn't help, and I encountered the same error.
[...]

PR #3098, opened 2026-06-10
iesys: Skip computing HMAC authorizations for pure policy sessions

matthewruffell:
This PR addresses an issue in using tpm2_unseal on systems running in FIPS mode, that implement FIPS 140-3 and also SP 800-131A [1] .

[1] https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar2.pdf

FIPS SP800-131A [1], Section 10, Table 9, which can be found on named page 20, but really page 26 in your pdf reader, mentions that for HMAC generations, key lengths < 112 bits are "Disallowed", and key lengths >= 112 bits are "Acceptable".

When a pure policy session is used, on both the command and response side, HMACs are computed with 0 length keys, which causes issues with openssl in FIPS mode.

This PR  changes the command side, by modifying iesys_compute_hmac() to return an empty authorization HMAC instead of computing a new HMAC with a 0 length key.

The response side is also changed, by modifying iesys_check_rp_hmacs() to extend the special case for POLICY_PASSWORD to also include pure policy sessions, instead of computing a new HMAC with a 0 length key.

Addresses: #2889

I'm not sure if you are interested in these changes, I just wanted to open the PR to see if you have any feedback on it, and it might end up useful as a vendor patch for other distros that also need to make FIPS SP800-131A work with their packages.

### DIFF
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -816,7 +816,10 @@ iesys_check_rp_hmacs(ESYS_CONTEXT *esys_context, TSS2L_SYS_AUTH_RESPONSE *rspAut
             continue;
 
         IESYS_SESSION *rsrc_session = &session->rsrc.misc.rsrc_session;
-        if (rsrc_session->type_policy_session == POLICY_PASSWORD) {
+        if (rsrc_session->type_policy_session == POLICY_PASSWORD
+            || (rsrc_session->sessionType == TPM2_SE_POLICY
+                && rsrc_session->type_policy_session != POLICY_AUTH
+                && rsrc_session->sizeHmacValue == 0)) {
             /* A policy password session has no auth value */
             if (rspAuths->auths[i].hmac.size != 0) {
                 LOG_ERROR("PolicyPassword session's HMAC must be 0-length.");
@@ -1165,6 +1168,19 @@ iesys_compute_hmac(ESYS_CONTEXT      *esys_context,
 
     if (session != NULL) {
         IESYS_SESSION *rsrc_session = &session->rsrc.misc.rsrc_session;
+
+        /* For pure policy sessions, return an empty authorization HMAC
+           instead of computing a HMAC with a zero length key. */
+        if (rsrc_session->sessionType == TPM2_SE_POLICY
+            && rsrc_session->type_policy_session != POLICY_AUTH
+            && rsrc_session->sizeHmacValue == 0) {
+            auth->hmac.size = 0;
+            auth->sessionHandle = session->rsrc.handle;
+            auth->nonce = rsrc_session->nonceCaller;
+            auth->sessionAttributes = rsrc_session->sessionAttributes;
+            return TSS2_RC_SUCCESS;
+        }
+
         r = iesys_crypto_hash_get_digest_size(rsrc_session->authHash, &authHash_size);
         return_if_error(r, "Initializing auth session");
 


moritzbuhl:
Looks good to me. But I am still new to the project so I would like @AndreasFuchsTPM or @JuergenReppSIT to take a look too.
Two questions I have after reading this change are:

1. is this the correct location or would an earlier or later check make more sense
2. are all the new statements required or would it make sense to never call these functions for 0 length keys

I'll see if I can answer these questions myself first.

moritzbuhl (dismissed):

AndreasFuchsTPM (changes requested):
Please fix the formatting according to our clang-format-rules:

```diff
--- a/src/tss2-esys/esys_iutil.c
+++ b/src/tss2-esys/esys_iutil.c
@@ -816,10 +816,10 @@ iesys_check_rp_hmacs(ESYS_CONTEXT *esys_context, TSS2L_SYS_AUTH_RESPONSE *rspAut
             continue;
 
         IESYS_SESSION *rsrc_session = &session->rsrc.misc.rsrc_session;
-        if (rsrc_session->type_policy_session == POLICY_PASSWORD ||
-            (rsrc_session->sessionType == TPM2_SE_POLICY &&
-             rsrc_session->type_policy_session != POLICY_AUTH &&
-             rsrc_session->sizeHmacValue == 0)) {
+        if (rsrc_session->type_policy_session == POLICY_PASSWORD
+            || (rsrc_session->sessionType == TPM2_SE_POLICY
+                && rsrc_session->type_policy_session != POLICY_AUTH
+                && rsrc_session->sizeHmacValue == 0)) {
             /* A policy password session has no auth value */
             if (rspAuths->auths[i].hmac.size != 0) {
                 LOG_ERROR("PolicyPassword session's HMAC must be 0-length.");
@@ -1171,14 +1171,14 @@ iesys_compute_hmac(ESYS_CONTEXT      *esys_context,
 
         /* For pure policy sessions, return an empty authorization HMAC
            instead of computing a HMAC with a zero length key. */
-        if (rsrc_session->sessionType == TPM2_SE_POLICY &&
-            rsrc_session->type_policy_session != POLICY_AUTH &&
-            rsrc_session->sizeHmacValue == 0) {
-                auth->hmac.size = 0;
-                auth->sessionHandle = session->rsrc.handle;
-                auth->nonce = rsrc_session->nonceCaller;
-                auth->sessionAttributes = rsrc_session->sessionAttributes;
-                return TSS2_RC_SUCCESS;
+        if (rsrc_session->sessionType == TPM2_SE_POLICY
+            && rsrc_session->type_policy_session != POLICY_AUTH
+            && rsrc_session->sizeHmacValue == 0) {
+            auth->hmac.size = 0;
+            auth->sessionHandle = session->rsrc.handle;
+            auth->nonce = rsrc_session->nonceCaller;
+            auth->sessionAttributes = rsrc_session->sessionAttributes;
+            return TSS2_RC_SUCCESS;
         }
 
         r = iesys_crypto_hash_get_digest_size(rsrc_session->authHash, &authHash_size);
```

matthewruffell:
At the moment I gate this behind a vendor patch that checks if FIPS mode is enabled, the patch is against the old formatted patch, but you get the idea.

```
From: Matthew Ruffell <matthew.ruffell@canonical.com>
Date: Tue, 9 Jun 2026 17:13:44 +1200
Subject: Add FIPS mode detection and use it to skip hmac for policy sessions
 We want to only skip calculation of the authorization HMAC for pure
 PolicyAuthValue / PolicyPassword policy sessions, since these will
 produce a HMAC length of 0.
 
 When FIPS mode is disabled, go back to the regular behaviour.
Origin: vendor
Bug-Ubuntu: https://bugs.launchpad.net/bugs/2074270
Last-Updated: 2026-06-10

Index: tpm2-tss-4.1.3/src/tss2-esys/esys_iutil.c
===================================================================
--- tpm2-tss-4.1.3.orig/src/tss2-esys/esys_iutil.c	2026-06-10 15:57:24.333003189 +1200
+++ tpm2-tss-4.1.3/src/tss2-esys/esys_iutil.c	2026-06-10 15:57:24.331589196 +1200
@@ -8,6 +8,7 @@
 #endif
 
 #include <inttypes.h>
+#include <stdio.h>
 
 #include "tss2_esys.h"
 #include "esys_mu.h"
@@ -965,6 +966,35 @@
     return TSS2_RC_SUCCESS;
 }
 
+/** Check whether the kernel crypto subsystem is running in FIPS mode.
+ *
+ * Reads /proc/sys/crypto/fips_enabled and reports whether it contains 1.
+ * On a FIPS 140-3 system the OpenSSL FIPS provider enforces SP 800-131A and
+ * rejects HMACs computed with a too-short (e.g. zero-length) key. This is used
+ * to gate the skipping of HMAC computation for pure policy sessions so that
+ * behaviour on non-FIPS systems is left unchanged.
+ * @retval true if FIPS mode is enabled, false otherwise.
+ */
+static bool
+iesys_fips_mode_enabled(void)
+{
+    /* FIPS mode cannot change at runtime, so the result is cached after the
+       first read of /proc/sys/crypto/fips_enabled. */
+    static int fips_mode = -1;
+
+    if (fips_mode < 0) {
+        int fips_enabled = 0;
+        FILE *fp = fopen("/proc/sys/crypto/fips_enabled", "r");
+        if (fp != NULL) {
+            if (fscanf(fp, "%d", &fips_enabled) != 1)
+                fips_enabled = 0;
+            fclose(fp);
+        }
+        fips_mode = (fips_enabled == 1) ? 1 : 0;
+    }
+    return fips_mode == 1;
+}
+
 /** Check the HMAC values of the response for all sessions.
  *
  * The HMAC values are computed based on the session secrets, the used nonces,
@@ -994,7 +1024,8 @@
 
         IESYS_SESSION *rsrc_session = &session->rsrc.misc.rsrc_session;
         if (rsrc_session->type_policy_session == POLICY_PASSWORD ||
-            (rsrc_session->sessionType == TPM2_SE_POLICY &&
+            (iesys_fips_mode_enabled() &&
+             rsrc_session->sessionType == TPM2_SE_POLICY &&
              rsrc_session->type_policy_session != POLICY_AUTH &&
              rsrc_session->sizeHmacValue == 0)) {
             /* A policy password session has no auth value */
@@ -1328,7 +1359,8 @@
 
         /* For pure policy sessions, return an empty authorization HMAC
            instead of computing a HMAC with a zero length key. */
-        if (rsrc_session->sessionType == TPM2_SE_POLICY &&
+        if (iesys_fips_mode_enabled() &&
+            rsrc_session->sessionType == TPM2_SE_POLICY &&
             rsrc_session->type_policy_session != POLICY_AUTH &&
             rsrc_session->sizeHmacValue == 0) {
                 auth->hmac.size = 0;
```

> 1. is this the correct location or would an earlier or later check make more sense

Its quite possible that an earlier check might be more appropriate. This is where we ended up, and made sense at the time. It can't be later as the 0 length HMAC would be computed and openssl would error out under FIPS mode.

> 2. are all the new statements required or would it make sense to never call these functions for 0 length keys

You still have to authenticate with the TPM somehow to send and receive commands right, so these functions are likely to be called anyway. It just comes down to how you want to treat the spec when openssl starts doing strange things under FIPS mode.

Like I said, you don't have to take these patches, I just opened up these MRs to see if you are interested in them, or to help any other distros if they implement FIPS SP 800-131A. 

Thanks for your time.
Matthew

moritzbuhl:
We want your patches! And I would like to integrate vendored patches wherever possible.

AndreasFuchsTPM (approved):

PR #3116, opened 2026-07-10
Correctly handle TPM_RH_NULL resources with TPM_RH_PW sessions

chrisccoulson:
Some TPM commands permit passing `TPM_RH_NULL` in a handle slot that
requires an authorization. An example of these are the various
attestation commands which allow passing `TPM_RH_NULL` for the signing
key. This is done by passing `ESYS_TR_NONE` at the ESAPI layer. In this
case, `esys_GetResourceObject` maps this to a NULL node which then gets
converted to `TPM_RH_NULL`. The NULL node is subsequently passed to
`iesys_gen_auths`. If the corresponding session slot is `ESYS_TR_PASSWORD`,
this function fails to set the session handle, resulting in the TPM
returning `TPM_RC_VALUE` for that session slot.

Regardless of what the handle is set to, the corresponding session slot
still needs to be valid. This change sets the session handle to `TPM_RH_PW`
whenever `ESYS_TR_PASSWORD` is supplied, regardless of what the
corresponding resource handle is.

Signed-off-by: Chris Coulson <chris.coulson@amutable.com>

### DIFF
--- a/Makefile-test.am
+++ b/Makefile-test.am
@@ -272,6 +272,7 @@ ESYS_TESTS_INTEGRATION_MANDATORY = \
     test/integration/esys-import.int \
     test/integration/esys-make-credential.int \
     test/integration/esys-make-credential-session.int \
+    test/integration/esys-null-auth-resource.int \
     test/integration/esys-nv-ram-counter.int \
     test/integration/esys-nv-ram-counter-session.int \
     test/integration/esys-nv-ram-counter-session-long-auth.int \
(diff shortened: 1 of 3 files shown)


AndreasFuchsTPM (changes requested):
Would you quickly fix the formatting report from CI:

```
--- a/test/integration/esys-null-auth-resource.int.c
+++ b/test/integration/esys-null-auth-resource.int.c
@@ -39,8 +39,8 @@ test_esys_get_time(ESYS_CONTEXT *esys_context) {
     TSS2_RC r;
     int     failure_return = EXIT_FAILURE;
 
-    TPM2B_ATTEST        *timeInfo = NULL;
-    TPMT_SIGNATURE      *signature = NULL;
+    TPM2B_ATTEST   *timeInfo = NULL;
+    TPMT_SIGNATURE *signature = NULL;
```

AndreasFuchsTPM (approved):

PR #3122, opened 2026-07-29
Size bound sub-types when unmarshalling TPM2B types

chrisccoulson:
When unmarshalling the subtype for a `TPM2B` sized structure type, the
unmarshalling is not bound by the `TPM2B` size field. If a TPM returns an
incorrectly formed payload, it's possible that the subtype may consume
more bytes than is indicated in the size field. In this case, it's
likely that unmarshalling will fail at some undefined point in the future,
but there is no explicit handling of this case.

This adds an explicit size bound for the subtype in the
`TPM2B_UNMARSHAL_SUBTYPE` macro, so that unmarshaling fails immediately if
the subtype tries to consume more bytes than what is indicated in the
size field. It also ensures that unmarshaling fails if the subtype
doesn't consume all bytes provided to it, as indicated by the size field,
which catches trailing bytes.

Signed-off-by: Chris Coulson <chris.coulson@amutable.com>

### DIFF
--- a/src/tss2-mu/tpm2b-types.c
+++ b/src/tss2-mu/tpm2b-types.c
@@ -193,7 +193,7 @@
 #define TPM2B_UNMARSHAL_SUBTYPE(type, subtype, member)                                             \
     TSS2_RC Tss2_MU_##type##_Unmarshal(uint8_t const buffer[], size_t buffer_size, size_t *offset, \
                                        type *dest) {                                               \
-        size_t  local_offset = 0;                                                                  \
+        size_t  local_offset = 0, local_buffer_size = 0;                                           \
         UINT16  size = 0;                                                                          \
         TSS2_RC rc;                                                                                \
                                                                                                    \
@@ -233,11 +233,16 @@
                                                                                                    \
         if (dest != NULL) {                                                                        \
             dest->size = size;                                                                     \
-            rc = Tss2_MU_##subtype##_Unmarshal(buffer, buffer_size, &local_offset, &dest->member); \
-            if (rc)                                                                                \
-                return rc;                                                                         \
-        } else {                                                                                   \
-            local_offset += size;                                                                  \
+        }                                                                                          \
+        local_buffer_size = local_offset + size;                                                   \
+        rc = Tss2_MU_##subtype##_Unmarshal(buffer, local_buffer_size, &local_offset,               \
+                                           dest ? &dest->member : NULL);                           \
+        if (rc)                                                                                    \
+            return rc;                                                                             \
+        if (local_offset != local_buffer_size) {                                                   \
+            LOG_DEBUG("Unmarshaled object size %zu does not match TPM2B size field %u",            \
+                      local_offset - (local_buffer_size - size), size);                            \
+            return TSS2_MU_RC_INSUFFICIENT_BUFFER;                                                 \
         }                                                                                          \
                                                                                                    \
         if (offset != NULL) {                                                                      \
--- a/test/unit/TPM2B-marshal.c
+++ b/test/unit/TPM2B-marshal.c
@@ -290,6 +290,40 @@ tpm2b_unmarshal_success_offset(void **state) {
 }
 
 /*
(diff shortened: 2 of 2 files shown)


AndreasFuchsTPM:
Nicely caught.
I actually wonder, why we do not unmarshal the subtypes in order to verify correctness in cases where dest==NULL.
We should do this in this PR or a different one, I guess...

What I means:
```
if (dest != NULL) {
    dest->size = size;
}
...
rc = Tss2_MU_##subtype##_Unmarshal(buffer, local_buffer_size, &local_offset, (dest)? &dest->member : NULL); \
```
and getting rid of the complete else block.

AndreasFuchsTPM:
Clang-Format-CI-Output:
```diff
diff --git a/src/tss2-mu/tpm2b-types.c b/src/tss2-mu/tpm2b-types.c
index 2ca72e3..bd9173d 100644
--- a/src/tss2-mu/tpm2b-types.c
+++ b/src/tss2-mu/tpm2b-types.c
@@ -234,7 +234,8 @@
         if (dest != NULL) {                                                                        \
             dest->size = size;                                                                     \
             local_buffer_size = local_offset + size;                                               \
-            rc = Tss2_MU_##subtype##_Unmarshal(buffer, local_buffer_size, &local_offset, &dest->member); \
+            rc = Tss2_MU_##subtype##_Unmarshal(buffer, local_buffer_size, &local_offset,           \
+                                               &dest->member);                                     \
             if (rc)                                                                                \
                 return rc;                                                                         \
             if (local_offset != local_buffer_size) {                                               \
diff --git a/test/unit/TPM2B-marshal.c b/test/unit/TPM2B-marshal.c
index 7fbcd61..b4e95c0 100644
--- a/test/unit/TPM2B-marshal.c
+++ b/test/unit/TPM2B-marshal.c
@@ -296,10 +296,10 @@ static void
 tpm2b_unmarshal_subtype_exceeds_size(void **state) {
     TPM2B_ECC_POINT point = { 0 };
     size_t          offset = 0;
-    uint8_t         buffer[] = { 0x00, 0x08,                           /* size = 8, real content is 12 */
+    uint8_t         buffer[] = { 0x00, 0x08, /* size = 8, real content is 12 */
                                  0x00, 0x04, 0xef, 0xbe, 0xad, 0xde,   /* x: 2 + 4 = 6 bytes */
                                  0x00, 0x04, 0x44, 0x33, 0x22, 0x11 }; /* y: 2 + 4 = 6 bytes */
-    TSS2_RC rc;
+    TSS2_RC         rc;
 
     rc = Tss2_MU_TPM2B_ECC_POINT_Unmarshal(buffer, sizeof(buffer), &offset, &point);
     assert_int_equal(rc, TSS2_MU_RC_INSUFFICIENT_BUFFER);
@@ -313,11 +313,11 @@ static void
 tpm2b_unmarshal_subtype_smaller_than_size(void **state) {
     TPM2B_ECC_POINT point = { 0 };
     size_t          offset = 0;
-    uint8_t         buffer[] = { 0x00, 0x10,                         /* size = 16, real content is 12 */
+    uint8_t         buffer[] = { 0x00, 0x10, /* size = 16, real content is 12 */
                                  0x00, 0x04, 0xef, 0xbe, 0xad, 0xde, /* x: 2 + 4 = 6 bytes */
                                  0x00, 0x04, 0x44, 0x33, 0x22, 0x11, /* y: 2 + 4 = 6 bytes */
                                  0x00, 0x00, 0x00, 0x00 };           /* 4 trailing padding bytes */
-    TSS2_RC rc;
+    TSS2_RC         rc;
 
     rc = Tss2_MU_TPM2B_ECC_POINT_Unmarshal(buffer, sizeof(buffer), &offset, &point);
     assert_int_equal(rc, TSS2_MU_RC_INSUFFICIENT_BUFFER);
```

chrisccoulson:
I was out last week, but I've pushed a change to fix the clang-format issue, and I implemented your suggestion to handle the `dest == NULL` case.

moritzbuhl (approved):

PR #3127, opened 2026-08-05
CI: Update container images and versions

moritzbuhl:
* use size optimized dev images
* move from Ubuntu 22.04 to 26.04
* move from Fedora 32 to 44
* move from Alpine 3.15 to 3.24
* add CentOS Stream 10
* update mbedtls from 3.1 to 3.6

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -8,6 +8,9 @@ set -eo pipefail
 source $DOCKER_BUILD_DIR/.ci/docker-prelude.sh
 
 export CONFIGURE_OPTIONS=
+# OpenSSL 3.5 disables SHA-1 signatures by default but the FAPI integration
+# suite contains a legacy SHA-1 signing check.
+export OPENSSL_ENABLE_SHA1_SIGNATURES=1
 
 if [ -d build ]; then
   rm -rf build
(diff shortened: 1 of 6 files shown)


.ci/docker.run
```diff
@@ -102,7 +105,7 @@ fi
 
 if [ "$SCANBUILD" == "yes" ]; then
   scan-build --status-bugs make -j
-elif [ "$CC" == "clang" -a "$DOCKER_IMAGE" == "ubuntu-24.04" ]; then
+elif [ "$CC" == "clang" -a "$DOCKER_IMAGE" == "ubuntu-24.04-dev" ]; then
   make -j compile_commands.json
   run-clang-tidy -header-filter $PWD $(find src include -name '*.c' -o -name '*.h') 2>&1 | tee clang-tidy.log
```
moritzbuhl, line 110: this is currently no longer running in CI because there is no ubuntu-24.04-dev image running the pipeline.
Is this the same as the clang-tidy github action or the clang-format action? Can I delete it then?

AndreasFuchsTPM: I would assume that this is just regular clang-tidy and we can remove this part.

AndreasFuchsTPM (approved):

PR #3128, opened 2026-08-06
fapi test: fix tests failing with physical TPMs.

JuergenReppSIT:
The check previously assumed that the list of current FAPI paths exactly matched the provided checklist.

However, physical TPMs may contain additional NV objects and therefore additional FAPI paths. These extra paths do not indicate an error.

Change the comparison so that it only verifies that every entry from the checklist is present in the current FAPI path list. Additional paths in the FAPI path list are now ignored.

Also one assertion is moved to ensure that a modified password is reset even when the test fails.

Remove the endorsement certificate paths from the checklist because their presence and exact paths may differ when tests are run with physical TPMs.

### DIFF
--- a/test/integration/main-fapi.c
+++ b/test/integration/main-fapi.c
@@ -172,24 +172,28 @@ cmp_jso(json_object *jso1, json_object *jso2) {
     }
 }
 
-/* Compare two delimiter separated token lists. */
+/*
+ * Compare two delimiter separated token lists.
+ * Every token from check_string must be found in computed string.
+ */
 bool
-cmp_strtokens(char *string1, char *string2, char *delimiter) {
+cmp_strtokens(char *computed_string, char *check_string, char *delimiter) {
     bool  found = false;
     char *token1 = NULL;
     char *token2 = NULL;
     char *end_token1;
     char *end_token2;
-    char *string2_copy;
+    char *computed_string_copy;
 
-    string1 = strdup(string1);
-    ASSERT(string1);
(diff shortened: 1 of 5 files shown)


test/integration/main-fapi.c
```diff
@@ -172,42 +172,47 @@ cmp_jso(json_object *jso1, json_object *jso2) {
...
+    token1 = strtok_r(check_string, delimiter, &end_token1);
     while (token1 != NULL) {
         found = false;
-        string2_copy = strdup(string2);
-        ASSERT(string2_copy);
-        token2 = strtok_r(string2_copy, delimiter, &end_token2);
+        computed_string_copy = strdup(computed_string);
+        ASSERT(computed_string_copy);
+        token2 = strtok_r(computed_string_copy, delimiter, &end_token2);
+        LOG_ERROR("XXX %s %s", token1, token2);
```
moritzbuhl: debug logging

moritzbuhl:
Not sure about the CIFuzz test and how my recent changes to the CI causes it to fail. But I'll look into it!

**UPDATE**: The CIFuzz docker image tried to download autoconf-archive from a mirror that was unavailable for the last three days. I created https://github.com/google/oss-fuzz/issues/15977 and https://github.com/google/oss-fuzz/issues/15978 but now the mirror is up again and the patch was just a hotfix to get our CI running again, so I dropped it.
To avoid this in the future and address other problems with CIFuzz I created #3129 for this.

moritzbuhl (approved):

PR #3133, opened 2026-08-11
dist: Fully lock the `tss` system user

dvzrv:
Use the `u!` notation to fully lock the system user. It is highly unlikely that the `tss` user should ever be logged into (e.g. via SSH).

This sysusers.d feature has been released with systemd 257 on 2024-12-10.

Related-to: https://github.com/systemd/systemd/releases/tag/v257


AndreasFuchsTPM:
2 years old systemd is quite recent, so no idea if this would fail for a certain amount of users.
Either we "ifdef" this somehow, or we wait a few more years ?

dvzrv:
> 2 years old systemd is quite recent, so no idea if this would fail for a certain amount of users.

Hm right, depends on the distro maybe.
As I'm packaging this for Arch Linux, it is not a concern there, but may be for others.

> Either we "ifdef" this somehow, or we wait a few more years ?

I'm not sure how something like that would look like, as I'm not an autotools expert (and also don't want to become one :sweat_smile: ).
If you have any suggestions on how to achieve this, please feel free to adjust this PR as needed.

Meanwhile, I have applied this for our distribution package, because we are striving to lock all system user accounts, that don't require login.

moritzbuhl:
I think we should check for the correct systemd version and if it is recent enough use a separate file that uses `u!`.

dvzrv:
Ah, excellent! :rocket: 

I guess that makes it mergeable also now?

AndreasFuchsTPM:
Can we squash this into a single commit ?

AndreasFuchsTPM (approved):

Issue #3132, opened 2026-08-10
Support vendor location for fapi-profiles and fapi-config


dvzrv:
Hi! 👋 

I'm packaging this project for Arch Linux.

I noticed, that you are still heavily relying on /etc/tpm2-tss/ to provide configuration files.

Would it be possible, to instead extend to be [UAPI configuration files specification](https://uapi-group.org/specifications/specs/configuration_files_specification/) compliant?

In general, it would be great to not require the files in /etc/tpm-tss/ to exist (but to be considered if they are there, to accommodate hermetic /usr use-cases and allow for the deletion of /etc as means of system reset.

moritzbuhl:
We can easily add an additional fallback check for `/usr/lib/...` if  `/etc/...` is not available.
Adding `conf.d` stuff is not really useful, adding `/run/...` is currently also not a target.
This way it also should not be an API break.

Personally, I am not sure about libeconf before it reaches a "stable" 1.0.0.

dvzrv:
I understand that adding an additional dependency on libeconf is maybe not that great.
Depending on what you want to do, it might not be necessary though.

I suppose that the most basic approach (i.e. first look into /etc/.. and if a particular file is not there, fall-back to the default in /usr/lib/.. - or even built-in default? - with the file in /etc/.. fully replacing the one in /usr/lib/.. if it exists) should suffice.

The .conf.d support is not really relevant, if it doesn't work for this use-case.
I don't know enough about this project to even have an educated guess on whether support for /run would make sense.

Either way, thanks for looking into this and considering it! 👍

moritzbuhl:
I experimented a bit with adding a default `fapi-config.json` and I came up with the diff below but I am not happy with this.

To me it feels odd to then just hardcode these locations that at the same time become more arbitrary by trying to add more locations other config files could be at.

There already is `ENV_FAPI_CONFIG` to set an arbitrary location and a fallback to /etc/ right now.
If we wanted to encode defaults into the program we should somehow communicate the users what these defaults are.
Further, we should add defaults for all the file path contents and avoid trying to get any of these from the file system.

I have some diff to try `ENV_FAPI_CONFIG` if set, otherwise fail. And if `ENV_FAPI_CONFIG` is not set, try `SYSCONF_DIR` and otherwise `LOCALSTATE_DIR`. I'll propose it as PR.

```diff
diff --git a/configure.ac b/configure.ac
index 42555c0d..07cc9a52 100644
[...]

PR #3139, opened 2026-08-24
FAPI: check SYSCONF_DIR and LOCALSTATE_DIR for fapi-config.json

moritzbuhl:
In case the config file does not exist in SYSCONF_DIR/..., try LOCALSTATE_DIR/...

Create a static copy of ifapi_io_read_async that accepts an already opened file descriptor for this multiple config file locations case.

This is a draft for feature request #3132

### DIFF
--- a/src/tss2-fapi/ifapi_config.c
+++ b/src/tss2-fapi/ifapi_config.c
@@ -5,19 +5,24 @@
  *******************************************************************************/
 
 #ifdef HAVE_CONFIG_H
-#include "config.h" // for SYSCONFDIR
+#include "config.h" // for SYSCONFDIR, LOCALSTATEDIR
 #endif
 
-#include <json.h>   // for json_object_put, json_object
-#include <stdint.h> // for uint8_t
-#include <stdlib.h> // for NULL, getenv, size_t
-#include <string.h> // for strncmp, strlen, memset, strdup
+#include <errno.h>    // for errno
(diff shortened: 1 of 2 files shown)


moritzbuhl:
we could also install the default config into the local state dir instead now

src/tss2-fapi/ifapi_config.c
```diff
@@ -140,6 +146,83 @@ ifapi_json_IFAPI_CONFIG_deserialize(json_object *jso, IFAPI_CONFIG *out) {
...
 
+/** Start reading a fd's complete content into memory in an asynchronous way.
+ *
+ * @param[in,out] io The input/output context being used for file I/O, stream already opened.
+ * @retval TSS2_RC_SUCCESS: if the function call was a success.
+ * @retval TSS2_FAPI_RC_IO_ERROR: if an I/O error was encountered; such as the file was not found.
+ * @retval TSS2_FAPI_RC_MEMORY: if memory could not be allocated to hold the read data.
+ */
+static TSS2_RC
+ifapi_config_read_async(struct IFAPI_IO *io) {
```
moritzbuhl, line 157: move the function to ifapi_io.c

moritzbuhl, line 157: could add a try_read_async that does not log error on open and may communicate a different failure in open

Issue #3126, opened 2026-08-03
--disable-weakcrypto by default


AndreasFuchsTPM:
We've had this switch for year now.
With the upcoming major release, it makes sense to change the default setting.

And if enabling, display a deprecation warning; maybe even mention that it will be removed even in a minor version update in the future.

moritzbuhl:
Just today I noticed that CentOS-stream10 requires setting `OPENSSL_ENABLE_SHA1_SIGNATURES` for `test/integration/fapi-ext-public-key.fint` to pass. I am not sure why I didn't see the failure with Fedora44 but here is some documentation on it:
https://fedoraproject.org/wiki/SHA1SignaturesGuidance

PR #3140, opened 2026-08-25
sys: move from disabling weak crypto to enabling deprecated crypto

moritzbuhl:
Roughly six years ago we introduced a flag that allows disabling weak crypto. Now we see that distros no longer provide certain weak algorithms so it is time to reverse the logic and disable weak crypto by default, require the --enable-deprecated-crypto configure flag to unblock it.

The wording "deprecated" add a slightly better explanation which algorithms are considered in comparison to "weak".

A first step to address #3126.

### DIFF
--- a/src/tss2-sys/sysapi_util.c
+++ b/src/tss2-sys/sysapi_util.c
@@ -303,13 +303,13 @@ GetNumResponseHandles(TPM2_CC commandCode) {
     return GetNumHandles(commandCode, 0);
 }
 
-#ifdef DISABLE_WEAK_CRYPTO
+#ifndef ENABLE_DEPRECATED_CRYPTO
 bool
-IsAlgorithmWeak(TPM2_ALG_ID algorithm, TPM2_KEY_SIZE key_size) {
+IsAlgorithmDeprecated(TPM2_ALG_ID algorithm, TPM2_KEY_SIZE key_size) {
     switch (algorithm) {
     case TPM2_ALG_RSA:
         if (key_size < 2048) {
-            LOG_ERROR("Error: weak algorithm");
+            LOG_ERROR("Error: deprecated algorithm");
             return true;
         }
         break;
@@ -318,12 +318,12 @@ IsAlgorithmWeak(TPM2_ALG_ID algorithm, TPM2_KEY_SIZE key_size) {
     case TPM2_ALG_CAMELLIA:
     case TPM2_ALG_SYMCIPHER:
         if (key_size < 128) {
-            LOG_ERROR("Error: weak algorithm");
+            LOG_ERROR("Error: deprecated algorithm");
             return true;
         }
         break;
     case TPM2_ALG_SHA1:
-        LOG_ERROR("Error: weak algorithm");
+        LOG_ERROR("Error: deprecated algorithm");
         return true;
         break;
     }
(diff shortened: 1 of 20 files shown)


src/tss2-sys/sysapi_util.c
```diff
@@ -303,13 +303,13 @@ GetNumResponseHandles(TPM2_CC commandCode) {
     return GetNumHandles(commandCode, 0);
 }
 
-#ifdef DISABLE_WEAK_CRYPTO
+#ifndef ENABLE_DEPRECATED_CRYPTO
```
AndreasFuchsTPM, line 306: Where is the `#else` case there the functions are defined if `ENABLE_DEPRECATED_CRYPTO` is set ?

AndreasFuchsTPM (changes requested):
I guess we should have a test-case in CI that tests `--enable-deprecated-crypto`

src/tss2-sys/sysapi_util.h
```diff
@@ -110,7 +111,7 @@ TSS2_RC ValidateTPML_PCR_SELECTION(const TPML_PCR_SELECTION *pcr_selection);
  * static inline is not portable, so make these empty defines to reduce generating functions
  * and thus binary size for them.
  */
-#define IsAlgorithmWeak(...)            TSS2_RC_SUCCESS
+#define IsAlgorithmDeprecated(...)      TSS2_RC_SUCCESS
```
moritzbuhl, line 114: here!

src/tss2-sys/sysapi_util.c
moritzbuhl: The header file defines them away:

```
#ifndef ENABLE_DEPRECATED_CRYPTO
bool    IsAlgorithmDeprecated(TPM2_ALG_ID algorith, TPM2_KEY_SIZE key_size);
TSS2_RC ValidatePublicTemplate(const TPM2B_PUBLIC *pub);
TSS2_RC ValidateNV_Public(const TPM2B_NV_PUBLIC *nv_public_info);
TSS2_RC ValidateTPML_PCR_SELECTION(const TPML_PCR_SELECTION *pcr_selection);
 * static inline is not portable, so make these empty defines to reduce generating functions
 * and thus binary size for them.
 */
#define IsAlgorithmDeprecated(...)      TSS2_RC_SUCCESS 
#define ValidatePublicTemplate(...)     TSS2_RC_SUCCESS
#define ValidateNV_Public(...)          TSS2_RC_SUCCESS
#define ValidateTPML_PCR_SELECTION(...) TSS2_RC_SUCCESS
#endif
```

moritzbuhl:
We should avoid SHA1 in `ifapi_policy_calculate.c` too. `fapi_crypto.c` too.
`ifapi_calculate_policy` should check the hash alg before doing anything.
Duplicate the supported check in fapi to avoid having to export the symbols from sapi. a static util lib that could be included in all would of course be better

AndreasFuchsTPM:
ifapi_policy_calculate.c:

ifapi_calculate_policy    
ifapi_compute_policy_pcr   no sha1 pcrs
ifapi_calculate_policy_signed  publickey ok ?
ifapi_calculate_policy_authorize_nv  no sha1
ifapi_calculate_policy_authorize   no sha1

PR #3149, opened 2026-09-15
Enable TCP_NODELAY in the 'swtpm' and 'mssim' TCTI modules

danieltrick:
This patch enables the [`TCP_NODELAY`](https://brooker.co.za/blog/2024/05/09/nagle.html) flag (turns off Nagle's algorithm) for TCP connections in the 'swtpm' and 'mssim' TCTI modules. We have observed significantly lower delays for TPM operations with this change in our tests.

It is safe to do this directly in `socket_connect()`, since that function is *only* used in `tcti-mssim.c` and `tcti-swtpm.c`.


src/util-io/io.c
```diff
@@ -182,6 +183,15 @@ socket_close(SOCKET *socket) {
     return TSS2_RC_SUCCESS;
 }
 
+static TSS2_RC
+socket_set_nodelay(SOCKET sock) {
```
moritzbuhl, line 187: I am confused why this is wrapped in another function.

danieltrick: It is not neccessary, but I think it makes the code more clean/readable.

(also: I did not want to re-use `ret` as `flag` or declare `flag` as a separate var inside `socket_connect()`)

moritzbuhl: I think this is the right approach as there already is
socket_set_nonblock
socket_recv_buf
socket_close
socket_connect
socket_poll

moritzbuhl (approved):

